PR #8, opened 2015-11-19
Changed Dockerfile to be based on Ubuntu

### LINKED ISSUE #6: Docker images should be based on Ubuntu images rather than Debian
CLI tooling currently only supports Ubuntu therefore the base image for the dotnet docker images should be switched from Debian based to Ubuntu.  (see https://github.com/dotnet/cli/issues/212)

MichaelSimons:
Changed Dockerfile to be based on Ubuntu since CLI doesn't yet have full support for Debian.  This is a short term change until CLI has full support for Debian.

Fix #6

### DIFF
--- a/0.0.1-alpha/Dockerfile
+++ b/0.0.1-alpha/Dockerfile
@@ -1,10 +1,10 @@
-FROM buildpack-deps:jessie-scm
+FROM buildpack-deps:trusty-scm
 
 # install dotnet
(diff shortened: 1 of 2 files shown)


dleeapho:
In hindsight it's a  obvious fix for #6. LGTM

ahmetalpbalkan:
LGTM

PR #9, opened 2015-11-19
Updating README.md files to be symbolic links.

### LINKED ISSUE #5: README.md at the repo root should be a symblink to the README.md next to the Dockerfile

MichaelSimons:
Modified the various README.md files to be symbolic links of a single README.md file located in the 0.0.1-alpha directory.  The multiple README.md files are required by Docker Hub as it picks up the README.md that is located next to the Dockerfile.

Fixes #5


ahmetalpbalkan:
Actually you shouldn't do like this. Docker Hub does not specify how it picks up README files and it is not predictable. Please have a single README file in the repo (that should be next to a dockerfile) and symlink from the root of the repo to that. I believe you don't have a reason to maintain multiple readme files. Also docker hub page of the image does not tell which Dockerfile will be shown in the Dockerfile tab. :-) so for these sort of things PTAL to aspnet-docker repo as we've figured all these out more than a year ago :-)

MichaelSimons:
@ahmetalpbalkan, Yes the docker hub behavior regarding which Dockerfile is shown is unfortunate and annoying.

Per your comments, the only thing I see to change is to remove the symlink README.md we have in the onbuild folder for the onbuild Dockerfile.  I believe everything else you stated is already in place.

0.0.1-alpha/onbuild/README.md
```diff
@@ -1,60 +0,0 @@
-Dotnet CLI Preview Docker Image
```
ahmetalpbalkan: I don't think you need this file at all. It'll use any README.md it would find next to a Dockerfile.

ahmetalpbalkan:
sorry to introduce resolve conflicts.

```
unlink ./0.0.1-alpha/onbuild/README.md
ln -s 0.0.1-alpha/README.md ./README.md
```

should fix the issue.

MichaelSimons:
@ahmetalpbalkan, I removed the readme in the onbuild dir, merged your root readme changes into the 0.0.1-alpha/readme and then kept the symlink readme I had already created at the root.

ahmetalpbalkan:
LGTM.

PR #10, opened 2015-11-19
Fix image name in readme

### LINKED ISSUE #7: documentation - FROM dotnet:<version> doesn't work when dotnet doesn't exist in library.
`FROM dotnet:<version>` in the documentation should be `FROM microsoft/dotnet:<version>` for those who don't already have the container in their library.

I'm assuming this is a general issue, but could possibly just be my instance. I'm running DockerToolbox on windows 8.1. After adding `microsoft/` in the FROM statement, seems to work fine.

ahmetalpbalkan:
Also unify dotnet, DotNet, Dotnet, .NET. Fixes #7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
-Dotnet CLI Preview Docker Image
+.NET CLI Preview Docker Image
 ====================
 
-This repository contains `Dockerfile` definitions for [Dotnet/cli] Docker images.
+This repository contains `Dockerfile` definitions for [dotnet/cli] Docker images.
 
-This project is part of .NET Core command-line (CLI) tools. You can find samples, documentation, and getting started instructions for .NET Core CLI tools at the [Dotnet/cli] repo.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,25 +36,25 @@ $ docker run -it --rm --name myRunningApp myDotNetApp
...
+$ docker run --rm -v "$PWD":/myapp -w /myapp microsoft/dotnet:0.0.1-alpha dotnet compile
 ```
 
 This will add your current directory as a volume to the container, set the working directory to the volume, and run the command `dotnet compile` which will tell dotnet to compile the project in the working directory.
 
-# Image Variants
+# Image variants
 
-The `dotnet` images come in many flavors, each designed for a specific use case.
+The `microsoft/dotnet` image come in many flavors, each designed for a specific use case:
```
MichaelSimons: ? '.' seems more correct.

ahmetalpbalkan: fixed

MichaelSimons:
LGTM

PR #13, opened 2015-11-20
Modified the base Dockerfile to install dotnet Debian repo

### LINKED ISSUE #12: Dockerfile should use Debian package repo to get the dotnet package
Currently the base Dockerfile obtains the dotnet Debian package via a public drop share.  This was done because there wasn't a public Debian package repo available.  There is one now available (http://apt-mo.trafficmanager.net/repos/dotnet/) and it should be used.  This eliminates the need to manually install dependent packages.

MichaelSimons:
Modified the base Dockerfile to install dotnet via the public Debian package repo.  Also removed the logic that set the DOTNET_HOME environment variable since .NET CLI addressed the issue that required this.

fixes #12

### DIFF
--- a/0.0.1-alpha/Dockerfile
+++ b/0.0.1-alpha/Dockerfile
@@ -1,13 +1,7 @@
 FROM buildpack-deps:trusty-scm
 
 # install dotnet
-RUN echo "deb [arch=amd64] http://llvm.org/apt/trusty/ llvm-toolchain-trusty-3.6 main" > /etc/apt/sources.list.d/llvm.list \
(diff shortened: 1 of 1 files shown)


0.0.1-alpha/Dockerfile
```diff
@@ -1,13 +1,7 @@
 FROM buildpack-deps:trusty-scm
 
 # install dotnet
```
ahmetalpbalkan: maybe even get rid of this?

ahmetalpbalkan:
LGTM

0.0.1-alpha/Dockerfile
MichaelSimons: Removed

PR #25, opened 2016-03-16
Temporarily work around LTTng failed assertion

naamunds:
Same change as in https://github.com/dotnet/dotnet-docker-preview/pull/2

cc: @MichaelSimons @dleeapho


MichaelSimons:
:ship:

PR #36, opened 2016-05-10
Added preview1/rc2 images for .NET Core and CLI Tools.

MichaelSimons:
These images are available under a [test Docker Hub repository](https://hub.docker.com/r/msimons/dotnet/) if you want to pull them.

@naamunds, @dleeapho @glennc


glennc:
This still requires @blackdwarf to add text and such to the .md right?

MichaelSimons:
@glennc correct the readme content is just a placeholder until @blackdwarf provides the official content.  My apologies for not calling this out.

1.0.0-preview1/onbuild/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM microsoft/dotnet:1.0.0-preview1
```
dleeapho: Should we be thinking of `onbuild` based off of `:core` instead? With `ENTRYPOINT ["dotnet"]` users would build in their publish output folder and run their (portable) app with `docker run microsoft/dotnet:onbuild myapp.dll` it's more lightweight and I'm not sure whether the typical onbuild scenario needs the full sdk.

MichaelSimons: @dleeapho I am not sure I am following?  onbuild as defined has to have the full sdk in order to restore and build.  Where/when do you see this happening?

dleeapho: I positing that there is another scenario for onbuild, instead of restoring and building in-container that onbuild could also be used to quickly run any portable app. The onbuild would copy the `dotnet publish` directory contents and `myapp.dll` would be the args to the `dotnet` (muxer) entrypoint. In any case we can wait and see if there is a need for what I'm proposing.

PR #49, opened 2016-06-02
Removed the liblldb package from the image

### LINKED ISSUE #48: llvm repo is down which prevents microsoft/dotnet image from building.
The llvm repo was taken down unexpectedly and there is no immediate signs of it coming back soon.  The microsoft/dotnet repo uses this to get the liblldb-3.6 package.  This package is only used if you try to use the sos plugin for lldb.  It is not a critical dependency and has been argued that it should be an optional dependency.  As a result the Dockerfile is failing to build.  Additionally if you pull the latest image, you will get errors when running apt-get update because this repo is registered in the image yet is unavailable.

MichaelSimons:
The llvm repo was taken down unexpectedly and there is no immediate signs of it coming back soon. The microsoft/dotnet repo uses this to get the liblldb-3.6 package. This package is only used if you try to use the sos plugin for lldb. It is not a critical dependency and has been argued that it should be an optional dependency. As a result the Dockerfile is failing to build. Additionally if you pull the latest image, you will get errors when running apt-get update because this repo is registered in the image yet is unavailable.

Fixes #48 

@dleeapho, @naamunds

### DIFF
--- a/1.0.0-preview1/Dockerfile
+++ b/1.0.0-preview1/Dockerfile
@@ -7,16 +7,13 @@ FROM buildpack-deps:jessie-scm
 ENV LTTNG_UST_REGISTER_TIMEOUT 0
 
 # Install .NET CLI dependencies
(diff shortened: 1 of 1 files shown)


naamunds:
:shipit:

PR #54, opened 2016-06-03
Added preview2/rc3 Dockerfiles

MichaelSimons:
Initial version of the preview2/rc3 Dockerfiles.  These are a copy of the preview1/rc2 images with the only changes being the SDK/Runtime versions installed and the necessary image name changes.  After this is merged, I will be making further changes but wanted to get this in first so that the other changes are easier to review.

@naamunds


naamunds:
Should there be a 1.0.0-preview2/README.md, analogous to 1.0.0-preview1/README.md?

1.0.0-rc2/core-deps/hooks/post_push
```diff
@@ -11,20 +11,9 @@ if [[ $DOCKERCFG == {\"auths\":* ]]; then
 fi
 
 
-echo "Pushing latest core-deps image"
+echo "Pushing core images"
```
naamunds: Should this be singular ("image")?

1.0.0-preview1/README.md
```diff
@@ -1 +0,0 @@
-../1.0.0-rc2/README.md
```
naamunds: Is this being updated to point to the rc3 README?  For some reason it's not clear from the diff, and when I click "View", it shows that it's the same as it is in the current version of the master branch of this repo (i.e. still pointing to rc2).

1.0.0-rc3/README.md
```diff
@@ -18,10 +18,14 @@ This project is part of .NET Core command-line (CLI) tools. You can find samples
 -       [`0.0.1-alpha-onbuild`, (*0.0.1-alpha/onbuild/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/0.0.1-alpha/onbuild/Dockerfile)
 -       [`1.0.0-preview1`, `latest` (*1.0.0-preview1/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview1/Dockerfile)
```
naamunds: Should the `latest` tag be removed from preview1?

1.0.0-preview1/README.md
MichaelSimons: That is because I removed the file with my changes.

1.0.0-rc3/core-deps/hooks/post_push
```diff
@@ -0,0 +1,30 @@
...
+    jsonStartIndex=$(expr index "$DOCKERCFG" :)
+    length=$((${#DOCKERCFG} - $jsonStartIndex - 1))
+    echo ${DOCKERCFG:jsonStartIndex:length} > /root/.dockercfg
+fi
+
+
+echo "Pushing latest core-deps image"
+tagStartIndex=$(expr index "$IMAGE_NAME" :)
+repoName=${IMAGE_NAME:0:tagStartIndex - 1}
+latestImageName=$repoName":core-deps"
```
naamunds: Why do we need to push an image with the `core-deps` tag in the post_push hook?  I thought that an image with that tag would automatically be pushed without this hook.  It looks like the post_push in 1.0.0-preview2 is being used to push the 1.0.0-preview2 image with a tag of `latest`, and to push the images that depend on it, so I expected that this post_push would be used similarly.  (In this case, the core-deps image shouldn't have the `latest` tag, so I expected that just the `core` images would be pushed from this post_push.)

MichaelSimons:
@naamunds, The README.md needs to be next to at a minimum once docker file that is auto built in order for it to get copied into the Docker Hub repo.  That is the only reason it exists outside of the root of the repo.  Also I believe it cannot be a link in order for Docker Hub to work.

naamunds:
LGTM, with some minor comments.

1.0.0-rc2/core-deps/hooks/post_push
MichaelSimons: Fixed here as well as the 1.0.0-preview2 post_push script.

1.0.0-rc3/README.md
MichaelSimons: yes, as well as the other non-versioned rc2/preview1 tags.

1.0.0-rc3/core-deps/hooks/post_push
MichaelSimons: All image variants should have a non-versioned tag.  The primary purpose of this hook is to trigger the building of the dependent core image.  Docker Hub does not have support for triggering the build of dependent images within the same repo.  Additionally, Docker Hub does not have support for multi-tagging an image.  Normally multi-builds would be used to get around the multi-tagging issue - in this case `<versioned>-core-deps` and  `core-deps`.  This does not work well however when you are using the build hooks to build and push additional images - they would get built and pushed multiple times unnecessarily.  Therefore I decided it would be best to use the build hooks to push the latest 'core-deps' with a build hook to get around this issue.

PR #55, opened 2016-06-06
Updated .NET CLI dependencies

### LINKED ISSUE #52: Update required packages to be in sync with CLI
See dotnet/cli#3088. The clang-3.5 dependency has been removed..

MichaelSimons:
Fixes #52 

@naamunds


naamunds:
Should the `libgssapi-krb5-2` package also be added to preview1 and rc2?  According to https://github.com/dotnet/cli/issues/3088#issuecomment-220092740, this dependency was added in mid-March, which I think was before those releases.

MichaelSimons:
@naamunds, I don't think it is necessary at this point.

naamunds:
:shipit:

PR #44, opened 2016-05-27
windows dockerfile

friism:

PR #56, opened 2016-06-08
Added windowsservercore and nanoserver images.

MichaelSimons:
The Nano Server SDK image is currently affected by https://github.com/dotnet/roslyn/issues/11592.  

@friism - This builds upon your https://github.com/dotnet/dotnet-docker/pull/44 and adds the full set of images for x86/x64, onbuild, core, and windowsservercore/nanoserver.

@blackdwarf - How should we communicate windowsservercore/nanoserver support within the readme?  One thing to call out is that the core-deps is only applicable to Linux

@naamunds, @dleeapho


MichaelSimons:
If you would like to give these a spin without building them yourself, I have them uploaded to my Docker Hub repo - https://hub.docker.com/r/msimons/dotnet/

MichaelSimons:
The latest [changes](https://github.com/dotnet/dotnet-docker/pull/56/commits/e29d62515de24907825ae111745b0f615559f955) have addressed the Nano Server issues.  

@naamunds, @dleeapho - these changes are ready to be merged into dev, please take a look.

1.0.0-preview2/nanoserver/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ENV DOTNET_SDK_VERSION 1.0.0-preview2-003041
+ENV DOTNET_SDK_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/$DOTNET_SDK_VERSION/dotnet-dev-win-x64.$DOTNET_SDK_VERSION.zip
+
+RUN powershell -NoProfile -Command \
+        $ErrorActionPreference = 'Stop'; \
+        $httpClient = New-Object System.Net.Http.HttpClient; \
+        $stream = $httpClient.GetAsync('%DOTNET_SDK_DOWNLOAD_URL%').Result.Content.ReadAsStreamAsync().Result; \
+        $file = [System.IO.File]::Create('dotnet.zip'); \
+        $stream.CopyTo($file); \
+        $file.Close(); \
```
naamunds: Should $stream also be closed?

friism: I'd also be curious to learn whether a less crazy way to download stuff on nano is on the way... :-)

1.0.0-preview2/nanoserver/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+        [System.IO.Compression.ZipFile]::ExtractToDirectory('dotnet.zip', '%ProgramFiles%\dotnet'); \
+        Remove-Item -Force 'dotnet.zip'
+
+RUN setx /M PATH "%PATH%;%ProgramFiles%\dotnet"
+
+# Copy forwarders assemblies to workaround https://github.com/dotnet/cli/issues/2967
```
naamunds: Should we log an issue to track removing this workaround when it's fixed in a version of Nano Server that's available in a Docker image?  My understanding from [https://github.com/dotnet/cli/issues/2967](https://github.com/dotnet/cli/issues/2967) is that it's fixed in version 14339 of Nano, but that that version isn't publicly available.

1.0.0-preview2/nanoserver/Dockerfile
MichaelSimons: For this particular scenario, I think it would be awesome if ADD could be used.  It would need to understand zip files and extract them just as it does for the other compression formats it currently supports.

1.0.0-preview2/nanoserver/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN powershell -NoProfile -Command \
+        $ErrorActionPreference = 'Stop'; \
+        Copy-Item -Path '%windir%\System32\forwarders\*' -Destination '%ProgramFiles%\dotnet\'; \
+        Copy-Item -Path '%windir%\System32\forwarders\*' -Destination '%ProgramFiles%\dotnet\shared\Microsoft.NETCore.App\1.0.0-rc3-004459-00\'
```
naamunds: I didn't see any mention of needing to copy the forwarders to both folders in [https://github.com/dotnet/cli/issues/2967](https://github.com/dotnet/cli/issues/2967).  Will the release notes mentioned there contain this information?

1.0.0-preview2/nanoserver/Dockerfile
MichaelSimons: Yes I was planning to log an issue to remove this once this PR is merged.

1.0.0-preview2/nanoserver/Dockerfile
friism: @MichaelSimons looking at the library linux dockerfiles the remote-fetch part of `ADD` is never used. This is typically because it cannot be combined in "fetch-archive -> extract -> delete-archive" `RUN`.

Details: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy

naamunds:
> FROM microsoft/dotnet:1.0.0-preview2

Are we going to have tags that include "debian" in their name, similar to the tags that include "nanoserver" and "windowsservercore"?  I don't see any on [https://hub.docker.com/r/msimons/dotnet/tags/](https://hub.docker.com/r/msimons/dotnet/tags/) , and I'm wondering if having them might be good for clarity, like how buildpack-deps has an explicit "debian" tag even though it appears that "latest" is equivalent to "debian" for that repo.

---
Refers to: 1.0.0-preview2/debian/onbuild/Dockerfile:1 in e29d625. [](commit_id = e29d62515de24907825ae111745b0f615559f955, deletion_comment = False)

1.0.0-preview2/nanoserver/Dockerfile
MichaelSimons: There is an active thread on the official recommendation from the nano team.  From this, it is looking like they are going to recommend coping the forwarders (maybe a specific subset) into the c:\windows\system32 folder.

Also the issue you mentioned is not up to date with the latest state of things.

friism:
@naamunds the image/registry machinery has support for resolving that automatically: https://blog.docker.com/2016/04/docker-windows-server-tp5/ (but not support for distinguishing Linux distros because Linux is just Linux to Docker).

1.0.0-preview2/windowsservercore/x64/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM windowsservercore
+
+# Install .NET Core SDK
+ENV DOTNET_SDK_VERSION 1.0.0-preview2-003041
+ENV DOTNET_SDK_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/$DOTNET_SDK_VERSION/dotnet-dev-win-x64.$DOTNET_SDK_VERSION.zip
+
+RUN powershell -NoProfile -Command \
+        $ErrorActionPreference = 'Stop'; \
+        Invoke-WebRequest '%DOTNET_SDK_DOWNLOAD_URL%' -OutFile 'dotnet.zip'; \
```
naamunds: I'm wondering if it would be better to use the same code for this as in the nanoserver Dockerfile in order to make the codebase easier to maintain, unless there are advantages to using Invoke-WebRequest instead of HttpClient beyond the code being a bit more readable.

1.0.0-preview2/nanoserver/Dockerfile
MichaelSimons: @naamunds, As this pattern is complex, I opted not to explicitly close the stream and will let it get disposed when powershell.exe exits.

1.0.0-rc3/windowsservercore/x64/core/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM windowsservercore
+
+# Install .NET Core
+ENV DOTNET_VERSION 1.0.0-rc3-004459-00
+ENV DOTNET_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/$DOTNET_VERSION/dotnet-win-x64.$DOTNET_VERSION.zip
+
+RUN @powershell -NoProfile -Command \
```
naamunds: Why is an @ symbol used here, but not in 1.0.0-preview2/windowsservercore/x64/Dockerfile?  Also, is there a reason for that file to have quotes around the URL in the Invoke-WebRequest command, but this one not to have quotes there?

I'm guessing neither of these differences have a behavioral impact, but I noticed them in a diff of the two files, so I figured I'd mention it to make sure.  (This applies to the x86 files also.)

naamunds:
:shipit:

1.0.0-preview2/windowsservercore/x64/Dockerfile
MichaelSimons: I went back and forth on this myself.  The reason I went with this approach is that I think this is the recommended way to install on Window Server Core as it is more concise, less error prone, and easier to read.  People will reference this Dockerfile as an example and therefore should use the recommended way.  If this was internal, I would agree with you that the two should use the same approach for maintenance reasons.

1.0.0-rc3/windowsservercore/x64/core/Dockerfile
MichaelSimons: There is no observable difference.  I will make them consistent.

1.0.0-preview2/nanoserver/Dockerfile
MichaelSimons: @friism, I'll ask around to see if there is going to be a more intuitive way to download on Nano.

PR #65, opened 2016-06-22
Triggered the building of local package cache in the SDK images

### LINKED ISSUE #58: Dockerfile should trigger the building of the local package cache.
With the introduction of the [offline feature](https://github.com/dotnet/cli/pull/3485), the dotnet CLI builds a local package cache on first use.  The dotnet Dockerfile should trigger this to happen.  This takes some time therefore it will be a savings for all users of the image.  Additionally having the package cache in the base image will save disc space for scenarios in which base image is used multiple times on the same machine as the base layer will be shared.

MichaelSimons:
Fixes #58 

@naamunds, @dleeapho


naamunds:
:shipit:

PR #76, opened 2016-06-27
Update readme for Windows Containers

MichaelSimons:
I manually ported https://github.com/MichaelSimons/dotnet-docker/pull/3 over to master and addressed the PR feedback.  I didn't want to duplicate all of the documentation for the windowsservercore and nanoserver variants - there are 6 of them.  A decision was made to take a different approach and left the `image variants` section alone since that is just intended to describe the 4 main variations (sdk, onbuild, core, core-deps) of the images.  The variations will remain the same as we move towards a docker manifest.  A Windows Containers section was then added which contains the differentiated verbiage.

@naamunds, @dleeapho, @blackdwarf, @PatrickLang

### DIFF
--- a/1.0.0/README.md
+++ b/1.0.0/README.md
@@ -19,17 +19,23 @@ This project is part of .NET Core command-line (CLI) tools. You can find samples
 -       [`1.0.0-preview1` (*1.0.0-preview1/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview1/Dockerfile)
 -       [`1.0.0-preview1-onbuild` (*1.0.0-preview1/onbuild/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview1/onbuild/Dockerfile)
 -       [`1.0.0-preview2-sdk`, `latest` (*1.0.0-preview2/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2/debian/Dockerfile)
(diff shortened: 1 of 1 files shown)


1.0.0/README.md
```diff
@@ -60,13 +66,25 @@ $ docker build -t my-dotnet-app .
...
 
+### `microsoft/dotnet:<version>-core`
+
+This image contains only .NET Core (runtime and libraries) and it is optimized for running [portable .NET Core applications](https://docs.microsoft.com/en-us/dotnet/articles/core/app-types). If you wish to run self-contained applications, please use the `core-deps` image described below. 
+
 ### `microsoft/dotnet:<version>-core-deps`
 
 This image contains the operating system with all of the native dependencies needed by .NET Core. Use this image to:
 
 1. Run a [self-contained](http://dotnet.github.io/docs/core-concepts/app-types.html) application.
```
naamunds: The link here will also need to be updated.  Because the doc page is fairly long, it may be useful to put a link directly to the intended section: https://docs.microsoft.com/en-us/dotnet/articles/core/app-types#self-contained-application.  The link for portable apps would be https://docs.microsoft.com/en-us/dotnet/articles/core/app-types#portable-applications.

naamunds:
LGTM (with one comment).

1.0.0/README.md
dleeapho: I like that the link goes to the beginning of the doc and not to the specific section-- it's good to read the entire doc to understand the entirety of the picture.

dleeapho:
LGTM

PatrickLang:
Took a brief look, looks good to me.

blackdwarf:
:shipit:

1.0.0/README.md
MichaelSimons: I updated the link but left it pointing to the beginning of the doc.

PR #80, opened 2016-06-30
Clarify onbuild image description

naamunds:
cc: @MichaelSimons @dleeapho

### DIFF
--- a/1.0.0/README.md
+++ b/1.0.0/README.md
@@ -49,11 +49,7 @@ applications. Use this image for your development process (developing, building 
 
 ### `microsoft/dotnet:<version>-onbuild`
 
-The most straightforward way to use this image is to use a .NET container as both the build and runtime environment. In your `Dockerfile`, writing something along the lines of the following will compile and run your project:
-
-```dockerfile
-FROM microsoft/dotnet:onbuild
-```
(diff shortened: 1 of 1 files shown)


1.0.0/README.md
```diff
@@ -49,11 +49,7 @@ applications. Use this image for your development process (developing, building
 
 ### `microsoft/dotnet:<version>-onbuild`
 
-The most straightforward way to use this image is to use a .NET container as both the build and runtime environment. In your `Dockerfile`, writing something along the lines of the following will compile and run your project:
-
-```dockerfile
-FROM microsoft/dotnet:onbuild
-```
+The most straightforward way to use this image is to use a Docker container as both the build and runtime environment for your application. A `Dockerfile` in the same directory as your project files which uses this image as its parent will compile and run your project.
```
MichaelSimons: What do you think about calling out more that the user's Dockerfile only needs the FROM? 

Creating a simple `Dockerfile` with the following content in the same directory as your project files will compile and run your project
```
FROM microsoft/dotnet:onbuild
```

naamunds: Good idea. I've made that change.

MichaelSimons:
:shipit:

dleeapho:
LGTM

PR #87, opened 2016-07-28
Removed the scripts to build the Windows images

MichaelSimons:
These can be removed since they have been replaced with a proper build definition.

@naamunds


naamunds:
:shipit:

PR #95, opened 2016-08-08
Added an ISSUE_TEMPLATE

MichaelSimons:
There has been an increase in the number of issues that do not include this basic information.  Hopefully by introducing a template, authors will include this information and the issues will get resolved quicker.


MichaelSimons:
cc @naamunds

naamunds:
:shipit:

PR #101, opened 2016-09-12
Add tags for major and minor version numbers

naamunds:
Also update folder names to use minor version numbers for released
products.

cc: @MichaelSimons


MichaelSimons:
:shipit:

MichaelSimons:
Replaced with Updated #103

PR #105, opened 2016-09-13
Removed alpha and preview1/rc2 images from the readme

MichaelSimons:
These cause clutter and confusion for newbies.

@dleeapho


dleeapho:
:shipit:

PR #106, opened 2016-09-13
Removed obsolete alpha artifacts

MichaelSimons:
@naamunds


MichaelSimons:
Also I should note that we are planning on deleting the alpha tags from docker hub since this is an obsolete pre-release.

naamunds:
:shipit:

PR #107, opened 2016-09-13
Removing obsolete 1.0.0-preview1 and 1.0.0-rc2 source

MichaelSimons:
These images are no longer being maintained and built.  The images (tags) will remain in Docker Hub however.

@naamunds


naamunds:
:ship:

PR #113, opened 2016-09-19
Add link to Docker images doc

spboyer:
Adding link to Building Docker Images for .NET Core Applications per @stevelasker request.

### DIFF
--- a/1.0/README.md
+++ b/1.0/README.md
@@ -10,6 +10,7 @@ This project is part of .NET Core command-line (CLI) tools. You can find samples
 [![Downloads from Docker Hub](https://img.shields.io/docker/pulls/microsoft/dotnet.svg)](https://registry.hub.docker.com/u/microsoft/dotnet)
 [![Stars on Docker Hub](https://img.shields.io/docker/stars/microsoft/dotnet.svg)](https://registry.hub.docker.com/u/microsoft/dotnet)
 
+See [Building Docker Images for .NET Core Applications](https://docs.microsoft.com/en-us/dotnet/articles/core/docker/building-net-docker-images) to get and understanding of the different Docker images that are offered and when is the right use case for them.
 
 ## Supported tags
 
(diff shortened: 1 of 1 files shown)


1.0/README.md
```diff
@@ -10,6 +10,7 @@ This project is part of .NET Core command-line (CLI) tools. You can find samples
 [![Downloads from Docker Hub](https://img.shields.io/docker/pulls/microsoft/dotnet.svg)](https://registry.hub.docker.com/u/microsoft/dotnet)
 [![Stars on Docker Hub](https://img.shields.io/docker/stars/microsoft/dotnet.svg)](https://registry.hub.docker.com/u/microsoft/dotnet)
 
+See [Building Docker Images for .NET Core Applications](https://docs.microsoft.com/en-us/dotnet/articles/core/docker/building-net-docker-images) to get and understanding of the different Docker images that are offered and when is the right use case for them.
```
MichaelSimons: >and  [](start = 152, length = 4)

Should this be 'an'?

MichaelSimons: The placement of this doesn't feel right to me.  One thing that seems odd is that it comes after the stars and downloads.  What do you think about moving this to the Image variants section?  It feels like that would be a better fit.

MichaelSimons: That comment is in reference to `to get and understanding`

naamunds: I agree that it would fit better in the `Image variants` section.

naamunds:
I think this documentation should ideally describe the same image tags as this repo's README; for the runtime images, that would be 1.0.1 rather than 1.0.0.  However, the Yeoman generator for Docker appears to use `1.0.0-core` as the base image, so perhaps the documentation shouldn't be updated until Yeoman is updated.

SteveLasker:
I don't know if we should wait for the yoeman generators are updated. The generators are scaffolders. Customers can update those on their own. They will also want/need to update existing projects so I wouldn't block on everything getting in sync. The main thing is we want to provide context for the various images and why

1.0/README.md
spboyer: fixed

spboyer: moved, makes sense

PR #122, opened 2016-09-28
Update to windows 10.0.14393.206

### LINKED ISSUE #75: Remove the reverse forwarders workaround that was added to the Nano Server images
In Windows Server 2016 TP5 there is an issue with the reverse forwarders feature.  To workaround this, the dotnet images for nano server copy the forwarders assemblies to the Shared Framework folder.  This workaround should be removed once the underlying issue is resolved in Windows Server.

### LINKED ISSUE #120: Update Dockerfile to support Win2016 release images
Latest is now at 10.0.14393.206 for NanoServer and WindowsServerCore images. Given that Windows Server 2016 has been released, it makes sense to have the Dockerfiles be based on at least the release version, if not just latest.

MichaelSimons:
With this new version the forwarders workaround is no longer needed therefore it was removed.

Fixes #120 
Fixes #75

cc @naamunds

### DIFF
--- a/1.0.0-preview2/nanoserver/Dockerfile
+++ b/1.0.0-preview2/nanoserver/Dockerfile
@@ -1,4 +1,4 @@
-FROM microsoft/nanoserver:10.0.14300.1030
+FROM microsoft/nanoserver:10.0.14393.206
 
 # Install .NET Core SDK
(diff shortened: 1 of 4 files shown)


naamunds:
:shipit:

PR #125, opened 2016-10-04
dotnet-sdk image installs `make`

colemickens:
I have a "development environment" container that only exists for the purpose of installing `make`. It would be helpful (and in my mind, consistent with an "SDK container")  if it contained `make`.

### DIFF
--- a/1.0.0-preview2/debian/Dockerfile
+++ b/1.0.0-preview2/debian/Dockerfile
@@ -9,6 +9,7 @@ ENV LTTNG_UST_REGISTER_TIMEOUT 0
 # Install .NET CLI dependencies
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


MichaelSimons:
@dleeapho, @glennc - I acknowledge `make` is useful in certain scenarios.  There are also **many** other tools that are useful in other various scenarios.  Not all people however will use and even want all or some of these tools.    What is the criteria we should be applying when determining if a given tool should be added to the SDK image?  How "useful" should the tool be?  How many people should want the tool added?  How do we measure all of this?

The current `microsoft/dotnet:1.0.0-preview2-sdk` image is 541.8 MB on disk.  We have been receiving negative feedback on its size already.  `make` would add 1.3 MB

colemickens:
Totally understood, I realize this may not be desired.

PR #128, opened 2016-10-07
Remove Windows Server Core images

richlander:
- Nano Server is a much better pairing for .NET Core.
- Early pulls data substantiates this. Very few pulls for the
  windowsservercore image.
- Windows Server Core and .NET Framework are a much better pairing.
- Developers are free to create their own images for Windows Server Core +
  .NET Core, as needed.
- This change significantly simplifies our offering and guidance on using
  Docker and .NET (both Core and Framework).


MichaelSimons:
Replacing with #129 since it includes the necessary script changes.

PR #134, opened 2016-10-21
Update onbuild to not 'run'

richlander:
/cc @MichaelSimons @kendrahavens

### DIFF
--- a/1.0.0-preview2/debian/onbuild/Dockerfile
+++ b/1.0.0-preview2/debian/onbuild/Dockerfile
@@ -3,7 +3,9 @@ FROM microsoft/dotnet:1.0.0-preview2-sdk
 RUN mkdir -p /dotnetapp
 WORKDIR /dotnetapp
 
-ENTRYPOINT ["dotnet", "run"]
+ENTRYPOINT ["dotnet", "out/dotnetapp"]
 
-ONBUILD COPY . /dotnetapp
+ONBUILD COPY project.json /dotnetapp
 ONBUILD RUN dotnet restore
+ONBUILD COPY . /dotnetapp
+ONBUILD RUN dotnet publish -c Release -o out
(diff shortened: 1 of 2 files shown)


MichaelSimons:
IMO we should not be making this change without revving the docker tag.  This is a behavior change and considered a breaking change.  Since our version number is tied to the CLI, that means we would have to wait until we pickup preview3.  This can be made in dotnet-nightly now however.

1.0.0-preview2/debian/onbuild/Dockerfile
```diff
@@ -3,7 +3,9 @@ FROM microsoft/dotnet:1.0.0-preview2-sdk
 RUN mkdir -p /dotnetapp
```
MichaelSimons: I think now would be a good time to remove this to simplify the Dockerfile.  Additionally all COPY destinations should be changed to `.` or `./`.

1.0.0-preview2/debian/onbuild/Dockerfile
```diff
@@ -3,7 +3,9 @@ FROM microsoft/dotnet:1.0.0-preview2-sdk
...
 WORKDIR /dotnetapp
 
-ENTRYPOINT ["dotnet", "run"]
+ENTRYPOINT ["dotnet", "out/dotnetapp"]
```
MichaelSimons: It looks like this broke CI.  I think it is missing `.dll`.  The same comment applies to the nanoserver version

MichaelSimons: This is not going to work for named projects.  By that I mean apps that specify a name in the `project.json`.  When a name is specified, dotnet creates the resulting assembly with that name.  We would need to retrieve this name dynamically.  I am not aware of any functionality the CLI provides for this so we would likely have to write our own logic.

richlander:
I'll close this PR and make it in dotnet-nightly. Good point on compat.

Ya, you can see I didn't actually test it. I wanted to use it to have the conversation. I'm happy with the answer.

For onbuild, I'm happy to give up on the name property. The general improvement seems to be significant enough to give up on that scenario. Basically, you've grown out of on-build at that point!

Cool?

PR #135, opened 2016-10-22
Update readme to match Docker norm

richlander:
/cc @MichaelSimons @kendrahavens @glennc @shirhatti @billwagner

### DIFF
--- a/1.0/README.md
+++ b/1.0/README.md
@@ -1,25 +1,10 @@
-![](https://avatars0.githubusercontent.com/u/9141961?v=3&amp;s=100)
+# Supported tags and respective `Dockerfile` links
 
-.NET Core Docker Image
-====================
-
-This repository contains `Dockerfile` definitions for [.NET Core](https://github.com/dotnet/core) Docker images. You can find samples, documentation, and getting started instructions for .NET Core in the [.NET Core documentation](https://docs.microsoft.com/dotnet/articles/core/).
-
-[![Downloads from Docker Hub](https://img.shields.io/docker/pulls/microsoft/dotnet.svg)](https://hub.docker.com/r/microsoft/dotnet)
-[![Stars on Docker Hub](https://img.shields.io/docker/stars/microsoft/dotnet.svg)](https://hub.docker.com/r/microsoft/dotnet)
-
-
-## Supported tags
-
-### Development images
 -       [`1.0.0-preview2-sdk`, `latest` (*1.0.0-preview2/debian/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2/debian/Dockerfile)
 -       [`1.0.0-preview2-nanoserver-sdk`, `nanoserver` (*1.0.0-preview2/nanoserver/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2/nanoserver/Dockerfile)
--       [`1.0.0-preview2-onbuild`, `onbuild` (*1.0.0-preview2/debian/onbuild/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2/debian/onbuild/Dockerfile)
--       [`1.0.0-preview2-nanoserver-onbuild`, `nanoserver-onbuild` (*1.0.0-preview2/nanoserver/onbuild/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2/nanoserver/onbuild/Dockerfile)
+/debian/onbuild/Dockerfile)
 -       [`1.0.0-preview2.1-sdk` (*1.0.0-preview2.1/debian/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2.1/debian/Dockerfile)
 -       [`1.0.0-preview2.1-nanoserver-sdk` (*1.0.0-preview2.1/nanoserver/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2.1/nanoserver/Dockerfile)
-
-### Runtime images
 -       [`1.0.1-core`, `1.0-core`, `1-core`, `core` (*1.0/debian/core/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0/debian/core/Dockerfile)
 -       [`1.0.1-nanoserver-core`, `1.0-nanoserver-core`, `1-nanoserver-core`, `nanoserver-core` (*1.0/nanoserver/core/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0/nanoserver/core/Dockerfile)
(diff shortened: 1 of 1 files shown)


1.0/README.md
```diff
@@ -1,54 +1,136 @@
-![](https://avatars0.githubusercontent.com/u/9141961?v=3&amp;s=100)
+# Supported tags and respective `Dockerfile` links
 
-.NET Core Docker Image
-====================
+-       [`1.0.0-preview2-sdk`, `latest` (*1.0.0-preview2/debian/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2/debian/Dockerfile)
+-       [`1.0.0-preview2-nanoserver-sdk`, `nanoserver` (*1.0.0-preview2/nanoserver/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2/nanoserver/Dockerfile)
+-       [`1.0.0-preview2-onbuild`, `onbuild` (*1.0.0-preview2/debian/onbuild/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-preview2/debian/onbuild/Dockerfile)
```
MichaelSimons: Per our decision earlier today, removing onbuild will help simplify this readme.

1.0/README.md
```diff
@@ -1,54 +1,136 @@
...
+ ```
+ The steps above are intended to show the basic functions of .NET Core tools. Try running `dotnet run` twice. You'll see that the second invocation skips compilation. The subsequent command after `dotnet run` demonstrates that you can run an application directly out of the bin folder, without the additional build logic that `dotnet run` adds. The last two commands demonstrate the publishing scenario, which prepares an app to be deployed on the same or other machine, with a requirement on only the .NET Core Runtime, not the larger SDK. Naturally, you don't have to exit immediately, but can continue to try out the product as long as you want.
+
+ On Windows, the experience is very similar. The commands should be the same, with the exception of the first command (specifically the image name), `ls` and the directory separators. Try the following command, to replace the first command above:
+
+ ```console
+  $ docker run -it --rm microsoft/dotnet:nanoserver
+ ```
+
+## Build and run an ASP.NET app within a .NET Core Container
```
MichaelSimons: I like this heading because a lot of people come here with this question.  I think we should work towards having this content in the asp.net readme and then this section will simply forward the readers onto it.

1.0/README.md
```diff
@@ -57,21 +139,98 @@ $ docker build -t my-dotnet-app .
 $ docker run -it --rm --name my-running-app my-dotnet-app
 ```
 
+You can learn more about how to use this image with the - [Development sample](https://github.com/dotnet/dotnet-docker-samples/tree/master/dotnetapp-development).
```
MichaelSimons: It looks like this is an invalid link  I think it is supposed to end with `dev` rather than `development`

1.0/README.md
```diff
@@ -57,21 +139,98 @@ $ docker build -t my-dotnet-app .
...
+
+## More Examples using these Images
+
+You can learn more about using .NET Core with Docker with [.NET Docker samples](https://github.com/dotnet/dotnet-docker-samples):
+
+- [ONBUILD](https://github.com/dotnet/dotnet-docker-samples/tree/master/dotnetapp-onbuild) sample using the `1.0.0-preview2-onbuild` .NET Core SDK image.
+- [Development](https://github.com/dotnet/dotnet-docker-samples/tree/master/dotnetapp-dev) sample using the `1.0.0-preview2-sdk` .NET Core SDK image.
+- [Production](https://github.com/dotnet/dotnet-docker-samples/tree/master/dotnetapp-prod) sample using the `1.0.1-core` .NET Core image.
+- [Self-contained](https://github.com/dotnet/dotnet-docker-samples/tree/master/dotnetapp-selfcontained) sample using the `1.0.1-core-deps` base OS image (with native dependencies added).
```
MichaelSimons: I am a little concerned that the description can easily get out of date since they reference specific versioned tags.  As we update the samples, will these descriptions get updated?  Could we get away with just saying `sdk`, `core` and `core-deps`?

1.0/README.md
```diff
@@ -57,21 +139,98 @@ $ docker build -t my-dotnet-app .
...
+
+Windows Container variants are provided at the same locations, above, and use slightly different image tags (for example, `1.0.0-preview2-nanoserver-onbuild`).
+
+See [Building Docker Images for .NET Core Applications](https://docs.microsoft.com/dotnet/articles/core/docker/building-net-docker-images) to learn more about the various Docker images and when to use each for them.
+
+See the following related repos for other application types:
```
MichaelSimons: Do you think it would help make these pop a little more if this were a `## Related Repos` section.  I feel like it would be beneficial to make them more noticeable.

1.0/README.md
```diff
@@ -1,54 +1,136 @@
...
+In your Dockerfile, include the following line to compile and run your project:
 
-### `microsoft/dotnet:<version>-onbuild`
+```dockerfile
+FROM microsoft/dotnet:1.0.0-preview2-onbuild
+```
+
+This image includes multiple ONBUILD triggers (effectively projected into your Dockerfile) which should cover simple applications (not intended for production and/or larger apps). The build will copy, restore, build and `dotnet run` your application.
+
+For Windows containers, you should instead include the following in your Dockerfile:
```
MichaelSimons: I think the first mention of Window Containers should link to their [documentation](https://msdn.microsoft.com/virtualization/windowscontainers/containers_welcome).  I know it is referenced in the nano server variant section but that seems too late.

richlander:
Made some more changes.

Changed all of the "core" references to "runtime", optimistically. Can we use now as an opportunity to make that change?

1.0/README.md
```diff
@@ -57,21 +139,98 @@ $ docker build -t my-dotnet-app .
...
-2. Build a custom copy of .NET Core by compiling [coreclr](https://github.com/dotnet/coreclr) and [corefx](https://github.com/dotnet/corefx).
+### `microsoft/dotnet:<version>-nanoserver`
 
-## Windows Containers
+There are multiple images for Windows Nanoserver. 
 
-Windows Containers images use the `microsoft/nanoserver` base OS image from Windows Server 2016.  For more information on Windows Containers and a getting started guide, please see: [Windows Containers Documentation](http://aka.ms/windowscontainers).
+Use the following tags for Windows container images:
 
 -       `1.0.0-preview2-nanoserver-sdk`
```
MichaelSimons: Does re-listing the tags here really help?  We don't list them for the other image variants.  This is why the tags section exists.  This list isn't even complete.  I would rather these not be listed from a maintenance perspective unless they are really adding value.

1.0/README.md
```diff
@@ -1,54 +1,149 @@
...
+
+For [Windows Containers](http://aka.ms/windowscontainers), you should instead include the Nanoserver version of the .NET Core SDK image:
+
+```dockerfile
+FROM microsoft/dotnet:nanoserver
+```
+
+Add the following additional lines to your Dockerfile, which will both build and run your application in the container. This Dockerfile has been optimized to take advantage of Docker layering, resulting in faster image building for iterative development. The file can be made shorter but wouldn't be better.
+
+``dockerfile
```
MichaelSimons: Missing ` this results in bad formatting

MichaelSimons:
Merged in - it will make it easier to update with the runtime tag changes.

PR #137, opened 2016-10-24
Added 1.1.0-preview1 core and 1.0.0-preview2-1 sdk Dockerfiles

MichaelSimons:
Because we are dropping support for the onbuild images, I removed the CI to test the image.  This also removed the need for the SDK build hooks in order to get Docker Hub auto-builds to work correctly.  Once the builds are all in place, I will make a follow up PR to update the README.md to include the new tags.

New tags (note the versionless tags `core`, `core-deps`, and `latest` will not be updated because this is not a stable release):
microsoft/dotnet:1.0.0-preview2-1-sdk
microsoft/dotnet:1.0.0-preview2-1-nanoserver-sdk
microsoft/dotnet:1.1.0-preview1-core-deps
microsoft/dotnet:1.1.0-preview1-core
microsoft/dotnet:1.1.0-preview1-nanoserver-core

@richlander, @naamunds


naamunds:
Why do the tags and folder structure use "preview2-1" instead of "preview2.1"?  It seems like "preview2.1" would be more consistent with the .NET branding of the release.

test/run-test.sh
```diff
@@ -46,17 +46,6 @@ for development_image_version in $( find . -path './.*' -prune -o -path '*/debia
...
-    echo "----- Testing ${development_tag_base}-onbuild -----"
-    pushd "${app_dir}" > /dev/null
-    echo "FROM ${development_tag_base}-onbuild" > Dockerfile
-    docker build -t "${app_name}-onbuild" .
-    popd > /dev/null
-    docker run -t ${optional_docker_run_args} --name "onbuild-test-${app_name}" "${app_name}-onbuild"
-
-    if [ -z "${DEBUGTEST}" ]; then
-        docker rmi "${app_name}-onbuild"
-    fi
```
naamunds: Because the onbuild testing is being removed, the logic in create-run-publish-app.sh for restoring project.json to specify framework-dependent deployment can be removed (i.e. lines 15, 23-25).

naamunds:
> The reason I used 2-1 is because that is the release-suffix the cli used - dotnet/cli#4468. When you run `dotnet --version` you see `1.0.0-preview2-1-003155`

I see.  The installer files still use "2.1", and that seems clearer to me when reading the tags, but I also agree that there's value in matching the CLI release-suffix, so I'm fine with whatever you and @richlander prefer.

MichaelSimons:
@richlander - can you provide your opinion on the 2-1 vs 2.1 discussion here?  I don't really have an opinion.  The inconsistencies in the various artifacts are unfortunate.

test/run-test.sh
MichaelSimons: Good point - removed.

MichaelSimons:
Thinking about the naming more.  The branded name should be 2.1 and that is what we should use.  It looks like 2-1 is being used in places to work around nugget package naming constraints.

PR #141, opened 2016-10-26
Fixed broken runtime-deps post_push build hook

MichaelSimons:


MichaelSimons:
Merging to unblock build.

PR #143, opened 2016-10-26
Fix tag in README

naamunds:
The tag `1.0.0-preview2` does not exist.  This should either be `1.0.0-preview2-sdk` or should just refer to the `latest` tag.  I think the latter makes more sense, because that is what is used in other examples in the README, e.g. `FROM microsoft/dotnet:nanoserver` and `docker run -it --rm microsoft/dotnet`

/cc @richlander @MichaelSimons


MichaelSimons:
:shipit:

PR #156, opened 2016-11-15
Add 1.1 images and refactor 1.0 sdk tags w/new runtime focused tagging pattern

MichaelSimons:
Fixes #42

This won't pass CI until 1.1 runtime packages are pushed to NuGet.

@richlander @naamunds @dleeapho

PR #158, opened 2016-11-16
Add 1.1-sdk-msbuild and 1.0-sdk-msbuild tags

MichaelSimons:
In regards to https://github.com/dotnet/dotnet-docker/pull/156#discussion_r88139619, we should still have a 1.X versioned tags.  I think it is reasonable to add them with the *-sdk-msbuild suffix.  Having these tags allow us to run our CI tests against the DH artifacts which is valuable.

@naamunds, @richlander


naamunds:
LGTM

PR #165, opened 2016-11-28
Re-order images as per versions

vivmishra:
Reordering the images in the ReadMe so that they appear ordered as per version first rather than as per product. This makes it easier to spot all the desired images for a specific version.

@richlander / @MichaelSimons  --- plz review

@dotnetbot skip CI please


richlander:
LGTM

PR #177, opened 2017-01-10
Update Dockerfile to preview4?

cdroulers:
Should there be a specific tag to get this preview or to keep preview3?


MichaelSimons:
@cdroulers, thanks for making this PR.  As you referenced, a change like this which picks up a new CLI would warrant a version increment of the Docker tags.  We also would like to keep the Docker images in sync for both LTS (1.0) and current (1.1).  In this case https://github.com/dotnet/dotnet-docker/blob/master/1.0/debian/sdk/msbuild/Dockerfile#L30 should get updated as well.  We are working on an update that is scheduled for next week therefore I am going to close this PR in lieu of that work.

cdroulers:
Perfect, thanks!

PR #181, opened 2017-01-22
Simplify windows dockerfiles

friism:
1. Add `ProgressPreference` - this will greatly speed up the build because of PowerShell bug: https://github.com/PowerShell/PowerShell/issues/2138#issuecomment-251261324
2. Use PowerShell throughout for consistency

### DIFF
--- a/1.0/nanoserver/runtime/Dockerfile
+++ b/1.0/nanoserver/runtime/Dockerfile
@@ -1,13 +1,13 @@
 FROM microsoft/nanoserver:10.0.14393.693
 
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
 # Install .NET Core
 ENV DOTNET_VERSION 1.0.3
 ENV DOTNET_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/$DOTNET_VERSION/dotnet-win-x64.$DOTNET_VERSION.zip
 
-RUN powershell -NoProfile -Command \
-        $ErrorActionPreference = 'Stop'; \
-        Invoke-WebRequest %DOTNET_DOWNLOAD_URL% -OutFile dotnet.zip; \
-        Expand-Archive dotnet.zip -DestinationPath '%ProgramFiles%\dotnet'; \
-        Remove-Item -Force dotnet.zip
+RUN Invoke-WebRequest $Env:DOTNET_DOWNLOAD_URL -OutFile dotnet.zip; \
+    Expand-Archive dotnet.zip -DestinationPath $Env:ProgramFiles\dotnet; \
+    Remove-Item -Force dotnet.zip
 
-RUN setx /M PATH "%PATH%;%ProgramFiles%\dotnet"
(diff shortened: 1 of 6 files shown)


build-and-test.ps1
```diff
@@ -44,7 +44,8 @@ Get-ChildItem -Recurse -Filter Dockerfile |
             $tagSuffix
         $tags.Add($tag) | Out-Null
         Write-Host "--- Building $tag from $($_.DirectoryName) ---"
-        docker build $optionalDockerBuildArgs -t $tag $_.DirectoryName
+        $dockerfilePath = $_.DirectoryName + '\Dockerfile'
```
MichaelSimons: Is this needed?  The nanoserver directories that contain the Dockerfiles already only contain a single Dockerfile.  The directory structure was designed the way it is explicitly for this reason.  It seems like it could get in the way if we ever need addition files in the context.

1.0/nanoserver/runtime/Dockerfile
```diff
@@ -1,13 +1,13 @@
...
-        $ErrorActionPreference = 'Stop'; \
-        Invoke-WebRequest %DOTNET_DOWNLOAD_URL% -OutFile dotnet.zip; \
-        Expand-Archive dotnet.zip -DestinationPath '%ProgramFiles%\dotnet'; \
-        Remove-Item -Force dotnet.zip
+RUN Invoke-WebRequest $Env:DOTNET_DOWNLOAD_URL -OutFile dotnet.zip; \
+    Expand-Archive dotnet.zip -DestinationPath $Env:ProgramFiles\dotnet; \
+    Remove-Item -Force dotnet.zip
 
-RUN setx /M PATH "%PATH%;%ProgramFiles%\dotnet"
+RUN setx /M PATH $($Env:PATH + ';' + $Env:ProgramFiles + '\dotnet');
```
MichaelSimons: What do you think about consolidating this into the previous RUN command?  I would also suggest removing the trailing ';' as it is not needed and is inconsistent with the rest of the run commands within the Dockerfiles (e.g. last cmd doesn't end with ';')

build-and-test.ps1
friism: @MichaelSimons You're right that it's not strictly needed, I just thought it was a nice way to make it explicit that the build cannot depend on local context. I've removed the change.

1.0/nanoserver/runtime/Dockerfile
friism: On Linux, it's customary to split this out, but that's because ENV is metadata and not something that's set with a command. Also note that they were separate before. I can collapse them if you want though.

friism:
@MichaelSimons Thanks for the quick review, comments addressed. Let me know if you want to collapse the path stuff with the setup commands

1.0/nanoserver/runtime/Dockerfile
MichaelSimons: Fair point - leave as is.

friism:
This PR took 21min to build: https://ci.dot.net/job/dotnet_dotnet-docker/job/master/job/NanoServer_prtest/22/

The previous one took 28minutes: https://ci.dot.net/job/dotnet_dotnet-docker/job/master/job/NanoServer_prtest/21/

PR #182, opened 2017-01-22
fix dotnet version

friism:
It's unclear to me whether updating the file in the `1.0` is correct. Lemme know

This uses a version of .NET that doesn't have really annoying Nuget problem that stops MusicStore from building: https://github.com/dotnet/cli/issues/4946#issuecomment-270219101

### DIFF
--- a/1.0/nanoserver/sdk/projectjson/Dockerfile
+++ b/1.0/nanoserver/sdk/projectjson/Dockerfile
@@ -1,7 +1,7 @@
 FROM microsoft/nanoserver:10.0.14393.693
 
 # Install .NET Core SDK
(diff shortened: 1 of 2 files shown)


MichaelSimons:
We have been only updating the bits in the official Dockerfiles on official releases of .NET.  We also rev the full versioned tags to ensure users have the ability to control which version they are on.  Because of that I would recommend against merging this change in at this time.  The https://github.com/dotnet/dotnet-docker-nightly has already picked up this change and will get merged into dotnet-nightly once a CLI release is made official.

Could transition to use the msbuild variants since that is the wave of the future?

friism:
Alright, thanks - I had forgotten about the nightlies, lemme try that.

PR #190, opened 2017-01-30
Remove the Platform param from nano CI

MichaelSimons:
Logic was added to the CI script to self detect the Platform therefore it is no longer being used and can be removed.  Once this PR is merged I will make a follow-up PR to remove the parameter from the build-and-test script.


MichaelSimons:
@dotnet-bot test CI please

PR #196, opened 2017-01-31
Remove shell scripts no longer needed after ci update

MichaelSimons:
Merging in changes from @friism to remove duplicated scripts - hooray!


MichaelSimons:
Dupe of 195

PR #202, opened 2017-02-09
Fix dotnet new -t

kendrahavens:
The `dotnet new` template commands have changed.

### DIFF
--- a/1.0/README.md
+++ b/1.0/README.md
@@ -132,7 +132,7 @@ docker run -it --rm microsoft/dotnet
 [now in the container]
 mkdir app
 cd app
-dotnet new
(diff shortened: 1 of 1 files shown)


kendrahavens:
@MichaelSimons Can you merge this is this looks good? I'll be in a recording the rest of the morning.

MichaelSimons:
@kendrahavens, these same changes should be made in dotnet-docker-nightly as well.

kendrahavens:
@MichaelSimons The dotnet new in the `Interactively build and run a simple .NET Core application` hasn't changed because it does not use any of the -t flags.

PR #209, opened 2017-02-16
Moved readme to root of repo

MichaelSimons:
When we first created the readme, Docker Hub required that it was next to a Dockerfile in order to pull it in correctly. Now they support having the readme at the root of the repo. This change will make the repo more "normal".

FYI Github diff doesn't show this change correctly.


dleeapho:
LGTM. Glad to see this change.

PR #226, opened 2017-04-04
Updates libcurl version for runtime dockerfiles

### LINKED ISSUE #225: Reponse time for HttpClient for POST requests are very slow due to old libcurl.
## Steps to reproduce the issue
1. Test performance of simple .NET Core app in Linux (with libcurl 7.38) https://github.com/damtur/dotnetcore-post-bug
2. Test the same application in Docker image (with libcurl 7.53.1) provided in https://gist.github.com/damtur/3519cb6e74ed570c8b4bbb5baddc86fe#file-dockerfile

## Expected behavior

Response time in Linux should be similar to windows and be around 1 ms per request.

## Actual behavior

For libcurl 7.38 I'm seeing response time ~40ms
For libcurl 7.53.1 I'm seeing response time ~1ms

## Additional information (e.g. issue happens only occasionally)

Plese see a blog post for more information: https://www.linkedin.com/pulse/post-net-core-damian-turczy%C5%84ski

## Output of `docker version`

```
Client:                                    
 Version:      17.03.1-ce                  
 API version:  1.27                        
 Go version:   go1.7.5                     
 Git commit:   c6d412e                     
 Built:        Tue Mar 28 00:40:02 2017    
 OS/Arch:      windows/amd64               
                                           
Server:                                    
 Version:      17.03.1-ce                  
 API version:  1.27 (minimum version 1.12) 
 Go version:   go1.7.5                     
 Git commit:   c6d412e                     
 Built:        Fri Mar 24 00:00:50 2017    
 OS/Arch:      linux/amd64                 
 Experimental: true                        
```

## Output of `docker info`

```
Containers: 12
 Running: 0
 Paused: 0
 Stopped: 12
Images: 51
Server Version: 17.03.1-ce
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 75
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host ipvlan macvlan null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 4ab9917febca54791c5f071a9d1f404867857fcc
runc version: 54296cf40ad8143b62dbcaa1d90e520a2136ddfe
init version: N/A (expected: 949e6facb77383876aeff8a6944dde66b3089574)
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.13-moby
Operating System: Alpine Linux v3.5
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 1.934 GiB
Name: moby
ID: 3K2R:A3HU:YMAH:OJ5V:7UKT:DLXD:7QYD:Z5U7:W5ZM:LPET:GJV4:M7BO
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): true
 File Descriptors: 13
 Goroutines: 22
 System Time: 2017-04-04T11:23:49.6104466Z
 EventsListeners: 0
Registry: https://index.docker.io/v1/
Experimental: true
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false
```

damtur:
This fixes a performance bug with POST requests.

Fix https://github.com/dotnet/dotnet-docker/issues/225

### DIFF
--- a/1.0/debian/runtime-deps/Dockerfile
+++ b/1.0/debian/runtime-deps/Dockerfile
@@ -17,3 +17,21 @@ RUN apt-get update \
         libuuid1 \
         zlib1g \
     && rm -rf /var/lib/apt/lists/*
+
+# Update libcurl to version 7.53.1
+RUN echo deb-src ftp://ftp.de.debian.org/debian/ stable main contrib >> /etc/apt/sources.list && \
+    apt-get update && \
+    apt-get -y build-dep curl && \
+    apt-get -y install wget
+
+RUN mkdir ~/curl && \ 
+    cd ~/curl && \ 
+    wget http://curl.haxx.se/download/curl-7.53.1.tar.bz2 && \ 
+    tar -xvjf curl-7.53.1.tar.bz2 && \ 
+    cd curl-7.53.1 && \ 
+    ./configure && \ 
+    make && \ 
+    make install && \ 
+    ldconfig && \
+    cd .. && \
+    rm -rf curl
--- a/1.1/debian/runtime-deps/Dockerfile
+++ b/1.1/debian/runtime-deps/Dockerfile
@@ -17,3 +17,21 @@ RUN apt-get update \
         libuuid1 \
         zlib1g \
     && rm -rf /var/lib/apt/lists/*
+
+# Update libcurl to version 7.53.1
+RUN echo deb-src ftp://ftp.de.debian.org/debian/ stable main contrib >> /etc/apt/sources.list && \
+    apt-get update && \
+    apt-get -y build-dep curl && \
+    apt-get -y install wget
+
+RUN mkdir ~/curl && \ 
+    cd ~/curl && \ 
+    wget http://curl.haxx.se/download/curl-7.53.1.tar.bz2 && \ 
+    tar -xvjf curl-7.53.1.tar.bz2 && \ 
+    cd curl-7.53.1 && \ 
+    ./configure && \ 
+    make && \ 
+    make install && \ 
+    ldconfig && \
+    cd .. && \
+    rm -rf curl


1.0/debian/runtime-deps/Dockerfile
```diff
@@ -17,3 +17,21 @@ RUN apt-get update \
         libuuid1 \
         zlib1g \
     && rm -rf /var/lib/apt/lists/*
+
+# Update libcurl to version 7.53.1
+RUN echo deb-src ftp://ftp.de.debian.org/debian/ stable main contrib >> /etc/apt/sources.list && \
+    apt-get update && \
+    apt-get -y build-dep curl && \
+    apt-get -y install wget
```
brantburnett: You probably want  to append "&& rm -rf /var/lib/apt/lists/*" here, like  the above command, to cleanup extra files so they aren't included in the layer of the image.

MichaelSimons:
@damtur, Thanks for making this PR and bringing attention to the performance issue.  As discussed in https://github.com/dotnet/corefx/issues/13551, the underlying .NET Core Fx issue has been [fixed](https://github.com/dotnet/corefx/pull/15354).  I think it would be preferable to wait for the product fix to be released.  What do you think about documenting this workaround in #225 so that people who are affected by this issue can add the workaround within their derived Dockerfiles until the product fix is released?  We could also make a few changes to the workaround to reduce the size.  This change increases the runtime-deps image from 166 MB to 485 MB.

damtur:
@MichaelSimons Yeah, I agree that increasing the image size is not great and installing developer dependency on production image is just super-bad. I was not happy with the solution either, but an alternative would be just to put a binary of libcurl into that image which might be even uglier... 

I've added a workaround documentation to #225 .

I still think that for .NET Core < 2.0 images we should do something to mitigate that problem :(

kendrahavens:
@damtur Thank you so much for adding the documentation and finding this in general! We hear you.

@MichaelSimons Yeah, I'm thinking publishing <2.0 images with new patch tags would be a fair amount of additional work with not as great a pay-off considering the larger image size. We'd also need to document that there is a trade off so I'm guessing we wouldn't want to roll this directly into the latest image. I'll keep an eye on the comments/issues on this though.

MichaelSimons:
@damtur, @kendrahavens - This CoreFx change was back ported to 1.1  -https://github.com/dotnet/corefx/pull/16895.  It will be in the next release and therefore will be picked up by the 1.1 Dockerfile.  It don't see the fix in 1.0.  Since the product did not pack port the fix, I do not think we should be adding a workaround for 1.0.

Because of the reasons discussed earlier, I'm closing this PR.  Again thanks @damtur!

PR #238, opened 2017-05-10
Add 2.0.0-preview1 images

MichaelSimons:


MichaelSimons:
FYI - The build definitions have also been updated accordingly as well.

2.0/debian/runtime-deps/hooks/post_push
```diff
@@ -0,0 +1,26 @@
...
+set -e
+
+echo "Pushing additional runtime-deps tags"
+imageName=$DOCKER_REPO":2.0.0-preview1-runtime-deps"
+docker tag $IMAGE_NAME $imageName
+docker push $imageName
+
+imageName=$DOCKER_REPO":2-runtime-deps"
+docker tag $IMAGE_NAME $imageName
+docker push $imageName
```
dagood: Just to confirm the pattern: there are no `:runtime-deps` or `:runtime` tags in this file because we want to leave the latest official on the old non-preview release, right?

MichaelSimons:
cc @natemcmaster

2.0/debian/runtime/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM microsoft/dotnet-nightly:2.0-runtime-deps
```
dagood: Should this be `dotnet` rather than `dotnet-nightly`? Sorry if I'm forgetting something about the build dependencies here.

MichaelSimons: Good catch - fixed

2.0/debian/runtime-deps/hooks/post_push
MichaelSimons: correct

Issue #239, opened 2017-05-16
.NET Core Docker images will move to multi-arch based tags


kendrahavens:
## Summary
Docker has a [multi-arch feature]( https://github.com/moby/moby/issues/15866) that [microsoft/dotnet-nightly]( https://hub.docker.com/r/microsoft/dotnet-nightly/) recently started utilizing.  The plan is to port this to the [official microsoft/dotnet repo]( https://hub.docker.com/r/microsoft/dotnet/) shortly.  The multi-arch feature allows a single tag to be used across multiple machine configurations.  Without this feature each architecture/OS/platform requires a unique tag.  For example, the microsoft/dotnet:1.0-runtime tag is based on Debian and microsoft/dotnet:1.0-runtime-nanoserver if based on Nano Server.  With multi-arch there will be one common microsoft/dotnet:1.0-runtime tag.  If you pull that tag from a Linux container environment you will get the Debian based image whereas if you pull that tag from a Windows container environment you will get the Nano Server based image.  This helps provide tag uniformity across Docker environments thus eliminating confusion.

Current microsoft/dotnet tags:
- [`1.0.5-runtime`, `1.0-runtime` (*1.0/debian/runtime/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0/debian/runtime/Dockerfile)
- [`1.0.5-runtime-nanoserver`, `1.0-runtime-nanoserver` (*1.0/nanoserver/runtime/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0/nanoserver/runtime/Dockerfile)
- [`1.0.5-runtime-deps`, `1.0-runtime-deps` (*1.0/debian/runtime-deps/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0/debian/runtime-deps/Dockerfile)
- [`1.0.5-sdk`, `1.0-sdk` (*1.0/debian/sdk/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0/debian/sdk/Dockerfile)
- [`1.0.5-sdk-nanoserver`, `1.0-sdk-nanoserver` (*1.0/nanoserver/sdk/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/1.0/nanoserver/sdk/Dockerfile)
[...]

PR #240, opened 2017-05-16
Add support for multi-arch tags

### LINKED ISSUE #99: Add manifest to better support Windows and Linux based platforms
With the current state of Docker and Windows Containers, a Windows based image cannot be run on Docker for Linux/Mac/Windows.  They can only be run from a Windows Containers environment.  Likewise a Linux based image cannot be run from a Windows Containers environment. The error experience when a user tries to pull/run an image in an unsupported environment is less than ideal.  A [manifest](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md) should be created which will help solve this problem.  A manifest will provide two benefits.  First, it will prevent an image from being pulled in an unsupported environment.  When a pull attempt is made from an unsupported environment, the error experience is meaningful and obvious as to why the operation cannot be performed.  Second, a manifest will allow a single tag (e.g. sdk, core, onbuild) to be pulled from multiple platforms.  When the tag is pulled, docker will inspect the manifest to figure out which image should be used for the current platform.

This [blog post](https://blog.docker.com/2016/04/docker-windows-server-tp5/) explains manifests and how they are useful.

This work is currently blocked on having docker tooling support for manifests.  For instance the microsoft/dotnet images are built with the Docker auto-build infrastructure within Docker Hub but it is not capable of producing a manifest.

MichaelSimons:
fixes #99 
related to #239


MichaelSimons:
@dotnet-bot test nanoserver please

PR #273, opened 2017-08-10
Update README

richlander:
I made some small formatting and content fixes to make the README better.

> This image is officially supported on Docker version 1.12.2.

Is this still the right version?


richlander:
Do we do anything in particular to support our images on 1.12 on a "best effort basis"?

PR #299, opened 2017-09-21
Update ReadMe to use multi-stage build sample

kendrahavens:
@MichaelSimons I addressed all of your comments from the stale PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,20 +57,34 @@ You can run a [sample application](https://hub.docker.com/r/microsoft/dotnet-sam
 docker run microsoft/dotnet-samples
 ```
 
-### Run a .NET Core application with the .NET Core Runtime image
+### Use multi-stage build with a .NET Core application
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,20 +57,34 @@ You can run a [sample application](https://hub.docker.com/r/microsoft/dotnet-sam
...
 FROM microsoft/dotnet:runtime
-WORKDIR /dotnetapp
-COPY out .
+WORKDIR /app
+COPY --from=build-env /app/out ./
 ENTRYPOINT ["dotnet", "dotnetapp.dll"]
 ```
 
-Build your application with the `dotnet` tools using the following commands:
+Change `dotnetapp.dll` to the DLL of your project. Build your application with the `dotnet` tools using the following commands:
```
MichaelSimons: With the use of the multi-stage build, I don't think there is any need for the steps to build the application on the host - this is accomplished by the above Dockerfile.  I would remove lines 87 - 94.  I think the note on line 101 is sufficient for renaming the application.

kendrahavens: Sounds good!

PR #303, opened 2017-09-29
Merge latest build definitions and tests from nightly repo

MichaelSimons:


test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -0,0 +1,102 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
```
ravimeda: This may not be needed since we are no longer read-write to Dockerfile to replace base image name.

MichaelSimons:
@dotnet-bot test ci please

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
MichaelSimons: Since this is not functional I will make this change in nightly and merge it next time.  I would like this to be a simple merge.

Issue #22, opened 2016-02-12
Create Alpine-based dotnet images


dleeapho:
Alpine is a small footprint distro that is gaining wide adoption in Official images. Prereq is to get Alpine support for the dotnet runtime, libraries, and CLI.

ctaggart:
This is one of the reasons I like golang, is that they have a [golang:alpine docker image](https://hub.docker.com/_/golang/) that is very small. There is now even a [rust-musl-builder](https://github.com/emk/rust-musl-builder) docker image. It would be nice if `dotnet --native` supported targeting `musl` too, so the binaries could be put on alpine images.

markrendle:
I know one of the things that's been blocking this is the absence of libttng on Alpine, but guess what? [It's available now!](http://lttng.org/download/#alpine)

RTodorov:
Hello guys!

Is this still on the roadmap? It's a bit hard to understand the current status of https://github.com/dotnet/coreclr/issues/917.

Thanks!

MichaelSimons:
@RTodorov, Yes Alpine support is being actively worked on although there is no set release date as of now.

MaximRouiller:
Just pinging the old thread to see if we could have a good ol' status update. :smile:

RTodorov:
Looking very much forward to it! It would be a huge improvement for those running dotnet core on Linux.

MichaelSimons:
While progress is being made on Alpine, it is not yet an [officially supported OS](https://github.com/dotnet/core/blob/master/roadmap.md#supported-os-versions).

mrnickel:
I'd like to add my voice to the chorus of people requesting alpine support for .NET.

Loving this support team! Keep up the good work!

skbergam:
Yes this would be a great addition.

ErikSchierboom:
@MichaelSimons I see that you've removed the "blocked dependency" label 14 days, does that mean there is no longer any blocking issue to prevent this from happening? Having a .NET Core Alpine image would be brilliant.

MichaelSimons:
@ErikSchierboom, I'm sorry but this work is still blocked on official support for Alpine.  The blocked dependency label was removed as dependencies are being tracked differently (on a KanBan board).

ErikSchierboom:
@MichaelSimons Ah, okay. Is there anywhere that I can track progress being done on that part?

richlander:
I was talking to @Petermarcu about Alpine support today. My understanding of the status:

[...]

PR #309, opened 2017-10-06
[WIP] add runtime-deps and runtime based on minideb

tuananh:
This PR will add a version of runtime and runtime-deps that are based on [bitnami/minideb](https://github.com/bitnami/minideb).

According to my tests, the runtime image size can reduce from ~ 233 MB down to ~ 172 MB (dotnet 2.0.0)

### DIFF
--- a/1.0/runtime-deps/minideb-jessie/Dockerfile
+++ b/1.0/runtime-deps/minideb-jessie/Dockerfile
@@ -1,0 +1,19 @@
+FROM bitnami/minideb:jessie
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        \
+# .NET Core dependencies
+        libc6 \
+        libcurl3 \
+        libgcc1 \
+        libgssapi-krb5-2 \
+        libicu52 \
+        liblttng-ust0 \
+        libssl1.0.0 \
+        libstdc++6 \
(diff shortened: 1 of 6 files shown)


tuananh:
one question: i want to add more images, how can i test this with CI?

MichaelSimons:
@tuananh - Thanks for the PR.  I understand the desire for slimmer runtime images.  Do you feel that an Alpine based image would be preferable over a bitnami/minideb based image?  .NET Core Alpine support is getting close and we are planning on [adding runtime images](https://github.com/dotnet/dotnet-docker/issues/22) for it.  I think it is preferable to limit the number of distro options provided (where it makes sense) as we already get feedback on the difficulty of deciphering the tags provided.

I am not familiar with bitnami/minideb, what is the benefit of it over debian:stretch-slim?  It is concerning that bitnami/minideb is not an official Docker image.

I would recommend that you open an issue for this.  I would like to get feedback on providing a slim debian variant in addition to Alpine.  I would also like to hear what is the preferred slim debian version.

@richlander

tuananh:
Thanks.

[minideb](https://github.com/bitnami/minideb) is a bit smaller than debian slim. Personally i would prefer debian over alpine but i think if possible, we can provide them as an option for user.

I'm going to close this PR and open an issue instead.

Issue #328, opened 2017-11-02
Stop producing <NanoServerVersion> suffixed tags


MichaelSimons:
Currently < NanoServerVersion > suffixed tags are built for every image variant - e.g. https://github.com/dotnet/dotnet-docker/blob/master/manifest.json#L76.  These tags are not documented.  These tags are not consistent with the Linux images as well as the other Windows images produced in the official Docker repos.  The tags should be removed instead of updating them during the next patch release.

MichaelSimons:
As part of this work, the FROM instructions should be updated to use the latest RS1/RS3 Nano Server tags - e.g. `microsoft/nanoserver:lstc2016` / `microsoft/nanoserver:1709`

cc @natemcmaster

PR #334, opened 2017-11-09
[Port] Stop producing NanoServerVersion suffixed tags

ravimeda:
Porting https://github.com/dotnet/dotnet-docker-nightly/pull/485

Generating images with nanoServerVersion suffixed tag is not consistent with (https://hub.docker.com/r/microsoft/nanoserver/). Produce only two variants - Semi Annual Channel (SAC) and 1709.

Refer dotnet/dotnet-docker#328


1.0/sdk/nanoserver/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # escape=`
 
-FROM microsoft/nanoserver:10.0.14393.1770
```
MichaelSimons: This file was somehow left behind as part of https://github.com/dotnet/dotnet-docker/pull/319.  It is not getting built.  Can you please delete it.

ravimeda: Fixed.

MichaelSimons:
Regarding the PR description.  The following is not a correct statement:

`Generating images with nanoServerVersion suffixed tag is not consistent with (https://hub.docker.com/r/microsoft/nanoserver/).`

The OS repos (including nanoserver) produce fully versioned tags along with stable tags.  The full patch versioned tags are not consistent with the linux based tags produced for microsoft/dotnet.  Going forward microsoft/dotnet will only include tags that include the stable OS suffix (e.g. stretch, nanoserver-1709, etc)

PR #362, opened 2018-02-12
Add .NET Core Samples

richlander:
- Moved samples from https://github.com/dotnet/dotnet-docker-samples
- Took a different approach from the original samples (which had many top-level directories). The new approach has many Dockerfiles in the same directory. I also took more opportunities to make the examples more OS-agnostic.
- I factored out instructions that have use for multiple scenarios.


samples/aspnetapp/README.md
```diff
@@ -0,0 +1,161 @@
...
+docker run --rm -p 8000:80 aspnetapp
+```
+
+After the application starts, visit `http://localhost:8000` in your web browser.
+
+Note: The `-p` argument maps port 8000 on your local machine to port 80 in the container (the form of the port mapping is `host:container`). See the [Docker run reference](https://docs.docker.com/engine/reference/commandline/run/) for more information on commandline parameters.
+
+## Build and run the sample for Linux ARM32 with Docker
+
+You need to build the [sample](Dockerfile.arm32) on Windows. This requirement is due to the .NET Core SDK not being currently supported on ARM32. The instructions assume that you are in the root of the repository.
```
MichaelSimons: MacOS would also work too?

samples/aspnetapp/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+* [microsoft/aspnetcore-build:2.0](https://hub.docker.com/r/microsoft/aspnetcore-build)
+* [microsoft/aspnetcore:2.0](https://hub.docker.com/r/microsoft/aspnetcore/)
+
+## Related Resources
+
+* [ASP.NET Core Getting Started Tutorials](https://www.asp.net/get-started)
+* [.NET Core Production Docker sample](../dotnetapp-prod/README.md)
+* [.NET Core Docker samples](../README.md)
+* [.NET Framework Docker samples](https://github.com/Microsoft/dotnet-framework-docker-samples)
```
MichaelSimons: Would it make sense to call out this includes ASP.NET samples?  In this context it seems useful.

samples/dotnetapp/Dockerfile.arm32
```diff
@@ -0,0 +1,33 @@
+FROM microsoft/dotnet:sdk AS build
```
MichaelSimons: It may be a good idea to include linux in the filename since the Dockerfile is tied to linux.  Trying to build it in a Windows container environment has a little strange behavior as the first stage will succeed.

samples/dotnetapp/README.md
```diff
@@ -0,0 +1,111 @@
...
+You can also [download the repository as a zip](https://github.com/dotnet/dotnet-docker/archive/master.zip).
+
+## Build and run the sample with Docker
+
+You can build and run the sample in Docker using the following commands. The instructions assume that you are in the root of the repository.
+
+```console
+cd samples
+cd dotnetapp
+docker build -t dotnetapp .
```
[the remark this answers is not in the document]
JRAlexander: Me too.

samples/dotnetapp/dotnet-docker-arm32.md
```diff
@@ -0,0 +1,58 @@
+# Build .NET Core Applications for Raspberry Pi with Docker
+
+You can build and run .NET Core apps with [Docker for Raspberry Pi and ARM32 devices](https://docs.docker.com/install/linux/docker-ce/debian), generally. These instructions are based on the [dotnetapp](README.md) sample.
+
+> Note: that Docker refers to ARM32 as `armhf` in documentation and other places.
+
+## Building the Sample with Docker
+
+You need to build the [sample](Dockerfile.arm32) on a 64-bit operating system, either Windows (in Linux container mode), macOS or Linux. This requirement is due to the .NET Core SDK not being currently supported on ARM32. The instructions assume that you are in the root of the repository.
```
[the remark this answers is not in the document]
JRAlexander: @MichaelSimons did you mean to say "If you are **not** on an arm32 host, the build stage will fail."?

samples/aspnetapp/README.md
JRAlexander: Do you mean the .NET Framework Samples?

MichaelSimons: Yes - the .NET Framework Docker samples include full fx ASP.NET samples.

samples/dotnetapp/dotnet-docker-arm32.md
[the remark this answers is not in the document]
MichaelSimons: My conjecture is this won't work on a Linux machines.  The `build` stage of the Dockerfile requires an amd64 linux host in order to build.  The `runtime` stage of the Dockerfile requires an amd32 linux host in order to build.  You can build on Windows and macOS because of how Docker utilizes virtualization.  In order to support this scenario on Linux, the SDK would need to support AMD32 which is not currently supported.

samples/dotnetapp/dotnetapp.sln
```diff
@@ -0,0 +1,71 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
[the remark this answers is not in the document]
richlander: I guess it is a question of what the SLN is should represent ... should it represent all of the assets (like you are suggesting) or be a more typical SLN with just typical assets (not duplicate Dockerfiles). Certainly, it should contain at least _a_ Dockerfile. Thoughts?

samples/aspnetapp/README.md
richlander: Is it just Linux that won't work?

richlander: Tested it everywhere. It works. I somehow got the idea it wouldn't. Updated text.

samples/dotnetapp/dotnet-docker-arm32.md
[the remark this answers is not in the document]
richlander: I tried this on Ubuntu 16.04 w/o issue.

samples/dotnetapp/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# copy csproj and restore as distinct layers
+COPY *.sln .
+COPY dotnetapp/*.csproj ./dotnetapp/
+COPY utils/*.csproj ./utils/
+COPY tests/*.csproj ./tests/
+RUN dotnet restore
+
+# copy and build everything else
+COPY dotnetapp/. ./dotnetapp/
```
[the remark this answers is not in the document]
richlander: Huh. I thought I tried that before and it didn't work. It worked now. Will fix that.

samples/dotnetapp/Dockerfile.selfcontained-linux-arm32
```diff
@@ -0,0 +1,21 @@
+FROM microsoft/dotnet:sdk AS build
+WORKDIR /app
+
+# copy csproj and restore as distinct layers
+COPY *.sln .
+COPY dotnetapp/*.csproj ./dotnetapp/
+COPY utils/*.csproj ./utils/
+COPY tests/*.csproj ./tests/
+RUN dotnet restore -s https://api.nuget.org/v3/index.json -s https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
```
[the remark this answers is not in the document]
richlander: Cause the linker is still only published to myget.

samples/aspnetapp/aspnetapp.sln
```diff
@@ -0,0 +1,34 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26124.0
+MinimumVisualStudioVersion = 15.0.26124.0
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "aspnetapp", "aspnetapp\aspnetapp.csproj", "{5FDCC1ED-9F59-47ED-969D-5E463CDD8D52}"
+EndProject
```
MichaelSimons: Should include .dockerignore and Dockerfile.

samples/aspnetapp/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM microsoft/dotnet:2.0-sdk AS build
```
MichaelSimons: Why not base the sample on the aspnetcore images?

samples/aspnetapp/README.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+Note: The `-c release` argument builds the application in release mode (the default is debug mode). See the [dotnet run reference](https://docs.microsoft.com/dotnet/core/tools/dotnet-run) for more information on commandline parameters.
+
+## Related Resources
+
+* [ASP.NET Core Getting Started Tutorials](https://www.asp.net/get-started)
+* [.NET Core Production Docker sample](../dotnetapp-prod/README.md)
+* [.NET Core Docker samples](../README.md)
+* [.NET Framework and and ASP.NET Docker samples](https://github.com/Microsoft/dotnet-framework-docker-samples)
```
MichaelSimons: Perhaps it would be more appropriate to use `and and and` here instead of just `and and` 😸

samples/dotnetapp/dotnetapp.sln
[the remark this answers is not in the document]
MichaelSimons: My viewpoint was not necessarily that a sln would ever have this many Dockerfiles. I do see customers as possibly having more than one in which case all Dockerfiles would be included. In this particular case, I don't want customers to open the sln and overlook the permutations of the Dockerfiles included in the sample.

MichaelSimons:
This PR fixes the following issues:
https://github.com/dotnet/dotnet-docker-samples/issues/99
https://github.com/dotnet/dotnet-docker-samples/issues/105
https://github.com/dotnet/dotnet-docker-samples/issues/108

samples/aspnetapp/Dockerfile
richlander: cause the final result is nearly 100MB diff, due to this line: https://github.com/aspnet/aspnet-docker/blob/master/2.0/stretch/runtime/Dockerfile#L8. This will only get worse for 2.0. For 2.1, it will be different.

samples/dotnetapp/dotnet-docker-selfcontained.md
```diff
@@ -0,0 +1,78 @@
...
+You first need to add a reference to the [linker package](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Illink.Tasks), using the following instructions. The instructions assume that you are in the root of the repository.
+
+```console
+cd samples
+cd dotnetapp
+cd dotnetapp
+dotnet add package ILLink.Tasks -v 0.1.4-preview-981901 -s https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
+```
+
+You can also update the Dockerfile to add the package reference, although that's not recommended as a long-term option (invalidates Docker cache with every build).
```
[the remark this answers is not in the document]
richlander: Oops. You are right. It wouldn't. I was thinking about that backwards.

samples/dotnetapp/push-image-to-dockerhub.md
```diff
@@ -0,0 +1,76 @@
...
+docker build -t dotnetapp .
+```
+
+You can test the image with the following instructions.
+
+```console
+docker run dotnetapp
+```
+
+## Push the Image
```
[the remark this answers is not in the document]
richlander: I don't grasp this. You'll likely have to explain this to me directly.

samples/dotnetapp/Dockerfile.selfcontained-linux-arm32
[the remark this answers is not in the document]
richlander: Made some more changes around that.

samples/dotnetapp/dotnet-docker-selfcontained.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Add a Reference to the IL Linker
+
+You first need to add a reference to the [linker package](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Illink.Tasks), using the following instructions. The instructions assume that you are in the root of the repository.
+
+```console
+cd samples
+cd dotnetapp
+cd dotnetapp
+dotnet add package ILLink.Tasks -v 0.1.4-preview-981901 -s https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
```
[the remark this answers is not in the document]
richlander: fixed.

richlander:
I think that I've resolved all of your feedback now. Thanks for all of it! I also made several other changes that you'll notice.

richlander:
I just added a README.DockerHub.md. I updated the tags to demonstrate what I think we need. I know that your manifest tool will generate the same content. This file is intended to be a spec and a placeholder.

samples/aspnetapp/.dockerignore
```diff
@@ -0,0 +1,9 @@
+# directories
+bin/
+obj/
+out/
```
MichaelSimons: I question if the two .dockerignore files have the correct paths specified for bin, obj, and out.  I think they should all be prefixed with `**/`.  Right now it you build the app locally, the bin and obj are getting copied into the image from the host because the app is in a subfolder of where the .dockerignore files reside.

samples/aspnetapp/Dockerfile
natemcmaster: Really, 100MB? I'm surprised. I would have thought it would be about equal because you end up publishing those assemblies in your local publish directory when setting /p:PublishWithAspNetCoreTargetManifest="false". 

Also, you lose out on crossgen and fast startup by not using the runtime store.

PR #376, opened 2018-02-15
Add Alpine SDK image

MichaelSimons:
_Copied from https://github.com/dotnet/dotnet-docker-nightly/pull/508 by @richlander_

This PR isn't quite ready. Work items:

* [x] Update URL format (to match other images)
* [x] Add hash/checksum matching (to match other images)
~* [ ] Consider globalization invariant mode support for the SDK.~
* [x] Update tests appropriately to verify Alpine SDK image
* [x] Update manifest.json
* [x] Update readme

Sample I used for testing: https://github.com/dotnet/dotnet-docker-samples/tree/alpine/dotnetapp-prod

I got the image to work as expected but needed to remove invariant-mode support for the SDK. it seems like the SDK fails to run with invariant mode.

This is the error that I saw before removing globalization invariant mode / adding ICU:

```console
Step 7/11 : RUN dotnet publish -c Release -o out
 ---> Running in 0f3ff19e0e0f
Microsoft (R) Build Engine version 15.6.22.57775 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 46.27 ms for /app/dotnetapp.csproj.
/usr/share/dotnet/sdk/2.2.0-preview1-007846/Microsoft.Common.CurrentVersion.targets(2051,5): error MSB3095: Invalid argument. SafeHandle cannot be null. [/app/dotnetapp.csproj]
/usr/share/dotnet/sdk/2.2.0-preview1-007846/Microsoft.Common.CurrentVersion.targets(2051,5): error MSB3095: Parameter name: pHandle [/app/dotnetapp.csproj]
The command '/bin/sh -c dotnet publish -c Release -o out' returned a non-zero code: 1
```

### DIFF
--- a/2.1/sdk/alpine/amd64/Dockerfile
+++ b/2.1/sdk/alpine/amd64/Dockerfile
@@ -1,12 +1,19 @@
 FROM microsoft/dotnet-nightly:2.1-runtime-deps-alpine
 
+# Disable the invariant mode (set in base image)
+RUN apk add --no-cache icu-libs
+
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT false
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+
 # Install .NET Core SDK
-ENV DOTNET_SDK_VERSION 2.1.300-preview1-008123
+ENV DOTNET_SDK_VERSION 2.1.300-preview1-008162
 
 RUN apk add --no-cache --virtual .build-deps \
         openssl \
     && wget -O dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DOTNET_SDK_VERSION/dotnet-sdk-$DOTNET_SDK_VERSION-alpine.3.6-x64.tar.gz \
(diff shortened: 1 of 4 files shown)


MichaelSimons:
Blocked on https://github.com/dotnet/cli/issues/8290

richlander:
I am having second thoughts on https://github.com/dotnet/cli/issues/8290. The SDK should not violate principle of least surprise. I think not enabling globalization would do that. Now, having the runtime and SDK not be symmetric would be very odd, but given the goals of Alpine, I think that's OK. Thoughts?

MichaelSimons:
I'm torn on this mainly because of the asymmetry.  I am concerned this will cause some issues at deployment.  We can try to call this out in documentation but I am sure it will get overlooked if read at all.

richlander:
So, how about we publish the sdk image as-is (with globalization present) since we have the opportunity to actually publish something now. We can work through the rest as we go. It's only preview1, so I think we have enough time to make changes some of which can still be breaking. Yes?

MichaelSimons:
I am good with that plan preview1 plan.

PR #382, opened 2018-02-20
Refactor pattern used to initial NuGet package cache

### LINKED ISSUE #84: Reevalute workaround for non-root user issue with NuGetScratch
As a workaround for https://github.com/dotnet/dotnet-docker/issues/78, a fix was submitted which removes the /tmp/NuGetScratch directory.  The root cause of this issue is https://github.com/NuGet/Home/issues/2793, so when that issue is fixed, the workaround here should be reevaluated.

Depending on the fix made in NuGet, we may want to remove the workaround in this repo.  Alternatively, we may want to keep this behavior and also remove the NuGetScratch directory in the Windows images to reduce the size of the Docker images, similar to the removal of the apt-get caches.

### LINKED ISSUE #381: Consider using a simpler pattern to initialize the NuGet package cache
Currently the SDK Dockerfiles use the following pattern to initialize the package cache.  

```
# Trigger the population of the local package cache
RUN mkdir warmup \
    && cd warmup \
    && dotnet new \
    && cd .. \
    && rm -rf warmup \
    && rm -rf /tmp/NuGetScratch
```

This pattern was necessary at the time it was written and can now be simplified.  Invoking any dotnet cmd will trigger the first run experience.  Additionally with the NuGetFallback folder, CLI no longer utilizes NuGet to initialize the package cache therefore the NuGetScratch directory is no longer created.

```
RUN dotnet help
```

MichaelSimons:
Fixes #381
Fixes #84

### DIFF
--- a/2.1/sdk/jessie/amd64/Dockerfile
+++ b/2.1/sdk/jessie/amd64/Dockerfile
@@ -27,11 +27,7 @@ RUN curl -SL --output dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotn
     && rm dotnet.tar.gz \
     && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
 
-# Trigger the population of the local package cache
 ENV NUGET_XMLDOC_MODE skip
-RUN mkdir warmup \
-    && cd warmup \
-    && dotnet new \
-    && cd .. \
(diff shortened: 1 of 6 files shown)


ravimeda:
2f03e95 LGTM.

PR #385, opened 2018-02-26
2.1 images

MichaelSimons:
- Switch to `stretch-slim` instead of `stretch` for all runtime images. Keep SDK as-is (no buildpack-deps for slim images)
- Keep using Stretch for MA tags like `2-sdk`, `2.1-runtime` and `latest`
- Drop Jessie SDK and Runtime (X64)
- Add Ubuntu:18.04 (bionic) for SDK and Runtime (X64 and ARM)


MichaelSimons:
cc @natemcmaster

MichaelSimons:
@dotnet-bot test ci please

PR #433, opened 2018-03-19
Combine ENV instructions to reduce number of layers

markrendle:
The SDK Dockerfiles use multiple ENV instructions (especially the Alpine images) which cause multiple layers to be created. [Docker's guidance](https://docs.docker.com/engine/reference/builder/#env) recommends using the multiple-variable form of the ENV instruction to reduce the number of intermediate layers.

I've changed the SDK Dockerfiles for 2.1 to use the recommended form.

I've only done this in the 2.1 preview images to start with, if it's wanted I can change the previous versions too.

### DIFF
--- a/2.1/sdk/alpine/amd64/Dockerfile
+++ b/2.1/sdk/alpine/amd64/Dockerfile
@@ -3,12 +3,13 @@ FROM microsoft/dotnet:2.1-runtime-deps-alpine
 # Disable the invariant mode (set in base image)
 RUN apk add --no-cache icu-libs
 
-ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT false
-ENV LC_ALL en_US.UTF-8
-ENV LANG en_US.UTF-8
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false \
+    LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8 \
+    DOTNET_SDK_VERSION=2.1.300-preview1-008174 \
+    NUGET_XMLDOC_MODE=skip
 
 # Install .NET Core SDK
-ENV DOTNET_SDK_VERSION 2.1.300-preview1-008174
 
 RUN apk add --no-cache --virtual .build-deps \
         openssl \
@@ -21,7 +22,6 @@ RUN apk add --no-cache --virtual .build-deps \
     && rm dotnet.tar.gz \
     && apk del .build-deps
 
-ENV NUGET_XMLDOC_MODE skip
 
 # Trigger first run experience by running arbitrary cmd to populate local package cache
 RUN dotnet help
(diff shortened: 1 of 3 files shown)


2.1/sdk/alpine/amd64/Dockerfile
```diff
@@ -3,12 +3,13 @@ FROM microsoft/dotnet:2.1-runtime-deps-alpine
 # Disable the invariant mode (set in base image)
 RUN apk add --no-cache icu-libs
 
-ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT false
-ENV LC_ALL en_US.UTF-8
-ENV LANG en_US.UTF-8
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false \
```
MichaelSimons: I am completely in favor of grouping ENV that are logically related and grouped in the Dockerfile today such as 

```
ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT false
ENV LC_ALL en_US.UTF-8
ENV LANG en_US.UTF-8
```

I am not convinced however that the other ENVs should be grouped.  This breaks the logical ordering.   I also look at the precedence set by the _official images_ on Docker.  They preserve logic groupings.

* https://github.com/docker-library/openjdk/blob/df9da1156dbc0e2ae0f87aeef9c79df8720d0811/10-jdk/slim/Dockerfile
* https://github.com/nodejs/docker-node/blob/ab12e9bab43bde5efd2e9fa278d941e4062c6f4a/9/alpine/Dockerfile
* https://github.com/docker-library/golang/blob/9bf6daddb324a4d8e82b0613cf348a6eff363f95/1.10/alpine3.7/Dockerfile

markrendle: And now I'm questioning whether this is the right approach given the impending `--squash` flag for `docker build`, which makes the whole layer-proliferation problem go away anyway.

MichaelSimons: Can you clarify `impending`?  Docker squash is currently available.  `official` images are not squashing layers.

I still see valuing in combining ENV instructions that are logically grouped within the .NET Dockerfiles.

richlander:
> I still see valuing in combining ENV instructions that are logically grouped within the .NET Dockerfiles.

That's my vote.

JRAlexander:
My vote as well.

MichaelSimons:
@markrendle - are you interesting in completing this work?  I would be happy to make the suggested changes if you'd like.

BTW - this PR should be made against the `nightly` branch.  The `nightly` branch is the staging environment for the next "release"

MichaelSimons:
I just realized why the multi-variable format is not more widely used.  

```
ENV Variable1 Value1
ENV Variable2=Value2 \
    Variable3=Value3+$Variable2+$Variable1
RUN echo $Variable3
```

`echo` emits `Value3++Value1`.  You cannot reference variables defined within a multi-variable ENV instructions.

richlander:
Closing this PR in favor of #445. Thanks @markrendle !

PR #446, opened 2018-03-24
Fix broken links on DH README

richlander:


richlander:
I don't have access to update the DH README. How should that happen?

MichaelSimons:
I just updated the README on DH.

PR #434, opened 2018-03-19
Add instructions for building with docker

richlander:
Like the developing in a docker container topic but specifically for building.

PR #450, opened 2018-03-26
Add environment variables to 2.0

richlander:
Same as #445. Replaces #447.

Also based on feedback given in #434


richlander:
I will let you handle the merging!

PR #456, opened 2018-03-29
Test refactoring

MichaelSimons:
* Renamed ImageDescriptor to ImageData.  The rename will help keep the test data declarations on one line.  This is important for readability to easily distinguish the differences.  More test data is coming soon which is what is driving this change.
* Renamed ImageDescriptor.DotNetCoreVersion to DotNetVersion
* Refactored GetDotNetImage to make it easier to call.


MichaelSimons:
@dotnet-bot test Debian - 2.1 Dockerfiles please

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -22,27 +22,27 @@ public class ImageTests
         private static string RepoOwner => Environment.GetEnvironmentVariable("REPO_OWNER") ?? "microsoft";
         private static string VersionFilter => Environment.GetEnvironmentVariable("IMAGE_VERSION_FILTER");
 
-        private static ImageDescriptor[] LinuxTestData = new ImageDescriptor[]
+        private static ImageData[] LinuxTestData = new ImageData[]
             {
```
dagood: When there's a brace on a next line, I think it's usually not indented. Helps a bit with the line length concern.

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -22,27 +22,27 @@ public class ImageTests
         private static string RepoOwner => Environment.GetEnvironmentVariable("REPO_OWNER") ?? "microsoft";
         private static string VersionFilter => Environment.GetEnvironmentVariable("IMAGE_VERSION_FILTER");
 
-        private static ImageDescriptor[] LinuxTestData = new ImageDescriptor[]
+        private static ImageData[] LinuxTestData = new ImageData[]
```
dagood: `new ImageData[]` is unnecessary. (While I'm here, this could be `readonly`.)

dagood: Hmm, the change isn't only readonly, but generates a new array each time it's called, right? It's only called once and I suppose it's more like immutability, so that's fine by me.

I'll confess I don't really care about `readonly` personally since it's so irritating to make proper immutable code in C# anyway, it's just that keeping these restrictive when there's no reason not to comes up in style guides.

MichaelSimons: You're right - it shouldn't evaluate each time.  I'll fix with my next PR.

PR #469, opened 2018-04-09
Refactor update-dependencies to support additional scenarios

MichaelSimons:
* Add support for reading local build info file.
* Add support for 2.0 sdk scenarios where the sdk version does not align with the dockerfile version folder layout.  Runtime version info is used to find Dockerfiles now.
* Add support for reading checksums from `https://dotnetcli.blob.core.windows.net/dotnet/checksums` if not found in `https://dotnetclichecksums.blob.core.windows.net`

@dotnet-bot skip ci please

### DIFF
--- a/scripts/update-dependencies/DockerfileShaUpdater.cs
+++ b/scripts/update-dependencies/DockerfileShaUpdater.cs
@@ -2,13 +2,13 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.DotNet.VersionTools.Dependencies;
-using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 
 namespace Dotnet.Docker
 {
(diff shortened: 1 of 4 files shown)


scripts/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -18,64 +18,136 @@ namespace Dotnet.Docker
...
+            $"ENV (?<{EnvNameGroupName}>(DOTNET|ASPNETCORE)_[\\S]*VERSION) (?<{ValueGroupName}>[\\S]*)";
+        private static readonly string s_envShaPattern =
+            $"ENV (?<{EnvNameGroupName}> DOTNET_[\\S]*DOWNLOAD_SHA) (?<{ValueGroupName}>[\\S]*)";
+        private static readonly string s_envDownloadUrlPattern = $"ENV (DOTNET_[\\S]*DOWNLOAD_URL) (?<{ValueGroupName}>[\\S]*)";
+        private static readonly string s_inlineUrlPattern = 
+            $"(?<{ValueGroupName}>https://dotnetcli.blob.core.windows.net/[^;\\s]*)";
+        private static readonly string s_varShaPattern =
+            $"[ \\$](?<{EnvNameGroupName}>(dotnet_|aspnetcore_)sha512)( )*=( )*'(?<{ValueGroupName}>[^'\\s]*)'";
+
+        private static readonly Regex s_downloadUrlRegex = new Regex($"({s_envDownloadUrlPattern})|{s_inlineUrlPattern}");
```
dagood: Unfortunate that GitHub's code coloring doesn't like that `s_varShaPattern` string. 😕 Not sure what's exactly causing it to break.

scripts/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -18,64 +18,136 @@ namespace Dotnet.Docker
...
-                    Trace.TraceInformation($"Downloading '{shaUrl}'.");
-                    using (Stream shaStream = new HttpClient().GetStreamAsync(shaUrl).Result)
-                    using (StreamReader reader = new StreamReader(shaStream))
-                    {
-                        sha = reader.ReadToEnd().ToLowerInvariant();
-                    }
+                    downloadUrl = SubstituteEnvRef(downloadUrl, versionInfo.EnvName, versionInfo.Version);
+                    sha = await GetDotNetCliChecksumsShaAsync(downloadUrl, versionInfo.EnvName)
+                        ?? await GetDotNetReleaseChecksumsShaAsync(downloadUrl, versionInfo.EnvName, versionInfo.Version);
+                    sha = sha?.ToLower();
```
dagood: `ToLowerInvariant`?

scripts/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -18,64 +18,136 @@ namespace Dotnet.Docker
...
+            {
+                if (response.IsSuccessStatusCode)
+                {
+                    sha = await response.Content.ReadAsStringAsync();
+                }
+                else
+                {
+                    Trace.TraceInformation($"Failed to find `dotnetclichecksums` sha");
+                }
+            };
```
dagood: Unnecessary semicolon (and more later on other `using`s).

scripts/update-dependencies/Options.cs
```diff
@@ -42,10 +42,11 @@ public void Parse(string[] args)
                     "GitHub user used to make PR (if not specified, a PR will not be created)");
                 GitHubUser = gitHubUser;
 
-                string buildInfoUrl = null;
+                Uri buildInfoUrl = null;
                 syntax.DefineParameter(
                     "build-info",
                     ref buildInfoUrl,
+                    (value) => new Uri(value),
                     "URL of the build info to update the Dockerfiles with");
```
dagood: Worthwhile to call out here that `file://` works?

scripts/update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: Switched

scripts/update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: Removed all

scripts/update-dependencies/Options.cs
MichaelSimons: Good suggestion.

scripts/update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: I had noticed that :(

MichaelSimons: If I don't put it on a separate line, the syntax coloring works...

dagood: Wow... that's pretty strange.

MichaelSimons: I wonder if this is by chance related to the same underlying cause as https://github.com/atom/language-csharp/issues/113?  Sounds like work is in play to get GH to use a more recent version of the grammar.  Will monitor.

kasper3: FYI, GitHub has just released and that issue along with many others around C# syntax highlighting are fixed 🎉

MichaelSimons: Thanks for the heads up @kasper3!

PR #472, opened 2018-04-09
[WIP] Misc test cleanup

MichaelSimons:
* Ensure CI is running with latest base images.  Solved by adding `--pull` option.
* DockerHelper.Kill has been sporadically failing which causes the logic to cleanup volumes to fail.  Refactored this to force delete the container in hopes of avoiding this problem.
* Added `ToString` override to ImageData in order to avoid truncation in test output - e.g. `Passed   Microsoft.DotNet.Docker.Tests.ImageTests.VerifyImages(imageDescriptor: ImageDescriptor { Architecture = "amd64", DotNetCoreVersion = "1.1", HasNoSdk = False, IsAlpine = False, IsArm = False, ... })`
* Shortened some long lines in the utests
* Switched base to using `dotnet test` instead of `dotnet xunit` in order to get detailed test output


test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -40,7 +39,7 @@ public class ImageTests
             new ImageData { DotNetVersion = "2.1", OsVariant = OS.Bionic, IsWeb = true },
             new ImageData { DotNetVersion = "2.1", OsVariant = OS.Alpine, IsWeb = true },
         };
-        private static ImageData[] WindowsTestData => new[]
+        private static readonly ImageData[] s_windowsTestData = new[]
```
dagood: Redundant `new[]`?

MichaelSimons: Removed

PR #475, opened 2018-04-10
Refactor test cleanup logic

MichaelSimons:
Builds have been failing frequently due to DockerHelper.Kill failing to remove the container.  This is an attempt to cleanup the container in a more robust fashion.

### DIFF
--- a/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
@@ -36,9 +36,9 @@ namespace Microsoft.DotNet.Docker.Tests
             ExecuteWithLogging($"build -t {tag} {buildArgsOption} -f {dockerfile} .");
         }
 
-        public static bool ContainerExists(string container)
+        public static bool ContainerExists(string name)
         {
-            return ResourceExists("container", container);
+            return ResourceExists("container", $"-f \"name={name}\"");
         }
 
         public void DeleteContainer(string container)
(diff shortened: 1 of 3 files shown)


test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -36,6 +36,20 @@ public void Build(string dockerfile, string tag, string fromImage, params string
...
+            return ResourceExists("container", container);
+        }
+
+        public void DeleteContainer(string container)
+        {
+            if (ContainerExists(container))
+            {
+                ExecuteWithLogging($"logs {container}", ignoreErrors: true);
+                ExecuteWithLogging($"container rm -f {container}");
+            }
```
dagood: Worth logging something when deletion is requested for a container that doesn't exist?

MichaelSimons: No because the non-aspnet flows run the container attached with `--rm` so the container is deleted right away and the call to DeleteContainer will no-op.  aspnet runs detatched and this is the code that is supposed to stop/clean it up.

MichaelSimons:
I made an additional change to remove the usage of a hard coded host port.  This is useful when the cleanup code fails (additional tests won't fail immediately when trying to use the port) and it is nice when running locally.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -36,9 +36,9 @@ public void Build(string dockerfile, string tag, string fromImage, params string
             ExecuteWithLogging($"build -t {tag} {buildArgsOption} -f {dockerfile} .");
         }
 
-        public static bool ContainerExists(string container)
+        public static bool ContainerExists(string name)
         {
-            return ResourceExists("container", container);
+            return ResourceExists("container", $"-f \"name={name}\"");
```
dagood: I'm not able to find docs on what a bare arg means to e.g. `docker container ls -q foo`, did this work before, or did I get the result command wrong? (New command looks good.)

MichaelSimons: It wasn't finding the container before.

PR #481, opened 2018-04-12
Refactor update-dependencies to not fail when sha's are not found

MichaelSimons:
There are times when the full set of sha's are not available on the checksum feeds.  Having update-dependencies complete all of changes is more beneficial than failing part way through.

### DIFF
--- a/scripts/update-dependencies/DockerfileShaUpdater.cs
+++ b/scripts/update-dependencies/DockerfileShaUpdater.cs
@@ -24,12 +24,12 @@ namespace Dotnet.Docker
         private static readonly string s_envVersionPattern =
             $"ENV (?<{EnvNameGroupName}>(DOTNET|ASPNETCORE)_[\\S]*VERSION) (?<{ValueGroupName}>[\\S]*)";
         private static readonly string s_envShaPattern =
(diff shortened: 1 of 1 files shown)


scripts/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -24,12 +24,12 @@ public class DockerfileShaUpdater : FileRegexUpdater
         private static readonly string s_envVersionPattern =
             $"ENV (?<{EnvNameGroupName}>(DOTNET|ASPNETCORE)_[\\S]*VERSION) (?<{ValueGroupName}>[\\S]*)";
         private static readonly string s_envShaPattern =
-            $"ENV (?<{EnvNameGroupName}> DOTNET_[\\S]*DOWNLOAD_SHA) (?<{ValueGroupName}>[\\S]*)";
+            $"ENV DOTNET_[\\S]*DOWNLOAD_SHA (?<{ValueGroupName}>[\\S]*)";
```
dagood: Isn't this still needed by `TryGetDotNetVersion`?

MichaelSimons: I don't think so - `TryGetDotNetVersion` uses the `s_versionRegex` which in turn uses the `s_envVersionPattern`.

dagood: I see, this cleanup makes sense then.

Issue #503, opened 2018-04-25
Unit Tests should be calling `dotnet publish` with `-c Release` to align with more common usage


MichaelSimons:
See 
* https://github.com/dotnet/dotnet-docker/blob/master/test/Microsoft.DotNet.Docker.Tests/ImageTests.cs#L204
* https://github.com/dotnet/dotnet-docker/blob/master/test/Microsoft.DotNet.Docker.Tests/ImageTests.cs#L255

PR #513, opened 2018-04-27
Misc test fixups

MichaelSimons:
1. Add a workaround for https://github.com/dotnet/cli/issues/9162 to get tests passing on arm.
2. Add exclusions to .dockerignore to avoid coping so much into testrunner image.
3. Fix for #503

### DIFF
--- a/test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
@@ -99,7 +99,12 @@ namespace Microsoft.DotNet.Docker.Tests
                 if (!imageData.HasNoSdk)
                 {
                     VerifySdkImage_PackageCache(imageData);
-                    await VerifySdkImage_RunApp(imageData, appSdkImage);
+
+                    // TODO: Skip running app in arm + web configuration to workaround https://github.com/dotnet/cli/issues/9162
+                    if (!(imageData.IsArm && imageData.IsWeb))
+                    {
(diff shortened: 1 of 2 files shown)


MichaelSimons:
@dleeapho - When you have time, can I get you to take a look?  Thanks

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -99,7 +99,12 @@ private static string GetFilterRegexPattern(string filterEnvName)
                 if (!imageData.HasNoSdk)
                 {
                     VerifySdkImage_PackageCache(imageData);
-                    await VerifySdkImage_RunApp(imageData, appSdkImage);
+
+                    // TODO: Skip running app in arm + web configuration to workaround https://github.com/dotnet/cli/issues/9162
+                    if (!(imageData.IsArm && imageData.IsWeb))
```
dleeapho: https://github.com/dotnet/cli/pull/9163 has been merged so dotnet/cli#9162 should be fixed going forward. Instead of skipping the ARM web tests in code is there no way to mark it as a known failure?

MichaelSimons: dotnet/cli#9162 is fixed but not in RC.  Because we are using [Theory] it is not practical to mark the aspnet arm scenarios as known failures.  Additionally, this causes the test scenario to fail immediately, while this workaround allows the rest of the scenario to run - e.g. testing the runtime images.

PR #550, opened 2018-05-21
Merge changes from master into nightly

MichaelSimons:


MichaelSimons:
@richlander - simpler how?  These are all the changes that have been checked into the 'master' branch.  I would like them merged into nightly as I am refactoring some build infrastructure that affects the samples build.

richlander:
I'm an idiot. I was thinking the other direction. It all makes sense now.

PR #573, opened 2018-06-06
Update 1.* tags to be consistent with 2.*

MichaelSimons:
Update 1.* tags to include <major>-<minor>-<image variant>-<os> tags.  This is to be consistent with the tags produced for 2.*.  I also updated the FROM instructions in the 1.* runtime images to utilize the new tags.  The previous references were somewhat confusing/incorrect as they utilizing a manifest tag.  This allows the build/test infrastructure to be consistent across all the version produced.

### DIFF
--- a/test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
@@ -368,12 +368,7 @@ namespace Microsoft.DotNet.Docker.Tests
                     throw new NotSupportedException($"Unsupported image type '{variantName}'");
             }
 
-            string imageName = $"{s_repoOwner}/{s_repoName}:{imageVersion}-{variantName}";
-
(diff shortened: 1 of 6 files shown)


MichaelSimons:
@dotnet-bot test NanoServer-1709 - 2.0 Dockerfiles please

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -368,12 +368,7 @@ public static string GetDotNetImage(DotNetImageType imageType, ImageData imageDa
                     throw new NotSupportedException($"Unsupported image type '{variantName}'");
             }
 
-            string imageName = $"{s_repoOwner}/{s_repoName}:{imageVersion}-{variantName}";
-
-            if (!imageData.DotNetVersion.StartsWith("1."))
```
dleeapho: 👍

PR #574, opened 2018-06-06
Fix GRPC issue - Add libc6-compat to base image

shubhamranjan:
Running a GRPC server on alpine throws an error as follows :

```
Error loading native library "/app/runtimes/linux/native/libgrpc_csharp_ext.x64.so"
   at Grpc.Core.Internal.UnmanagedLibrary..ctor(String[] libraryPathAlternatives)
   at Grpc.Core.Internal.NativeExtension.LoadUnmanagedLibrary()
   at Grpc.Core.Internal.NativeExtension.LoadNativeMethods()
   at Grpc.Core.Internal.NativeExtension..ctor()
   at Grpc.Core.Internal.NativeExtension.Get()
   at Grpc.Core.GrpcEnvironment.GrpcNativeInit()
   at Grpc.Core.GrpcEnvironment..ctor()
   at Grpc.Core.GrpcEnvironment.AddRef()
   at Grpc.Core.Channel..ctor(String target, ChannelCredentials credentials, IEnumerable`1 options)
   at WebApplication2.Startup.ConfigureServices(IServiceCollection services) in /WebApplication2/Startup.cs:line 16
```

Dependency  **libc6-compat** has been added which fixes the same.

### DIFF
--- a/2.1/runtime-deps/alpine3.7/amd64/Dockerfile
+++ b/2.1/runtime-deps/alpine3.7/amd64/Dockerfile
@@ -12,6 +12,7 @@ RUN apk add --no-cache \
         tzdata \
         userspace-rcu \
         zlib \
+        libc6-compat \
     && apk -X https://dl-cdn.alpinelinux.org/alpine/edge/main add --no-cache \
         lttng-ust
 


MichaelSimons:
@shubhamranjan, what is the justification for adding this to the microsoft/dotnet images?  From the looks of things this is a native dependency GRPC adds.

shubhamranjan:
GRPC works out of of the box with other dotnet 2.1 linux flavors like debian/bionic. This is something  required by gRPC for alpine and users anyhow will have to install this to run their grpc servers.
We'll save users a lot of time  by just installing it ourselves.

MichaelSimons:
The philosophy we are taking with the Alpine image is to keep the base image as reasonably small as possible.  Components that are not required for all use cases are not included by default.  This is why globalization is disabled by default (https://github.com/dotnet/dotnet-docker/blob/master/samples/dotnetapp/README.md#build-and-run-the-sample-for-alpine-with-docker).  I see the libc6-compat lib GRPC requires as falling into this category.

shubhamranjan:
Cool , In that case this can be declined / closed.Devs (like me) wanting grpc support for alpine may create a custom image from this as base image.

MichaelSimons:
...and we are certainly open to hearing feedback on cases like this.  If there are large number of folks needing certain components, please let us know.  We may decide to add variants to better serve certain scenarios.

PR #581, opened 2018-06-11
Add samples script

richlander:
- Start of a customer-oriented script for samples
- Shows one possible approach to unit testing
- Implemented in PowerShell
- Next step is pushing to a registry (requires more thought)


samples/dotnetapp/build.ps1
```diff
@@ -0,0 +1,54 @@
...
+
+# Run Docker Build
+Write-Output "## Running Docker build"
+docker build --pull -t dotnetapp .
+
+# Capture tests if Docker build fails
+if ($LASTEXITCODE -ne 0)
+{
+    Write-Output "## Docker build failed. Attempting to collect test results."
+    docker build --pull --target testrunner -t dotnetapp:test .
```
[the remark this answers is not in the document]
richlander: Ya. I was going back and forth on that. The old scheme made some sense in the absence of a script. With a script, then testing only as part of docker run seems like the way to go. I'll do that.

samples/dotnetapp/build.ps1
```diff
@@ -4,51 +4,47 @@
...
 $DockerOS = docker version -f "{{ .Server.Os }}"
+$ImageName = "dotnetapp"
+$TestImageName = "dotnetapp:test"
+
+Log "Running Docker build"
+docker build --pull -t $ImageName .
+if ($LASTEXITCODE -ne 0) { throw "## docker build failed" }
+
+Log "Run tests and collect results."
+docker build --pull --target testrunner -t $TestImageName .
```
MichaelSimons: The `--pull` doesn't seem desirable here.  You really don't want to test with a different base image that what you built with on line 18.

samples/dotnetapp/build.ps1
```diff
@@ -4,51 +4,47 @@
...
+    Write-Output "###### $s"
+}
+
 $IsRunningOnUnix = $PSVersionTable.contains("Platform") -and $PSVersionTable.Platform -eq "Unix"
 $DockerOS = docker version -f "{{ .Server.Os }}"
+$ImageName = "dotnetapp"
+$TestImageName = "dotnetapp:test"
+
+Log "Running Docker build"
+docker build --pull -t $ImageName .
```
MichaelSimons: To be more robust `.` should be changed to `$PSScriptRoot`.  This applies to line 22 as well.

samples/dotnetapp/Dockerfile.basic-preview
```diff
@@ -1,23 +0,0 @@
-# This Dockerfile uses nightly preview builds for .NET Core
```
MichaelSimons: Aren't there some MD references to the two removed Dockerfile.basic files?  e.g https://github.com/dotnet/dotnet-docker/blob/e433af439bccc24d7152cd4d8d23a901348cf83a/README.md#L137

samples/dotnetapp/Dockerfile
```diff
@@ -2,34 +2,28 @@ FROM microsoft/dotnet:2.1-sdk AS build
...
 WORKDIR /app/dotnetapp
-RUN dotnet build
+RUN dotnet publish -c Release -o out
 
 
+# optional stage - build to stage to run tests
+# https://github.com/dotnet/dotnet-docker/blob/master/samples/dotnetapp/dotnet-docker-unit-testing.md
 FROM build AS testrunner
 WORKDIR /app/tests
+COPY tests/. .
```
MichaelSimons: I'm wondering if restore and build makes sense in this stage?  That would cache the packages useful for iterative dev.  It would also surface any compiler errors.  Running the resulting image would should just be about running the tests.

samples/dotnetapp/Dockerfile
```diff
@@ -1,7 +1,6 @@
+# copy csproj and restore as distinct layers
```
MichaelSimons: Curious why this change?  This phase does more than just copy and restore.

samples/dotnetapp/Dockerfile.alpine-x64-selfcontained
```diff
@@ -1,37 +1,29 @@
...
-# copy everything else and build app
-COPY . .
+# copy and build app and libraries
+WORKDIR /app/
+COPY dotnetapp/. ./dotnetapp/
+COPY utils/. ./utils/
 WORKDIR /app/dotnetapp
-RUN dotnet build
+# add IL Linker package
+RUN dotnet add package ILLink.Tasks -v 0.1.4-preview-981901 -s https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
```
MichaelSimons: I ran into a scenario the other day where there was a bug with this version.  We should consider upgrading to the latest -  0.1.5-preview-1461378

PR #583, opened 2018-06-11
Add edge releases for Alpine and Debian

richlander:
- Starting process of adding next set of releases.
- alpine-edge and buster-slim work as expected, per initial testing.
- Consider pushing images to dotnet-nightly DH repo for new Dockerfiles.

The benefit of adding releases more aggressively is two-fold:
- First class samples if people want to run ahead that they can build themselves.
- Early testing for us.


richlander:
I am out, curling. Merge when you are happy.

PR #585, opened 2018-06-11
Restructure .vsts.pipelines structure and increase build leg timeout

MichaelSimons:
- Folder structuring changes are a result general guidance on structuring from vsts.
- Timeout increase is result of DH being slow at times in conjunction with #583.


.vsts.pipelines/phases/build-linux-amd64.yml
```diff
@@ -1,11 +1,13 @@
 parameters:
   imageBuilderImage: null
   manifest: null
+  timeoutInMinutes: null
```
dseefeld: I see that the timeout is passed in with build-all.  If one isn't passed in, what is the behavior?

MichaelSimons: If null is passed in, it will get the default behavior (60 minutes).  

The use of the parameter is a temporary workaround until the global variable feature is available/deployed.  The parameter allows the value to be defined in a single location which is what I am after.

PR #586, opened 2018-06-12
Removed LibUV from aspnetcore-runtime image

### LINKED ISSUE #582: Is LibUV needed as a required dependency?
See: https://github.com/dotnet/dotnet-docker/blob/master/2.1/aspnetcore-runtime/alpine3.7/amd64/Dockerfile#L3-L6

I thought LibUV was now optional given the adoption of a .NET sockets-based implementation.

MichaelSimons:
Fixes #582


richlander:
What about this file? https://github.com/dotnet/dotnet-docker/blob/42115f52c598305e219b7032793f027abb80baa0/2.1/aspnetcore-runtime/alpine-edge/amd64/Dockerfile

MichaelSimons:
Updated the new alpine-edge image as well.  I forgot to sync and pull in the latest...

PR #600, opened 2018-06-18
Use ContainerAdministrator USER in nanoserver containers

ilyalukyanov:
This should fix the issue (https://github.com/dotnet/corefx/issues/25717) when X509Certificate2 instances cannot be created inside Nanoserver containers.

The issue was found by the Microsoft's Support team and changes in the PR are based on a workaround that was suggested.


MichaelSimons:
I'm not in favor of changing the default user to ContainerAdministrator within the NanoServer images.  Windows switched from away from having the ContainerAdministrator be the default in the base OS images for "secure by default" reasons.  The same reasoning should exist for the microsoft/dotnet images.  Scenarios which require running as admin should have to explicitly elevate permissions.

MichaelSimons:
The `microsoft/dotnet` images are not changing the user from what the base Nano Server images have.  For 1709 and greater they are following the least privilege principle.  For scenarios that need elevated privileges, you would need to construct your own Dockerfile and change the USER or `docker run` with the `-u, --user string` option.   Following the least privilege principle seems like the correct thing here.

Closing, but feel free to continue the conversation here or better yet log an issue.

PR #612, opened 2018-07-02
Rename build queues and the 'buildPath' variable

MichaelSimons:
* Renamed the `buildPath` variable to `dotnetVersion` as it more accurately describes what the variable contains.  It is used for more than the imageBuilder `path` option.
* Dropped the `Build_` prefix from the build queue names.  The `Build_` prefix was confusing to see in the build output for the Test and Copy phases - e.g. `Test_Linux_amd64 Build_2_0_Jessie`


MichaelSimons:
The failing `Debian - 2.1. Dockerfiles` CI is a separate issue - https://github.com/dotnet/dotnet-docker/pull/614

PR #614, opened 2018-07-02
Update libicu on buster images to libicu60

MichaelSimons:
`libicu57` is no longer in the default `buster` package cache, upgrading to `libicu60`.


MichaelSimons:
This issue is currently blocking builds in the `nightly` branch.

PR #619, opened 2018-07-10
[update-dependencies] Add checksum cache and compute logic

MichaelSimons:
1. Added a checksum cache which helps speed up the app since several of the Dockerfiles download the same dotnet components.
1. Added logic to compute the checksum which is useful in special cases were the published checksums are not available.

@dotnet-bot skip ci please

### DIFF
--- a/scripts/update-dependencies/DockerfileShaUpdater.cs
+++ b/scripts/update-dependencies/DockerfileShaUpdater.cs
@@ -3,10 +3,12 @@
 
 using Microsoft.DotNet.VersionTools.Dependencies;
 using System.Collections.Generic;
+using System.Security.Cryptography;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
@@ -26,7 +28,7 @@ namespace Dotnet.Docker
         private static readonly string s_envShaPattern =
             $"ENV DOTNET_[\\S]*DOWNLOAD_SHA (?<{ValueGroupName}>[\\S]*)";
         private static readonly string s_envDownloadUrlPattern = $"ENV (DOTNET_[\\S]*DOWNLOAD_URL) (?<{ValueGroupName}>[\\S]*)";
-        private static readonly string s_inlineUrlPattern = 
+        private static readonly string s_inlineUrlPattern =
             $"(?<{ValueGroupName}>https://dotnetcli.blob.core.windows.net/[^;\\s]*)";
         private static readonly string s_varShaPattern =
             $"[ \\$](dotnet_|aspnetcore_)sha512( )*=( )*'(?<{ValueGroupName}>[^'\\s]*)'";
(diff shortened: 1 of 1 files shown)


MichaelSimons:
@dagood - would you have time to take a look at this PR?

dagood:
I do now, taking a look.

scripts/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -61,9 +65,19 @@ public DockerfileShaUpdater(string dockerfilePath) : base()
                 if (TryGetDotNetDownloadUrl(dockerfile, out string downloadUrl))
                 {
                     downloadUrl = SubstituteEnvRef(downloadUrl, versionInfo.EnvName, versionInfo.Version);
-                    sha = await GetDotNetCliChecksumsShaAsync(downloadUrl, versionInfo.EnvName)
-                        ?? await GetDotNetReleaseChecksumsShaAsync(downloadUrl, versionInfo.EnvName, versionInfo.Version);
-                    sha = sha?.ToLowerInvariant();
+                    if (!s_shaCache.TryGetValue(downloadUrl, out sha))
+                    {
+                        sha = await GetDotNetCliChecksumsShaAsync(downloadUrl, versionInfo.EnvName)
+                            ?? await GetDotNetReleaseChecksumsShaAsync(downloadUrl, versionInfo.EnvName, versionInfo.Version)
```
dagood: It looks maybe ok to ignore `versionInfo.EnvName` in the cache key, but `versionInfo.Version` seems more likely to cause a false hit.

scripts/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -78,6 +92,38 @@ public DockerfileShaUpdater(string dockerfilePath) : base()
...
+                    {
+                        byte[] hashedInputBytes = hash.ComputeHash(httpStream);
+
+                        StringBuilder stringBuilder = new StringBuilder(128);
+                        foreach (byte b in hashedInputBytes)
+                        {
+                            stringBuilder.Append(b.ToString("X2"));
+                        }
+                        sha = stringBuilder.ToString();
+                    };
```
dagood: Unnecessary `;`

scripts/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -35,6 +37,8 @@ public class DockerfileShaUpdater : FileRegexUpdater
         private static readonly Regex s_shaRegex = new Regex($"({s_envShaPattern})|({s_varShaPattern})");
         private static readonly Regex s_versionRegex = new Regex(s_envVersionPattern);
 
+        private static Dictionary<string, string> s_shaCache = new Dictionary<string, string>();
```
dagood: Could also be `readonly`, right?

scripts/update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: The download url is the item we are retrieving the checksum for.  The envName and Version have already been substituted in the url that is read from the Dockerfile.

scripts/update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: removed

scripts/update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: yes - added.

scripts/update-dependencies/DockerfileShaUpdater.cs
dagood: Ok, sounds fine. I guess in theory the cache could be broken if there were a broken Dockerfile included in auto-update, but that's already a broken situation anyway.

PR #622, opened 2018-07-10
Updates for July Servicing

MichaelSimons:


joshhubers:
Is anyone of aware of issues with this merge? 

I'm seeing some odd things with the Microsoft Dotnet dockerhub.
https://hub.docker.com/r/microsoft/dotnet/

All of the tags say they were updated at the same time, and following any of the tag links for older versions of dotnet lead to a dockerfile that was updated to 2.1.

For example, the tag 2.0-sdk for 2.0/sdk/stretch/amd64/Dockerfile leads to this: 
https://github.com/dotnet/dotnet-docker/blob/master/2.0/sdk/stretch/amd64/Dockerfile

Which clearly shows:
```
# Install .NET Core SDK
ENV DOTNET_SDK_VERSION 2.1.202
```
And looking at the git blame, it points to this pull?

Whenever I go to pull microsoft/dotnet:2.0-sdk I get an image with 2.1 installed.

MichaelSimons:
@joshhubers - Regarding the 2.0-sdk image containing the 2.1.202 SDK.  This is expected.  This tag represents the latest SDK version to develop a 2.0 application.  The 2.0 release notes may help clarify -  https://github.com/dotnet/core/tree/master/release-notes/2.0.

The version number disparity between the runtime and sdk is confusing and has been addressed in 2.1.  In 2.1, the major and minor numbers are aligned between the two.

The tags listing on Docker Hub all having the same time simply means they were all pushed at the same time.  All of the images were patched as seen in this PR.

MichaelSimons:
@joshhubers - Are you having concrete issues with the latest images?  If so, I would like to hear about them.  Thanks

joshhubers:
@MichaelSimons 

No, sorry, I think the error I'm seeing is unrelated, I was just thrown off by seeing "2.1". Did not realize there was a runtime and sdk version separation.

Issue #625, opened 2018-07-11
Nanoserver selfcontained sample does not build


batkuip:
## Steps to reproduce the issue
1. cd samples/dotnetapp
2. docker build --pull -t dotnetapp:nanoserver-selfcontained -f Dockerfile.nanoserver-x64-selfcontained 

## Expected behavior
builds docker image

## Actual behavior
fails with 
```
/usr/share/dotnet/sdk/2.1.302/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.CrossGen.targets(72,5): error : Unable to find resolved path for '/root/.nuget/packages/runtime.win-x64.microsoft.netcore.app/2.1.2/runtimes/win-x64/native/../../../tools/crossgen'. [/app/dotnetapp/dotnetapp.csproj]
```

## Output of `docker version`

```
Client:
 Version:      18.05.0-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   f150324
 Built:        Wed May  9 22:12:05 2018
 OS/Arch:      darwin/amd64
 Experimental: true
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.05.0-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.10.1
  Git commit:   f150324
  Built:        Wed May  9 22:20:16 2018
  OS/Arch:      linux/amd64
  Experimental: true
```

## Output of `docker info`

```
Containers: 4
 Running: 0
 Paused: 0
 Stopped: 4
Images: 54
Server Version: 18.05.0-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host ipvlan macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 773c489c9c1b21a6d78b5c538cd395416ec50f88
runc version: 4fc53a81fb7c994640722ac585fa9ca548971871
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.93-linuxkit-aufs
Operating System: Docker for Mac
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 1.952GiB
[...]

PR #627, opened 2018-07-11
Fix broken copy from reference in nano selfcontained Dockerfile

MichaelSimons:
I noticed this issue while investigating #625.

@dotnet-bot skip CI please


MichaelSimons:
Merging in order to fix the broken sample.

PR #656, opened 2018-08-14
Refactor CI job names

MichaelSimons:
Rename the `Debian - * Dockerfiles` CI jobs to `Linux - * Dockerfiles`.  The usage of `Debian` is from the time when the repo only supported Debian distro based images.  The repo now supports multiple distros so a more abstract name is appropriate.


MichaelSimons:
@dotnet-bot test ci please

PR #665, opened 2018-08-20
Shorten README.md to workaround 25k limit in Docker Hub

MichaelSimons:
This is a workaround for the 25k limit Docker Hub has on the full description field (e.g. the contents of the README.md).  We are working with Docker to try to get the full description size increased or a better long term solution.

Introduced a new TAGS.md file that contains the full set of documented tags.  Updated the README.md to reference the TAGS.md within the preview tags sections.

@dotnet-bot skip ci please.


MichaelSimons:
@dleeapho - You bring up a fair point - I changed it to `Tags`.

Issue #677, opened 2018-08-27
Consistent environment variable names for DOTNET_RUNNING_IN_CONTAINER


fbeltrao:
## Steps to reproduce the issue
1. Build .NET Core application based on microsoft/dotnet:2.1-aspnetcore-runtime-nanoserver-1709
2. Build .NET Core application based on microsoft/dotnet:2.1-aspnetcore-runtime
3. Verify that nano server uses DOTNET_RUNNING_IN_CONTAINERS while the later uses DOTNET_RUNNING_IN_CONTAINER

## Expected behavior
Both base image define the same environment variable name

## Actual behavior
They have distinct environment variable names

## Reference files
https://github.com/dotnet/dotnet-docker/blob/master/2.1/aspnetcore-runtime/nanoserver-1709/amd64/Dockerfile#L36
https://github.com/dotnet/dotnet-docker/blob/master/2.1/runtime-deps/stretch-slim/amd64/Dockerfile#L21

My proposed solution would be to change all windows based images to use DOTNET_RUNNING_IN_CONTAINER (singular). See them [here](https://github.com/dotnet/dotnet-docker/search?q=DOTNET_RUNNING_IN_CONTAINERS&type=Code)

I have created [PR 678](https://github.com/dotnet/dotnet-docker/pull/678) as an optional solution.

MichaelSimons:
@fbeltrao - Thanks for logging this issue.  When fixing this, I don't think we should be simply renaming the existing variable in the nanoserver images to be singular.  As you noted in the PR this is considered a breaking change.  

Since 2.1 is going to be LTSC, I am in favor of adding the singular version.  In the 2.2 images (which are in preview) and going forward, we can make the correction so that there is only on variable.

PR #678, opened 2018-08-27
(Replaced by #680) Consistent environment variable names for DOTNET_RUNNING_IN_CONTAINER

fbeltrao:
This PR is a potential fix for issue #677 

I am however, not sure about the side effects of this change. It will potentially break any existing application that expects the old behavior. We could provide both for backward compatibility giving time for people to make code changes.

Another option is to fix this problem only in newer versions.

### DIFF
--- a/2.1/aspnetcore-runtime/nanoserver-1803/amd64/Dockerfile
+++ b/2.1/aspnetcore-runtime/nanoserver-1803/amd64/Dockerfile
@@ -34,4 +34,4 @@ USER ContainerUser
 # Configure web servers to bind to port 80 when present
 ENV ASPNETCORE_URLS=http://+:80 `
     # Enable detection of running in a container
-    DOTNET_RUNNING_IN_CONTAINERS=true
+    DOTNET_RUNNING_IN_CONTAINER=true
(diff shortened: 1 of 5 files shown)


MichaelSimons:
@fbeltrao - Thanks for making this PR.  As I noted in the #677 - I'm of the opinion that we should not make this a breaking change.  My preference would be to add the singular variable in the 2.1 images and then fix the problem going forward.  Lastly can you make this change in the `nightly` branch?  Nightly includes 2.2 and 3.0 images as well.  The changes will be available immediately in nightly and then get merged into master during the next patch Tuesday.

fbeltrao:
@MichaelSimons thanks for the feedback. Just to make sure I understand your suggestions.

- To fix 2.2+ versions (currently in preview) we can make the breaking change, having only DOTNET_RUNNING_IN_CONTAINER for all images. This PR should be made on **nightly** branch.

- To fix 2.1 version (with LTSC) we cannot introduce a breaking change, therefore we should define both values (singular and plural) on windows images. The PR for this change should go to **$master** branch.

MichaelSimons:
@fbeltrao - One small correction - both changes should go into `nightly`.  The 2.1 changes will get merged into `master` during the next .NET servicing event (e.g. patch Tuesday).

fbeltrao:
@MichaelSimons new PR made on nightly branch: https://github.com/dotnet/dotnet-docker/pull/680

PR #717, opened 2018-10-01
Refactor manifest.json and readme templates for easier maintainability

MichaelSimons:
Refactored the manifest.json to utilize variables for the full version numbers.  This will make it easier to make and review the changes to introduce a new patch version.  Similarly the readme templates were refactored to use the `major.minor` tags where possible.  There are a few places that need to list out explicit tags due to the fact that `runtime-deps` images are shared across versions.


MichaelSimons:
Withdrawing as this is going to require more work.  The build-and-test script needs updating to handle this change :(

PR #732, opened 2018-10-08
Added some Dockerfiles for creating a .NET Core on Windows Server Core

mattleibow:
Just a nice small PR to add the initial bits to get a Windows Server Core image with .NET Core.

Not sure of the reason they don't exist, but I am sure someone will post a comment in a minute 😺 

Anyway, this is very useful so I don't have to download an install the Core SDK each time on my images.

A practical use case is this issue: https://github.com/mono/SkiaSharp/issues/591. As a result, I had to create a longer Dockerfile to first create a container, install .NET Core SDK, and then build my app, create a new container, install the Core Runtime and then run my app: https://gist.github.com/mattleibow/32ff6d124e7b5b93e94f1eb45d243aba

I don't mind installing the vc_redist at this point, as I only have to do it on the runtime container, and I hope to be removing this step, but I hope to be able to just pull the latest .NET Core image that can to font subsetting :)

### DIFF
--- a/2.1/aspnetcore-runtime/windowsservercore-1709/amd64/Dockerfile
+++ b/2.1/aspnetcore-runtime/windowsservercore-1709/amd64/Dockerfile
@@ -1,0 +1,23 @@
+# escape=`
+FROM microsoft/windowsservercore:1709
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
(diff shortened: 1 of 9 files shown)


mattleibow:
I see my docker files are all wrong and I shouldn't have copy-pasted 😜 so I opened an issue for better discussion.

MichaelSimons:
Closing as we have no plans to support Windows Server Core based images.  See discussion in #741.

PR #751, opened 2018-10-15
The ports should be same for consistency

iamalsaher:


jespersh:
isn't it because port 80 inside the docker container is forwarded to port 8000 outside?
Has this been changed everywhere else or does it propagate down inside the container image as config?

iamalsaher:
Yeah, I realised this, feel free to close the request

PR #757, opened 2018-10-16
Update build yaml to share build and copy jobs across all platforms

MichaelSimons:
With this change there is not a single job for each build and copying images regardless of the platform.  Right now there is too much different between testing Linux and Windows images but I have a long term plan to eliminate the differences. Overall this greatly helps with the maintenance and onboarding of new platforms.  

Azure DevOps is considering a feature which would greatly help cleanup the duplicated pool definitions - https://github.com/Microsoft/azure-pipelines-yaml/pull/3.  They are also considering support for yaml anchors which would help as well.

@dotnet-bot skip ci please


.vsts-pipelines/jobs/build-test-publish-repo.yml
```diff
@@ -57,67 +78,115 @@ jobs:
...
+    pool: # linuxArm32v7Pool
+      name: DotNetCore-Infra
+      demands:
+      - agent.os -equals linux
+      - RemoteDockerServerOS -equals linux
+      - RemoteDockerServerArch -equals arm32v7
     matrix: $[ dependencies.GenerateMatrices.outputs['buildMatrix.buildMatrixLinuxArm32v7'] ]
-- template: copy-images-windows.yml
+    architecture: arm
+    dockerClientOS: linux
```
dseefeld: Do you need "useRemoteDockerServer: true" here?

dseefeld: Never mind, I found the answer when I reviewed copy-images.yml

MichaelSimons: There is a mistake here.  This should be using the linuxAmd64Pool.  You can use Linux amd64 machines to perform the copy which is going to be much faster than the arm machines.  This is what was being done before.

PR #776, opened 2018-10-29
Improve maintainablity of readme templates

MichaelSimons:
Updated the readme templates to reference the <major>.<minor> tags versus the <major>.<minor>.<patch> tags.  This eases the maintenance as the templates don't have to be updated when picking up new patch versions.  

In order to support this, a new version of the image-builder is used which has dotnet/docker-tools#119.  Along with getting this change comes the new change to not include the Dockerfile path in the readme.  This is to reduce this size of the readme to fit into Docker Hub which has a 25 K limit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,29 +38,29 @@ See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotn
 
 # Linux amd64 tags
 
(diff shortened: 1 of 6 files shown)


MichaelSimons:
@dotnet-bot test NanoServer-1709 - 2.1-2.2 Dockerfiles please
@dotnet-bot test NanoServer-1803 - 2.1-2.2 Dockerfiles please

PR #778, opened 2018-10-29
Updates to ease/automate on-boarding new patch versions

MichaelSimons:
There are three parts to these changes:

1. Update the manifest to use variables to encapsulate the full version part of the tag.  This makes it easier to on-board new patches as only one place within the manifest needs to be updated.
2. In order to support variable usage within the manifest tag names, the build-and-test script needs to be updated to use the ImageBuilder tool which supports variable substitution.
3. By using manifest variables to encapsulate the full version, the update-dependencies tool can easily update the versions and regenerate the tags readme.

The end result of these changes is that Maestro will now automatically update everything needed when a new `patch` version is created.

### DIFF
--- a/scripts/update-dependencies/ManifestUpdater.cs
+++ b/scripts/update-dependencies/ManifestUpdater.cs
@@ -1,0 +1,62 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.VersionTools.Dependencies;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Dotnet.Docker
+{
+    /// <summary>
+    /// An IDependencyUpdater that will update the full versioned tags within the manifest to align with the
+    /// current product version.
+    /// </summary>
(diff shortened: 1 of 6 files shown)


MichaelSimons:
@dotnet-bot test NanoServer-1803 - 3.0 Dockerfiles please

scripts/update-dependencies/ManifestUpdater.cs
```diff
@@ -0,0 +1,62 @@
...
+        private string _tagVersion;
+
+        public ManifestUpdater(string imageVariantName, string version, string repoRoot) : base()
+        {
+            _tagVersion = version;
+
+            int firstDashIndex = version.IndexOf('-');
+            if (firstDashIndex != -1)
+            {
+                int secondDashIndex = version.IndexOf('-', firstDashIndex + 1);
```
dseefeld: When would a version have a second dash?  A comment with an example would be helpful here.

MichaelSimons: `2.2.0-rtm-35586`

Does the following comment help?

```
// Derive the Docker tag version from the product build version.
// Example: Product build version 2.2.0-rtm-35586 => Docker tag version 2.2.0.
```

dseefeld: Yes.  That's helpful.

PR #784, opened 2018-11-04
Install sudo

AArnott:
Without this, Azure Pipelines builds that use this container cannot possibly install any additional packages. They can't sudo to install them because sudo isn't in the image, and since Azure Pipelines runs build steps as non-root, `apt install` without `sudo` fails.


MichaelSimons:
I think this warrants a discussion which is best done within an issue.  @AArnott - would you mind opening an issue?

MichaelSimons:
Closing until #788 is triaged.  If approved, the change to the Dockerfile in this PR will need to be propagated to all of the SDK images.

PR #796, opened 2018-11-08
[TEST] Refactor ImageData to be more strongly typed

MichaelSimons:
Refactor the ImageData class used in the tests.
- version properties are of type `Version`
- architecture is an enum
- compute sdkOs versus defining it
- shorten member names
- reformat test data declaration to be more readable
- misc cleanup

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/Dockerfile.linux.testapp
+++ b/tests/Microsoft.DotNet.Docker.Tests/Dockerfile.linux.testapp
@@ -19,6 +19,5 @@ RUN echo '<?xml version="1.0" encoding="utf-8"?>' >> NuGet.config \
     && echo '  </packageSources>' >> NuGet.config \
     && echo '</configuration>' >> NuGet.config \
     && echo '' >> NuGet.config
-RUN cat NuGet.config
 RUN dotnet restore $optional_restore_args
 RUN dotnet build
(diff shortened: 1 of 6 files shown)


tests/Microsoft.DotNet.Docker.Tests/Dockerfile.linux.testapp
```diff
@@ -19,6 +19,5 @@ RUN echo '<?xml version="1.0" encoding="utf-8"?>' >> NuGet.config \
     && echo '  </packageSources>' >> NuGet.config \
     && echo '</configuration>' >> NuGet.config \
     && echo '' >> NuGet.config
-RUN cat NuGet.config
```
dseefeld: Why was this here and why is it being removed?

MichaelSimons: It was there for debugging purposes and shouldn't have been checked in.

tests/Microsoft.DotNet.Docker.Tests/ImageVersion.cs
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public static class ImageVersion
```
dseefeld: What are these for?

MichaelSimons: There are used when defining the ImageData instances. I thought it was cleaner to define them here than inline in ImageTests.

MichaelSimons: See https://github.com/dotnet/dotnet-docker/blob/refactor-TestData/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs#L32

Issue #742, opened 2018-10-11
Add Windows Arm32 images


MichaelSimons:
Add support for Windows Arm32.  This work should be done for 2.2

MichaelSimons:
This work is complete.  The following Windows arm32 tags are supported.

## Windows Server, version 1809 arm32 tags

- [`2.2.1-sdk-nanoserver-1809-arm32`, `2.2-sdk-nanoserver-1809-arm32`, `2.2.103-sdk`, `2.2-sdk`, `sdk`, `latest` (*Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/2.2/sdk/nanoserver-1809/arm32/Dockerfile)
- [`2.2.1-aspnetcore-runtime-nanoserver-1809-arm32`, `2.2-aspnetcore-runtime-nanoserver-1809-arm32`, `2.2.1-aspnetcore-runtime`, `2.2-aspnetcore-runtime`, `aspnetcore-runtime` (*Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/2.2/aspnetcore-runtime/nanoserver-1809/arm32/Dockerfile)
- [`2.2.1-runtime-nanoserver-1809-arm32`, `2.2-runtime-nanoserver-1809-arm32`, `2.2.1-runtime`, `2.2-runtime`, `runtime` (*Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/2.2/runtime/nanoserver-1809/arm32/Dockerfile)

PR #810, opened 2018-11-14
Adding runtime and aspnet-runtime images for Windows arm32

MichaelSimons:
Related #742

SDK builds are not yet available.  They will be added in a follow-up PR.


README.md
```diff
@@ -139,6 +139,15 @@ See the [complete set of tags](https://github.com/dotnet/dotnet-docker/blob/nigh
 
 See the [complete set of tags](https://github.com/dotnet/dotnet-docker/blob/nightly/TAGS.md).
 
+# Windows arm32 tags
```
dleeapho: Maybe it's the introduction of Windows arm32 tags that triggered this but I'm wondering if the top-level grouping of the `README.md` might be better with OS then arch: "Linux/amd64/arm32" followed by "Windows/amd64/arm32". Not suggesting we need to change it now but throwing it out there for consideration.

MichaelSimons: The current order is based on usage #s.  I think you raise a good point to reconsider.  Let's discuss this with Rich at stand-up.

PR #852, opened 2018-12-05
Refactor tests

MichaelSimons:
This is a pretty substantial change to the tests.

1. Modified the e2e scenario to use the sdk image in a standalone fashion to invoke `dotnet new` then have a _normal_ multi-stage Dockerfile that copies in an existing project, builds and publishes it.  Previously we would invoke dotnet new and dotnet build within a single Dockerfile which isn't a realistic usage.  This new patterns supports edits to the generated artifacts (e.g. csproj) prior to building which has been a challenge for us in the past.
2. Split out the single test case apart.  We were duplicating some tests before between the console and web scenarios which was a waste.  The previous pattern also was problematic as it required a separate _row_ in the test data for each web scenario.  I plan to add more test cases in subsequent changes to increase our testing coverage.


tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -33,19 +38,25 @@ public void Build(string dockerfile, string tag, string fromImage, params string
...
-        public static bool ContainerExists(string name)
-        {
-            return ResourceExists("container", $"-f \"name={name}\"");
+            ExecuteWithLogging($"build -t {tag}{targetArg}{buildArgsOption}{dockerfileArg} {contextDir}");
         }
 
-        public void DeleteContainer(string container)
+        public static bool ContainerExists(string name) => ResourceExists("container", $"-f \"name={name}\"");
+
+        public void CP(string src, string dest) => ExecuteWithLogging($"cp {src} {dest}");
```
dseefeld: (nit) Seems like this could just be named "Copy".

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -0,0 +1,224 @@
...
+                _outputHelper.WriteLine("Tests are blocked on https://github.com/dotnet/corefx/issues/33563");
+                return;
+            }
+
+            string appDir = CreateTestAppWithSdkImage(_isWeb ? "web" : "console");
+            List<string> tags = new List<string>();
+
+            try
+            {
+                if (!_imageData.HasNoSdk)
```
dseefeld: There's a lot of negatives going on here.  Change this to HasSdk and remove all of the negativity?! ;)

MichaelSimons: You're a glass half full type of guy aren't you?  lol

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -0,0 +1,224 @@
...
+            finally
+            {
+                _dockerHelper.DeleteContainer(containerName);
+            }
+        }
+
+        private async Task VerifyHttpResponseFromContainer(string containerName)
+        {
+            var retries = 30;
+            var url = !Config.IsRunningInContainer && DockerHelper.IsLinuxContainerModeEnabled
```
dseefeld: I don't understand checking for IsLinuxContinerModeEnabled here.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
MichaelSimons: Good suggestion - Changes

tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
MichaelSimons: I was originally thinking it should be aligned with the Docker command, but that is not readable as you noted and it is not consistent with the other APIs in this class (e.g. DeleteImage).  Renamed

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
MichaelSimons: Because there is a difference in behavior between windows and linux containers in addition to if you are running within a container.  You can't use `localhost` with Windows is the key difference here.

dseefeld: Maybe a comment would be helpful here.

MichaelSimons: Done

PR #860, opened 2018-12-11
Use repo ARG to reference runtime-deps images

MichaelSimons:
It is becoming increasingly difficult to ensure the correct set of changes get merged from nightly to master when we release. One thing that makes this prone to human error is that the two branches cannot be easily diffed/merged. One of the primary reasons for this is the Dockerfiles that are based on the runtime-deps images - FROM microsoft/dotnet-nightly:1.1-runtime-deps-stretch vs FROM microsoft/dotnet:1.1-runtime-deps-stretch. To eliminate this friction, we are introducing a ARG for the REPO to pull the base images from. The default value would be overridden while building the nightly branch.


MichaelSimons:
This is ready for review now.

PR #876, opened 2019-01-04
Remove LTTng Dependency from Images

brianrob:
Liblttng (and it's dependency userspacercu) are not actually required to build or run .NET Core applications.  They are only required when building the CoreCLR product.

Liblttng is an optional component that can be used alongside lttng-tools to collect tracing data from running .NET Core applications.  Removing liblttng saves some disk space (~1MB on alpine).  Given that it can't be used on its own without lttng-tools, there isn't a reason to have it built into the image.  We should remove it to help those who are striving for super-small images.

cc: @richlander

### DIFF
--- a/1.0/runtime-deps/jessie/amd64/Dockerfile
+++ b/1.0/runtime-deps/jessie/amd64/Dockerfile
@@ -10,7 +10,6 @@ RUN apt-get update \
         libgcc1 \
         libgssapi-krb5-2 \
         libicu52 \
-        liblttng-ust0 \
         libssl1.0.0 \
         libstdc++6 \
         libunwind8 \
--- a/1.1/runtime-deps/stretch/amd64/Dockerfile
+++ b/1.1/runtime-deps/stretch/amd64/Dockerfile
@@ -10,7 +10,6 @@ RUN apt-get update \
         libgcc1 \
         libgssapi-krb5-2 \
         libicu57 \
-        liblttng-ust0 \
         libssl1.0.2 \
         libstdc++6 \
         libunwind8 \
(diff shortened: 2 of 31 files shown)


MichaelSimons:
I'm not sure we should be making this change to our released images.  I consider this to be a breaking change and therefore should be done only at a major/minor version change.

The runtime-deps deb packages defined in core-setup should reflect this change as well. cc @rakeshsinghranchi

brianrob:
@MichaelSimons I can see what you're saying.  Are you proposing just modifying the 3.0 Dockerfiles?

I'd like to also cover 2.1 and 2.2 if possible - do we have separate images for 2.1.x and 2.2.x or do we just update the image and folks automatically get the latest version of 2.x?

MichaelSimons:
Yes I am proposing only modifying the 3.0 Dockerfiles.  

We have separate images for each product release e.g. 2.1.x and 2.2.x.  We only release Dockerfile changes when the product releases as to not cause churn to our users.  We have tags specific to each product version e.g. 2.2.0 and 2.2.1.  We also have floating tags e.g. 2.2.  These floating tags reference different images over time.  For example when 2.2.0 was released, the 2.2 tag referenced the 2.2.0 image.  Once 2.2.1 is released the 2.2 tag will be updated to reference the 2.2.1 image.  This allows customers  to automatically get the latest patches if desired.  If customers want to be explicit then they would use the appropriate 2.2.x tag.

Making this change during a patch release doesn't seem appropriate to me.  We could cause disruption to users of the <major>.<minor> tags which do not expect breaking changes to be made.

As a side, can you make the PR against the nightly branch.  That will make the change available immediately in the https://hub.docker.com/r/microsoft/dotnet-nightly Docker Hub repo.  It will be merged to master during the next release.

brianrob:
Understood.  Thank you for the explanation.  Yes, I will make the change to just the 3.0 images and open a new PR against nightly.

PR #877, opened 2019-01-04
Remove LTTng Dependency from Images

brianrob:
Liblttng (and it's dependency userspacercu) are not actually required to build or run .NET Core applications. They are only required when building the CoreCLR product.

Liblttng is an optional component that can be used alongside lttng-tools to collect tracing data from running .NET Core applications. Removing liblttng saves some disk space (~1MB on alpine). Given that it can't be used on its own without lttng-tools, there isn't a reason to have it built into the image. We should remove it to help those who are striving for super-small images.

cc: @richlander

### DIFF
--- a/3.0/runtime-deps/alpine3.8/amd64/Dockerfile
+++ b/3.0/runtime-deps/alpine3.8/amd64/Dockerfile
@@ -10,9 +10,7 @@ RUN apk add --no-cache \
     libssl1.0 \
     libstdc++ \
     tzdata \
-    userspace-rcu \
-    zlib \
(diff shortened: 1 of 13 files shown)


brianrob:
@MichaelSimons, this is the follow-up PR from #876.

richlander:
Nice!

brianrob:
@MichaelSimons, is there anything else that you think should be changed here, or are we ready to merge?

brianrob:
Ok, thanks much @MichaelSimons.

Issue #878, opened 2019-01-04
Nuget credential provider is built in with SDK docker images


shubham90:
## Steps to reproduce the issue
The[ artifacts credential provider ](https://github.com/Microsoft/artifacts-credprovider) should be included in SDK docker images by default. 
## Expected behavior
Dotnet sdk image download already contains credential provider. 

## Actual behavior
Today, we need to add a separate step adding cred provider

## Additional information (e.g. issue happens only occasionally)
Only in images where SDK version is > 2.1.500
I am happy to contribute if this is triaged.

wesley-pattison:
Any update on this? I was expecting with the release of 2.2 SDK image this would be resolved.

moisoiu:
@shubham90 Can you please tell me how did you manage to make it work on your implementation?

shubham90:
@moisoiu what scenario are you trying to implement? Acquisition of cred provider on dotnet docker images or use of cred provider in docker containers generally?

moisoiu:
> @moisoiu what scenario are you trying to implement? Acquisition of cred provider on dotnet docker images or use of cred provider in docker containers generally?

Hi @shubham90  I would say both, and somehow I've managed to install the microsoft cred provided but still it seems that I can't connect to my private TFS nugget packager.

I've post the docker image and nuget.config on the link below
[https://github.com/Microsoft/artifacts-credprovider/issues/63](url)

shubham90:
I have replied to you on that thread. let's use that thread to discuss your issue.

pleandre:
Hello,

Our build server doesn't have direct access to internet, and we need to use a nexus proxy to download all nuget packages, docker images, etc.

Therefore it seems difficult to execute the following script in our Dockerfile: 
https://raw.githubusercontent.com/Microsoft/artifacts-credprovider/master/helpers/installcredprovider.sh

Do you plan to include this plugin by default in the next release of dotnet core sdk docker image ?
Regards,
Pierre

MichaelSimons:
This was discussed at length and a decision was made to not include the _Azure Artifacts Credential Provider_ within the .NET Core SDK images at this time.  There is active work on improving the overall installation experience for the Cred Provider including adding a .NET Core global tool experience which will make it easier to install in a derived image.  

Some factors that went into the decision were:

* Adding the Cred Provider isn't something everyone would want.  We want to maintain a high bar for what goes into the image in order to keep the size manageable and generally appealing as a base image for everyone to use.
[...]

PR #891, opened 2019-01-09
Adding artifacts nuget credential provider to be included by default …

shubham90:
…in all sdk docker images


MichaelSimons:
I've have a couple high level comments.
1. What is the UX if you are not running as the default container user?
2. I am concerned about the unused netfx artifacts that are getting installed within the windows images.  What is the value in having the contents of plugins/netfx?
3. I the tests (scenario based) should be updated to have a smoke test of the plugin.

MichaelSimons:
> Users won't be able to access the plugin if not a default container user. They will have to place the plugin in $HOME/.nuget manually. Nuget expects the plugin to be present at very specific user location.

Is there an open issue to address this limitation?  The image fully supports running as the admin or custom defined users.  It would be ideal if these scenarios would just work "out of the box" with the credential provider.  Would it be plausible to define an env var that defines a custom plugin location.  Then we could place the plugin in C:\Users\Public.

> Nuget.exe cannot use the netcore version and has to use the netfx version. So, while netfx is not required when using dotnet restore, it is required with nuget.exe restore on windows.

@shubham90 - If you are not familiar with Nano Server, it doesn't include or support the .NET Framework.  Nano Server and .NET Core are intended go together.  Given this, I'm not making the connection on the scenario in which this is going to work.  I tried running the `CredentialProvider.Microsoft.exe` and it fails to launch (no visible output) and returns `-1073741515`

shubham90:
Fixed with the latest commit: 
1. Updated locations for cred provider in both Linux and win. 
2. Added an environment variable to control cred provider locations.
2. Removing netfx from nanoservers.
3. Fixing build for sac2016

Pending: Still working on tests. Will send out a PR shortly

shubham90:
@MichaelSimons I just discovered that the plugin fails for dotnet 3.0.100-preview sdk. Created a bug here: https://github.com/Microsoft/artifacts-credprovider/issues/51

I will add 3.0 support once this is resolved. I also discovered that the testing I am doing is basically running "dotnet restore" to force the plugin to start up and run. But, that is already included in the existing tests. Do you still want me to add that?

MichaelSimons:
shubham90, regarding the tests, the part I that I feel is not tested correctly is the positive confirmation that the cred provider is installed and hooked up correctly.  Performing a dotnet restore as is already done doesn't achieve this.  In other words if something mythically changes such that a different nuget plugin paths variable was suddenly used, a test should fail.  If something breaks where no cred provider assemblies are copied into the plugin dir, a test should fail.

shubham90:
Sounds reasonable. I will add those tests and update the PR today. Thanks for the feedback. Do you also want to add the files (tar.gz and zip) to "dotnetcli.blob.core.windows.net"? Having to download from GitHub adds another dependency.

2.1/sdk/alpine3.7/amd64/Dockerfile
```diff
@@ -22,10 +22,22 @@ RUN apk add --no-cache --virtual .build-deps \
...
+    && tar -C /usr/share/credentialprovider -xzf microsoft.nuget.credentialprovider.tar.gz plugins/netcore \
+    && rm microsoft.nuget.credentialprovider.tar.gz
+
 # Enable correct mode for dotnet watch (only mode supported in a container)
 ENV DOTNET_USE_POLLING_FILE_WATCHER=true \ 
     # Skip extraction of XML docs - generally not useful within an image/container - helps performance
-    NUGET_XMLDOC_MODE=skip
+    NUGET_XMLDOC_MODE=skip \
+    # Nuget Credential provider path
+    NUGET_PLUGIN_PATHS=/usr/share/credentialprovider/plugins/netcore/CredentialProvider.Microsoft/CredentialProvider.Microsoft.dll
```
MichaelSimons: {minor} - We have been keeping the ENVs in alphabetical order.  It helps when searching.  The list is continuing to grow over time.  Can you please put the new ENV in alphabetical order?

3.0/sdk/nanoserver-1709/amd64/Dockerfile
```diff
@@ -18,7 +18,6 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
     Expand-Archive dotnet.zip -DestinationPath dotnet; `
     Remove-Item -Force dotnet.zip
 
-
```
MichaelSimons: Please add these lines back, by convention we use a double blank line in between stages.

2.2/sdk/nanoserver-sac2016/amd64/Dockerfile
```diff
@@ -17,6 +17,20 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
...
+ENV NUGET_CREDENTIAL_PROVIDER_VERSION 0.1.8
+
+RUN Invoke-WebRequest -OutFile microsoft.nuget.credentialprovider.zip https://github.com/Microsoft/artifacts-credprovider/releases/download/$Env:NUGET_CREDENTIAL_PROVIDER_VERSION/Microsoft.NuGet.CredentialProvider.zip; `
+    $credprovider_sha512 = '853e2b0b348742520e87ea557e6c1f38306fed3544d87dc10ec2823942caaea4d41b2c3912b8c4802e7211a38707706d5e6fdd3806ebf4b42e939960c0ca5d35'; `
+    if ((Get-FileHash microsoft.nuget.credentialprovider.zip -Algorithm sha512).Hash -ne $credprovider_sha512) { `
+        Write-Host 'CHECKSUM VERIFICATION FAILED FOR CREDENTIAL PROVIDER!'; `
+        exit 1; `
+    }; `
+    `
+    Expand-Archive microsoft.nuget.credentialprovider.zip -DestinationPath C:\Users\Public\credentialprovider; `
```
MichaelSimons: I see three different patterns for extracting the zip/tarball that product two different results.  Ideally there would be consistency.

Patterns
1. Linux expands the tarball with a filter so that only the netcore plugin folder is included.  The LICENSE and ThirdPartyNotices **are not included**.
2. Some of the Windows Dockerfiles like this one expands the entire zip and then deletes only the netfx content.  The LICENSE and ThirdPartyNotices **are included**.
3. The 2.2/sdk/nanoserver-1809/amd64/Dockerfile expands the entire zip in the installer phase and then only copies the netcore plugins to the final stage.  The LICENSE and ThirdPartyNotices **are not included**.

The first question to answer it what files are required.  I would assume the LICENSE and ThirdPartyNotices are required.  

On a related question, depending on what files are included, is there a way to simplify the directory structure?  Can some of the hierarchy be collapsed?

2.1/sdk/alpine3.7/amd64/Dockerfile
shubham90: Sure, will fix.

2.1/sdk/nanoserver-1709/amd64/Dockerfile
```diff
@@ -36,7 +51,9 @@ ENV ASPNETCORE_URLS=http://+:80 `
     # Enable correct mode for dotnet watch (only mode supported in a container)
     DOTNET_USE_POLLING_FILE_WATCHER=true `
     # Skip extraction of XML docs - generally not useful within an image/container - helps performance
-    NUGET_XMLDOC_MODE=skip
+    NUGET_XMLDOC_MODE=skip `
+    # Setting environment variable for credential provider
+    NUGET_PLUGIN_PATHS="C:\Users\Public\credentialprovider\\plugins\\netcore\CredentialProvider.Microsoft\CredentialProvider.Microsoft.dll"
```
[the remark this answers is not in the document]
shubham90: I think it's a typo. Thanks for catching it. I will fix these.

3.0/sdk/nanoserver-1709/amd64/Dockerfile
shubham90: sure

2.2/sdk/nanoserver-sac2016/amd64/Dockerfile
shubham90: I am talking with the team to just release the netcore version of the plugin which contains license and relevant ThirdPartyNotices and add it to the dotnet blobstore account. This will alleviate all these additional step of deleting the netfx folder and managing the license files explicitly. This will also simplify the directory structure. Thanks for pointing this out.

2.1/sdk/alpine3.7/amd64/Dockerfile
```diff
@@ -22,8 +22,20 @@ RUN apk add --no-cache --virtual .build-deps \
...
+    && credprovider_sha512='3c4c10d7aa5f9ca005b455fb380c2bd549432df5f4352eb58d1325de5c14db7f0d59fad6eef1c1f7a076804c32fd255ee84a1ac1d790f5a83adc537b74eee7c0' \
+    && echo "$credprovider_sha512  microsoft.nuget.credentialprovider.tar.gz" | sha512sum -c - \
+    && mkdir -p /usr/share/credentialprovider \
+    && tar -C /usr/share/credentialprovider -xzf microsoft.nuget.credentialprovider.tar.gz plugins/netcore \
+    && rm microsoft.nuget.credentialprovider.tar.gz
+
 # Enable correct mode for dotnet watch (only mode supported in a container)
-ENV DOTNET_USE_POLLING_FILE_WATCHER=true \ 
+ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
+    # Nuget Credential provider path
```
zarenner: nit: odd capitalization

2.1/sdk/nanoserver-1709/amd64/Dockerfile
```diff
@@ -18,12 +18,28 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
     Expand-Archive dotnet.zip -DestinationPath dotnet; `
     Remove-Item -Force dotnet.zip
 
+# Install artifacts credential provider
+ENV NUGET_CREDENTIAL_PROVIDER_VERSION 0.1.8
+
+RUN [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; `
+    Invoke-WebRequest -OutFile microsoft.nuget.credentialprovider.zip https://github.com/Microsoft/artifacts-credprovider/releases/download/$Env:NUGET_CREDENTIAL_PROVIDER_VERSION/Microsoft.NuGet.CredentialProvider.zip; `
```
zarenner: Why is setting the security protocol necessary?

shubham90: GitHub only supports TLS 1.2 and .net default is tls 1.0. This will go away though when we move the files to some blobstore account.

2.2/sdk/nanoserver-sac2016/amd64/Dockerfile
shubham90: This PR is blocked on couple of issues:
1) PR build failing on Linux 2.1 and 2.2. Actively working and testing this.
2) Plugin's file location needs to be changed from GitHub to some blobstore account. Next in todo list.
3) Create a netcore plugin output file and remove unnecessary computation. This will be done as part of 2).
4) Non-blocker: Cred provider does not work where only sdk 3.0 is installed. We have identified the problem and fix will go in after we do 1, 2 and 3.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -158,5 +158,39 @@ public async Task VerifyAspNetCoreRuntimeImage_AppScenario(ImageData imageData)
             ImageScenarioVerifier verifier = new ImageScenarioVerifier(imageData, _dockerHelper, _outputHelper, isWeb: true);
             await verifier.Execute();
         }
+
```
MichaelSimons: I am still concerned that the tests don't positively confirm the Credential Provider is functional.  A test should fail if the plugin is not getting loaded/used by NuGet.  It is important that the rests would catch this type of regression.

shubham90: Do you mean test to verify if nuget is honoring the plugins environment variable or not?

MichaelSimons: That is part of it.  The tests should be verifying in some way that the plugin is hooked up correctly and functional.   For example, hypothetically speaking, if the environment variable were misspelled, the tests as they are now would still pass.

shubham90: That's a valid scenario if nuget had a bug around this. I can test nuget is hooked up to the plugin using "dotnet restore" by grepping some logs. For checking functionality, I can simply run the plugin dll in help mode. If that is sufficient, let me know. If not, let me know if you have some other ideas. Thanks!

2.2/sdk/nanoserver-sac2016/amd64/Dockerfile
MichaelSimons: @shubham90, what was the resolution to simplifying the directory structure?

`/usr/share/credentialprovider/plugins/netcore/CredentialProvider.Microsoft/CredentialProvider.Microsoft.dll`

Can this be simplified to `/usr/share/credentialprovider/CredentialProvider.Microsoft.dll`? What value does the plugins and netcore layers add for this scenario?

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
MichaelSimons: Are you expecting to complete the 3.0 work before merging this PR?  To make sure we are on the same page, I don't see the 3.0 work as a requirement to merging as 3.0 is still in preview.  We would want 3.0 supported for preview 3.  I do think however, the testing needs to be complete before we merge this PR as we cannot release the 2.* changes without having the appropriate tests in place to prevent regressions.

satbai: As you're probably aware, @shubham90 will be OOF until early February. I believe the plan was to not wait for the 3.0 work, hence the [bug ](https://github.com/Microsoft/artifacts-credprovider/issues/51) was created with "Need to revert [this ](https://github.com/dotnet/dotnet-docker/pull/891/commits/b60953809a4b0fd37e5d9412eb1bcd10741873a4) commit once resolved."

MichaelSimons: @satbai - I am fine with the 3.0 work coming in later.  The part I am concerned about is the test coverage in the 2.* Dockerfiles.  These are production images and therefore require the appropriate regression tests be in place as these get rebuilt whenever there is a patch to .NET or the base image is patched.

2.2/sdk/nanoserver-sac2016/amd64/Dockerfile
satbai: The folder structure is what [NuGet expects](https://github.com/NuGet/Home/wiki/NuGet-Cross-Plat-Credential-Plugin#plugin-installation-and-discovery). Since in this PR it looks like we use the environment variable NUGET_PLUGIN_PATHS the folder structure doesn't have to follow that pattern and can be changed if needed. I'm not sure if it should block this PR or if it can be done in a follow-up PR when @shubham90 is back.

shubham90: Simplifying the directory structure is coming at the cost of more processing (copying all the license files to the flat dir structure). The package itself cannot be flattened because nuget expects the dlls to be in a specific location. So, it's better to leave it as is.

shubham90:
I accidently rebased the wrong branch, fix coming soon.

shubham90:
@MichaelSimons The tests we discussed are passing now. Ended up doing everything in one command(dotnet new + add package + restore + check cache). If you feel that the test coverage is good enough now, please merge the PR. Let me know if there is something else that I missed.

2.1/sdk/nanoserver-1709/amd64/Dockerfile
```diff
@@ -18,12 +18,27 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
...
+    Expand-Archive microsoft.nuget.credentialprovider.zip -DestinationPath credentialprovider; `
+    Remove-Item -Force microsoft.nuget.credentialprovider.zip
+
 
 # SDK image
 FROM mcr.microsoft.com/windows/nanoserver:1709
 
 COPY --from=installer ["/dotnet", "/Program Files/dotnet"]
 
+COPY --from=installer ["credentialprovider", "C:\\Users\\Public\\credentialprovider"]
```
MichaelSimons: Please update the source and destination formatting to match the previous COPY instruction.  This was a recent change we made to simplify the formatting and make things consistent (e.g the source isn't rooted but the dest is).

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -157,5 +157,70 @@ public async Task VerifyAspNetCoreRuntimeImage_AppScenario(ImageData imageData)
...
+        [Theory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifySdkImage_NugetCredentialProviderIsPresent(ImageData imageData)
+        {
+            const string linuxFilePath = "/usr/share/credentialprovider/plugins/netcore/CredentialProvider.Microsoft/CredentialProvider.Microsoft.dll";
+            const string windowsFilePath = "C:\\Users\\Public\\credentialprovider\\plugins\\netcore\\CredentialProvider.Microsoft\\CredentialProvider.Microsoft.dll";
+            
+            string verifyCredProviderCommand = null;
+            if (imageData.Version.Major == 2)
```
MichaelSimons: Consider reducing the if else complexity and nesting by using the following pattern.  This can also be applied to the next test case as well.

```
            if (imageData.Version.Major != 2)
            {
                _outputHelper.WriteLine("Nuget Credential provider v2 is not supported in 1.* or 3.* images. Skipping test.");
                return;
            }

            string verifyCredProviderCommand;
            if (DockerHelper.IsLinuxContainerModeEnabled)
            {
                verifyCredProviderCommand = $"sh -c \"test -f $NUGET_PLUGIN_PATHS && test $NUGET_PLUGIN_PATHS = {linuxFilePath}\"";
            }
            else
            {
                verifyCredProviderCommand = $"CMD /S /C IF NOT EXIST %NUGET_PLUGIN_PATHS% ( exit 1 ) ELSE ( IF %NUGET_PLUGIN_PATHS% NEQ {windowsFilePath} ( exit 1 ) ELSE ( exit 0 ) )";
            }
            
            // Simple check to verify the NuGet credential provider was installed
            _dockerHelper.Run(
                image: imageData.GetImage(DotNetImageType.SDK, _dockerHelper),
                command: verifyCredProviderCommand,
                name: imageData.GetIdentifier("NugetCredentialProviderPresent"));
```

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -157,5 +157,70 @@ public async Task VerifyAspNetCoreRuntimeImage_AppScenario(ImageData imageData)
...
+        [Theory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifySdkImage_NugetCredentialProviderRuns(ImageData imageData)
+        {
+            string verifyCredProviderRestore = null;
+            if (imageData.Version.Major == 2)
+            {
+                if (DockerHelper.IsLinuxContainerModeEnabled)
+                {
+                    verifyCredProviderRestore = $"sh -c \"dotnet new console --force --framework netcoreapp{imageData.Version} && dotnet add package newtonsoft.json --no-restore && dotnet restore --no-cache && ls $HOME/.local/share/NuGet/plugins-cache | grep -q CredentialProvider.Microsoft.dll\"";
```
MichaelSimons: Is `--force` necessary in the `dotnet new` call?

MichaelSimons: What do you think about factoring `dotnet new {appType} --framework netcoreapp{_imageData.Version}` into a static helper method on ImageScenarioVerifier?  It feels like it would be beneficial to share this with the ImageScenarioVerifier logic.  This has been something that has evolved over time therefore having it shared would be good.

2.1/sdk/alpine3.7/amd64/Dockerfile
```diff
@@ -22,8 +22,20 @@ RUN apk add --no-cache --virtual .build-deps \
...
+    && echo "$credprovider_sha512  microsoft.nuget.credentialprovider.tar.gz" | sha512sum -c - \
+    && mkdir -p /usr/share/credentialprovider \
+    && tar -C /usr/share/credentialprovider -xzf microsoft.nuget.credentialprovider.tar.gz \
+    && rm microsoft.nuget.credentialprovider.tar.gz
+
 # Enable correct mode for dotnet watch (only mode supported in a container)
-ENV DOTNET_USE_POLLING_FILE_WATCHER=true \ 
+ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
+    # NuGet Credential provider path
+    NUGET_PLUGIN_PATHS=/usr/share/credentialprovider/plugins/netcore/CredentialProvider.Microsoft/CredentialProvider.Microsoft.dll \
```
MichaelSimons: @richlander - this is another case of where we need to provide guidance on installing tools in our Docker images.  While this isn't strictly a global tool, it is similar in nature.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
shubham90: It's not necessary but I wanted to be sure that we create a new app every time we run this. If this test is running on a container that is previously been used, then there will be no restore because newtonsoft.json would already be added from previous run. --force makes sure that newtonsoft.json is added to the csproj every single time.

shubham90: Good idea, I will do that.

shubham90: There are differences in how ImageScenarioVerifier is using this vs cred provider. Also, I will have to go back to use volume mount in the tests because of separate docker run commands. It was causing me some issues in windows environment. For now, I am going to leave it as is.

2.1/sdk/nanoserver-1709/amd64/Dockerfile
shubham90: done, thanks for pointing this out.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
shubham90: done, somehow my resharper didn't prompted me for this.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
MichaelSimons: All I meant was to encapsulate the command string within a method.

```
public static string GetDotnetNewCmd(string appType, string netCoreAppVersion) => return $"dotnet new {appType} --framework netcoreapp{netCoreAppVersion}";
```

MichaelSimons: IMO this is adding confusion and implying something more is going on.  All tests run in a new container always so previous state should never be an issue.

shubham90: fair enough, If we are sure that all tests run in new containers, then I can remove it.

shubham90: ah, I thought you were asking to separate out the docker run logic. I will do this right away.

2.1/sdk/alpine3.7/amd64/Dockerfile
shubham90: @richlander @MichaelSimons I would like to merge this PR if there are no blockers. We can accommodate the tool location changes whenever that gets locked. It won't be a breaking change because the location are stored in an env var.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -14,6 +14,8 @@ namespace Microsoft.DotNet.Docker.Tests
 {
     public class ImageScenarioVerifier
     {
+        public static string GetDotnetNewCmd(string appType, string netCoreAppVersion) => $"dotnet new {appType} --framework netcoreapp{netCoreAppVersion}";
```
MichaelSimons: NIT - Could you put this inline with the other methods versus at the top of the class before the fields?  Also in the c# code of this repo, we try to keep the line length less that 120 to make it easier to view in GitHub.  This is also an issue within the test cases.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -157,5 +157,62 @@ public async Task VerifyAspNetCoreRuntimeImage_AppScenario(ImageData imageData)
...
+            if (imageData.Version.Major != 2)
+            {
+                _outputHelper.WriteLine("Nuget Credential provider v2 is not supported in 1.* or 3.* images. Skipping test.");
+                return;
+            }
+            string dotnetNewCmd = ImageScenarioVerifier.GetDotnetNewCmd("console", imageData.VersionString);
+            string verifyCredProviderRestore;
+            if (DockerHelper.IsLinuxContainerModeEnabled)
+            {
+                verifyCredProviderRestore = $"sh -c \"{dotnetNewCmd} && dotnet add package newtonsoft.json --no-restore && dotnet restore --no-cache && ls $HOME/.local/share/NuGet/plugins-cache | grep -q CredentialProvider.Microsoft.dll\"";
```
MichaelSimons: Consider factoring out the duplicated code into a single place.  This would also help reduce the extremely long lines.

```
string dotnetNewCmd = ImageScenarioVerifier.GetDotnetNewCmd("console", imageData.VersionString) +
string dotnetCmds = $"{dotnetNewCmd} && dotnet add package newtonsoft.json --no-restore && dotnet restore --no-cache";
```

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -157,5 +157,62 @@ public async Task VerifyAspNetCoreRuntimeImage_AppScenario(ImageData imageData)
...
+        [Theory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifySdkImage_NugetCredentialProviderRuns(ImageData imageData)
+        {
+            if (imageData.Version.Major != 2)
+            {
+                _outputHelper.WriteLine("Nuget Credential provider v2 is not supported in 1.* or 3.* images. Skipping test.");
+                return;
+            }
+            string dotnetNewCmd = ImageScenarioVerifier.GetDotnetNewCmd("console", imageData.VersionString);
```
MichaelSimons: Consider a blank line after the if here.

2.2/sdk/nanoserver-1809/arm32/Dockerfile
```diff
@@ -10,6 +10,14 @@ RUN curl -SL --output dotnet.zip https://dotnetcli.blob.core.windows.net/dotnet/
     && tar -zxf dotnet.zip -C "%ProgramFiles%\dotnet" `
     && del dotnet.zip
 
+# Install artifacts credential provider
+ENV NUGET_CREDENTIAL_PROVIDER_VERSION 0.1.11
+
+RUN curl -SL --output microsoft.nuget.credentialprovider.zip https://azureartifactassets.blob.core.windows.net/credentialprovider/releases/download/%NUGET_CREDENTIAL_PROVIDER_VERSION%/Microsoft.NetCore2.NuGet.CredentialProvider.zip `
```
MichaelSimons: This fails to build for me.

```
Step 5/10 : RUN curl -SL --output microsoft.nuget.credentialprovider.zip https://azureartifactassets.blob.core.windows.net/credentialprovider/releases/download/%NUGET_CREDENTIAL_PROVIDER_VERSION%/Microsoft.NetCore2.NuGet.CredentialProvider.zip     && mkdir -p "C:\Users\Public\credentialprovider"     && tar -C "C:\Users\Public\credentialprovider" -xzf microsoft.nuget.credentialprovider.zip     && del microsoft.nuget.credentialprovider.zip
 ---> Running in 411da0b16e55
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current‌
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0‌
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--‌     0‌
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0‌
  1 1230k    1 15971    0     0   7985      0  0:02:37  0:00:02  0:02:35  5908‌
100 1230k  100 1230k    0     0   410k      0  0:00:03  0:00:03 --:--:-‌-  399k‌
A subdirectory or file -p already exists.‌
Error occurred while processing: -p.‌
The command &#x27;cmd &#x2F;S &#x2F;C curl -SL --output microsoft.nuget.credentialprovider.zip https:&#x2F;&#x2F;azureartifactassets.blob.core.windows.net&#x2F;credentialprovider&#x2F;releases&#x2F;download&#x2F;%NUGET_CREDENTIAL_PROVIDER_VERSION%&#x2F;Microsoft.NetCore2.NuGet.CredentialProvider.zip     &amp;&amp; mkdir -p &quot;C:\Users\Public\credentialprovider&quot;     &amp;&amp; tar -C &quot;C:\Users\Public\credentialprovider&quot; -xzf microsoft.nuget.credentialprovider.zip     &amp;&amp; del microsoft.nuget.credentialprovider.zip&#x27; returned a non-zero code: 1‌
System.InvalidOperationException: Failed to execute docker build -t dotnetdocker.azurecr.io/build-staging/dotnet-nightly/2452262/dotnet/core-nightly/sdk:2.2.104 -t dotnetdocker.azurecr.io/build-staging/dotnet-nightly/2452262/dotnet/core-nightly/sdk:2.2 -t dotnetdocker.azurecr.io/build-staging/dotnet-nightly/2452262/dotnet/core-nightly/sdk:latest -t dotnetdocker.azurecr.io/build-staging/dotnet-nightly/2452262/dotnet/core-nightly/sdk:2.2.104-nanoserver-1809-arm32 -t dotnetdocker.azurecr.io/build-staging/dotnet-nightly/2452262/dotnet/core-nightly/sdk:2.2-nanoserver-1809-arm32 -f 2.2/sdk/nanoserver-1809/arm32/Dockerfile 2.2/sdk/nanoserver-1809/arm32
   at Microsoft.DotNet.ImageBuilder.ExecuteHelper.Execute(ProcessStartInfo info, Func`2 executor, Boolean isDryRun, String errorMessage, String executeMessageOverride) in /image-builder/src/ExecuteHelper.cs:line 69
   at Microsoft.DotNet.ImageBuilder.ExecuteHelper.ExecuteWithRetry(String fileName, String args, Boolean isDryRun, String errorMessage, String executeMessageOverride) in /image-builder/src/ExecuteHelper.cs:line 46
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.BuildImages() in /image-builder/src/Commands/BuildCommand.cs:line 60
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.ExecuteAsync() in /image-builder/src/Commands/BuildCommand.cs:line 28
   at Microsoft.DotNet.ImageBuilder.ImageBuilder.Main(String[] args) in /image-builder/src/ImageBuilder.cs:line 59
```

2.1/sdk/alpine3.7/amd64/Dockerfile
MichaelSimons: We are giving a thumbs up to the locations in the PR.  Thanks for being so patient on this!

shubham90: @MichaelSimons No worries! I have fixed all the comments from your latest review. Please feel free to merge whenever you have time. Also, it would be nice if you could verify the ARM32 build once more after my change.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
shubham90: done

2.2/sdk/nanoserver-1809/arm32/Dockerfile
shubham90: It was "-p" which was causing the issue. mkdir does not support -p on windows and creates a directory named -p. Since, the dotnet step was already creating that -p dir, this step failed.

MichaelSimons:
@shubham90 - I ran a couple builds and am seeing the following intermittent test failure on Linux amd64.

[Test Results](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2457552&view=ms.vss-test-web.test-result-details)
```
/usr/share/dotnet/sdk/2.2.104/NuGet.targets(114,5): error : Could not find a part of the path '/root/.local/share/NuGet/plugins-cache/7a7ff1b21f69ab23c9f21123251bb7e17eb0433c$CredentialProvider.Microsoft.dll/https:_api.nuget.org_v3_index.json.dat-new'. [/app/app.csproj]
```

I am also see the following consistent failure on Linux arm32 

[Test Results](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2452927&view=ms.vss-test-web.test-result-details)
```
/usr/share/dotnet/sdk/2.1.504/NuGet.targets(114,5): error : A task was canceled. [/app/app.csproj]
```

shubham90:
> /usr/share/dotnet/sdk/2.1.504/NuGet.targets(114,5): error : A task was canceled. [/app/app.csproj]

shubham90:
> /usr/share/dotnet/sdk/2.1.504/NuGet.targets(114,5): error : A task was canceled. [/app/app.csproj]

I am checking this now

shubham90:
@MichaelSimons We have added a fix for one of the issues and nuget is adding additional logging in their next patch for the other issue(transient issue on Linux which we couldn't repro on our boxes). I have also added verbosity to one of the "dotnet restores" to capture anything interesting in case of failures. Could you please verify the fix on your build env? Thanks.

MichaelSimons:
Closing per [this discussion](https://github.com/dotnet/dotnet-docker/issues/878#issuecomment-513978346).

@shubham90 - I appreciate your effort on this and it did surface a number of issues which is definitely goodness - thanks.

Issue #892, opened 2019-01-11
Installing dotnet 2.1 core runtime in other Alpine images


priyaananthasankar:
## Steps to reproduce the issue
1. Have a Python 3.6 alpine base image
2. Want to install Dotnet 2.1 core runtime in it using apk command or any other way

## Expected behavior
Run dotnet commands

## Actual behavior
Cannot find any documentation or reference that shows such an example. 

## Additional information (e.g. issue happens only occasionally)
Dotnet core 2.1 runtime installations depends on debian packages. If I want to install dotnet core 2.1 runtime in a python alpine base image, it seems like not possible today?

Motivations:

- Installing python on a dotnet alpine is possible and works, but the reverse would render me a smaller image. Python installations on existing docker base files that are non python, always tend to occupy space. 

- People prefer a python base image for data science related work but when they want to run any dotnet based functions or DLL's this becomes a roadblock. 

Any suggestions appreciated to install dotnet core 2.1 in python alpine base image. (NOT looking to get a dotnet core 2.1 base image alpine and installing python on it - this exists)

MichaelSimons:
.NET Core does not currently support installation via apk packages.  If you want to install .NET Core on alpine, please reference how it is done in the [Dockerfiles](https://github.com/dotnet/dotnet-docker/blob/master/2.1/runtime/alpine3.7/amd64/Dockerfile) for .NET Core Alpine images.

MichaelSimons:
cc @dleeapho, @leecow - I'm going to move this issue to core-setup as an ask for apk support.

MichaelSimons:
@priyaananthasankar - I'm going to close this issue as I copied the ask for apk support to core-setup.  Feel free to continue any conversations specific to Docker here.

priyaananthasankar:
@MichaelSimons : so in the example you linked, I see this:
FROM $REPO:2.1-runtime-deps-alpine3.7 
Can I use a python alpine image as a base? I want to start with a python alpine image and install the dotnet core runtime in there. What method do you suggest?

kriscoleman:
All of these links are broken, how is this done?

mthalman:
> All of these links are broken, how is this done?

Since this issue was originally asking about 2.1, you can find the current 2.1 runtime-deps Dockerfile for Alpine here: https://github.com/dotnet/dotnet-docker/blob/master/2.1/runtime/alpine3.10/amd64/Dockerfile.  

[...]

PR #893, opened 2019-01-12
Sample for dotnet core 2.2 build dependencies + dotnet core compile o…

priyaananthasankar:
This is a sample dockerfile that pulls in a Python 3.6  alpine image and builds dotnet dependencies + dotnet core and tests running dotnet + Python3 from a Python file as entrypoint.

A sample Dockerfile that is based off a non dotnet alpine image is required to demonstrate:

1. Building dotnet core + deps from a Python alpine image (since apk is not supported)

This is required by folks who base off their work on open source alpine images but at some point need a dotnet runtime installed and preferably alpine for smaller sizes. This will help external customers looking for such small sized samples.

### DIFF
--- a/samples/dotnetapp/Dockerfile.python.alpine-x64
+++ b/samples/dotnetapp/Dockerfile.python.alpine-x64
@@ -1,0 +1,66 @@
+FROM microsoft/dotnet:2.2-sdk-alpine AS build
+WORKDIR /app
+
+# copy csproj and restore as distinct layers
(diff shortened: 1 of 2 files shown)


MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/issues/892.

We have not received any requests for any samples to show how to install python in a runtime image.  In an effort to keep our samples focused, we are going to hold off merging this sample at this time.

PR #913, opened 2019-01-31
Update the readmes with the move to MCR

MichaelSimons:
@dotnet-bot skip ci please.


MichaelSimons:
I waffle on whether or not the product page readme should call out `Featured Tags` I decided against it as I just couldn't find a clear way to present them.  I landed as it is better done in the concrete repos.  Any feedback is appreciated.

MichaelSimons:
cc @rotata

MichaelSimons:
Merging to update broken links.  Will make follow on PR for any feedback provided.

Issue #938, opened 2019-02-22
[Infrastructure] nano-server-1809 build legs started failing when trying to login to Docker.


MichaelSimons:
Here is the official build logic which first surfaced this issue - https://github.com/dotnet/dotnet-docker/blob/nightly/.vsts-pipelines/jobs/test-images-windows-client.yml#L27 - and here is the corresponding build failure - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2421536

I contacted AzDO folks on this and they believe this is an issue with PS.  They are following up on the underlying issue.

MichaelSimons:
The blocking issue was addressed.  We still need to investigate if the builds should be using a credential provider.

MichaelSimons:
Credential warning is being tracked with https://github.com/dotnet/docker-tools/issues/187.  

Closing as this issue has been addressed.

PR #939, opened 2019-02-22
Fix regression in docker login within build test legs

MichaelSimons:
This is a workaround for #938.  We will need a better long term solution that preserves the retry logic.

While investigating this, I realized docker logout was not being called correctly so I corrected that issue as well.

@dotnet-bot skip ci please

### DIFF
--- a/.vsts-pipelines/jobs/test-images-windows-client.yml
+++ b/.vsts-pipelines/jobs/test-images-windows-client.yml
@@ -24,10 +24,11 @@ jobs:
   - template: ../steps/init-docker-windows.yml
     parameters:
       setupImageBuilder: false
-  - powershell: >
-      ./scripts/Invoke-WithRetry.ps1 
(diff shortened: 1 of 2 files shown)


.vsts-pipelines/jobs/test-images-windows-client.yml
```diff
@@ -24,10 +24,11 @@ jobs:
   - template: ../steps/init-docker-windows.yml
     parameters:
       setupImageBuilder: false
-  - powershell: >
-      ./scripts/Invoke-WithRetry.ps1 
-      "docker login -u $(acr.userName) -p $(BotAccount-dotnet-docker-acr-bot-password) $(acr.server)"
+  - task: CmdLine@1
```
mthalman: Wouldn't it be simpler to just use script here instead of CmdLine@1 since script will just execute cmd.exe anyway?
- script: docker login -u $(acr.userName) -p $(BotAccount-dotnet-docker-acr-bot-password) $(acr.server)

MichaelSimons: Yes - changed.

PR #944, opened 2019-02-25
Fix Invoke-ImageBuilder for Windows scenario

mthalman:
In the Windows scenario, a container may not be created when Invoke-ImageBuilder.ps1 is invoked.  So it should not attempt to remove a container in the `finally` block.  Fixed this by adding a flag which is set when the script has created the container and checking the flag before removing the container.


MichaelSimons:
@dotnet-bot test please

MichaelSimons:
@dotnet-bot test this please

PR #948, opened 2019-02-27
Add dockerignore file for projects being tested

mthalman:
When projects are generated from the SDK to be tested, they automatically have an obj folder created for them with some artifact files in it.  These files end up getting copied when the project files are copied in when building the Docker image for that project.  But the file in the obj folder are completely unnecessary for the image.

These changes dynamically add a .dockerignore file to the project that gets created which prevents the obj folder contents from being copied to the image.


tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/.dockerignore
```diff
@@ -0,0 +1 @@
+**/obj
```
MichaelSimons: Thoughts on adding `bin` as well to make it a more canonical .dockerignore?

mthalman: Might as well.

Issue #360, opened 2018-01-22
Please support PowerShell core on SDK images


AceHack:
Please add powershell core on both linux and windows SDK images

AceHack:
Please also make this the default cmd handler in the image

MichaelSimons:
We would like to hear more requests from the community on this before making a change.  Making the Powershell the default `SHELL` would not be something we would consider as we feel it is not what the Linux community would expect.

AceHack:
Would it be possible to create an alternative tag/image that has PowerShell as the default?  This would make it so I would not have to create my own base image for all my apps that does the same.  Thanks.

MichaelSimons:
If there is a strong demand for such images then it would be something we would consider.  In general we try to limit the number of images permutations as deciphering the differences between all of the images can be difficult.

AceHack:
Just to give a little context, I use PowerShell so I can create one Dockerfile that works on both Windows and Linux containers even though in many cases the builds are complicated and require scripting.  Before I had to maintain two different Dockerfiles and script the same actions in two different languages.  Thanks.

richlander:
Assume we did add powershell to our SDK images. Why would you need the default shell to be powershell in the dotnet images to enable your scenario?

AceHack:
Because on Windows the executable name is powershell and on Linux it’s pwsh.  So at a minimum I would have to have two base images that set the default she'll accordingly and then the multiarch tag.  Then all my services would need to reference the multiarch tag.

richlander:
I can see how that would be nice, but it's well within your capability to solve. It's not so challenging that it's worth us creating a whole series of additional images to update. We'd have to create ~20 new images to make that work.

AceHack:
If you add powershell to the base images for the sdk but don't make it the default could you run the following

#!/bin/bash
ln -s /usr/bin/pwsh /usr/bin/powershell

This is alias pwsh as powershell and I should be able to set the default myself in a way that would work on both OSs in a single docker file.

Thanks.

richlander:
That may be a middle ground.

First, we need more requests to add powershell. This request is the first one I am aware of.

itz1508:
Same request to add powershell :)
________________________________

yuhuazhong:
I second @AceHack on supporting powershell on Linux. It just enable a lot of possibility for development. It also help us make a stronger business case to adopt container technology in our company.

kz1000fan:
[...]

PR #960, opened 2019-03-05
Add PowerShell global to docker-sdk images

adityapatwardhan:
Add PowerShell global tool as part of the SDK image.

* Download the tool from the fwlink
* Validate checksum where possible
* Install tool
* Clean up dowloaded package
* Setup symbolic on Linux or add to PATH on Windows.

Related to #360

Our release pipeline tests installation of global tool and executing some very basic commands on Windows and Linux.

### DIFF
--- a/3.0/sdk/alpine3.9/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.9/amd64/Dockerfile
@@ -30,10 +30,12 @@ RUN dotnet help
 # Install PowerShell global tool
 ENV POWERSHELL_VERSION 6.2.0-rc.1
 
-RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg 'https://go.microsoft.com/fwlink/?linkid=2080856' \
+RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg "https://pwshtool.blob.core.windows.net/v6-2-0-rc-1/PowerShell.Linux.$POWERSHELL_VERSION.nupkg" \
     && powershell_sha512='0bb710ed0ffba267c6d82253abae5427a771861d6b50c3574dfe77d746f8d78d309397abbd1ff300d5b6b70a9d794943836822158d7de6095b61df3fba490c08' \
     && echo "$powershell_sha512  PowerShell.Linux.$POWERSHELL_VERSION.nupkg" | sha512sum -c - \
     && mkdir -p /usr/share/powershell \
     && dotnet tool install --add-source / --tool-path /usr/share/powershell --version $POWERSHELL_VERSION PowerShell.Linux \
     && rm PowerShell.Linux.$POWERSHELL_VERSION.nupkg \
-    && ln -s /usr/share/powershell/pwsh /usr/bin/pwsh
+    && ln -s /usr/share/powershell/pwsh /usr/bin/pwsh \
+    # Add ncurses-terminfo-base to resolve psreadline dependency
+    && apk add --no-cache ncurses-terminfo-base
--- a/3.0/sdk/alpine3.9/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.9/amd64/Dockerfile
@@ -30,6 +30,7 @@ RUN dotnet help
 # Install PowerShell global tool
 ENV POWERSHELL_VERSION 6.2.0-rc.1
 
+# The global tool package PowerShell.Linux only supports linux-x64, linux-musl-x64, unix, linux, linux-arm64, osx runtimes.
 RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg "https://pwshtool.blob.core.windows.net/v6-2-0-rc-1/PowerShell.Linux.$POWERSHELL_VERSION.nupkg" \
     && powershell_sha512='0bb710ed0ffba267c6d82253abae5427a771861d6b50c3574dfe77d746f8d78d309397abbd1ff300d5b6b70a9d794943836822158d7de6095b61df3fba490c08' \
     && echo "$powershell_sha512  PowerShell.Linux.$POWERSHELL_VERSION.nupkg" | sha512sum -c - \
--- a/3.0/sdk/alpine3.9/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.9/amd64/Dockerfile
@@ -40,10 +40,5 @@ RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg https://pwshtool.blob.cor
     && ln -s /usr/share/powershell/pwsh /usr/bin/pwsh \
     # To reduce image size, remove the copy nupkg that nuget keeps.
     && find /usr/share/powershell -print | grep -i '.*[.]nupkg$' | xargs rm \
-    # Remove unnecessary runtimes
-    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/linux-arm \
-    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/linux-arm64 \
-    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/linux-x64 \
-    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/osx \
     # Add ncurses-terminfo-base to resolve psreadline dependency
     && apk add --no-cache ncurses-terminfo-base
--- a/3.0/sdk/alpine3.9/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.9/amd64/Dockerfile
@@ -20,9 +20,20 @@ RUN wget -O dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DO
     && rm dotnet.tar.gz
 
 # Enable correct mode for dotnet watch (only mode supported in a container)
-ENV DOTNET_USE_POLLING_FILE_WATCHER=true \ 
+ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
     # Skip extraction of XML docs - generally not useful within an image/container - helps performance
     NUGET_XMLDOC_MODE=skip
 
 # Trigger first run experience by running arbitrary cmd
 RUN dotnet help
+
+# Install PowerShell global tool
+ENV POWERSHELL_VERSION 6.2.0-rc.1
+
+RUN curl -SL --output PowerShell.Linux.$POWERSHELL_VERSION.nupkg 'https://go.microsoft.com/fwlink/?linkid=2080856' \
+    && powershell_sha512='0bb710ed0ffba267c6d82253abae5427a771861d6b50c3574dfe77d746f8d78d309397abbd1ff300d5b6b70a9d794943836822158d7de6095b61df3fba490c08' \
+    && echo "$powershell_sha512  PowerShell.Linux.$POWERSHELL_VERSION.nupkg" | sha512sum -c - \
+    && mkdir -p /usr/share/powershell \
+    && dotnet tool install --add-source / --tool-path /usr/share/powershell --version $POWERSHELL_VERSION PowerShell.Linux \
+    && rm PowerShell.Linux.$POWERSHELL_VERSION.nupkg \
+    && ln -s /usr/share/powershell/pwsh /usr/bin/pwsh
--- a/3.0/sdk/alpine3.9/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.9/amd64/Dockerfile
@@ -31,7 +31,7 @@ RUN dotnet help
 ENV POWERSHELL_VERSION 6.2.0-rc.1
 
 # The global tool package PowerShell.Linux only supports linux-x64, linux-musl-x64, unix, linux, linux-arm64, osx runtimes.
-RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg https://pwshtool.blob.core.windows.net/v6-2-0-rc-1/PowerShell.Linux.$POWERSHELL_VERSION.nupkg \
+RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg https://pwshtool.blob.core.windows.net/$POWERSHELL_VERSION/PowerShell.Linux.$POWERSHELL_VERSION.nupkg \
     && powershell_sha512='0bb710ed0ffba267c6d82253abae5427a771861d6b50c3574dfe77d746f8d78d309397abbd1ff300d5b6b70a9d794943836822158d7de6095b61df3fba490c08' \
     && echo "$powershell_sha512  PowerShell.Linux.$POWERSHELL_VERSION.nupkg" | sha512sum -c - \
     && mkdir -p /usr/share/powershell \
--- a/3.0/sdk/alpine3.9/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.9/amd64/Dockerfile
@@ -31,7 +31,7 @@ RUN dotnet help
 ENV POWERSHELL_VERSION 6.2.0-rc.1
 
 # The global tool package PowerShell.Linux only supports linux-x64, linux-musl-x64, unix, linux, linux-arm64, osx runtimes.
-RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg "https://pwshtool.blob.core.windows.net/v6-2-0-rc-1/PowerShell.Linux.$POWERSHELL_VERSION.nupkg" \
+RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg https://pwshtool.blob.core.windows.net/v6-2-0-rc-1/PowerShell.Linux.$POWERSHELL_VERSION.nupkg \
     && powershell_sha512='0bb710ed0ffba267c6d82253abae5427a771861d6b50c3574dfe77d746f8d78d309397abbd1ff300d5b6b70a9d794943836822158d7de6095b61df3fba490c08' \
     && echo "$powershell_sha512  PowerShell.Linux.$POWERSHELL_VERSION.nupkg" | sha512sum -c - \
     && mkdir -p /usr/share/powershell \
--- a/3.0/sdk/alpine3.9/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.9/amd64/Dockerfile
@@ -28,11 +28,11 @@ ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
 RUN dotnet help
 
 # Install PowerShell global tool
-ENV POWERSHELL_VERSION 6.2.0-rc.1
+ENV POWERSHELL_VERSION 6.2.0-rc.5
 
 # The global tool package PowerShell.Linux only supports linux-x64, linux-musl-x64, unix, linux, linux-arm64, osx runtimes.
 RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg https://pwshtool.blob.core.windows.net/tool/$POWERSHELL_VERSION/PowerShell.Linux.$POWERSHELL_VERSION.nupkg \
-    && powershell_sha512='0bb710ed0ffba267c6d82253abae5427a771861d6b50c3574dfe77d746f8d78d309397abbd1ff300d5b6b70a9d794943836822158d7de6095b61df3fba490c08' \
+    && powershell_sha512='6ac43ce637f5b098fda79f621a7f1ccbb5288422a40a8603a55ee5b2dbddad704fac12b2fb6f8729432769a0077e0caad8c2bb0cba9ce5f49eda077d7b91be9f' \
     && echo "$powershell_sha512  PowerShell.Linux.$POWERSHELL_VERSION.nupkg" | sha512sum -c - \
     && mkdir -p /usr/share/powershell \
     && dotnet tool install --add-source / --tool-path /usr/share/powershell --version $POWERSHELL_VERSION PowerShell.Linux \
@@ -40,5 +40,10 @@ RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg https://pwshtool.blob.cor
     && ln -s /usr/share/powershell/pwsh /usr/bin/pwsh \
     # To reduce image size, remove the copy nupkg that nuget keeps.
     && find /usr/share/powershell -print | grep -i '.*[.]nupkg$' | xargs rm \
+    # Remove unnecessary runtimes
+    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/linux-arm \
+    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/linux-arm64 \
+    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/linux-x64 \
+    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/osx \
     # Add ncurses-terminfo-base to resolve psreadline dependency
     && apk add --no-cache ncurses-terminfo-base
--- a/3.0/sdk/alpine3.9/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.9/amd64/Dockerfile
@@ -30,7 +30,6 @@ RUN dotnet help
 # Install PowerShell global tool
 ENV POWERSHELL_VERSION 6.2.0
 
-# The global tool package PowerShell.Linux.Alpine only supports linux-musl-x64, unix, linux runtimes.
 RUN wget -O PowerShell.Linux.Alpine.$POWERSHELL_VERSION.nupkg https://pwshtool.blob.core.windows.net/tool/$POWERSHELL_VERSION/PowerShell.Linux.Alpine.$POWERSHELL_VERSION.nupkg \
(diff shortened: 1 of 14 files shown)


adityapatwardhan:
@MichaelSimons  @richlander 

I have a couple of questions:

1. Should I make `pwsh` the default `CMD`?
1. In most cases `RUN dotnet help` is not needed anymore since I use `dotnet` to install the tool. Should I remove it?

MichaelSimons:
@adityapatwardhan - Thanks for making this PR.  To answer your questions

1.  I don't think we should be doing this.
2. I'm on the fence on this.  Ultimately I think the CLI should have a way to explicitly trigger the first run experience.  If this existed I would say definitely no.  So given the intent, I lean towards leaving it.

Regarding your changes, can you describe the type of automated tests you plan on adding for PS?

3.0/sdk/alpine3.9/amd64/Dockerfile
```diff
@@ -20,9 +20,20 @@ RUN wget -O dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DO
...
     # Skip extraction of XML docs - generally not useful within an image/container - helps performance
     NUGET_XMLDOC_MODE=skip
 
 # Trigger first run experience by running arbitrary cmd
 RUN dotnet help
+
+# Install PowerShell global tool
+ENV POWERSHELL_VERSION 6.2.0-rc.1
+
+RUN curl -SL --output PowerShell.Linux.$POWERSHELL_VERSION.nupkg 'https://go.microsoft.com/fwlink/?linkid=2080856' \
```
MichaelSimons: Is there a reason why the nupkg can't be pulled directly from NuGet?

adityapatwardhan: The `PowerShell.Linux` and `PowerShell.Windows` are special packages with reduced size that were created for this purpose. We do not plan to ship these on `NuGet`. We will be shipping a package called `PowerShell` on `NuGet` which is larger since it has `Windows` and `Linux` assemblies and a shim layer.

3.0/sdk/nanoserver-1809/amd64/Dockerfile
```diff
@@ -18,15 +18,29 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
...
+ENV POWERSHELL_VERSION 6.2.0-rc.1
+
+RUN Invoke-WebRequest -OutFile PowerShell.Windows.$POWERSHELL_VERSION.nupkg 'https://go.microsoft.com/fwlink/?linkid=2081065'; `
+    $powershell_sha512 = '404b81c7d1cdd79556db560eaf1ab46c80d7d0d006e91a1fa3eccd00032a0415b63c6c3c2a6f386881b9de7b62ca4dc46f11ab3d941690564578961b947716ce'; `
+    if ((Get-FileHash PowerShell.Windows.$POWERSHELL_VERSION.nupkg -Algorithm sha512).Hash -ne $powershell_sha512) { `
+        Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
+        exit 1; `
+    }; `
+    `
+    \dotnet\dotnet tool install --add-source . --tool-path '\powershell' --version $POWERSHELL_VERSION PowerShell.Windows; `
```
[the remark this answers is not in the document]
adityapatwardhan: There should be no undesirable side effects on installing the global tool. Should I move it out of the installer stage and move it down?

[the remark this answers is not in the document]
adityapatwardhan: @MichaelSimons If i move things around a bit, I can have powershell show up as a global tool? Is that the intent? I made some changes to nanoserver-1709 to protype this:

```
Microsoft Windows [Version 10.0.16299.967]
(c) 2017 Microsoft Corporation. All rights reserved.

C:\>dotnet tool list -g
Package Id              Version         Commands
------------------------------------------------
powershell.windows      6.2.0-rc.1      pwsh

C:\>pwsh
PowerShell 6.2.0-rc.1
Copyright (c) Microsoft Corporation. All rights reserved.

https://aka.ms/pscore6-docs
Type 'help' to get help.

PS C:\> exit
```

Let me know if this is was is required, i will make those changes.

3.0/sdk/nanoserver-1809/amd64/Dockerfile
```diff
@@ -18,15 +18,29 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
     Expand-Archive dotnet.zip -DestinationPath dotnet; `
     Remove-Item -Force dotnet.zip
 
+# Install PowerShell global tool
+ENV POWERSHELL_VERSION 6.2.0-rc.1
+
+RUN Invoke-WebRequest -OutFile PowerShell.Windows.$POWERSHELL_VERSION.nupkg 'https://go.microsoft.com/fwlink/?linkid=2081065'; `
```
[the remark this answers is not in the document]
adityapatwardhan: Fixed this.

3.0/sdk/alpine3.9/amd64/Dockerfile
MichaelSimons: Our Dockerfiles are more than just the source for the Docker images we produce.  They often get referenced as the way to do things by our customers.  I am pretty certain you will see this pattern copied and referenced as the way to install PS as a dotnet tool.  This feels like a potential source of confusion.  Would it make sense to have a differentiated name for these packages and publish them to NuGet?

adityapatwardhan: These packages support a small set of runtimes tailored towards platforms supported by dotnet-docker images. This was done to reduce the size of the package.
Hence, even a `PowerShell.Linux` package would not work on all supported Linux platforms and would create confusion as a global tool. The tool we plan to publish on NuGet has all the runtimes we support. 
Would adding a comment explaining this and mentioning the package is internal and not for public consumption directly be sufficient?

richlander: I have less issue with this pattern. I think it is fine for us/anyone to publish specialized to other NuGet feeds. My issue is the fwlink. I'm not at all comfortable with that. There are at least three problems with it:

* It is not OSS-friendly. Users have no idea what that points to (beyond traversing the link).
* It is not good for versioning. Ideally, I'd like to see each Dockerfile point to a specific version of PowerShell, just like we do with .NET Core. This means that you can go back and re-build an old Dockerfile and get the same result. That old Dockerfile might result in images with security vulnerabilities, but that's actually a good thing in this case.
* It's not good for engineering but it is unclear from looking at this source what the version coupling is between .NET Core and PowerShell.

In short, fwlinks are opaque.

We have a tool called DARC that you can likely use to help with this. It enables you to publish LKG versions to a database, as part of your build, that this repo can subscribe and then versions flow automatically w/o you doing anything. That would be my recommendation.

adityapatwardhan: I have updated the URLs to a publicly accessible blob.

adityapatwardhan: @MichaelSimons Should I add a comment saying the packages are not for public consumption?

MichaelSimons: They are for public consumption by including them in our open source Dockerfiles.  An expected use case is that there will be a set of users that will copy our Dockerfiles and build their own images for various reasons.  What may be useful to clarify is what is different about this NuGet versus your general purpose NuGet package.

MichaelSimons:
@adityapatwardhan - Can you answer this question I asked earlier?

> can you describe the type of automated tests you plan on adding for PS?

In this repo we mostly test at a scenario level.  It seems like it should be easy to execute a simple pwsh command to smoke test PS is installed and functional.

3.0/sdk/alpine3.9/amd64/Dockerfile
adityapatwardhan: Agreed. I will add a comment explaing the runtimes it supports and also mention that there is global tool published on NuGet.org which supports all runtimes.

adityapatwardhan:
@MichaelSimons  We have automated tests at the end of the build to install the tool on Linux and Windows. Also run some simple validation to verify version and run a couple of cmdlets.

MichaelSimons:
@adityapatwardhan - yes but we need to verify pwsh works within the scope of these Docker images.  Is pwsh on the path, does the shared runtime and pwsh work together, are the required native dependencies present, etc.  These are the things that should be validated.  One scenario test should be able to cover this.

adityapatwardhan:
@MichaelSimons  Should I add similar test to validate PowerShell here: 

https://github.com/dotnet/dotnet-docker/blob/1e84193426671c8ad957087d21a4ff002e1c6b4e/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs#L148-L158

MichaelSimons:
@adityapatwardhan, yes.  Here is an outline of what comes to mind.

```
        [Theory]
        public void VerifySdkImage_PowerShell(ImageData imageData)
        {
        {
            if (imageData.Version.Major < 3)
            {
                _outputHelper.WriteLine("PowerShell does not exist in pre-3.0 images, skip testing");
                return;
            }

            // TODO:  Modify DockerHelper.Run method to return output by `return ExecuteWithLogging...`
            string output = _dockerHelper.Run(
                image: _imageData.GetImage(DotNetImageType.SDK, _dockerHelper),
                name: _imageData.GetIdentifier($"pwsh"),
                command: $"pwsh -c xyz"); // TODO: Replace xyz with cmd
            
            // TODO: Validate output against expected?
        }
```

adityapatwardhan:
@MichaelSimons I have added the test. Please have a look. The CI seems stuck from last night or is the queue pretty long?

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -156,5 +156,26 @@ public async Task VerifyAspNetCoreRuntimeImage_AppScenario(ImageData imageData)
             ImageScenarioVerifier verifier = new ImageScenarioVerifier(imageData, _dockerHelper, _outputHelper, isWeb: true);
             await verifier.Execute();
         }
+
+        [Theory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifySDKImage_PowerShellScenario(ImageData imageData)
```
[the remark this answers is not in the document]
adityapatwardhan: Fixed.

3.0/sdk/alpine3.9/amd64/Dockerfile
```diff
@@ -20,9 +20,23 @@ RUN wget -O dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DO
...
     NUGET_XMLDOC_MODE=skip
 
 # Trigger first run experience by running arbitrary cmd
 RUN dotnet help
+
+# Install PowerShell global tool
+ENV POWERSHELL_VERSION 6.2.0-rc.1
+
+# The global tool package PowerShell.Linux only supports linux-x64, linux-musl-x64, unix, linux, linux-arm64, osx runtimes.
+RUN wget -O PowerShell.Linux.$POWERSHELL_VERSION.nupkg "https://pwshtool.blob.core.windows.net/v6-2-0-rc-1/PowerShell.Linux.$POWERSHELL_VERSION.nupkg" \
```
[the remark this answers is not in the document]
adityapatwardhan: Unfortunately, Azure blobs do not support having `.` in the container names. 
[Container Naming Rules](https://docs.microsoft.com/en-us/rest/api/storageservices/Naming-and-Referencing-Containers--Blobs--and-Metadata#container-names)

3.0/sdk/stretch/arm32v7/Dockerfile
```diff
@@ -34,3 +34,15 @@ ENV ASPNETCORE_URLS=http://+:80 \
 
 # Trigger first run experience by running arbitrary cmd
 RUN dotnet help
+
+# Install PowerShell global tool
+ENV POWERSHELL_VERSION 6.2.0-rc.1
+
+# The global tool package PowerShell.Linux only supports linux-x64, linux-musl-x64, unix, linux, linux-arm64, osx runtimes.
```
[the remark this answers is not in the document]
adityapatwardhan: It is supposed to work. I will validate this.

3.0/sdk/alpine3.9/amd64/Dockerfile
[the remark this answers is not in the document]
MichaelSimons: Could you have a common container that contains all versions?  e.g. Could you use a pattern that follows what dotnet uses - https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$Env:DOTNET_SDK_VERSION/dotnet-sdk-$Env:DOTNET_SDK_VERSION-win-x64.zip?

adityapatwardhan:
@MichaelSimons  There is definately an error in package creation. The Linux package should not have Windows runtimes and vice versa. We have a `6.2.0` package coming out by end of month. Can this be fixed then?

MichaelSimons:
@adityapatwardhan - yes the size issue can be addressed at the release.  I would however like to have a good idea of what the layer size will be when addressed because the current size is going to be a difficult sell.

adityapatwardhan:
@MichaelSimons I should be able to create a private package fixing the packaging error by Monday. We can have a look then.

3.0/sdk/alpine3.9/amd64/Dockerfile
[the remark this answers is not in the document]
adityapatwardhan: That might work. Let me give that a try.

3.0/sdk/stretch/arm32v7/Dockerfile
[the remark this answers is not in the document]
adityapatwardhan: There seems to be some issue when tool gets installed. If directly try to call `dotnet pwsh.dll` from the tool install location it works fine.

```
root@82e2bd4ffbe3:/usr/share/powershell/.store/powershell.linux/6.2.0-rc.1/powershell.linux/6.2.0-rc.1/tools/netcoreapp2.1/any# dotnet ./pwsh.dll
PowerShell 6.2.0-rc.1
Copyright (c) Microsoft Corporation. All rights reserved.

https://aka.ms/pscore6-docs
Type 'help' to get help.

PS /usr/share/powershell/.store/powershell.linux/6.2.0-rc.1/powershell.linux/6.2.0-rc.1/tools/netcoreapp2.1/any> uname -a
Linux 82e2bd4ffbe3 4.14.71-v7+ #1145 SMP Fri Sep 21 15:38:35 BST 2018 armv7l armv7l armv7l GNU/Linux
```

@MichaelSimons  Any ideas?

adityapatwardhan:
@MichaelSimons I can potentially delete the copy of the nupkg that nuget keeps. That will save somwhere around 20 MBs in the size.

3.0/sdk/nanoserver-1803/amd64/Dockerfile
```diff
@@ -18,15 +18,30 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
...
+
+# The global tool package PowerShell.Windows only supports win10-x64, win-arm, win-x64 runtimes.
+RUN Invoke-WebRequest -OutFile "PowerShell.Windows.$ENV:POWERSHELL_VERSION.nupkg" "https://pwshtool.blob.core.windows.net/v6-2-0-rc-1/PowerShell.Windows.$ENV:POWERSHELL_VERSION.nupkg"; `
+    $powershell_sha512 = '404b81c7d1cdd79556db560eaf1ab46c80d7d0d006e91a1fa3eccd00032a0415b63c6c3c2a6f386881b9de7b62ca4dc46f11ab3d941690564578961b947716ce'; `
+    if ((Get-FileHash "PowerShell.Windows.$ENV:POWERSHELL_VERSION.nupkg" -Algorithm sha512).Hash -ne $powershell_sha512) { `
+        Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
+        exit 1; `
+    }; `
+    `
+    \dotnet\dotnet tool install --add-source . --tool-path '\powershell' --version "$ENV:POWERSHELL_VERSION" PowerShell.Windows; `
```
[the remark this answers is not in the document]
adityapatwardhan: Fixed.

3.0/sdk/stretch/arm32v7/Dockerfile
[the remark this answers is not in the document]
MichaelSimons: This looks to be an issue to `dotnet tools`.  I logged an issue for it - https://github.com/dotnet/cli/issues/10985

MichaelSimons:
@adityapatwardhan - yes I was thinking the nupkg copy should be cleaned up.

3.0/sdk/alpine3.9/amd64/Dockerfile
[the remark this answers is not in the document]
adityapatwardhan: Resolved.

adityapatwardhan:
@MichaelSimons 

I built the alpine 3.9 container locally to check the layer sizes. The reduction in the size will be part of our next release expected near end of month.

*Layer size before reductions:*

| IMAGE | CREATED | CREATED BY | SIZE  |
|----------|-----------|---------------|------|
| 245a2bf6f1b6 | 20 hours ago |  /bin/sh -c wget -O PowerShell.Linux.$POWERSH…   | 104MB |

*After reductions:*

| IMAGE | CREATED | CREATED BY | SIZE  |
|----------|-----------|---------------|------|
| 0a7610b47997 | About a minute ago |  /bin/sh -c wget -O PowerShell.Linux.$POWERSH…   | 35.5MB |

adityapatwardhan:
@MichaelSimons I have addressed all the review comments. Please have another look. I have also put some layer size information above. It is from a private of the next release of PowerShell `v6.2.0`, which contains further reduction in size.

3.0/sdk/alpine3.9/amd64/Dockerfile
```diff
@@ -28,15 +28,22 @@ ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
...
+    && powershell_sha512='6ac43ce637f5b098fda79f621a7f1ccbb5288422a40a8603a55ee5b2dbddad704fac12b2fb6f8729432769a0077e0caad8c2bb0cba9ce5f49eda077d7b91be9f' \
     && echo "$powershell_sha512  PowerShell.Linux.$POWERSHELL_VERSION.nupkg" | sha512sum -c - \
     && mkdir -p /usr/share/powershell \
     && dotnet tool install --add-source / --tool-path /usr/share/powershell --version $POWERSHELL_VERSION PowerShell.Linux \
     && rm PowerShell.Linux.$POWERSHELL_VERSION.nupkg \
     && ln -s /usr/share/powershell/pwsh /usr/bin/pwsh \
+    # To reduce image size, remove the copy nupkg that nuget keeps.
+    && find /usr/share/powershell -print | grep -i '.*[.]nupkg$' | xargs rm \
+    # Remove unnecessary runtimes
+    && rm -fr /usr/share/powershell/.store/powershell.linux/$POWERSHELL_VERSION/powershell.linux/$POWERSHELL_VERSION/tools/netcoreapp2.1/any/runtimes/linux-arm \
```
MichaelSimons: I thought the plan was to create rid specific packages.  These packages aren't rid specific if they contain more than one runtime.

adityapatwardhan: The end result is the same. I just makes your builds easier and faster. Are having rid specific packages a must?

adityapatwardhan: Even for `linux-x64`, we have some dependencies coming from `unix` and some from `linux`.

Assemblies from `unix`:
System.Drawing.Common.dll
System.Runtime.Caching.dll
System.Data.SqlClient.dll
Microsoft.Management.Infrastructure.dll
Microsoft.Management.Infrastructure.Native.dll
System.Private.ServiceModel.dll

Assemblies from `linux`:
System.Data.Odbc.dll

MichaelSimons: The value is the simplicity of the Dockerfile.  The cleanliness makes it easy for the end user to see what the image contains.  Having logic here to know what runtime folders are required is not desirable.  rid specific packages are going to provide the best experience here.

adityapatwardhan: @MichaelSimons  I have created RID specific packages now, which will be available with the next release. I will remove the code for deleting runtime folders from the docker files. In my next PR for updating the version, I will also update the package names, is that ok?

`PowerShell.Linux.Alpine.$POWERSHELL_VERSION.nupkg`
`PowerShell.Linux.arm32.$POWERSHELL_VERSION.nupkg`
`PowerShell.Linux.arm64.$POWERSHELL_VERSIONnupkg`
`PowerShell.Linux.x64.$POWERSHELL_VERSION.nupkg`

`PowerShell.Windows.arm32.$ENV:POWERSHELL_VERSION.nupkg`
`PowerShell.Windows.x64.$ENV:POWERSHELL_VERSION.nupkg`

MichaelSimons: Yes, looks great.  We can't merge however until the CLI issue is address as that is a blocker.

adityapatwardhan: Yes of course.

MichaelSimons:
@adityapatwardhan - it looks as thought he underlying .NET CLI issue was resolved with the latest bits.  I pulled in the latest 3.0 build.  Can you update this PR?  FYI - the nanoserver-1709 images were recently removed (https://github.com/dotnet/dotnet-docker/pull/1023).

adityapatwardhan:
I am currently updating the branch for changing version to `6.2.0`. We released today. I will make the change for nanoserver 1709.

adityapatwardhan:
@MichaelSimons PR updated for PowerShell `v6.2.0` and updated to use the rid specific packages. Also rebased to remove nanoserver-1709. Please have a look.

3.0/sdk/nanoserver-1809/arm32v7/Dockerfile
```diff
@@ -10,9 +10,16 @@ RUN curl -SL --output dotnet.zip https://dotnetcli.blob.core.windows.net/dotnet/
     && tar -zxf dotnet.zip -C "%ProgramFiles%\dotnet" `
     && del dotnet.zip
 
+# The global tool package PowerShell.Windows.arm32 only supports win-arm runtimes.
```
MichaelSimons: This Dockerfile is missing the following:

```
ENV POWERSHELL_VERSION 6.2.0
```

3.0/sdk/nanoserver-1803/amd64/Dockerfile
```diff
@@ -18,15 +18,31 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
     Expand-Archive dotnet.zip -DestinationPath dotnet; `
     Remove-Item -Force dotnet.zip
 
+# Install PowerShell global tool
+ENV POWERSHELL_VERSION 6.2.0
+
+# The global tool package PowerShell.Windows.x64 only supports win10-x64, win-x64 runtimes.
```
MichaelSimons: Now that there are true rid specific packages, I feel like this comment is no longer necessary.  Thoughts on removing it?

3.0/sdk/nanoserver-1809/arm32v7/Dockerfile
adityapatwardhan: Good catch.

adityapatwardhan: Added.

3.0/sdk/nanoserver-1803/amd64/Dockerfile
adityapatwardhan: Removed.

adityapatwardhan:
@MichaelSimons Updated. Please have another look.

adityapatwardhan:
@MichaelSimons an unrelated test is failing. There were no changes in 2.1-2.2 docker files.

MichaelSimons:
@adityapatwardhan - I am still seeing a failure while building the Windows arm32 image.

```
C:\Program Files\dotnet\sdk\3.0.100-preview4-010965\NuGet.targets(119,5): error : End of Central Directory record could not be found. [C:\Windows\TEMP\zcrrhos2.z54\restore.csproj]
The tool package could not be restored.‌
Tool 'powershell.windows.arm32' failed to install. This failure may have been caused by:

* You are attempting to install a preview release and did not use the --version option to specify the version.
* A package by this name was found, but it was not a .NET Core tool.
* The required NuGet feed cannot be accessed, perhaps because of an Internet connection problem.
* You mistyped the name of the tool.‌
The command &#x27;cmd &#x2F;S &#x2F;C curl -SL --output PowerShell.Windows.arm32.%POWERSHELL_VERSION%.nupkg https:&#x2F;&#x2F;pwshtool.blob.core.windows.net&#x2F;tool&#x2F;$ENV:POWERSHELL_VERSION&#x2F;PowerShell.Windows.arm32.$ENV:POWERSHELL_VERSION.nupkg     &amp;&amp; mkdir &quot;%ProgramFiles%\powershell&quot;     &amp;&amp; &quot;%ProgramFiles%\dotnet\dotnet&quot; tool install --add-source . --tool-path &quot;%ProgramFiles%\powershell&quot; --version %POWERSHELL_VERSION% PowerShell.Windows.arm32     &amp;&amp; del PowerShell.Windows.arm32.%POWERSHELL_VERSION%.nupkg     &amp;&amp; del &quot;%ProgramFiles%\powershell\.store\powershell.windows.arm32.\%POWERSHELL_VERSION%\powershell.windows.arm32\%POWERSHELL_VERSION%\powershell.windows.arm32.%POWERSHELL_VERSION%.nupkg&quot;&#x27; returned a non-zero code: 1‌
```

3.0/sdk/nanoserver-1809/arm32v7/Dockerfile
```diff
@@ -10,9 +10,18 @@ RUN curl -SL --output dotnet.zip https://dotnetcli.blob.core.windows.net/dotnet/
     && tar -zxf dotnet.zip -C "%ProgramFiles%\dotnet" `
     && del dotnet.zip
 
+# Install PowerShell global tool
+ENV POWERSHELL_VERSION 6.2.0
+
+RUN curl -SL --output PowerShell.Windows.arm32.%POWERSHELL_VERSION%.nupkg https://pwshtool.blob.core.windows.net/tool/$ENV:POWERSHELL_VERSION/PowerShell.Windows.arm32.$ENV:POWERSHELL_VERSION.nupkg `
```
MichaelSimons:
```suggestion
RUN curl -SL --output PowerShell.Windows.arm32.%POWERSHELL_VERSION%.nupkg https://pwshtool.blob.core.windows.net/tool/%POWERSHELL_VERSION%/PowerShell.Windows.arm32.%POWERSHELL_VERSION%.nupkg `
```

adityapatwardhan: I believe this should fix the arm32 build failure.

MichaelSimons:
I updated my previous comment - I meant that we are only on-boarding Alpine arm64 images.

PR #965, opened 2019-03-05
Add 2.1 & 2.2 Alpine 3.9 images

### LINKED ISSUE #921: Add Alpine 3.9 images
Alpine 3.9 has been released and the microsoft/dotnet alpine images should be updated to based on it.

The following is the proposal.
1. 3.0 will only have 3.9 images of Alpine.
2. Add 3.9 Alpine images for 2.1 and 2.2.  The floating `alpine` tag will be updated to reference 3.9.  Keep the older alpine images for a period (3 months) before being deprecated.

Related issues:
https://github.com/dotnet/corefx/issues/33870
https://github.com/dotnet/corefx/pull/34789

MichaelSimons:
Fixes #921.


manifest.json
```diff
@@ -112,9 +112,41 @@
...
+                },
+                "2.2-alpine3.9": {
+                  "documentationGroup": "2.2"
+                },
+                "$(2.2-RuntimeVersion)-alpine": {
+                  "documentationGroup": "2.2"
+                },
+                "2.2-alpine": {
+                  "documentationGroup": "2.2"
+                }
```
mthalman: Are these references to 2.2 supposed to be here?

MichaelSimons: yes - The runtime-deps Dockerfiles didn't change between 2.1 and 2.2.  We share the Dockerfiles and images between the two version however we dual tag the images with both 2.1. and 2.2 tags.  This provides a nice user experience in that it makes it obvious which tags users should use regardless if they are building a 2.1 or 2.2 app.  The `documentationGroup` is some image-builder special sauce used as part of the infra to generate the tags readme.  The one downside to this pattern is that astute users may be confused by the 2.2 Dockerfile link in the readme leads them to a 2.1 Dockerfile.  The primary upside/driver behind doing this is for layer sharing.

PR #1006, opened 2019-03-21
Rewrite Dockerfile for arm32 Linux samples

mthalman:
The samples build is failing for arm32 Linux leg because the build agents have an architecture of aarch64 and the multi-arch manifest description doesn't have a matching architecture for the 2.2 SDK and runtime images that we reference.  To fix this, we eventually want to make use of the `--platform` option on the Docker CLI to explicitly set what platform we want to get but that's only available in experimental mode.  To workaround this, we are going to modify the Dockerfile dynamically so it uses the platform-specific tag for the image instead of the 2.2 flag.

### DIFF
--- a/.vsts-pipelines/jobs/build-images.yml
+++ b/.vsts-pipelines/jobs/build-images.yml
@@ -20,12 +20,12 @@ jobs:
   variables:
     osVersion: ${{ parameters.osVersion }}
   steps:
-  - ${{ if eq(parameters.name, 'Build_Linux_arm32v7') }}:
+  - ${{ if and(contains(variables.manifest, 'samples'), eq(parameters.name, 'Build_Linux_arm32v7')) }}:
     - script: >
         job="$(Agent.JobName)" &&
         sampleFolder=$(echo "$job" | grep -o -E "aspnetapp|dotnetapp") &&
         dockerfilePath="samples/$sampleFolder/Dockerfile" &&
-        sed 's/:2.2 AS/:2.2-bionic-arm32v7 AS/g' $dockerfilePath > Dockerfile.tmp &&
+        sed 's/:2.2 AS/:2.2-stretch-slim-arm32v7 AS/g' $dockerfilePath > Dockerfile.tmp &&
         mv Dockerfile.tmp $dockerfilePath
       displayName: Update Sample Dockerfile
   - template: ${{ format('../steps/init-docker-{0}.yml', parameters.dockerClientOS) }}
--- a/.vsts-pipelines/jobs/build-images.yml
+++ b/.vsts-pipelines/jobs/build-images.yml
@@ -20,6 +20,14 @@ jobs:
   variables:
     osVersion: ${{ parameters.osVersion }}
   steps:
+  - ${{ if eq(parameters.name, 'Build_Linux_arm32v7') }}:
+    - script: >
+        job="$(Agent.JobName)" &&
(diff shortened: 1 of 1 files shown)


.vsts-pipelines/jobs/build-images.yml
```diff
@@ -20,6 +20,14 @@ jobs:
   variables:
     osVersion: ${{ parameters.osVersion }}
   steps:
+  - ${{ if eq(parameters.name, 'Build_Linux_arm32v7') }}:
+    - script: >
+        job="$(Agent.JobName)" &&
+        sampleFolder=$(echo "$job" | grep -o -E "aspnetapp|dotnetapp") &&
+        dockerfilePath="samples/$sampleFolder/Dockerfile" &&
+        sed 's/:2.2 AS/:2.2-bionic-arm32v7 AS/g' $dockerfilePath > Dockerfile.tmp &&
```
MichaelSimons: This should using the `2.2-stretch-slim-arm32v7` tag as that is what is included in the multi-arch tag previously used.
```suggestion
        sed 's/:2.2 AS/2.2-stretch-slim-arm32v7 AS/g' $dockerfilePath > Dockerfile.tmp &&
```

MichaelSimons: Is the tmp file necessary here?

.vsts-pipelines/jobs/build-images.yml
```diff
@@ -20,6 +20,14 @@ jobs:
   variables:
     osVersion: ${{ parameters.osVersion }}
   steps:
+  - ${{ if eq(parameters.name, 'Build_Linux_arm32v7') }}:
```
MichaelSimons: I understand this condition is sufficient but do you think adding a samples check would help the readability here - contains(variables['manifest'], 'samples')

.vsts-pipelines/jobs/build-images.yml
mthalman: If you try to simply output it to same file as the input file, you end up with an empty file.  After reading up on it, this is the most platform-agnostic way to do this.  There is an option you can pass to sed to handle outputting back to the same file but it's not supported on all flavors.

.vsts-pipelines/jobs/build-images.yml
mthalman: In order to do this and have changes isolated, I've had to define this as a condition instead.  We can't reference custom variables in template expressions.

.vsts-pipelines/jobs/build-images.yml
```diff
@@ -20,6 +20,20 @@ jobs:
...
+      sed 's/:2.2 AS build/:2.2-stretch-arm32v7 AS build/g' $dockerfilePath > Dockerfile.tmp &&
+      mv Dockerfile.tmp $dockerfilePath &&
+      sed 's/:2.2 AS runtime/:2.2-stretch-slim-arm32v7 AS runtime/g' $dockerfilePath > Dockerfile.tmp &&
+      mv Dockerfile.tmp $dockerfilePath
+    displayName: Update Sample Dockerfile
+    condition: "
+      and(
+        succeeded(),
+        contains(variables['manifest'], 'samples'),
+        eq('${{ parameters.name }}', 'Build_Linux_arm32v7'))"
```
MichaelSimons: Consider a comment explaining why this is needed.

> Workaround for building x-platform samples on an aarch64 machine.  Docker doesn't have support to specify a target platform to build for (https://github.com/moby/moby/pull/37350).

PR #1009, opened 2019-03-22
Add version monitor build pipeline

MichaelSimons:
This is a port of a modeled build that exists today to monitor the Docker version dependencies this repo depends on so that automatic builds will get triggered as base images are updated.


mthalman:
I'm assuming that the original designer-based version of the pipeline should never have been configured to run two distinct jobs since one job is sufficient?

MichaelSimons:
The old build definition had two jobs because at the time it was created the, image builder didn't have an option for --architecture='*'.  This was added as part of the MCR changes.

PR #1014, opened 2019-03-26
Reference new version of image builder

mthalman:
Reference the newest version of ImageBuilder which has renamed the matrix types when generating a matrix.  As part of these changes, I've factored out the identifiers for those matrix types into parameters such that they map to the particular jobs that are being execute (i.e. build, test, publish).  The matrix values are then referenced dynamically by referencing the appropriate parameter.  This was done in order to better position things for better support for the PR build refactoring that will require the ability to dynamically select which matrix type will be used for the build jobs.  I would have preferred using variables defined in common.yml for defining these matrix type mappings instead of parameters but it wasn't possible due to limitations in AzDO with being able to use expressions on variables.

### DIFF
--- a/.vsts-pipelines/jobs/build-test-publish-repo.yml
+++ b/.vsts-pipelines/jobs/build-test-publish-repo.yml
@@ -15,12 +15,15 @@ jobs:
   - template: ../steps/generate-matrix.yml
     parameters:
       matrixType: ${{ parameters.buildMatrixType }}
+      name: buildMatrix
   - template: ../steps/generate-matrix.yml
     parameters:
       matrixType: ${{ parameters.testMatrixType }}
+      name: testMatrix
   - template: ../steps/generate-matrix.yml
     parameters:
       matrixType: ${{ parameters.publishMatrixType }}
+      name: publishMatrix
   - template: ../steps/cleanup-docker-linux.yml
 
   ################################################################################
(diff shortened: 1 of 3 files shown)


.vsts-pipelines/jobs/build-test-publish-repo.yml
```diff
@@ -7,11 +12,29 @@ jobs:
     name: Hosted Ubuntu 1604
   steps:
   - template: ../steps/init-docker-linux.yml
-  - script: $(runImageBuilderCmd) generateBuildMatrix --manifest $(manifest) --type build --os-type '*' --architecture '*' $(imageBuilder.generateMatrixQueueArgs)
+  - script: >
+      $(runImageBuilderCmd) generateBuildMatrix
```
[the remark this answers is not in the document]
mthalman: Yeah, that makes sense.  I'll do that.

.vsts-pipelines/steps/generate-matrix.yml
```diff
@@ -0,0 +1,13 @@
...
+steps:
+  - script: >
+      $(runImageBuilderCmd) generateBuildMatrix
+      --manifest $(manifest)
+      --type ${{ parameters.matrixType }}
+      --os-type '*'
+      --architecture '*'
+      $(imageBuilder.generateMatrixQueueArgs)
+    displayName: Generate ${{ parameters.matrixType }} Matrix
+    name: ${{ parameters.matrixType }}
```
MichaelSimons: I am somewhat in favor of leaving this with a name on the intended usage.  To me, it makes the build output easier to understand in that you don't have to make the mental translation.

MichaelSimons: Missing Line Feed

mthalman:
@dotnet-bot skip ci please

PR #1032, opened 2019-04-03
Refactor pwsh test to reduce fragility

### LINKED ISSUE #1045: Update PowerShell test to not rely on Get-Date
The VerifySDKImage_PowerShellScenario test attempts to verify that PowerShell works in the container by calling Get-Date and comparing the output value to the current date from the test code.  There can be a difference in time between when the value is calculated in the container versus when the value is calculated in test code.  If the time difference crosses over midnight, then the two values will correspond to different dates and fail the test.  Another possibility is that the container and the test runner are running in different time zones and also cross the midnight time boundary.

The test should be updated to make use of a PowerShell command that is not date/time-dependent.

MichaelSimons:
Fixes #1045

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
@@ -142,15 +142,14 @@ namespace Microsoft.DotNet.Docker.Tests
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


MichaelSimons:
cc @adityapatwardhan

PR #1046, opened 2019-04-10
Refactor pipeline for PR builds that can build/test in same job

mthalman:
Adds a build pipeline for validating PR builds in the nightly branch.

This required substantial refactoring of our pipeline jobs since they were designed to split apart building and testing into separate jobs.  For PR builds, we can't do this because it would require pushing images to the private ACR which we don't want to do.

The test steps were factored out of a job template and into a steps template that allows them to be referenced by either a build or test job, depending on whether it's a PR build or official build.  This also requires a different type of build matrix in order to produce the appropriate legs.  For PR builds, we use the platformVersionedOs matrix instead for building the images instead of platformDependencyGraph.  This allows all the appropriate images to be available in one job and perform testing on them.

And since we do not have ARM machines for public builds, we're also skipping ARM legs in the PR build pipeline.


.vsts-pipelines/steps/test-images-windows-client.yml
```diff
@@ -0,0 +1,44 @@
...
+- powershell: |
+    $optionalTestArgs=""
+    if ("${{ parameters.isPullRequestBuild }}" -eq "false") {
+      $optionalTestArgs="-Registry $env:ACR_SERVER -RepoPrefix $env:STAGINGREPOPREFIX"
+    } else {
+      $optionalTestArgs="-IsLocalRun"
+    }
+    echo "##vso[task.setvariable variable=optionalTestArgs]$optionalTestArgs"
+  displayName: Set Test Variables
+- ${{ if eq(parameters.cleanupDocker, 'true') }}:
```
MichaelSimons: I'm not seeing this as a declared parameter.

.vsts-pipelines/steps/test-images-windows-client.yml
```diff
@@ -0,0 +1,44 @@
...
+- ${{ if eq(parameters.isPullRequestBuild, 'false') }}:
+  - script: docker login -u $(acr.userName) -p $(BotAccount-dotnet-docker-acr-bot-password) $(acr.server)
+    displayName: Docker login
+- powershell: Get-ChildItem -Path tests -r | Where {$_.Extension -match "trx"} | Remove-Item
+  displayName: Cleanup Old Test Results
+- powershell: >
+    ./tests/run-tests.ps1
+    -VersionFilter $(dotnetVersion)*
+    -OSFilter $(osVariant)
+    $(optionalTestArgs)
```
MichaelSimons: Curious if you can do an `- ${{ if eq(parameters.isPullRequestBuild, 'false') }}:` condition right here and eliminate the dynamic variable?

.vsts-pipelines/steps/test-images-windows-client.yml
```diff
@@ -0,0 +1,44 @@
+parameters:
+  isPullRequestBuild: false
+steps:
+- powershell: |
+    $optionalTestArgs=""
```
MichaelSimons: Is this necessary?

.vsts-pipelines/steps/test-images-linux-client.yml
```diff
@@ -0,0 +1,78 @@
+parameters:
```
MichaelSimons: Can you explain the meaning of the 'client' suffix in the template name?

.vsts-pipelines/jobs/build-test-publish-repo.yml
```diff
@@ -16,15 +18,19 @@ jobs:
...
-      matrixType: ${{ parameters.publishMatrixType }}
-      name: publishMatrix
-  - template: ../steps/cleanup-docker-linux.yml
+      customBuildLegGrouping: ${{ parameters.customBuildLegGrouping }}
+  - ${{ if eq(parameters.isPullRequestBuild, 'false') }}:  
+    - template: ../steps/generate-matrix.yml
+      parameters:
+        matrixType: ${{ parameters.testMatrixType }}
+        name: testMatrix
+        customBuildLegGrouping: ${{ parameters.customBuildLegGrouping }}
```
MichaelSimons: Does the `customBuildLegGrouping` do anything when generating the `platformVersionedOS` matrix?  I feel like it might be best to just specify the `customBuildLegGrouping` for the `buildMatrix` path since that is the supported/intended usage.  Specifying it elsewhere is just adding complexity albeit minor for paths that aren't used.

.vsts-pipelines/builds/dotnet-core-nightly-pr.yml
```diff
@@ -0,0 +1,14 @@
...
+trigger: none
+
+variables:
+- template: ../variables/common-pr.yml
+- name: manifest
+  value: manifest.json
+jobs:
+- template: ../jobs/build-test-publish-repo.yml
+  parameters:
+    isPullRequestBuild: true
```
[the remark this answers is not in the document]
mthalman: I would have loved to do that but it can't be done with how AzDO pipelines currently work.  The isPullRequestBuild value needs to be referenced for conditional expressions.  This can't be done when it's defined as a variable.

.vsts-pipelines/steps/test-images-windows-client.yml
mthalman: Nope, I had tried that originally.  AzDO doesn't like it:
`The directive 'if' is not allowed in this context. Directives are not supported for expressions that are embedded within a string. Directives are only supported when the entire value is an expression.`

.vsts-pipelines/builds/dotnet-core-nightly-pr.yml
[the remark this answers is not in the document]
MichaelSimons: Sure - makes sense

.vsts-pipelines/steps/test-images-linux-client.yml
mthalman: I just copied these names from the original job files which had 'client' in the name.

.vsts-pipelines/jobs/build-test-publish-repo.yml
mthalman: `customBuildLegGrouping` applies to any of the matrix types.  It's just a general purpose mechanism for customizing the matrix.  You're correct in that it's only being used for the build matrix but I was just being thorough and applying it everywhere since it technically can be used in all cases.  I'm open to changing it.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -86,7 +86,7 @@ private static string GetFilterRegexPattern(string filterEnvName)
             return filter != null ? $"^{Regex.Escape(filter).Replace(@"\*", ".*").Replace(@"\?", ".")}$" : null;
         }
 
-        [Theory]
+        //[Theory]
```
MichaelSimons: Are the changes to this file intentional?

.vsts-pipelines/jobs/build-images.yml
```diff
@@ -21,14 +21,15 @@ jobs:
   variables:
     osVersion: ${{ parameters.osVersion }}
     ${{ if eq(parameters.isPullRequestBuild, 'false') }}:
-      imageBuilderBuildArgs: --registry-override $(acr.server)
+      imageBuilderBuildArgs:
```
MichaelSimons: I was expecting a `>` here.  How does this look in the build output?  Is it one line?

```
folded_style: >
    This entire block of text will be the value of 'folded_style', but this
    time, all newlines will be replaced with a single space.

    Blank lines, like above, are converted to a newline character.

        'More-indented' lines keep their newlines, too -
        this text will appear over two lines.
```

.vsts-pipelines/steps/test-images-windows-client.yml
```diff
@@ -2,15 +2,14 @@ parameters:
   isPullRequestBuild: false
 steps:
 - powershell: |
-    $optionalTestArgs=""
```
MichaelSimons: Can this change (removing `$optionalTestArgs=""`)  be made to the linux file as well?

mthalman: No, it's needed in the Linux file because the logic will append to the existing value of optionalTestArgs so it needs to be set to a default value to start out with.  In the Windows case, there's no appending going on so defaulting it isn't necessary.

.vsts-pipelines/jobs/build-images.yml
mthalman: So it looks like things run as they should but it doesn't output it very nicely in the log.  I've fixed it.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
mthalman: Whoops, nope.

.vsts-pipelines/jobs/build-test-publish-repo.yml
```diff
@@ -45,24 +45,33 @@ jobs:
   parameters:
     name: Build_NanoServer1803_amd64
     pool: # windows1803Amd64Pool
-      name: DotNetCore-Docker
+      ${{ if eq(variables['System.TeamProject'], 'public') }}:
```
MichaelSimons: From a readability perspective, it is easier to read the following IMO.  Thoughts on always using the `eq` operator with the target project everywhere?

```
        name: DotNetCore-Docker-Public
      ${{ if eq(variables['System.TeamProject'], 'internal') }}:
        name: DotNetCore-Docker
```

.vsts-pipelines/jobs/build-test-publish-repo.yml
```diff
@@ -111,7 +126,10 @@ jobs:
       name: Test_Linux_arm64v8
       buildDependencies: Build_Linux_arm64v8
       pool: # linuxArm64v8Pool
-        name: DotNetCore-Docker
+        ${{ if eq(variables['System.TeamProject'], 'public') }}:
```
MichaelSimons: Are these conditional pools necessary within the `Test Images` section?  It seems like a dead code path because of the outer `if`

.vsts-pipelines/variables/common.yml
```diff
@@ -1,5 +1,5 @@
 variables:
 - template: docker-images.yml
-- ${{ if ne(variables['System.TeamProject'], 'internal') }}:
+- ${{ if ne(variables['System.TeamProject'], 'public') }}:
   - group: DotNet-Docker-Common
-  - group: DotNet-Docker-Secrets
+  - group: DotNet-Docker-Secrets
```
MichaelSimons: Missing line feed

.vsts-pipelines/steps/test-images-windows-client.yml
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+      $optionalTestArgs="-IsLocalRun"
+    }
+    echo "##vso[task.setvariable variable=optionalTestArgs]$optionalTestArgs"
+  displayName: Set Test Variables
+- ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+  - template: init-docker-windows.yml
+    parameters:
+      setupImageBuilder: false
+- ${{ if eq(variables['System.TeamProject'], 'internal') }}:
```
MichaelSimons: Is this `if` necessary?  Can the login get combined into the body of the previous if?

.vsts-pipelines/jobs/build-test-publish-repo.yml
MichaelSimons: Thinking about the future, I think we should only specify the `customBuildLetGrouping` for the `buildMatrix` (e.g. the scenario it is required for).  It will simplify the code and make it easier to change as we move forward.  We know the publishMatrixType is going to change in order to reduce the number of legs.

.vsts-pipelines/builds/dotnet-core.yml
```diff
@@ -1,8 +1,14 @@
...
 variables:
 - template: ../variables/common.yml
 - name: manifest
   value: manifest.json
 jobs:
 - template: ../jobs/build-test-publish-repo.yml
+  parameters:
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      buildMatrixType: platformVersionedOs
+      customBuildLegGrouping: pr-build
```
MichaelSimons: Missing line feed.

.vsts-pipelines/builds/dotnet-core.yml
```diff
@@ -1,8 +1,14 @@
 trigger: none
-pr: none
+pr:
+- nightly
```
MichaelSimons: Should this be master?

mthalman: doh

Issue #1054, opened 2019-04-12
're-exec error: exit status -1073741819' when building Windows arm32 Dockerfile


MichaelSimons:
This issue started happening with the latest 3.0 builds 

## Steps to reproduce the issue
1. Build this Dockerfile (.NET Core SDK) - https://github.com/MichaelSimons/dotnet-docker/blob/3204d3bc5e488140b63b0e2edc46953f1030cc3d/3.0/sdk/nanoserver-1809/arm32v7/Dockerfile
1. Build the following Dockerfile based on the image produced from step 1

```
FROM dotnet/core-nightly/sdk:3.0-nanoserver-1809-arm32v7

RUN dotnet new web --framework netcoreapp3.0 --no-restore
```

**Results:**

PS D:\temp\3.0test\webApp131993925978252812> docker build -t 3.0-build-131993931078657512 --no-cache -f .\Dockerfile.repo .
Sending build context to Docker daemon  14.34kB
Step 1/2 : FROM dotnetdocker.azurecr.io/build-staging/150340/dotnet/core-nightly/sdk:3.0-nanoserver-1809-arm32v7 as new
---> 87e897465a90
Step 2/2 : RUN dotnet new web --framework netcoreapp3.0 --no-restore
---> Running in 796bbd8a51c1
Getting ready...
The template "ASP.NET Core Empty" was created successfully.
re-exec error: exit status -1073741819: output:

**Notes:**
Running the Dockerfile instruction interactively within a container works as expected.

MichaelSimons:
This issue appears to happen when a new container layer does not create any registry hive.  The underlying Windows issue is being tracked by https://microsoft.visualstudio.com/OS/_workitems/edit/21168329 (internal Microsoft link)

MichaelSimons:
One workaround that will trigger the creation of the registry hive is to switch users.

```
FROM dotnet/core-nightly/sdk:3.0-nanoserver-1809-arm32v7

USER ContainerAdministrator

RUN dotnet new web --framework netcoreapp3.0 --no-restore
```

MichaelSimons:
Closing since support for Windows ARM is being dropped (#1840).

PR #1055, opened 2019-04-12
Update 3.0 Windows Arm32 images and Add test workaround for Docker issue

MichaelSimons:
These changes include a test workaround for issue #1054.  Once the underlying Windows/Docker issue gets resolved, the workaround can be removed.


tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -131,8 +131,14 @@ private string CreateTestAppWithSdkImage(string appType)
 
                 ApplyProjectCustomizations(_imageData, Path.Combine(appDir, "app.csproj"));
 
+                string sourceDockerfileName = $"Dockerfile.{DockerHelper.DockerOS.ToLower()}";
```
mthalman: Add a comment here referencing the logged issue and a TODO to remove it when a fix is available.

PR #1068, opened 2019-04-18
Enable PR validation for Arm32 images

### LINKED ISSUE #373: Build and test arm images as part of CI
@MichaelSimons commented on [Fri Jan 12 2018](https://github.com/dotnet/dotnet-docker-nightly/issues/520)

Currently the arm images aren't getting built and tested as part of CI.  They are only built and tested in official builds.  The reason CI is not in place is because the necessary machines don't exist in CI.

MichaelSimons:
Fixes #373 

The 2.2 Windows Arm leg tags ~90 minutes.  Getting this checked in will allow us to more easily iterate on trying out changes to improve this time in addition to getting better hardware (hummingboards).

### DIFF
--- a/.vsts-pipelines/jobs/build-images.yml
+++ b/.vsts-pipelines/jobs/build-images.yml
@@ -18,6 +18,8 @@ jobs:
   pool: ${{ parameters.pool }}
   strategy:
     matrix: $[ ${{ parameters.matrix }} ]
+  ${{ if and(eq(variables['System.TeamProject'], 'public'), eq(parameters.useRemoteDockerServer, 'true')) }}:
+    timeoutInMinutes: 120
   variables:
     osVersion: ${{ parameters.osVersion }}
     ${{ if eq(variables['System.TeamProject'], 'internal') }}:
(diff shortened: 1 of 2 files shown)


.vsts-pipelines/jobs/build-images.yml
```diff
@@ -18,6 +18,8 @@ jobs:
   pool: ${{ parameters.pool }}
   strategy:
     matrix: $[ ${{ parameters.matrix }} ]
+  ${{ if and(eq(variables['System.TeamProject'], 'public'), eq(parameters.useRemoteDockerServer, 'true')) }}:
```
mthalman: Just to keep things simple, would it be better to just remove the condition and have the timeout always set to this value regardless of the TeamProject?  Do we really want the build to fail if it happens to go beyond 2 hrs on the internal build?

MichaelSimons: The reason I went with this approach is that from time to time docker hangs occur and I don't want jobs to run for 2 hours before timing out.

mthalman: Ok, sounds reasonable.

Issue #1081, opened 2019-04-22
Add Ubuntu 19.04 images for 3.0


MichaelSimons:
1. This only applies to 3.0 and will not be back ported to 2.x
2. Once 19.10 is released, we will add 19.10 images.  19.04 images will be supported for 3 more months before being dropped.  The same thing will happen once 20.04 (LTS) is released - 20.04 images will be added and the 19.10 images will be supported for 3 more months before being dropped.
3. Support will remain for Bionic (18.04) throughout the lifetime of .NET Core 3.0.

PR #1083, opened 2019-04-22
Add support for Ubuntu Disco (19.04) in 3.0

MichaelSimons:
Related to #1081


3.0/aspnet/disco/amd64/Dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG REPO=mcr.microsoft.com/dotnet/core/runtime
+FROM $REPO:3.0-disco
+
+# Install .NET Core
```
mthalman: .NET -> ASP.NET

MichaelSimons: Fixed here and three other places.

PR #1090, opened 2019-04-24
Add pipeline for updating .NET Core dependencies from drop location

mthalman:
Extracts the runtime, SDK, and ASP.NET versions from the SDK.zip file and passes those values to the update dependencies tool.  The channel to target is defined as a build argument.

### DIFF
--- a/.vsts-pipelines/builds/update-dependencies-from-drop.yml
+++ b/.vsts-pipelines/builds/update-dependencies-from-drop.yml
@@ -1,0 +1,43 @@
+trigger: none
+pr: none
+variables:
+- group: DotNet-Maestro
+
+jobs:
+- job: UpdateDependencies
+  pool: Hosted Ubuntu 1604
+  steps:
+  - script: |
+      artifactDir=$BUILD_STAGINGDIRECTORY
+      curl -SLo $artifactDir/sdk.zip https://dotnetcli.blob.core.windows.net/dotnet/Sdk/release/$(channel)/dotnet-sdk-latest-win-x64.zip
+      unzip $artifactDir/sdk.zip -d $artifactDir/sdk
+
+      cd $artifactDir/sdk/sdk
+      sdkVersionsPath=$(find . -name .version)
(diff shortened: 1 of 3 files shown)


.vsts-pipelines/builds/update-dependencies-from-drop.yml
```diff
@@ -0,0 +1,43 @@
...
+pr: none
+variables:
+- group: DotNet-Maestro
+
+jobs:
+- job: UpdateDependencies
+  pool: Hosted Ubuntu 1604
+  steps:
+  - script: |
+      artifactDir=$BUILD_STAGINGDIRECTORY
```
MichaelSimons: I feel like there is value in encapsulating this as a standalone script (checked into the scripts folder).  It would certainly make it easier to test as a standalone component in the future.

.vsts-pipelines/builds/update-dependencies-from-drop.yml
```diff
@@ -0,0 +1,43 @@
...
+  - script: >
+      docker run --rm -v /var/run/docker.sock:/var/run/docker.sock update-dependencies
+      --user dotnet-maestro-bot
+      --email dotnet-maestro-bot@microsoft.com
+      --password $(BotAccount-dotnet-maestro-bot-PAT)
+      --runtime-version $(runtimeVer)
+      --sdk-version $(sdkVer)
+      --aspnet-version $(aspnetVer)
+    displayName: Run Update Dependencies
+  - script: docker system prune -a -f --volumes
```
MichaelSimons: Thoughts on removing this in both definitions since we are using hosted?

.vsts-pipelines/builds/update-dependencies-from-drop.yml
mthalman: done

.vsts-pipelines/builds/update-dependencies-from-drop.yml
mthalman: agreed

Issue #1080, opened 2019-04-22
Publish build steps get skipped when building a subset of images


MichaelSimons:
The publish build steps get skipped whenever a build is queued that does not build the entire set of images.  e.g. imageBuilder.queueArgs = --path '3.0'.  This is because the publish build steps have a dependency on the tests succeeding.  If not all of the images are built, the corresponding tests are skipped therefore causing publish to be skipped.  To workaround this issue, a second publish only build must be run to publish the build images.  It would be convenient and less error prone if one build could be utilized to build, test and publish a subset of the images.

PR #1092, opened 2019-04-24
Introduce build pipeline stages

MichaelSimons:
The driver for this is described in #1080.  Pipeline stages makes it easier to declare dependencies on logical sections of the build (e.g. build, test and publish).  This eliminates the complexity and complications we were facing around publishing.  Publishing should only run once all testing is complete and successful.  The complications come in when the build is run on a subset of the graph.

The one downside to this change is that now no tests will start until the build stage is complete (for official builds only).  It was nice that tests would start immediately after the corresponding build leg just as a way to get feedback sooner.  This is not a big deal since this is only for official builds.  PR validation runs tests within the build stage.  The wall clock time of the build shouldn't be affected by this change.


.vsts-pipelines/stages/build-test-publish-repo.yml
```diff
@@ -0,0 +1,198 @@
...
+
+stages:
+
+################################################################################
+# Build Images
+################################################################################
+- stage: Build
+  condition: or(eq(variables['singlePhase'], ''), eq(variables['singlePhase'], 'build'))
+  jobs:
+  - template: ../jobs/generate-matrix.yml
```
mthalman: You haven't integrated the component detection task.  I would expect that would go here.

MichaelSimons: It doesn't look like those changes where integrated into nightly.  I will cherry-pick them into this PR.

PR #1149, opened 2019-05-31
Add ENTRYPOINT when container is running

impaulapple:
We use dockerfile to build an image, cause we want to put a application in it. so I thing we should add a ENTRYPOINT to make it easier.


mthalman:
@impaulapple - Your proposed change is for a Dockerfile that produces a .NET Core image.  That is a general purpose image that can be used for any application.  Your changes are application-specific and not appropriate for this image.  The commands you've added should belong in your own application's Dockerfile.  As an example, you can look at the Dockerfile of our sample ASP.NET application: https://github.com/dotnet/dotnet-docker/blob/master/samples/aspnetapp/Dockerfile.

PR #1155, opened 2019-06-03
Update to latest image-builder

MichaelSimons:
This brings in the fix for https://github.com/dotnet/docker-tools/pull/190.


MichaelSimons:
Merging to unblock builds.

PR #1162, opened 2019-06-07
Publish image info

mthalman:
Updates the build pipeline so that the built images have their image info collected and published to the master image info file.

Here's how the flow works:
1. In the build stage, each build job writes to a file its own image info containing metadata about only the images it just built.  This file is published as a pipeline artifact with a unique name specific to that job so it doesn't conflict with the other jobs.
2. During the publish stage, all of the pipeline artifacts are downloaded and fed into the ImageBuilder command to merge all of the content from the files into the master image info file.

Notes:
* In order to get files into and out of the container that ImageBuilder runs in, a volume mount is used that maps to the $(Build.ArtifactStagingDirectory) of the build agent.
* In order to support scenarios where the build is run for a specific stage like test or publish where the build hasn't been run, we need the ability to target the pipeline artifacts from the source build (the build that produced the images).  The DownloadPipelineArtifact task is capable of referencing artifacts from other builds.  A sourceBuildId variable is being introduced as a variable that can be set at queue time to provide this value.  This variable replaces the original stagingRepoPrefix variable which is now computed from the sourceBuildId variable.

### DIFF
--- a/.vsts-pipelines/jobs/build-images.yml
+++ b/.vsts-pipelines/jobs/build-images.yml
@@ -57,9 +57,15 @@ jobs:
       --os-type $(osType)
       --os-version "$(osVersion)"
       --architecture $(architecture)
+      --image-info-output-path "$(imageBuilderArtifactsPath)/$(legName)-image-info.json"
       --retry
       $(imageBuilderBuildArgs)
     displayName: Build Images
+  - task: PublishPipelineArtifact@0
+    inputs:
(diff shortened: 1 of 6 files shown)


mthalman:
> LGTM - What is the time impact of the changes to publishing? It looks as though it should be marginal.

Yeah, about 30 secs to download the pipeline artifacts from a full build and then 4 secs to update the file in Git.

PR #1180, opened 2019-06-18
 Remove wildcards from filters in test steps

mthalman:
As part of merging the common engineering infra to dotnet-framework-docker, I discovered a difference between the two repos in how filters are specified in tests.  The dotnet-framework-docker repo had the wildcard character removed from the VersionFilter parameter.  This ensured that you could run tests for just '4.7' and not end up including '4.7.x'.  So I'm incorporating that difference here.


mthalman:
Needed to add back the wildcard for the OSFilter parameter on Linux because it needs to be able to match "stretch" or "stretch-slim", for example.

Issue #1187, opened 2019-06-19
Stop producing .NET Core Docker images for 1.* once it is EOL.


MichaelSimons:
.NET Core 1.0 and 1.1 will reach EOL on June 27th, 2019 ([announcement](https://devblogs.microsoft.com/dotnet/net-core-1-0-and-1-1-will-reach-end-of-life-on-june-27-2019/)).  At this time, the 1.0 and 1.1 Dockerfiles should be removed and the corresponding tags should no longer be documented within the readme.  All 1.0 and 1.1 images WILL REMAIN on Docker Hub and MCR indefinitely as to not impact any customers still using them.

- [ ] Remove Dockerfiles
- [ ] Remove tags from readmes
- [ ] Remove manifest entries
- [ ] Remove tests (including special cases to handle 1.* differences)
- [ ] Remove MCR Docker Hub image syndication 
- [ ] Verify no sample references
- [ ] Remove image size baseline data
- [ ] Remove update-dependency specializations to handle 1.* differences

PR #1191, opened 2019-06-21
Remove all support for 1.0 and 1.1 images (1.0 and 1.1 EOL)

MichaelSimons:
Related to #1187 

While 1.0 and 1.1 EOL isn't until 6/27, this work in being done in the nightly branch ahead of time in preparation.

**Changes made**
1. Removed Dockerfiles
1. Removed tags from readmes
1. Removed manifest entries
1. Removed tests (including special cases to handle 1.* differences)
1. Removed sample references
1. Removed image size baseline data
1. Removed update-dependency specializations to handle 1.* differences


tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -15,12 +15,11 @@ public class ImageData
         private List<string> _pulledImages = new List<string>();
         private Version _runtimeDepsVersion;
         private string _sdkOS;
-        private Version _sdkVersion;
 
         public Arch Arch { get; set; }
         public Version Version { get; set; }
         public string VersionString => Version.ToString(2);
-        public bool HasSdk => SdkVersion == Version;
+        public bool HasSdk => _sdkOS != null;
```
MichaelSimons: This was a bug I discovered while making this change.  The alpine arm64 tests make use of the Buster SDK because there isn't an alpine arm64 sdk image.  As a result of this, they were running the sdk tests against the Buster image unnecessarily.  In essence the arm64 buster SDK image was being tested multiple times.

eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -23,24 +23,21 @@ public class DockerfileShaUpdater : FileRegexUpdater
         private const string EnvNameGroupName = "envName";
         private const string ValueGroupName = "value";
 
-        private static readonly string s_envVersionPattern =
```
MichaelSimons: The 1.* Dockerfiles used an `ENV` to store the download URL.  This can be cleaned up so there is now one code path.  As part of this I also cleaned up an old pattern that was used in 2.0 where the SHAs were stored in an `ENV`.

PR #1214, opened 2019-07-10
Integrate eng/common folder from docker-tools

mthalman:


mthalman:
> LGTM - I assume you will back port the tweaks you made into docker-tools once you confirm everything is working smoothly?

Yep, that's the plan

PR #1217, opened 2019-07-11
Update tests to support dynamically pulling missing images

mthalman:
In order to support https://github.com/dotnet/docker-tools/issues/218, these changes update the test code to account for not having all images required by the test available in the staging ACR location.  In that case, the test will pull them directly from MCR.

Also needed to make some changes to the pipeline.  These are ugly but necessary due to issues dealing with multi-line variables in the pipeline YAML stuff.

### DIFF
--- a/eng/common/templates/steps/init-docker-linux.yml
+++ b/eng/common/templates/steps/init-docker-linux.yml
@@ -8,6 +8,13 @@ steps:
 - script: echo "##vso[task.setvariable variable=artifactsPath]/artifacts"
   displayName: Define Artifacts Path Variable
 
+- ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+  - script: echo '##vso[task.setvariable variable=imageBuilderImageInfoArg]--image-info-output-path $(artifactsPath)/$(legName)-image-info.json'
+    displayName: Set Image Info Arg for Image Builder
+- ${{ if eq(variables['System.TeamProject'], 'public') }}:
+  - script: echo '##vso[task.setvariable variable=imageBuilderImageInfoArg]'
+    displayName: Set Image Info Arg for Image Builder
+
   ################################################################################
   # Cleanup Docker Resources
   ################################################################################
@@ -69,10 +76,3 @@ steps:
   - ${{ if eq(parameters.setupRemoteDockerServer, 'true') }}:
     - script: echo "##vso[task.setvariable variable=dockerClientImage]$(imageNames.testRunner.withrepo)"
       displayName: Define dockerClientImage Variable
-
-- ${{ if eq(variables['System.TeamProject'], 'internal') }}:
-  - script: echo '##vso[task.setvariable variable=imageBuilderImageInfoArg]--image-info-output-path $(artifactsPath)/$(legName)-image-info.json'
-    displayName: Set Image Info Arg for Image Builder
-- ${{ if eq(variables['System.TeamProject'], 'public') }}:
-  - script: echo '##vso[task.setvariable variable=imageBuilderImageInfoArg]'
-    displayName: Set Image Info Arg for Image Builder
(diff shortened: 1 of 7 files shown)


tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -15,11 +18,21 @@ public class DockerHelper
         public static string DockerOS => GetDockerOS();
         public static string ContainerWorkDir => IsLinuxContainerModeEnabled ? "/sandbox" : "c:\\sandbox";
         public static bool IsLinuxContainerModeEnabled => string.Equals(DockerOS, "linux", StringComparison.OrdinalIgnoreCase);
+
+        public JArray ImageInfoRepos { get; }
```
MichaelSimons: This doesn't feel like the correct place.  DockerHelper's primary purpose is providing access to the Docker concepts (e.g. CLI).  What do you think about a static member on ImageData?

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -123,7 +124,30 @@ private string GetImageName(DotNetImageType imageType)
                 arch = "-arm64v8";
             }
 
-            return $"{Config.Registry}/{Config.RepoPrefix}dotnet/core{repoSuffix}/{variantName}:{imageVersion.ToString(2)}-{os}{arch}";
+            string tag = $"{imageVersion.ToString(2)}-{os}{arch}";
+            string repo = $"dotnet/core{repoSuffix}/{variantName}";
+
+            bool imageExistsInStaging = true;
+            if (dockerHelper.ImageInfoRepos != null)
```
MichaelSimons: I think a comment would be helpful here explaining why no ImageInfoRepos means the image exists in staging.

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -87,7 +88,7 @@ public string GetImage(DotNetImageType imageType, DockerHelper dockerHelper)
             return imageName;
         }
 
-        private string GetImageName(DotNetImageType imageType)
+        private string GetImageName(DotNetImageType imageType, DockerHelper dockerHelper)
```
MichaelSimons: This method is getting fairly lengthy.  It may help the readability to break it apart into GetTagName and GetRegistryName.

tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
mthalman: Makes sense

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -69,6 +71,18 @@ public string SdkOS
             set { _sdkOS = value; }
         }
 
+        private static JArray GetImageInfoRepos()
+        {
+            string imageInfoPath = Environment.GetEnvironmentVariable("IMAGE_INFO_PATH");
+            if (!String.IsNullOrEmpty(imageInfoPath))
+            {
+                string imageInfoContents = File.ReadAllText(imageInfoPath);
+                return JsonConvert.DeserializeObject<JArray>(imageInfoContents);
```
MichaelSimons: Thoughts on storing this in a static field?  In most build leg scenarios we will end up loading the image info file 4 times.

eng/common/templates/steps/init-docker-linux.yml
```diff
@@ -69,3 +69,10 @@ steps:
   - ${{ if eq(parameters.setupRemoteDockerServer, 'true') }}:
     - script: echo "##vso[task.setvariable variable=dockerClientImage]$(imageNames.testRunner.withrepo)"
       displayName: Define dockerClientImage Variable
+
```
MichaelSimons: To me it feels like this would better belong at the top right after the `artifactsPath` variable is set.

PR #1220, opened 2019-07-13
README typo correction

RyanWarnock:
I believe there is a typo on the README.md, I've offered the correction. If it is intentional please disregard this.


MichaelSimons:
Thanks Ryan for fixing this typo!

PR #1230, opened 2019-07-25
Minor tweaks to speed up tests

### LINKED ISSUE #1078: Investigate optimizing tests
There are a few test optimizations that should be investigated to see if we can speed up tests.  They take a fair amount of time during the arm build legs which run on lower end hardware.

- [ ] Utilizing the --no-restore option.  This is particularly relevant when creating the test projects.
- [ ] Avoid port exposure except for web scenarios.

MichaelSimons:
The primary change here is to utilize the `--no-restore` option in several places where restoring is unnecessary.

Fixes #1078

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
@@ -212,17 +212,17 @@ namespace Microsoft.DotNet.Docker.Tests
             string name,
             string command = null,
             string workdir = null,
-            string publishArgs = " -p 80",
+            string optionalRunArgs = null,
             bool detach = false,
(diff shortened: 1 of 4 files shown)


mthalman:
Any rough numbers for how much this improves things?

MichaelSimons:
It is hard to measure because of the amount of variance we see in our runs.  It is fairly insignificant.  The removal of the `dotnet restore -r $rid` is the big one.  This wasn't actually correct to begin with because this restore brings in the minimum package version which doesn't align with the version of the runtime the image includes.  The subsequent publish call was bringing in the correct version.  For ASP.NET apps this fix saves ~15s to 1 min.  I'm expecting a bigger savings on Windows ARM.

PR #1275, opened 2019-08-20
Update PowerShell version to `v7.0.0-preview.3`

adityapatwardhan:
Update to the latest PowerShell release and add new environment variable which will be used by PowerShell telemetry.

### DIFF
--- a/3.0/sdk/alpine3.10/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.10/amd64/Dockerfile
@@ -28,11 +28,11 @@ ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
 RUN dotnet help
 
 # Install PowerShell global tool
-ENV POWERSHELL_VERSION=7.0.0-preview.3 \
+ENV POWERSHELL_VERSION=7.0.0-previewa.3 \
     POWERSHELL_DISTRIBUTION_CHANNEL=DOTNETSDK-Alpine-3.10
 
 RUN wget -O PowerShell.Linux.Alpine.$POWERSHELL_VERSION.nupkg https://pwshtool.blob.core.windows.net/tool/$POWERSHELL_VERSION/PowerShell.Linux.Alpine.$POWERSHELL_VERSION.nupkg \
-    && powershell_sha512='f42acb9a200c56b0fa5c94ee4f9fb09711fe43bab4764dba71ae912797eb6d42e442bc1a6e83a54679004138bf8f44596b0a075b30e4a91f53fb5939ffcb0153' \
+    && powershell_sha512='9dc84593409b1a0248ed86c61b2130567ecdef282034e37eb51a2fb756dc601ace994aa94b5a20ab6a3c57fcbd32b764b79e3d809cb47167e89392f0b7686236' \
     && echo "$powershell_sha512  PowerShell.Linux.Alpine.$POWERSHELL_VERSION.nupkg" | sha512sum -c - \
     && mkdir -p /usr/share/powershell \
     && dotnet tool install --add-source / --tool-path /usr/share/powershell --version $POWERSHELL_VERSION PowerShell.Linux.Alpine \
--- a/3.0/sdk/alpine3.10/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.10/amd64/Dockerfile
@@ -29,7 +29,7 @@ RUN dotnet help
 
 # Install PowerShell global tool
 ENV POWERSHELL_VERSION=7.0.0-previewa.3 \
-    POWERSHELL_DISTRIBUTION_CHANNEL=DOTNETSDK-Alpine-3.10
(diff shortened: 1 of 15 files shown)


mthalman:
@adityapatwardhan - The test failures in the build for Alpine look related to your changes.  Please investigate.  The build failure for NanoServer looks like it just needs a retry.

3.0/sdk/alpine3.10/amd64/Dockerfile
```diff
@@ -28,10 +28,11 @@ ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
 RUN dotnet help
 
 # Install PowerShell global tool
-ENV POWERSHELL_VERSION 7.0.0-preview.2
+ENV POWERSHELL_VERSION=7.0.0-preview.3 \
+    POWERSHELL_DISTRIBUTION_CHANNEL=DOTNETSDK-Alpine-3.10
```
MichaelSimons: 1. I find the casing of this value to be strange.  Does `DOTNETSDK` need to be upper cased? 
1. Are you concerned about `DOTNETSDK` being ambiguous?  Should `Docker` be included as was done in the PS Docker images?
1. Does PS telemetry not collect info from the running OS environment?  I am wondering why the OS needs to be included.
1. It is a common practice for .NET customers to clone our Dockerfiles as necessary, tweak, build and use the resulting image.  This will skew the telemetry results to a degree if the value is not updated.  To help combat this, the telemetry logic can self detect some information such as the previously mentioned OS info in the case a user changes the base image to a different distro.

adityapatwardhan: Here are the details that we collect: https://devblogs.microsoft.com/powershell/new-telemetry-in-powershell-7-preview-3/

- Not stuck on what the casing or the name should look like. Would something lilke `PSDocker-DotnetSDK-Alpine-3.10` be better?
- What we collect is `$PSVersionTable.OS`, which gives us something like: `Linux 4.9.184-linuxkit #1 SMP Tue Jul 2 22:58:16 UTC 2019`. This does not tell us which OS is used.
- Agreed that this might skew the results, but i believe the number of people using the image or creating an image based of this would be higher compared to people copying and modifying the `FROM` statement, but i could be wrong. Since we collect telemetry at start we don't want to do much, to have least amount of effect on start up times.

adityapatwardhan:
@mthalman I have root caused the issue for Alpine. It was due to overaggressive package size reduction. Working on a fix.

3.0/sdk/alpine3.10/amd64/Dockerfile
MichaelSimons: I like the casing on your proposal.  I wonder if `PSDocker-DotnetCoreSDK-Alpine-3.10` would be better?  It is more precise.

adityapatwardhan: Changed it.

adityapatwardhan:
@MichaelSimons @mthalman All issues fixed. Please re-review.

adityapatwardhan:
@mthalman Any ideas when this change goes live?

mthalman:
@adityapatwardhan - Build is done.  Should be good.  The build pipeline is located [here](https://dev.azure.com/dnceng/internal/_build?definitionId=359) (internal MSFT link).

MichaelSimons:
```
PS D:\repos\docker-tools\eng\file-pusher> docker pull mcr.microsoft.com/dotnet/core-nightly/sdk:3.0-buster
3.0-buster: Pulling from dotnet/core-nightly/sdk
4ae16bd47783: Pull complete                                                                                          bbab4ec87ac4: Pull complete                                                                                          2ea1f7804402: Pull complete                                                                                          96465440c208: Pull complete                                                                                          78f0dfdbd224: Pull complete                                                                                          cbb7ca314bb9: Pull complete                                                                                          4121f1526301: Pull complete                                                                                          a79b74c8466b: Pull complete                                                                                          Digest: sha256:4a7b483c84c70ed29b2ed3f32fa0a56f912c8090e128eef6b3b2fbb2c719ccba
Status: Downloaded newer image for mcr.microsoft.com/dotnet/core-nightly/sdk:3.0-buster
mcr.microsoft.com/dotnet/core-nightly/sdk:3.0-buster
PS D:\repos\docker-tools\eng\file-pusher> docker run --rm mcr.microsoft.com/dotnet/core-nightly/sdk:3.0-buster pwsh --version
PowerShell 7.0.0-preview.3
```

adityapatwardhan:
i meant when will this move from the nightly branch to `mcr.microsoft.com/dotnet/core/sdk:3.0-buster`? During next release?

mthalman:
> i meant when will this move from the nightly branch to `mcr.microsoft.com/dotnet/core/sdk:3.0-buster`? During next release?

Correct, for the 3.0 Preview 9 release.

PR #1312, opened 2019-09-16
Remove tzdata from Alpine runtime-deps

### LINKED ISSUE #1311: Remove tzdata package from Alpine-based runtime-deps images
In order to maintain consistency among the `runtime-deps` images, we've decided that none of the `runtime-deps` images should have the `tzdata` package installed.  This package is currently installed in the Alpine images and should be removed.  Consumers of these images that need to call into the `TimeZoneInfo` API will need to explicitly install the `tzdata` package in their own Dockerfile.

mthalman:
Fixes #1311

### DIFF
--- a/2.1/runtime-deps/alpine3.10/amd64/Dockerfile
+++ b/2.1/runtime-deps/alpine3.10/amd64/Dockerfile
@@ -10,6 +10,7 @@ RUN apk add --no-cache \
         libssl1.1 \
         libstdc++ \
         lttng-ust \
+        tzdata \
         userspace-rcu \
         zlib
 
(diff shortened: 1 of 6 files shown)


2.1/runtime-deps/alpine3.10/amd64/Dockerfile
```diff
@@ -10,7 +10,6 @@ RUN apk add --no-cache \
         libssl1.1 \
         libstdc++ \
         lttng-ust \
-        tzdata \
```
MichaelSimons: We can't make this change to 2.*.  It would be a breaking change.

PR #1316, opened 2019-09-17
update for latest 3.0 drop

mthalman:


eng/mcr-tags-metadata-templates/aspnet-tags.yml
```diff
@@ -1,4 +1,9 @@
 $(McrTagsYmlRepo:aspnet)
+$(McrTagsYmlTagGroup:3.0-buster-slim)
```
MichaelSimons: All of the template changes don't seem to be reflected in the readmes.  The readmes still have the 3.0 Preview sub-sections.

mthalman: Fixed

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config.nightly
```diff
@@ -3,6 +3,7 @@
   <packageSources>
     <clear />
     <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+    <add key="dotnet-core-3.0-rc2" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/3.0.100-rc2-014265/nuget/v3/index.json" />
```
MichaelSimons: This doesn't feel like the best long term solution.  Having version specific feeds in the NuGet.config that is shared across all versions doesn't seem like the best way to ensure the expected packages are being used within the tests.  It also requires another artifact to be updated when new versions are pushed.  Would it be feasible to dynamically detect the versioned feeds and modify the config?

mthalman: How about we get these changes in for now and I'll log a separate issue to address this?

MichaelSimons: Yes - I am fine with that.

PR #1322, opened 2019-09-18
Remove ASPNETCORE_URLS env var from 3.0 SDK

### LINKED ISSUE #1199: ASPNETCORE_URLS environment variable set in SDK image
I was wondering why the ASPNETCORE_URLS environment variable is being set for the SDK images. I can see that this was added with #428, and discussed in #426. The rationale there was that one shouldn't use dotnet run inside of the container in the first place, but if you do at least it would work.

With the availability of remote development features in VS Code, especially with containers, I would suggest this makes less sense now. Personally I think that using the SDK image as a base image for creating a development container is a really nice way to setup consistent development environments for teams to use.

Now this does beg the question whether it makes sense that ASPNETCORE_URLS is set so http://+:80 for the SDK image, since if you would install the SDK directly on your machine the environment variable wouldn't be set and Kestrel will default to using ports 5000 and 5001 (for http and https respectively). Therefore I would suggest removing this from the SDK image to create consistency with a native install.

Additionally, the ASPNETCORE_Environment variable isn't set within the SDK image. To be consistent with the native install I would imagine this variable be set to Development in the SDK image.

mthalman:
Fixes #1199

### DIFF
--- a/3.0/sdk/alpine3.10/amd64/Dockerfile
+++ b/3.0/sdk/alpine3.10/amd64/Dockerfile
@@ -19,12 +19,12 @@ RUN wget -O dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DO
     && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet \
     && rm dotnet.tar.gz
 
-# Enable correct mode for dotnet watch (only mode supported in a container)
-ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
+# Unset the value from the base image
+ENV ASPNETCORE_URLS= \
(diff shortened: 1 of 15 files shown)


3.0/sdk/alpine3.10/amd64/Dockerfile
```diff
@@ -22,7 +22,9 @@ RUN wget -O dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DO
 # Enable correct mode for dotnet watch (only mode supported in a container)
 ENV DOTNET_USE_POLLING_FILE_WATCHER=true \
```
MichaelSimons: The common convention we have been following is to alphabetize the ENVs.

PR #1325, opened 2019-09-23
Update for 3.0 release

mthalman:
* Updates 3.0 to GA release
* Remove `tzdata` from Alpine `runtime-deps`.
* Remove `ASPNETCORE_URLS` environment variable from 3.0 SDK
* Update PowerShell version

### DIFF
--- a/3.0/aspnet/nanoserver-1809/arm32v7/Dockerfile
+++ b/3.0/aspnet/nanoserver-1809/arm32v7/Dockerfile
@@ -6,6 +6,6 @@ FROM $REPO:3.0-nanoserver-1809-arm32v7
 # Install ASP.NET Core Runtime
 ENV ASPNETCORE_VERSION 3.0.0
 
-RUN curl -SL --output dotnet.zip https://dotnetcli.blob.core.windows.net/dotnet/aspnetcore/Runtime/$ASPNETCORE_VERSION/aspnetcore-runtime-%ASPNETCORE_VERSION%-win-arm.zip `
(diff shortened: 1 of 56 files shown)


3.0/aspnet/nanoserver-1809/arm32v7/Dockerfile
```diff
@@ -4,8 +4,8 @@ ARG REPO=mcr.microsoft.com/dotnet/core/runtime
 FROM $REPO:3.0-nanoserver-1809-arm32v7
 
 # Install ASP.NET Core Runtime
-ENV ASPNETCORE_VERSION 3.0.0-rc1.19457.4
+ENV ASPNETCORE_VERSION 3.0.0
 
-RUN curl -SL --output dotnet.zip https://dotnetcli.blob.core.windows.net/dotnet/aspnetcore/Runtime/%ASPNETCORE_VERSION%/aspnetcore-runtime-%ASPNETCORE_VERSION%-win-arm.zip `
+RUN curl -SL --output dotnet.zip https://dotnetcli.blob.core.windows.net/dotnet/aspnetcore/Runtime/$Env:ASPNETCORE_VERSION/aspnetcore-runtime-%ASPNETCORE_VERSION%-win-arm.zip `
```
MichaelSimons: I don't understand this change?  Windows arm doesn't include PS.  I would expect this to remain using `%`.  Also the two ENV references within this single line are inconsistent.

mthalman: It was set to `$ASPNETCORE_VERSION` before, which was incorrect, and then incorrectly set to `$Env:ASPNETCORE_VERSION`. Now fixed to be `%ASPNETCORE_VERSION%`.

PR #1339, opened 2019-09-25
Add 3.1 preview1 Dockerfiles

### LINKED ISSUE #1299: Produce nightly images for 3.1
Now that builds of 3.1 are available, we should be producing images for them in the nightly branch.

mthalman:
Adds Dockerfiles and updates the infrastructure to build, test, and publish Docker images for .NET Core 3.1

Fixes #1299


3.1/runtime-deps/alpine3.10/amd64/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM alpine:3.10
```
MichaelSimons: Unless there is a change in the native dependencies in 3.1, I think we should be re-tagging/sharing the runtime-deps with 3.0.  This supports more layer sharing/re-use.

3.1/sdk/nanoserver-1809/arm32v7/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# Install .NET Core SDK
+ENV DOTNET_SDK_VERSION 3.1.100-preview1-014065
+
+RUN curl -SL --output dotnet.zip https://dotnetcli.blob.core.windows.net/dotnet/Sdk/%DOTNET_SDK_VERSION%/dotnet-sdk-%DOTNET_SDK_VERSION%-win-arm.zip `
+    && mkdir "%ProgramFiles%\dotnet" `
+    && tar -zxf dotnet.zip -C "%ProgramFiles%\dotnet" `
+    && del dotnet.zip
+
+# Install PowerShell global tool
+ENV POWERSHELL_VERSION=7.0.0-preview.3 `
```
MichaelSimons: The sdk files appear out of sync with 3.0.

1. PS should be preview 4
1. The ASPNETCORE_URLS ENV should be removed.

Issue #1340, opened 2019-09-25
pwsh not executable by non-root users


robmccready:
## Steps to reproduce the issue
1. Launch container as non-root user: 
    docker run -it -u 123:123 mcr.microsoft.com/dotnet/core/sdk:3.0
2. Inside container attempt to launch pwsh
3. bash: /usr/bin/pwsh: Permission denied
4. ls -al /usr/share/powershell shows permissions on pwsh
    -rwxr--r-- 1 root root 86584 Sep 23 15:06 pwsh

## Expected behavior
pwsh should be executable by non-root user. Jenkins passes the jenkins user to docker which forces a non-root user.

## Actual behavior
pwsh permission denied 

## Work around
I created my own image inheriting from mcr.microsoft.com/dotnet/core/sdk:3.0 with
   RUN chmod 755 /usr/share/powershell/pwsh

MichaelSimons:
@adityapatwardhan - This seems like something we should fix.  Is this something you can create a PR for?

robmccready:
I looked at two of the docker files and it looks like the powershell download/install is consistent in each so I think I should be able to put a PR together in the next few days.

MichaelSimons:
@robmccready - If you make a PR, please make it against the nightly branch.

mthalman:
Finished roll out with https://github.com/dotnet/dotnet-docker/pull/1473

PR #1345, opened 2019-09-26
Fix for bug 1340 - powershell not executable by non-root users

robmccready:
[Bug 1340](https://github.com/dotnet/dotnet-docker/issues/1340)

* Add "chmod 755 /usr/share/powershell/pwsh" to all linux Dockerfiles that pull in powershell
* Update Image Test app to execute powershell test as non-root user

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
@@ -135,6 +135,7 @@ namespace Microsoft.DotNet.Docker.Tests
             string output = _dockerHelper.Run(
                 image: imageData.GetImage(DotNetImageType.SDK, _dockerHelper),
                 name: imageData.GetIdentifier($"pwsh"),
+                optionalRunArgs: "-u 12345:12345", // force test to run as non-root user inside container
                 command: $"pwsh -c (Get-Childitem env:DOTNET_RUNNING_IN_CONTAINER).Value"
             );
 
--- a/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
@@ -135,14 +135,14 @@ namespace Microsoft.DotNet.Docker.Tests
 
         [Theory]
         [MemberData(nameof(GetImageData))]
-        public void VerifySDKImage_PowerShellScenario_Installed(ImageData imageData)
+        public void VerifySDKImage_PowerShellScenario_DefaultUser(ImageData imageData)
         {
-            VerifySDKImage_PowerShell_Execute(imageData, null);
+            VerifySDKImage_PowerShellScenario_Execute(imageData, null);
         }
 
         [Theory]
         [MemberData(nameof(GetImageData))]
-        public void VerifySDKImage_PowerShellScenario_Executable_As_AlternativeUser(ImageData imageData)
+        public void VerifySDKImage_PowerShellScenario_NonDefaultUser(ImageData imageData)
         {
             var optRunArgs = "-u 12345:12345"; // Linux containers test as non-root user
             if (imageData.OS.Contains("nanoserver", StringComparison.OrdinalIgnoreCase))
@@ -150,7 +150,7 @@ namespace Microsoft.DotNet.Docker.Tests
                 optRunArgs = "-u ContainerAdministrator "; // windows containers test as Admin, default execution is as ContainerUser
             }
 
-            VerifySDKImage_PowerShell_Execute(imageData, optRunArgs);
+            VerifySDKImage_PowerShellScenario_Execute(imageData, optRunArgs);
         }
 
         private void VerifySDKImage_PowerShellScenario_Execute(ImageData imageData, string optionalArgs)
--- a/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
@@ -153,7 +153,7 @@ namespace Microsoft.DotNet.Docker.Tests
             VerifySDKImage_PowerShell_Execute(imageData, optRunArgs);
         }
 
-        private void VerifySDKImage_PowerShell_Execute(ImageData imageData, string optionalArgs)
+        private void VerifySDKImage_PowerShellScenario_Execute(ImageData imageData, string optionalArgs)
         {
             if (imageData.Version.Major < 3)
             {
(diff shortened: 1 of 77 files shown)


tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -135,6 +135,7 @@ public void VerifySDKImage_PowerShellScenario(ImageData imageData)
             string output = _dockerHelper.Run(
                 image: imageData.GetImage(DotNetImageType.SDK, _dockerHelper),
                 name: imageData.GetIdentifier($"pwsh"),
+                optionalRunArgs: "-u 12345:12345", // force test to run as non-root user inside container
```
MichaelSimons: I'm not in favor of this test change.  These scenario based tests are intended to validate the canonical usage scenario.  This should be using the default user.  I think it would be better to add a new true unit test case that would validate the execute permissions.

MichaelSimons:
@daxian-dbw, @adityapatwardhan - could one of you review this change please?

mthalman:
New Dockerfiles were just added for 3.1: https://github.com/dotnet/dotnet-docker/pull/1339.  Please merge those changes in and update those files appropriately with these changes.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
robmccready: Not a problem. I will split it out and add logic to handle the difference with the windows servers (looks like needs -u ContainerUser instead).

robmccready:
No problem.

> New Dockerfiles were just added for 3.1: #1339. Please merge those changes in and update those files appropriately with these changes.

daxian-dbw:
The permission changes look good to me.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -153,6 +153,33 @@ public void VerifySDKImage_PowerShellScenario(ImageData imageData)
...
+            if (imageData.Version.Major < 3)
+            {
+                _outputHelper.WriteLine("PowerShell does not exist in pre-3.0 images, skip testing");
+                return;
+            }
+
+            var optRunArgs = "-u 12345:12345"; // force test to run as non-root (linux) user inside container
+            if (imageData.OS.Contains("nanoserver", StringComparison.OrdinalIgnoreCase))
+            {
+                optRunArgs = "-u ContainerUser"; // non-admin user on windows
```
MichaelSimons: The default user for NanoServer based images is the ContainerUser.  Given this, I think this test should target running with a non-default user, `-u 12345:12345` on the linux platform and `-u ContainerAdministrator` on windows.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -153,6 +153,33 @@ public void VerifySDKImage_PowerShellScenario(ImageData imageData)
             Assert.Equal(output, bool.TrueString, ignoreCase: true);
         }
 
+        [Theory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifySDKImage_PowerShellScenario_NonRoot_Executable(ImageData imageData)
```
MichaelSimons: Consider factoring out the common logic here to share it with the `VerifySDKImage_PowerShellScenario` method

```
        private void ExecuteSDKPowerShellScenario(ImageData imageData, string optionalRunArgs = null)
        {
            if (imageData.Version.Major < 3)
            {
                _outputHelper.WriteLine("PowerShell does not exist in pre-3.0 images, skip testing");
                return;
            }

            // A basic test which executes an arbitrary command to validate PS is functional
            string output = _dockerHelper.Run(
                image: imageData.GetImage(DotNetImageType.SDK, _dockerHelper),
                name: imageData.GetIdentifier($"pwsh"),
                optionalRunArgs: optionalRunArgs,
                command: $"pwsh -c (Get-Childitem env:DOTNET_RUNNING_IN_CONTAINER).Value"
            );

            Assert.Equal(output, bool.TrueString, ignoreCase: true);
        }
```

robmccready: Test methods refactored.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
robmccready: Updated to use -u ContainerAdministrator as the alternative user test for windows

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -135,7 +135,25 @@ public void VerifySdkImage_PackageCache(ImageData imageData)
...
+            var optRunArgs = "-u 12345:12345"; // Linux containers test as non-root user
+            if (imageData.OS.Contains("nanoserver", StringComparison.OrdinalIgnoreCase))
+            {
+                optRunArgs = "-u ContainerAdministrator "; // windows containers test as Admin, default execution is as ContainerUser
+            }
+
+            VerifySDKImage_PowerShell_Execute(imageData, optRunArgs);
+        }
+
+        private void VerifySDKImage_PowerShell_Execute(ImageData imageData, string optionalArgs)
```
MichaelSimons:
```suggestion
        private void VerifySDKImage_PowerShellScenario_Execute(ImageData imageData, string optionalArgs)
```

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -135,7 +135,25 @@ public void VerifySdkImage_PackageCache(ImageData imageData)
 
         [Theory]
         [MemberData(nameof(GetImageData))]
-        public void VerifySDKImage_PowerShellScenario(ImageData imageData)
+        public void VerifySDKImage_PowerShellScenario_Installed(ImageData imageData)
```
MichaelSimons: What do you think about the names?

```
VerifySDKImage_PowerShellScenario_DefaultUser
VerifySDKImage_PowerShellScenario_NonDefaultUser
``` 

It provides uniformity yet makes it easy to distinguish the difference in the two tests?

robmccready: I like it and I think it probably is a little clearer for non native english speakers as well.

robmccready:
Is it normal for the PR build pipeline to have errors that do not seem related to the PR changes? None of the builds have succeeded 100% but all the errors are things like docker failing to clean up after the tests. I didn't see any failures that look like they are caused by the powershell tests.

MichaelSimons:
@robmccready - Unfortunately we have been plagued with instability within the windows build agents.  I re-ran the failing job and it is now green.

PR #1387, opened 2019-10-15
Add 3.1 preview1 Dockerfiles

mthalman:


3.0/runtime-deps/alpine3.10/amd64/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM alpine:3.10
```
MichaelSimons: I think these 3.10 runtime-deps dockerfiles should be moved to 3.1 until 3.0 alpine 3.10 is released.  It is a potential source of confusion to have them in the 3.0 directory when no 3.0 tags are produced.

Issue #1404, opened 2019-10-23
Snippet doc for installing .NET Core


mthalman:
There can be cases where the .NET community wants a .NET image that we don't produce, such as .NET installed on a particular base image.  In those cases, it's up to the them to author and build their own image.  It would be nice to have a snippet document that demonstrated how .NET can be installed in an image in this way.

Different scenarios would be:
* For application-specific images, the [dotnet-install scripts](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) can be used for simplicity and easier servicing.
* For platform images, follow the pattern as is done in this repo to provide full transparency of the source of the installation.

mthalman:
Fixed with https://github.com/dotnet/dotnet-docker/pull/1424

PR #1413, opened 2019-10-28
Add snippet doc for system.drawing.common

### LINKED ISSUE #1401: Snippet documentation for using System.Drawing API
Provide a sample snippet showing how to consume the System.Drawing API in a Docker scenario.

See https://github.com/dotnet/dotnet-docker/issues/618

mthalman:
Fixes #1401


samples/snippets/using-system-drawing-common.md
```diff
@@ -0,0 +1,43 @@
+# Using the System.Drawing.Common Package in a Docker Container
+
+If your application makes use of the System.Drawing.Common NuGet package, you'll need to ensure some additional dependencies are installed in your Docker container for both Linux and Windows. See #618 (Linux) and #1098 (Windows) for more info on this issue.
+
+For Linux, it's just a matter of installing a few extra packages in your Dockerfile:
```
MichaelSimons: Consider adding `Linux` and `Windows` subsections.  This will make it easier for the target audience to skip to the relevant parts depending on their usage.  It also provides the ability to deep link.

samples/snippets/using-system-drawing-common.md
```diff
@@ -0,0 +1,43 @@
...
+
+```Dockerfile
```
MichaelSimons: This is specific to debian/ubuntu.  Since we produce Alpine images, I think we should include a snippet specific to Alpine.

samples/snippets/using-system-drawing-common.md
```diff
@@ -0,0 +1,43 @@
...
+        libx11-dev \
+     && rm -rf /var/lib/apt/lists/*
+```
+
+For Windows, it's going to be more work because the required dependencies are not available in Windows Nano Server, the OS type which the official .NET Core images for Windows are based on. Instead, you'll need to use the [Windows Server Core images](https://hub.docker.com/_/microsoft-windows-servercore) that are installed with the necessary dependencies. Because there are no official .NET Core images based on Server Core, you'll need to author a Dockerfile that installs .NET Core yourself.
+
+The quickest way to produce an image with .NET Core installed is to locate the relevant Dockerfiles in the [dotnet-docker](https://github.com/dotnet/dotnet-docker) repo that match the .NET Core and Windows version you want to work with. Copy those Dockerfiles and replace `mcr.microsoft.com/windows/nanoserver` with `mcr.microsoft.com/windows/servercore`. You can then build images from those Dockerfiles and tag them with your own tags (this example tags them with `servercore/sdk:4.8` and `servercore/runtime:4.8`) that you can then reference from your application's Dockerfile.
+
+Example application Dockerfile:
```
MichaelSimons: I don't find this example to be much help.  Instead do you think there is value in including cmds that illustrate what is discussed in the previous paragraph?  You could download a specific Dockerfile, search and replace the from, and then build the result.

You don't need a custom SDK if you aren't going to run unit tests.  Perhaps this should be called out for both Windows and Linux.

samples/snippets/using-system-drawing-common.md
```diff
@@ -0,0 +1,43 @@
...
+        libc6-dev \
+        libgdiplus \
+        libx11-dev \
+     && rm -rf /var/lib/apt/lists/*
+```
+
+For Windows, it's going to be more work because the required dependencies are not available in Windows Nano Server, the OS type which the official .NET Core images for Windows are based on. Instead, you'll need to use the [Windows Server Core images](https://hub.docker.com/_/microsoft-windows-servercore) that are installed with the necessary dependencies. Because there are no official .NET Core images based on Server Core, you'll need to author a Dockerfile that installs .NET Core yourself.
+
+The quickest way to produce an image with .NET Core installed is to locate the relevant Dockerfiles in the [dotnet-docker](https://github.com/dotnet/dotnet-docker) repo that match the .NET Core and Windows version you want to work with. Copy those Dockerfiles and replace `mcr.microsoft.com/windows/nanoserver` with `mcr.microsoft.com/windows/servercore`. You can then build images from those Dockerfiles and tag them with your own tags (this example tags them with `servercore/sdk:4.8` and `servercore/runtime:4.8`) that you can then reference from your application's Dockerfile.
```
MichaelSimons: This would be a good place to reference the https://github.com/dotnet/dotnet-docker/issues/1404 snippet once created.  That snippet should provide a more in-depth description of the available options (e.g. utilizing the dotnet-install script).

MichaelSimons: Should this call out that the Server Core images are going to be substantially larger?

mthalman: I'm having second thoughts about #1404.  It seems that is more of a thing covered by https://github.com/dotnet/core/tree/master/release-notes

samples/snippets/using-system-drawing-common.md
```diff
@@ -0,0 +1,98 @@
...
+ENTRYPOINT ["dotnetapp"]
+```
+
+### Framework-dependent deployment
+
+For a framework-dependent deployment, you're relying on .NET Core being installed in the base image you'll be referencing. Because there are no official .NET Core images based on Server Core, you'll need to author a Dockerfile that installs .NET Core yourself. Note that Server Core images are substantially larger than Nano Server images.
+
+The quickest way to produce an image with .NET Core installed is to locate the relevant Dockerfiles in the [dotnet-docker](https://github.com/dotnet/dotnet-docker) repo that match the .NET Core and Windows version you want to work with. Copy those Dockerfiles and replace `mcr.microsoft.com/windows/nanoserver` with `mcr.microsoft.com/windows/servercore`. You can then build images from those Dockerfiles and tag them with your own tags (this example tags them with `servercore/sdk:4.8` and `servercore/runtime:4.8`) that you can then reference from your application's Dockerfile.
+
+These commands illustrate copying one of the .NET Core runtime Dockerfiles and modifying it to be based on Server Core instead of Nano Server:
```
MichaelSimons:
```suggestion
These PowerShell commands illustrate copying one of the .NET Core runtime Dockerfiles and modifying it to be based on Server Core instead of Nano Server:
```

samples/snippets/using-system-drawing-common.md
```diff
@@ -0,0 +1,98 @@
...
+COPY --from=build /app/out ./
+
+ENTRYPOINT ["dotnetapp"]
+```
+
+### Framework-dependent deployment
+
+For a framework-dependent deployment, you're relying on .NET Core being installed in the base image you'll be referencing. Because there are no official .NET Core images based on Server Core, you'll need to author a Dockerfile that installs .NET Core yourself. Note that Server Core images are substantially larger than Nano Server images.
+
+The quickest way to produce an image with .NET Core installed is to locate the relevant Dockerfiles in the [dotnet-docker](https://github.com/dotnet/dotnet-docker) repo that match the .NET Core and Windows version you want to work with. Copy those Dockerfiles and replace `mcr.microsoft.com/windows/nanoserver` with `mcr.microsoft.com/windows/servercore`. You can then build images from those Dockerfiles and tag them with your own tags (this example tags them with `servercore/sdk:4.8` and `servercore/runtime:4.8`) that you can then reference from your application's Dockerfile.
```
MichaelSimons: > (this example tags them with `servercore/sdk:4.8` and `servercore/runtime:4.8`)

This doesn't appear to correspond with the example.

samples/snippets/using-system-drawing-common.md
```diff
@@ -0,0 +1,98 @@
...
+
+These commands illustrate copying one of the .NET Core runtime Dockerfiles and modifying it to be based on Server Core instead of Nano Server:
+
+```PowerShell
```
MichaelSimons: Because of the length and structure of these commands, I find that the formatting makes difficult to tease apart the number of commands.  I don't have any great suggestions here.  Adding a `PS C:\>` to the beginning of each line would help but then hurts the copy/paste action.

samples/snippets/using-system-drawing-common.md
```diff
@@ -0,0 +1,98 @@
...
+# install System.Drawing native dependencies
+RUN apk add \
+        --no-cache \
+        --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ \
+        libgdiplus
+```
+
+## Windows
+
+For Windows, it's going to be more work because the required dependencies are not available in Windows Nano Server, the OS type which the official .NET Core images for Windows are based on. Instead, you'll need to use the [Windows Server Core images](https://hub.docker.com/_/microsoft-windows-servercore) that are installed with the necessary dependencies. You have two options: you can choose to use a [framework-dependent](https://docs.microsoft.com/dotnet/core/deploying/#framework-dependent-deployments-fdd) or [self-contained](https://docs.microsoft.com/dotnet/core/deploying/#self-contained-deployments-scd) deployment. A framework-dependent deployment requires that the .NET Core runtime be installed on the base image whereas with a self-contained deployment, the app and all of its dependencies (including .NET Core) are deployed together.
```
MichaelSimons: > Instead, you'll need to use the [Windows Server Core images](https://hub.docker.com/_/microsoft-windows-servercore) that are installed with the necessary dependencies.

The full Windows image is also an option here.

samples/snippets/using-system-drawing-common.md
MichaelSimons: I was expecting that we would be able to just reference the `Installing .NET Core in a Dockerfile` snippet here.  I see the reason you included the commands but I don't like how nothing discusses the maintenance you take on by doing this - e.g. your Dockerfile is fixed to a particular .NET Core version.

PR #1424, opened 2019-10-30
Snippets for installing .NET Core

mthalman:


samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,103 @@
+# Installing .NET Core in a Dockerfile
+
+There can be times where you need .NET Core installed on a base image than is unavailable amongst the set of official .NET Core Docker images, such as a boutique Linux distro for example. In that case, you'll need to author your own Dockerfile which installs .NET Core. The snippets below describe how to do this.
```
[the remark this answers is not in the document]
dagood: > Wouldn't this most often require a special source built .NET Core to begin with?

This is a question of degree, but I would say no. The major things that I know of that require a rebuild are using something other than glibc, and using an exotic CPU architecture. There are a huge number of distros and distro versions that are compatible with the `linux-x64` portable build if you install the right dependencies.

So I wouldn't call it an edge case. However, because a distro Microsoft doesn't provide Docker images for *is* a specialized base image, it's reasonable IMO to only call out specialized base images as you suggested.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Custom Application Images
+
+If you're building an image for a custom application that is not to be intended to be publicly consumed as a platform for other applications, you can get by with a simpler Dockerfile implementation compared to [general platform images](#general-platform-images) if you choose. Because the image is only intended for your own organization's purposes, the need for transparency in the Dockerfile is lessened. Convenience can trump clarity in this case.
+
+For custom application images, it is recommended that you install .NET Core by [package manager](#installing-from-a-linux-package-manager)(Linux only) or [dotnet-install script](#installing-from-dotnet-install-script) (Linux/Windows).
```
dagood:
```suggestion
For custom application images, it is recommended that you install .NET Core by [package manager](#installing-from-a-linux-package-manager) (Linux only) or [dotnet-install script](#installing-from-dotnet-install-script) (Linux/Windows).
```

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Installing from a Linux Package Manager
+
+For Linux, you may prefer to use your Linux distro's package manager to install .NET Core rather than directly from a binary archive. Each .NET Core release provides [release notes](https://github.com/dotnet/core/tree/master/release-notes) that contains details on how to install from the various distro package managers. Using a package manager allows for easier maintenance since you only have to reference the major/minor version of the release and you'll get servicing releases "for free". (This is true as long as you are mindful of Docker's caching functionality. You may need to build with the `--no-cache` option to force the build to re-execute the commands that install .NET Core in order to get an updated servicing release.)
```
dagood: > Each .NET Core release provides [release notes](https://github.com/dotnet/core/tree/master/release-notes) that contains details on how to install from the various distro package managers.

I didn't know there were instructions here until you pointed them out to me, and I wouldn't recommend them because they're out of date. (In particular https://github.com/dotnet/core/blob/master/Documentation/linux-setup.md#installation-using-a-package-manager. Missing distros/versions.)

The ones I'm familiar with are at https://dotnet.microsoft.com/download/dotnet-core, selecting a version, then going to a `Package Manager Instructions: x64` link. These are the ones I keep an eye on and poke people to get fixed.

It looks like the steps below do follow the instructions I know about, so good there. 😄

dagood: > (This is true as long as you are mindful of Docker's caching functionality. You may need to build with the `--no-cache` option to force the build to re-execute the commands that install .NET Core in order to get an updated servicing release.)

Maybe worth mentioning that you *can* use the package manager's syntax for specifying an exact version, then you can use the same `ENV DOTNET_VERSION 3.0.0` approach to avoid cache issues and have more repeatability.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+    && apt-get install -y --no-install-recommends \
+        dotnet-runtime-3.0 \
+    \
+    # Cleanup
+    && rm -rf /var/lib/apt/lists/*
+```
+
+## Installing from dotnet-install script
+
+A set of [installation scripts](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) are provided to conveniently install .NET Core on Linux with Bash or Windows with PowerShell. These scripts can be thought of as a happy medium between the two previously mentioned approaches (binary archive and package manager). They fill a gap on systems where the desired .NET Core release is not available through a package manager. With the installation script, you have flexibility in specifying which version gets installed. You can install a specific version such as 3.0.1, the latest of a release channel such as the latest 3.0 patch, etc.
```
dagood: > They fill a gap on systems where the desired .NET Core release is not available through a package manager.

Filling the gap is really what the tarballs/zips are about, IMO... the script just makes them easier to use. With it, you don't have to write your own tarball grabber/extractor, and the API is about the same across OSes.

(Since this approach also doesn't require elevation, it's particularly good for dev and CI scenarios even if the official Deb/RPMs do support your system, but that's not relevant here.)

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
+# Installing .NET Core in a Dockerfile
+
+There can be times where you need .NET Core installed on a base image than is unavailable amongst the set of official [.NET Core Docker images](https://hub.docker.com/_/microsoft-dotnet-core), such as a different Linux distro version or a Windows Server Core image. In that case, you'll need to author your own Dockerfile which installs .NET Core. The snippets below describe how to do this.
```
dagood:
```suggestion
There can be times where you need .NET Core installed on a base image that is unavailable amongst the set of official [.NET Core Docker images](https://hub.docker.com/_/microsoft-dotnet-core), such as a different Linux distro version or a Windows Server Core image. In that case, you'll need to author your own Dockerfile which installs .NET Core. The snippets below describe how to do this.
```

samples/snippets/installing-dotnet.md
mthalman: @dagood - Regarding the out-of-date info at https://github.com/dotnet/core/blob/master/Documentation/linux-setup.md#installation-using-a-package-manager, could you log a bug for that and specify what should be expected there or to remove it entirely in favor of the docs page?

dagood: Sure, filed https://github.com/dotnet/core/issues/3728.

samples/snippets/installing-dotnet.md
mthalman: I've clarified the language around filling the gap.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+
+In some cases, you can workaround dependencies by publishing your .NET Core application as a [self-contained app](https://docs.microsoft.com/en-us/dotnet/core/deploying) in which case all of your app's dependencies are packaged with the app. This reduces the dependencies that need to be installed separately on the base image. For example, a Windows app may require dependencies that only exist in Windows Server Core but only .NET Core images on Nano Server are available. In that case, the app could be deployed as a self-contained app and operate just fine in the Nano Server image. Example Dockerfiles that demonstrate publishing a self-contained app are available in the samples for [Linux](https://github.com/dotnet/dotnet-docker/blob/master/samples/dotnetapp/Dockerfile.debian-x64-selfcontained) and [Windows](https://github.com/dotnet/dotnet-docker/blob/master/samples/dotnetapp/Dockerfile.nanoserver-x64-selfcontained).
```
MichaelSimons: >In that case, the app could be deployed as a self-contained app and operate just fine in the Nano Server image

Shouldn't this say Windows Server Core instead?

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+
+There are two scenarios you should consider depending on how your Docker image is to be consumed: [general platform images](#general-platform-images) and [custom application images](#custom-application-images). But before getting to that, let's first consider whether you actually need to go down this road.
+
+## Determine whether a different base image is needed
+
+Before doing all the work of authoring and maintaining a Dockerfile that installs .NET Core, it's worthwhile to stop and thoroughly analyze whether you actually do need a different base image than those provided as part of the set of official [.NET Core Docker images](https://hub.docker.com/_/microsoft-dotnet-core).
+
+If there's a platform that you require that is available in its own Docker image, ask yourself whether it would be better to use that image and add .NET Core to it or would it be better to use the .NET Core image as the base and add the platform to it. An example scenario is using .NET Core with MySQL; determine whether you would prefer to start with the MySQL image and install .NET Core onto it or start with the .NET Core image and install MySQL onto it.
```
MichaelSimons: I realize this is an example, but MySQL doesn't feel like a good one.  The recommendation here would be to run MySQL in a separate container.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Custom Application Images
+
+If you're building an image for a custom application that is not to be intended to be publicly consumed as a platform for other applications, you can get by with a simpler Dockerfile implementation compared to [general platform images](#general-platform-images) if you choose. Because the image is only intended for your own organization's purposes, the need for transparency in the Dockerfile is lessened. Convenience can trump clarity in this case.
```
MichaelSimons: I feel like this paragraph is someone out of place because there isn't the background which explains the value of `clarity`.  Event though I feel this is the more common case, I think the article read better with `General Platform Images` first.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+
+For custom application images, it is recommended that you install .NET Core by [package manager](#installing-from-a-linux-package-manager) (Linux only) or [dotnet-install script](#installing-from-dotnet-install-script) (Linux/Windows).
+
+## General Platform Images
+
+If you're building an image for a platform that is intended to be publicly consumed, you should strongly consider following the Docker [guidelines](https://github.com/docker-library/official-images) for official images. Those guidelines will inject a high level of quality in your images and instill trust and confidence by those consuming it. One of the guidelines is on [clarity](https://github.com/docker-library/official-images#clarity) which can be accomplished by installing .NET Core by [binary archive](#installing-from-a-binary-archive) (Linux/Windows) or [package manager](#installing-from-a-linux-package-manager) (Linux only), the two recommended approaches for general platform images.
```
MichaelSimons: I feel like a sentence on the meaning of clarity here would be helpful so that the reader can get a basic understanding without having to traverse the link.

> One of the guidelines is on [clarity](https://github.com/docker-library/official-images#clarity), write the Dockerfile in such a way that it is obvious what components are installed and where they are retrieved from.  This can be accomplished by installing .NET Core by [binary archive](#installing-from-a-binary-archive) (Linux/Windows) or [package manager](#installing-from-a-linux-package-manager) (Linux only), the two recommended approaches for general platform images.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+
+When authoring your Dockerfiles, you can look to the official [.NET Core Dockerfiles](https://github.com/dotnet/dotnet-docker) as a template for the install steps. There are several variations depending on the .NET Core version, OS type, and architecture being used.
+
+In addition to installing .NET Core, you'll also need to ensure that the [prerequisites](https://github.com/dotnet/core/blob/master/Documentation/prereqs.md) are installed. The [.NET Core Dockerfiles](https://github.com/dotnet/dotnet-docker) also demonstrate how that can be done.
+
+In the spirit of [clarity](https://github.com/docker-library/official-images#clarity), the Dockerfiles for the official .NET Docker images do not use a general purpose script for installing .NET Core. Rather, they explicity provide each step of the installation process and reference the exact URL of the binary archive.
+
+Example:
+
+```Dockerfile
```
MichaelSimons: I think we should be including a Windows variant here as well.  I think it is fair to assume a large number of the our Windows users will not have an understanding of and will be intimidated by the Linux example.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+
+For Linux, you may prefer to use your Linux distro's package manager to install .NET Core rather than directly from a binary archive. Each .NET Core release provides on how to install .NET Core from the various distro package managers. To locate the instructions, visit the [.NET Core download page](https://dotnet.microsoft.com/download/dotnet-core), select the desired version, click the `Package Manager Instructions: x64` link, and then select the desired distro. As an example, you can find the package manager instructions for .NET Core 3.0 [here](https://dotnet.microsoft.com/download/linux-package-manager/rhel7/sdk-3.0.100). Using a package manager allows for easier maintenance since you only have to reference the major/minor version of the release and you'll get servicing releases "for free". (This is true as long as you are mindful of Docker's caching functionality. You may need to build with the `--no-cache` option to force the build to re-execute the commands that install .NET Core in order to get an updated servicing release. Alternatively, you can specify the full version (major/minor/build) and increment that with each service release.)
```
MichaelSimons:
```suggestion
For Linux, you may prefer to use your Linux distro's package manager to install .NET Core rather than directly from a binary archive. Each .NET Core release provides instructions on how to install .NET Core from the various distro package managers. To locate the instructions, visit the [.NET Core download page](https://dotnet.microsoft.com/download/dotnet-core), select the desired version, click the `Package Manager Instructions: x64` link, and then select the desired distro. As an example, you can find the package manager instructions for .NET Core 3.0 [here](https://dotnet.microsoft.com/download/linux-package-manager/rhel7/sdk-3.0.100). Using a package manager allows for easier maintenance since you only have to reference the major/minor version of the release and you'll get servicing releases "for free". (This is true as long as you are mindful of Docker's caching functionality. You may need to build with the `--no-cache` option to force the build to re-execute the commands that install .NET Core in order to get an updated servicing release. Alternatively, you can specify the full version (major/minor/build) and increment that with each service release.)
```

MichaelSimons: I feel like it is worthwhile to call out that the prereqs in this case get installed for you.

MichaelSimons: I feel the the () section at the end is too long for parens.  The last sentence doesn't feel like it should be within the parens.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,135 @@
...
+    && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+```
+
+Example (Windows):
+
+```Dockerfile
+# escape=`
+
+FROM mcr.microsoft.com/windows/servercore:1903
+RUN powershell -Command `
```
MichaelSimons: Should this be specifying `$ErrorActionPreference = 'Stop'`?

samples/snippets/installing-dotnet.md
mthalman: Actually, my understanding of this was incorrect after thinking about it some more.  I was mistakenly thinking that an app that depended on files contained only in Server Core could run on Nano if it was published as self-contained.  But that's not the case.  So really the scenario that this is useful in is for Linux.

samples/snippets/installing-dotnet.md
mthalman: In that case, I may just omit an example entirely.  My intent was to provide an example of a well-known platform that is fairly easy to install; MySQL fits that well.  That makes it a decision that could go either way as to which base image to use.  Another example would be Node but that requires fairly advanced setup so it would make more sense to just use that as the base image.  I can't find a better example than MySQL, any ideas?

samples/snippets/installing-dotnet.md
MichaelSimons: I'm not following you.  I see this as a possible solution for the System.Drawing scenario.  .NET Core only has Nano Server based images which do not include the gliplus assemblies.  In this case, you can use the existing SDK images to publish a standalone app and then copy the app into an image based directly on windows/servercore.  This will work because windows/servercore includes the gdiplus assemblies.

samples/snippets/installing-dotnet.md
MichaelSimons: The Node scenario crossed my mind as well.  PowerShell Core in a runtime image?

samples/snippets/installing-dotnet.md
mthalman: Oh, gotcha.  I wasn't catching on that you were referring to use windows/servercore directly as the base image.

samples/snippets/installing-dotnet.md
```diff
@@ -0,0 +1,159 @@
...
+RUN curl -SL --output dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet/Runtime/$DOTNET_VERSION/dotnet-runtime-$DOTNET_VERSION-linux-x64.tar.gz \
+    && dotnet_sha512='0cabf85877eb3ee0415e6f8de9390c95ec90fa8f5a0fdb104f1163924fd52d89932a51c2e07b5c13a6b9802d5b6962676042a586ec8aff4f2a641d33c6c84dec' \
+    && echo "$dotnet_sha512 dotnet.tar.gz" | sha512sum -c - \
+    && mkdir -p /usr/share/dotnet \
+    && tar -zxf dotnet.tar.gz -C /usr/share/dotnet \
+    && rm dotnet.tar.gz \
+    && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+```
+
+Example (Windows):
```
MichaelSimons: To make this a more universally applicable sample, consider wrapping the `RUN` content within `powershell -Command...`.

PR #1444, opened 2019-11-05
Fix the setting of PowerShell telemetry environment variable on Windows SDK images

### LINKED ISSUE #1402: POWERSHELL_DISTRIBUTION_CHANNEL env var is not set correctly in the Windows SDK images.
The POWERSHELL_DISTRIBUTION_CHANNEL env variable is set in the builder image when it should be set in the sdk image.  I believe the result of this is that the PowerShell telemetry is not logging a distribution channel.

adityapatwardhan:
Fixes #1402 

Sets the variable in SDK image instead of the builder image.

### DIFF
--- a/3.0/sdk/nanoserver-1903/amd64/Dockerfile
+++ b/3.0/sdk/nanoserver-1903/amd64/Dockerfile
@@ -19,7 +19,8 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
     Remove-Item -Force dotnet.zip
 
 # Install PowerShell global tool
-ENV POWERSHELL_VERSION=7.0.0-preview.4
+ENV POWERSHELL_VERSION=7.0.0-preview.4 `
+    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetCoreSDK-NanoServer-1903
 
 RUN Invoke-WebRequest -OutFile PowerShell.Windows.x64.$ENV:POWERSHELL_VERSION.nupkg https://pwshtool.blob.core.windows.net/tool/$ENV:POWERSHELL_VERSION/PowerShell.Windows.x64.$ENV:POWERSHELL_VERSION.nupkg; `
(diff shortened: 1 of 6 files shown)


3.1/sdk/nanoserver-1903/amd64/Dockerfile
```diff
@@ -50,7 +49,9 @@ ENV DOTNET_RUNNING_IN_CONTAINER=true `
     # Enable correct mode for dotnet watch (only mode supported in a container)
     DOTNET_USE_POLLING_FILE_WATCHER=true `
     # Skip extraction of XML docs - generally not useful within an image/container - helps performance
-    NUGET_XMLDOC_MODE=skip
```
MichaelSimons: Please sort ENVs to help readability and maintenance.

3.0/sdk/nanoserver-1903/amd64/Dockerfile
```diff
@@ -19,8 +19,7 @@ RUN Invoke-WebRequest -OutFile dotnet.zip https://dotnetcli.blob.core.windows.ne
     Remove-Item -Force dotnet.zip
 
 # Install PowerShell global tool
```
MichaelSimons: I wasn't thinking we would back port this to 3.0.  It is technically a breaking change from an image perspective.  3.0 isn't LTS so it will go out of support in the near future once 3.1 is released.

3.1/sdk/nanoserver-1903/amd64/Dockerfile
adityapatwardhan: It seems sorted by variable name, am I missing something?

3.0/sdk/nanoserver-1903/amd64/Dockerfile
adityapatwardhan: I can revert the changes from 3.0.

3.1/sdk/nanoserver-1903/amd64/Dockerfile
MichaelSimons: You are correct, excuse me - I don't know what I saw when I made that comment.

PR #1446, opened 2019-11-06
Add unit tests to verify the environment variables within the images

MichaelSimons:
Add unit test to validate environment variables are set uniformly across all of the images.


tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -102,6 +102,31 @@ private static string GetFilterRegexPattern(string filterEnvName)
             return filter != null ? $"^{Regex.Escape(filter).Replace(@"\*", ".*").Replace(@"\?", ".")}$" : null;
         }
 
+        [Theory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifySdkImage_EnvironmentVariables(ImageData imageData)
```
mthalman: What is the performance impact of these tests?  Each call to `VerifyEnvironmentVariable` ends up executing a `docker run` command.  Things could be improved performance-wise by validating all the environment variables as a set rather than individually.

MichaelSimons: Each call is 2 to 3 seconds in Linux and 3 to 5 seconds on Windows.

mthalman: So it would seem a fair amount of time could be saved by consolidating things into one call.  This could be implemented by having a data structure that contains the set of env. var. names and their expected values which `GetEnvironmentVariable` consumes to construct the command that would output all the variable values on separate lines.

MichaelSimons: I refactored to make one docker run per image.

tests/Microsoft.DotNet.Docker.Tests/EnvironmentVariableInfo.cs
```diff
@@ -0,0 +1,102 @@
...
+                delimiterEscape='\\';
+            }
+            else
+            {
+                echoParts = variables.Select(envVar => $"%{envVar.Name}%");
+                invokeCommand = $"CMD /S /C";
+                delimiterEscape='^';
+            }
+
+            string echoArgs = echoParts.Aggregate((working, next) => $"{working}{delimiterEscape}{delimeter}{next}");
```
mthalman:
```suggestion
            string echoArgs = String.Join(delimiterEscape + delimiter, echoParts);
```

This achieves the same result less verbosely.

tests/Microsoft.DotNet.Docker.Tests/EnvironmentVariableInfo.cs
```diff
@@ -0,0 +1,102 @@
...
+            this.AllowAnyValue = allowAnyValue;
+        }
+
+        public static void Verify(
+            IEnumerable<EnvironmentVariableInfo> variables,
+            DotNetImageType imageType,
+            ImageData imageData,
+            DockerHelper dockerHelper)
+        {
+            GetActualValues(variables, imageType, imageData, dockerHelper);
```
mthalman: To me it seems weird to have a method named `Verify` populating the values of a bunch of objects.  From the point of view of code that consumes `EnvironmentVariableInfo`, it exposes an `ActualValue` value that isn't populated right away and isn't done so until some static method on the class gets called.  Just seems off to me.  But do you even need to have an `ActualValue` property?  Can't you do the `Assert` calls at the point where you're currently setting `ActualValue`?  To me, that would be a cleaner design because you're not exposing state that isn't set in a clear manner.  And that way you also only need to do one loop through the variables.

MichaelSimons: Good feedback, adjusted per your feedback.

PR #1453, opened 2019-11-07
Use blob hostname for accessing SHAs

mthalman:
The changes from #1450 caused the update-dependencies tool to fail when attempting to calculate the checksums.  It constructs a URL based off of `dotnetcli.blob.core.net` and ends up trying to access `dotnetclichecksums.azureedge.net` which doesn't exist.  To keep the logic simple, I've just reverted the change in that file back to using `dotnetcli.blob.core.net`.

### DIFF
--- a/eng/update-dependencies/DockerfileShaUpdater.cs
+++ b/eng/update-dependencies/DockerfileShaUpdater.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.DotNet.VersionTools.Dependencies;
+using System;
 using System.Collections.Generic;
 using System.Security.Cryptography;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -26,7 +26,7 @@ public class DockerfileShaUpdater : FileRegexUpdater
         private static readonly string s_versionPattern =
             $"ENV (?<{EnvNameGroupName}>(DOTNET|ASPNETCORE)_[\\S]*VERSION) (?<{ValueGroupName}>[\\S]*)";
         private static readonly string s_urlPatternFormat =
-            $"(?<{ValueGroupName}>https://dotnetcli.azureedge.net/[^;\\s]*{{0}})";
+            $"(?<{ValueGroupName}>https://dotnetcli.blob.core.net/[^;\\s]*{{0}})";
```
MichaelSimons: Doesn't this need to remain `dotnetcli.azureedge.net`?  This is used in the regex used on the Dockerfiles so it needs to use the new URL.

mthalman: Good point.  Fixed

Issue #1447, opened 2019-11-06
Refactor Dockerfiles to consolidate number of ENV layers in the resulting images


MichaelSimons:
Per the [Dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers) it is a good practice to minimize the number of Docker layers.  As the .NET Core Images have evolved over time, there have been several ENV instructions added for various reasons that should get consolidated.  The majority of times the separate ENV instructions were added for readability reasons.  

The [3.1 Alpine SDK dockerfile](https://github.com/dotnet/dotnet-docker/blob/master/3.1/sdk/alpine3.10/amd64/Dockerfile) is a prime example as the are four ENV layers that could easily be consolidated to one yet maintain the readability.

1. Version variables should be removed (`DOTNET_SDK_VERSION` and `POWERSHELL_VERSION`).  The intended use is within the Dockerfile itself.  The same values are available via other means within the resulting image (e.g. `dotnet --version`).
2. Remaining variables should be consolidated into primary ENV layer.  The existing comment convention maintains the readability.

PR #1459, opened 2019-11-12
Refactor to reduce layers and consolidate ENVs

MichaelSimons:
* Consolidate the number of ENV layers
* Reformat ENV instruction to eliminate top level comment which can be confusing since the top level comment is not applicable to the entire layer
* Consolidated the first run trigger into the dotnet install layer.  This layer used to be separate because of its size when it included the NuGet package cache.  Now that the layer is trivially small, it makes sense to consolidate it.
* Moved the static layers (e.g. ENV) to the top.  This will be beneficial when we start utilizing a build cache.

Related to #1447


mthalman:
I like these changes.

It makes sense to avoid exposing to the container any named values that are only meant for the Dockerfile.

I also like the consolidation of all the environment variables in a single location within the Dockerfile; makes for easier readability.

I always did think it did look a bit wonky having the initial environment variable's comment appear above the `ENV` declaration.  Your proposal is a better solution.

I'm all for getting these changes into 3.1.

eng/update-dependencies/VariableHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+
+            string envPattern = $"ENV {upperCasedVariableGroup} (?<{ValueGroupName}>[^\r\n]*)";
+            string linuxPattern = $"\\b{lowerCasedVariableGroup}=(?<{ValueGroupName}>[^\\s]*)";
+            string windowsPowerShellPattern = $"\\${lowerCasedVariableGroup} = '(?<{ValueGroupName}>[^\\s]*)'";
+            string windowsCmdPattern = $"\\bset \"{lowerCasedVariableGroup}=(?<{ValueGroupName}>[^\\s]*)\"";
+
+            return new Regex($"{envPattern}|{linuxPattern}|{windowsPowerShellPattern}|{windowsCmdPattern}");
+        }
+    }
+}
```
mthalman: EOF character

eng/update-dependencies/VariableHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+            string joinedVariableNames = string.Join('|', variableNames);
+            string upperCasedVariableGroup = string.Format(variableGroupFormat, joinedVariableNames.ToUpperInvariant());
+            string lowerCasedVariableGroup = string.Format(variableGroupFormat, joinedVariableNames.ToLowerInvariant());
+
+            string envPattern = $"ENV {upperCasedVariableGroup} (?<{ValueGroupName}>[^\r\n]*)";
+            string linuxPattern = $"\\b{lowerCasedVariableGroup}=(?<{ValueGroupName}>[^\\s]*)";
+            string windowsPowerShellPattern = $"\\${lowerCasedVariableGroup} = '(?<{ValueGroupName}>[^\\s]*)'";
+            string windowsCmdPattern = $"\\bset \"{lowerCasedVariableGroup}=(?<{ValueGroupName}>[^\\s]*)\"";
+
+            return new Regex($"{envPattern}|{linuxPattern}|{windowsPowerShellPattern}|{windowsCmdPattern}");
```
mthalman: Is it even necessary to account for `envPattern` still?  This would apply to `upperCasedVariableGroup` as well, as the sole dependency of `envPattern`.

MichaelSimons: Yes because the tool is useful to run to update 2.1, 2.2, and 3.0 Dockerfiles during servicing releases.

3.1/sdk/nanoserver-1809/arm32v7/Dockerfile
```diff
@@ -2,14 +2,22 @@
...
+    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetCoreSDK-NanoServer-1809-arm32
+
 # Install .NET Core SDK
 RUN set "dotnet_sdk_version=3.1.100-preview3-014645" `
     && call curl -SL --output dotnet.zip https://dotnetcli.azureedge.net/dotnet/Sdk/%dotnet_sdk_version%/dotnet-sdk-%dotnet_sdk_version%-win-arm.zip `
     && mkdir "%ProgramFiles%\dotnet" `
     && tar -zxf dotnet.zip -C "%ProgramFiles%\dotnet" `
-    && del dotnet.zip `
-    # Trigger first run experience by running arbitrary cmd to populate local package cache
-    && "%ProgramFiles%\dotnet\dotnet" help
```
mthalman: No more `dotnet help`?

MichaelSimons: This is technically not needed because the PS layer does a `dotnet tool install`.  The presence of this, triggers the Windows ARM32 issue.

PR #1470, opened 2019-11-16
to see if build has been triggered

Motaz-Al-Zoubi:


MichaelSimons:
@Motaz-Al-Zoubi, Can you explain the purpose of this PR?

PR #1477, opened 2019-11-19
Fix regression in Invoke-ReadmeGeneration.ps1

MichaelSimons:
Add the missing `

I also changed the parameter order to group the options.


mthalman:
Wouldn't it be better to fix this in docker-tools and let it propagate from there?

MichaelSimons:
I was trying to get a quick fix in to unblock my work to add Ubuntu 20.04 images.  Made PR in docker-tools - https://github.com/dotnet/docker-tools/pull/336

Issue #1454, opened 2019-11-08
SDK & ASP.Net images contain files & directories with insecure permissions


paulgear:
## Steps to reproduce the issue
1. docker run -ti mcr.microsoft.com/dotnet/core/sdk:2.1-bionic
2. find / -xdev -type d \( -perm -0002 -a ! -perm -1000 \) -ls  # <-- find world-writable directories without the sticky bit set
3. find / -xdev -type f -perm -o+w -ls  # <-- find world-writable files
4. find / -xdev \( -nouser -o -nogroup \) -ls  # <-- find files without a registered user or group

## Expected behavior
Should be no output

## Actual behavior
Here is a non-exhaustive list of dangerous files/directories:
https://pastebin.ubuntu.com/p/P7Kgq6WTf3/

Some of the files in question are .Net DLLs, which I suspect could be used to gain a foothold on a container running externally-facing services and then used to pivot to gain greater access.

## Additional information (e.g. issue happens only occasionally)

This affects -aspnet and -sdk variants of the Docker images, but didn't affect the -runtime versions when I tested.

## Output of `docker version`

Not terribly relevant, but here you go:
```
Client: Docker Engine - Community
 Version:           19.03.4
 API version:       1.40
 Go version:        go1.12.10
 Git commit:        9013bf583a
 Built:             Fri Oct 18 15:52:22 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.4
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.10
  Git commit:       9013bf583a
  Built:            Fri Oct 18 15:50:54 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.5
  GitCommit:        bb71b10fd8f58240ca47fbb579b9d1028eea7c84
 runc:
  Version:          1.0.0-rc6+dev
  GitCommit:        2b18fe1d885ee5083ef9f0838fee39b62d653e30
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

## Output of `docker info`

```
Client:
 Debug Mode: false

Server:
 Containers: 356
  Running: 1
  Paused: 0
  Stopped: 355
 Images: 1413
 Server Version: 19.03.4
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
[...]

PR #1493, opened 2019-11-22
Clean up file ownership issues

mthalman:
Installing PowerShell in SDK image leaves behind world-writable directories without sticky bit set, namely `/tmp/NuGetScratch`.  These changes fix this in the 3.1 SDK images by clearing the local NuGet resources.

In addition, the files extracted from the .NET tar files end up not having ownership by the user.

Adding a unit test to verify these issues in the 3.1 SDK images.

Related: #1454

### DIFF
--- a/3.1/aspnet/alpine3.10/amd64/Dockerfile
+++ b/3.1/aspnet/alpine3.10/amd64/Dockerfile
@@ -6,5 +6,5 @@ RUN aspnetcore_version=3.1.0 \
     && wget -O aspnetcore.tar.gz https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz \
     && aspnetcore_sha512='fa5e4ae71134a8a6db4ad6a247d3e31406673e03f0a64f7faaad3d84cfb3b70d2cf69e9d9abc1f8688138907d4ddd37cd908669999d85a87892e164053c63847' \
     && echo "$aspnetcore_sha512  aspnetcore.tar.gz" | sha512sum -c - \
-    && tar -zxf aspnetcore.tar.gz -C /usr/share/dotnet ./shared/Microsoft.AspNetCore.App \
+    && tar --no-same-owner -zxf aspnetcore.tar.gz -C /usr/share/dotnet ./shared/Microsoft.AspNetCore.App \
     && rm aspnetcore.tar.gz
--- a/3.1/aspnet/alpine3.10/amd64/Dockerfile
+++ b/3.1/aspnet/alpine3.10/amd64/Dockerfile
@@ -6,5 +6,5 @@ RUN aspnetcore_version=3.1.0 \
     && wget -O aspnetcore.tar.gz https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz \
     && aspnetcore_sha512='fa5e4ae71134a8a6db4ad6a247d3e31406673e03f0a64f7faaad3d84cfb3b70d2cf69e9d9abc1f8688138907d4ddd37cd908669999d85a87892e164053c63847' \
     && echo "$aspnetcore_sha512  aspnetcore.tar.gz" | sha512sum -c - \
-    && tar --no-same-owner -zxf aspnetcore.tar.gz -C /usr/share/dotnet ./shared/Microsoft.AspNetCore.App \
+    && tar -ozxf aspnetcore.tar.gz -C /usr/share/dotnet ./shared/Microsoft.AspNetCore.App \
     && rm aspnetcore.tar.gz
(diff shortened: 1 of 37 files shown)


tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -105,6 +105,28 @@ private static string GetFilterRegexPattern(string filterEnvName)
...
+        public void VerifySDKImage_WorldWritableFilesCheck(ImageData imageData)
+        {
+            if (imageData.Version < new Version("3.1") || !DockerHelper.IsLinuxContainerModeEnabled)
+            {
+                return;
+            }
+
+            string worldWritableDirectoriesWithoutStickyBitCmd = @"find / -xdev -type d \( -perm -0002 -a ! -perm -1000 \)";
+            string worldWritableFilesCmd = "find / -xdev -type f -perm -o+w";
+            string command = $"/bin/sh -c \"{worldWritableDirectoriesWithoutStickyBitCmd} && {worldWritableFilesCmd}\"";
```
MichaelSimons: Can the tests also check for files without a registered user or group?

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -105,6 +105,28 @@ private static string GetFilterRegexPattern(string filterEnvName)
             return filter != null ? $"^{Regex.Escape(filter).Replace(@"\*", ".*").Replace(@"\?", ".")}$" : null;
         }
 
+        [Theory]
```
MichaelSimons: This test should get run for all of the image variants, not just the `sdk`.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
mthalman: That test would fail.  Basically all of the files in `/user/share/dotnet` would be the culprits.  These changes wouldn't address that.

MichaelSimons: I missed that earlier.  Shouldn't we be setting the owner to root?

MichaelSimons: To fix that we could extract with `--no-same-owner`

3.1/aspnet/alpine3.10/amd64/Dockerfile
```diff
@@ -6,5 +6,5 @@ RUN aspnetcore_version=3.1.0 \
     && wget -O aspnetcore.tar.gz https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz \
     && aspnetcore_sha512='fa5e4ae71134a8a6db4ad6a247d3e31406673e03f0a64f7faaad3d84cfb3b70d2cf69e9d9abc1f8688138907d4ddd37cd908669999d85a87892e164053c63847' \
     && echo "$aspnetcore_sha512  aspnetcore.tar.gz" | sha512sum -c - \
-    && tar -zxf aspnetcore.tar.gz -C /usr/share/dotnet ./shared/Microsoft.AspNetCore.App \
+    && tar --no-same-owner -zxf aspnetcore.tar.gz -C /usr/share/dotnet ./shared/Microsoft.AspNetCore.App \
```
MichaelSimons: I know I mentioned `--no-same-owner`, but I just noticed `-o` is the supported short form.  Since we use the short form for the other options, what do you think about using it here to be consistent?

3.1/runtime/alpine3.10/arm64v8/Dockerfile
```diff
@@ -7,6 +7,6 @@ RUN dotnet_version=3.1.0 \
     && dotnet_sha512='417858bb24e1ec7337c7b85dd2c16ba3b16182bb0839e7c67d46b4ba38d430dd95a58e5ec533586c0cd4b69fc0f82323297eb4c5fbd8c2e8a5f9a1d36a3658f9' \
     && echo "$dotnet_sha512  dotnet.tar.gz" | sha512sum -c - \
     && mkdir -p /usr/share/dotnet \
-    && tar -C /usr/share/dotnet -xzf dotnet.tar.gz \
+    && tar --no-same-owner -C /usr/share/dotnet -oxzf dotnet.tar.gz \
```
MichaelSimons:
```suggestion
    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz \
```

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -109,7 +109,8 @@ private static string GetFilterRegexPattern(string filterEnvName)
         [MemberData(nameof(GetImageData))]
         public void VerifyImage_InsecureFilesCheck(ImageData imageData)
         {
-            if (imageData.Version < new Version("3.1") || !DockerHelper.IsLinuxContainerModeEnabled)
+            if (imageData.Version < new Version("3.1") || !DockerHelper.IsLinuxContainerModeEnabled ||
+                (imageData.OS.Contains("alpine") && imageData.IsArm))
```
MichaelSimons: There is no way to check validate on Alpine arm?

PR #1495, opened 2019-11-23
WIP: Add Amazon Linux 2

mattlorimor:
### What?
This PR is a first stab at adding [Amazon Linux 2](https://aws.amazon.com/amazon-linux-2/) variants of the images.

### Why?
* Amazon has shown consistent strong commitment to portability, security, and performance of of their Amazon Linux 2 distro and Docker image.
* Security issues in Amazon Linux 2 and its packages are fixed at a cadence that helps when needing a container to run in a high-compliance environment.

### Testing
I locally test built the 3.0 `runtime-deps`, `runtime`, and `sdk`; added a test `Dockerfile` to `samples\`; built it; and ran the container. This was the output:

image

So, it looks like the entire chain built okay and ran the sample `dotnetapp` project.

### Known Issues
Unless I'm missing something, the only thing I have seen that will be an issue is that there is no equivalent for `liblttng-ust0` on Amazon Linux 2 that I can find. As far as I can tell, [this will result in `perfcollect` not behaving properly](https://github.com/dotnet/corefx-tools/issues/91) - resulting in a specific method of tracing not working.

### OS Packages
I did my best to match package-for-package what it looked like was required in the `runtime-deps`. Below is a table that shows the breakdown and a comparison to Amazon Linux 2 for what packages are installed to `runtime-deps\bionic\amd64\Dockerfile`. Extra scrutiny around this is definitely needed.

| package_in_stretch | already_in_al2 | package_in_al2 |
|--------------------|----------------|----------------|
| libc6              | yes            | glibc          |
| libgcc1            | yes            | libgcc         |
| libgssapi-krb5-2   | yes            | krb5-libs      |
| libicu60           | no             | libicu         |
| liblttng-ust0      | no             | ? (n/a)        |
| libssl1.0.0        | no             | libssl-devel   |
| libstdc++6         | yes            | libstdc++      |
| zlib1g             | yes            | zlib1g         |

### Misc
I realize that this PR may be unwanted as it is entirely possible that you may not wish to add another OS into the mix at this time. I can doubly understand it given the issue I mentioned around `perfcollect`. If it is the case that this is an unwanted addition, I'll probably just end up building something like this for our internal use case. I simply figured that if we were running into a desire to have this that it was likely there are others out there as well.


mattlorimor:
I am unsure why the build is throwing an `A value was not found for the ARG '$REPO'` error when trying to `Generate platformVersionedOs Matrix`. I must have screwed up something in the `manifest.json`, but I'm not sure what.

mthalman:
While we very much appreciate the work you put into this and your desire to contribute, we're going to decline this PR.  There is engineering/compute/storage cost associated with each distro that is supported.  We thoroughly evaluate whether a particular distro is right to be included in the official images.  At this time, we're not looking to take on another distro.  The only one that's been considered right now is CentOS.

In the future, it's suggested to submit an issue for requests like this prior to putting in all the work of a PR.

mattlorimor:
@mthalman - Makes sense! As mentioned in the PR, I totally get it. It's really not a big deal.

Doing the PR got me to the point that I can give input internally on whether we could look at building our own AL2 images to use or whether we should use the Alpine images you already provide.

It's good to know that CentOS is being looked at as well.

Thanks for taking the time to reply.

PR #1501, opened 2019-11-26
Add the standard .NET contributing guide

MichaelSimons:


mthalman:
This is fine as a temporary solution but I think there could be a lot more content here specific to the Docker domain.  Some examples I can think of:
* Log an issue before submitting any major PRs so there can be discussion prior to the work.
* Where to find the PR builds and notes on the admittedly flakiness involved there.
* Any changes to a Dockerfile should be applied across the board where relevant (i.e. don't just make a change to a Dockerfile for a specific product version/platform when it's relevant to more cases than that).
* Functional PRs should be submitted to `nightly` branch first.  Readme PRs can be submitted to `master`.

MichaelSimons:
Agreed.  The first one is already covered in the common contribution guide.

PR #1533, opened 2019-12-17
Prevent force-pushes on PR in update-dependencies

### LINKED ISSUE #1114: Update dependencies infrastructure is causing PRs to get updated repeatedly
The update dependencies infrastructure is causing the PRs it creates to get updated repeatedly with the same changes via forced pushes.  In turn this triggers the PR validation to be re-run which is wasting machine usage.  Additionally the force push update is at times happening before the previous PR validation completes.  This can lead to cases where we are never able to get a "green" PR.  An example of this is https://github.com/dotnet/dotnet-docker/pull/1113.

Ideally the PR should not be getting updated when there are no new changes.

mthalman:
The current behavior of update-dependencies is that it will update an existing PR by force-pushing.  The consequence of this is that it triggers another build.  This happens even when there were no changes that were made from the previous state of the PR's branch.  That's because there's no logic to compare what the state of the PR's branch is.  All it does is detect what files have changed in the _local_ repo and produced a commit that contains those files and force-pushes them blindly into the PR.

Making use of `PullRequestCreator` isn't a good option in this case.  It has to do a force push because it doesn't know about or require a local repo to compare files against.  It's up to the caller to produce a correct commit.  In theory, we could implement logic in update-dependencies that requests the file content for every file in the PR's branch and compares that against the contents of the local repo.  But that would be incredibly inefficient and use up REST API quotas.

A better way is to simply clone the PR's repo/branch to the computer and let git do the comparison for us.  After cloning the repo to a temp location, update-dependencies clears out the contents to get it into a clean state.  Then it copies the contents of the repo where the changes were made into the temp repo.  Now we can use git to easily determine the diffs.  Just stage the files that changed and commit/push them back to the PR's branch.

Fixes #1114

### DIFF
--- a/eng/update-dependencies/Program.cs
+++ b/eng/update-dependencies/Program.cs
@@ -49,7 +49,6 @@ namespace Dotnet.Docker
                         await CreatePullRequestAsync();
                     }
                 }
-                await CreatePullRequestAsync();
             }
             catch (Exception e)
             {
@@ -214,14 +213,7 @@ namespace Dotnet.Docker
                 Repository.Clone($"https://github.com/{gitHubAuth.User}/{Options.GitHubProject}", tempRepoPath, cloneOptions);
 
                 // Remove all existing directories and files from the temp repo
-                foreach (string file in Directory.GetFiles(tempRepoPath))
-                {
-                    File.Delete(file);
-                }
-                foreach (DirectoryInfo dir in new DirectoryInfo(tempRepoPath).GetDirectories().Where(dir => dir.Name != ".git"))
-                {
-                    Directory.Delete(dir.FullName, true);
-                }
+                ClearRepoContents(tempRepoPath);
 
                 // Copy contents of local repo changes to temp repo
                 DirectoryCopy(".", tempRepoPath);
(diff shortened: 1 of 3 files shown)


eng/update-dependencies/update-dependencies.csproj
```diff
@@ -7,7 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="2.2.0-prerelease-02431-01" />
+    <PackageReference Include="LibGit2Sharp" Version="0.26.2" />
+    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="1.0.0-beta.19426.12" />
```
MichaelSimons: Is this the latest version from the Arcade feed?  I see 5.0.0-beta.19616.4.  It looks like the NuGet.config is not using the Arcade feed.

eng/update-dependencies/Program.cs
```diff
@@ -45,9 +46,10 @@ public static async Task Main(string[] args)
                     }
                     else
                     {
-                        await CreatePullRequestAsync(buildInfos);
+                        await CreatePullRequestAsync();
                     }
                 }
+                await CreatePullRequestAsync();
```
MichaelSimons: I don't understand this call.  Why is this called outside of `if (updateResults.ChangesDetected())` as well as inside?

eng/update-dependencies/Program.cs
```diff
@@ -157,14 +159,149 @@ private static async Task CreatePullRequestAsync(IEnumerable<IDependencyInfo> bu
...
+                        prOptions);
+                }
+                else
+                {
+                    UpdateExistingPullRequest(gitHubAuth, prOptions, commitMessage, upstreamBranch);
+                }
+            }
+        }
+
+        private static void UpdateExistingPullRequest(
```
MichaelSimons: Consider breaking this method apart to make it a little smaller.  It is well over 100 lines.  The pieces that jumped out at me were the two chunks of delete directory logic.  Having these encapsulated into a method would help simplify this method and would be consistent with how the CopyDirectory logic is encapsulated.

eng/update-dependencies/update-dependencies.csproj
mthalman: I just copied from file-pusher:
https://github.com/dotnet/docker-tools/blob/3788c781de06e973905cadae0ab80c0cecd09c05/eng/file-pusher/file-pusher.csproj#L11

Can you point me to the arcade feed?  I'm not finding it.

eng/update-dependencies/Program.cs
mthalman: That shouldn't be there.  Clean up error.

eng/update-dependencies/update-dependencies.csproj
MichaelSimons: According to the [arcade readme](https://github.com/dotnet/arcade)

> Packages are published daily to our tools feed:
>
> https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json
> 
> This feed is browsable from here:
> 
> https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-eng

Issue #391, opened 2018-02-27
Add unit tests for the samples pushed to Docker Hub


MichaelSimons:
@MichaelSimons commented on [Mon Apr 03 2017](https://github.com/dotnet/dotnet-docker-samples/issues/41)

Recently a bad image was built and pushed - sha256:2bd25764afa2c699c3d2ddd80d6b59c197fbc121c5ee58737cb0ec40e981a5c3.  

```
PS > docker run --rm microsoft/dotnet-samples@sha256:2bd25764afa2c699c3d2ddd80d6b59c197fbc121c5ee58737cb0ec40e981a5c3
The specified framework 'Microsoft.NETCore.App', version '1.0.4' was not found.
  - Check application dependencies and target a framework version installed at:
      C:\Program Files\dotnet\shared\Microsoft.NETCore.App
  - The following versions are installed:
      1.0.3
  - Alternatively, install the framework version '1.0.4'.
```

This was caused by the way the image was built.  The machine the build was performed on was had an older cached version of the 1.0.3 runtime image that was being used.  Unit tests should be created and run as part of the build process to prevent regressions like this from happening in the future.

---

@MichaelSimons commented on [Mon Apr 03 2017](https://github.com/dotnet/dotnet-docker-samples/issues/41#issuecomment-291283110)

cc @dleeapho, @kendrahavens

PR #1543, opened 2019-12-30
Unit test refactoring

mthalman:
Several unit test refactoring changes in order to support #391:
* Adds Category trait to unit tests so that specific categories of tests can be executed.
* Factors `ImageData` into an abstract class with a derived `ProductImageData` class.  `ImageData` is meant for generic images in general, providing support for sample images in the future.  `ProductImageData` provides functionality specific to the "product" images (SDK, runtime, etc).
* Upgrades the test project to .NET Core 3.1

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
@@ -19,7 +19,8 @@ namespace Microsoft.DotNet.Docker.Tests
 
         [Theory]
         [MemberData(nameof(GetImageData))]
-        public async Task VerifyAppScenario(ImageData imageData)
+        [Trait("Category", "aspnet")]
+        public async Task VerifyAppScenario(ProductImageData imageData)
         {
             ImageScenarioVerifier verifier = new ImageScenarioVerifier(imageData, DockerHelper, OutputHelper, isWeb: true);
             await verifier.Execute();
@@ -27,7 +28,8 @@ namespace Microsoft.DotNet.Docker.Tests
 
         [Theory]
         [MemberData(nameof(GetImageData))]
-        public void VerifyEnvironmentVariables(ImageData imageData)
+        [Trait("Category", "aspnet")]
+        public void VerifyEnvironmentVariables(ProductImageData imageData)
         {
             base.VerifyCommonEnvironmentVariables(imageData);
         }
--- a/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
@@ -8,6 +8,7 @@ using Xunit.Abstractions;
 
 namespace Microsoft.DotNet.Docker.Tests
 {
+    [Trait("Category", "aspnet")]
     public class AspnetImageTests : CommonRuntimeImageTests
     {
         public AspnetImageTests(ITestOutputHelper outputHelper)
@@ -19,7 +20,6 @@ namespace Microsoft.DotNet.Docker.Tests
 
         [Theory]
         [MemberData(nameof(GetImageData))]
-        [Trait("Category", "aspnet")]
         public async Task VerifyAppScenario(ProductImageData imageData)
         {
             ImageScenarioVerifier verifier = new ImageScenarioVerifier(imageData, DockerHelper, OutputHelper, isWeb: true);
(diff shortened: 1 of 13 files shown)


tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
```diff
@@ -19,15 +19,17 @@ public AspnetImageTests(ITestOutputHelper outputHelper)
 
         [Theory]
         [MemberData(nameof(GetImageData))]
-        public async Task VerifyAppScenario(ImageData imageData)
+        [Trait("Category", "aspnet")]
```
MichaelSimons: It looks like the TraitAttribute can be applied at the class level.  This would make the code easier to maintain and eliminate the potential risk of forgetting the attribute which may cause tests to get skipped in automation.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -19,9 +19,8 @@ protected ImageTests(ITestOutputHelper outputHelper)
         
         protected DockerHelper DockerHelper { get; }
         protected ITestOutputHelper OutputHelper { get; }
-        protected abstract DotNetImageType ImageType { get; }
```
MichaelSimons: Why is this being removed here?  Do you envision that the sample test classes would derive from this class?  If so, it feels like it would be valuable to have a ProductImageTests base class.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -178,14 +178,14 @@ private async Task RunTestAppImage(string image, bool runAsAdmin = false, string
             }
         }
 
-        private async Task VerifyHttpResponseFromContainer(string containerName)
+        public static async Task VerifyHttpResponseFromContainerAsync(string containerName, DockerHelper dockerHelper, ITestOutputHelper outputHelper)
```
MichaelSimons: Why is this being made public?

tests/run-tests.ps1
```diff
@@ -13,9 +13,27 @@ param(
     [string]$RepoPrefix,
     [switch]$DisableHttpVerification,
     [switch]$PullImages,
-    [string]$ImageInfoPath
+    [string]$ImageInfoPath,
+    [ValidateSet('runtime', "runtime-deps", 'aspnet', 'sdk')]
+    [string[]]$TestCategories = @('runtime', "runtime-deps", 'aspnet', 'sdk')
```
MichaelSimons: Once the samples are added, is the idea that they would get skipped by default?

tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
mthalman: I didn't realize that could be done.

I also realized that I can't define a single base `VerifyInsecureFiles` method in `CommonRuntimeImageTests` with multiple Trait attributes. Doing so will cause other execution of the test for test categories that were not specified due to inheritance.  For example, if you specify to only exist the `runtime` category, the `VerifyInsecureFiles` method will get "activated" because it has the `runtime` category.  But now that the method is activated and it is being inherited by other test classes, those other test classes will also run their logic on that test method.  So it gets run for `runtime`, `runtime-deps`, and `aspnet`.  To mitigate this, I have to move that method down into each of the concrete test classes.

tests/Microsoft.DotNet.Docker.Tests/ImageTests.cs
mthalman: I've reverted these changes.  The intention is not to have the sample test class derive from `ImageTests`.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
mthalman: Reverted for now since it doesn't make sense in the current changes.  It will be used by sample tests though.

tests/run-tests.ps1
mthalman: Yes

tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
MichaelSimons: Does the class level attribute help here?  Does the trait get applied to base methods?

mthalman: Indeed it does help.  After moving the trait to the base class, it filters appropriately.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -178,14 +178,14 @@ private async Task RunTestAppImage(string image, bool runAsAdmin = false, string
             }
         }
 
-        private async Task VerifyHttpResponseFromContainer(string containerName)
+        private static async Task VerifyHttpResponseFromContainerAsync(string containerName, DockerHelper dockerHelper, ITestOutputHelper outputHelper)
```
MichaelSimons: Given these set of changes, it doesn't make sense to me that this was changed to a static method.

PR #1547, opened 2020-01-03
Tests for published sample images

### LINKED ISSUE #391: Add unit tests for the samples pushed to Docker Hub

mthalman:
Adds tests to verify that the sample console and ASP.NET projects that are published to Docker Hub work correctly when run in a container.

Note that these tests will only be executed as part of the existing samples build pipeline which only gets triggered on commits.  There will be no sample tests run for PRs.  This is because we can't use path filtering for an AzDO pipeline trigger when the source is GitHub, which means we can't run the sample pipeline on the condition when a sample project file is changed.  The only other option would be to always run the sample pipeline for all PRs and I don't think we want to do that, especially considering that the building of the sample images would be based on the images in Docker Hub, not the locally built images.

Fixes #391

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
@@ -10,18 +10,7 @@ namespace Microsoft.DotNet.Docker.Tests
     {
         public string GetImage(SampleImageType imageType, DockerHelper dockerHelper)
         {
-            string tagPrefix;
-            switch (imageType)
-            {
-                case SampleImageType.Dotnetapp:
-                    tagPrefix = "dotnetapp";
-                    break;
-                case SampleImageType.Aspnetapp:
-                    tagPrefix = "aspnetapp";
-                    break;
-                default:
-                    throw new NotSupportedException($"Unsupported image type '{imageType}'");
-            }
+            string tagPrefix = Enum.GetName(typeof(SampleImageType), imageType).ToLowerInvariant();
 
             string tag = GetTagName(tagPrefix, OS);
             string imageName = GetImageName(tag, "samples");
--- a/tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
@@ -13,10 +13,10 @@ namespace Microsoft.DotNet.Docker.Tests
             string tagPrefix;
             switch (imageType)
             {
-                case SampleImageType.Console:
+                case SampleImageType.Dotnetapp:
                     tagPrefix = "dotnetapp";
                     break;
-                case SampleImageType.Aspnet:
+                case SampleImageType.Aspnetapp:
                     tagPrefix = "aspnetapp";
                     break;
                 default:
--- a/tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
@@ -1,0 +1,34 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class SampleImageData : ImageData
+    {
+        public string GetImage(SampleImageType imageType, DockerHelper dockerHelper)
+        {
+            string tagPrefix;
+            switch (imageType)
+            {
+                case SampleImageType.Console:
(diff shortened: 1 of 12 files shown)


tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public static IEnumerable<object[]> GetImageData()
+        {
+            return TestData.GetSampleImageData()
+                .Select(imageData => new object[] { imageData });
+        }
+
+        [Theory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifyConsoleSampleImage(SampleImageData imageData)
```
MichaelSimons: The `SampleImage` suffix feels redundant in the test case names since these are contained within the `SampleImageTests`

MichaelSimons: I think the tests would be easier to understand if the test case names and `SampleImageType` names aligned with the tag names - `dotnetapp` and `aspnetapp`.

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    [Trait("Category", "sample")]
+    public class SampleImageTests
```
MichaelSimons: Since this class doesn't derive from `ImageTests` would it make sense to rename `ImageTests` to `ProductImageTests`?

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            return TestData.GetSampleImageData()
+                .Select(imageData => new object[] { imageData });
+        }
+
+        [Theory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifyDotnetSample(SampleImageData imageData)
+        {
+            string image = imageData.GetImage(SampleImageType.Console, DockerHelper);
```
MichaelSimons: I think the `SampleImageType.Console` name hurts the readability.  I think it would be easier to understand if it were named `dotnetapp`.

tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
```diff
@@ -13,10 +13,10 @@ public string GetImage(SampleImageType imageType, DockerHelper dockerHelper)
             string tagPrefix;
```
MichaelSimons: This could be simplified.

```
string tagPrefix = Enum.GetName(typeof(SampleImageType), imageType).ToLowerInvariant();
```

tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class SampleImageData : ImageData
+    {
+        public string GetImage(SampleImageType imageType, DockerHelper dockerHelper)
+        {
+            string tagPrefix = Enum.GetName(typeof(SampleImageType), imageType).ToLowerInvariant();
+
```
MichaelSimons: (nit) this blank line doesn't seem necessary.

PR #1554, opened 2020-01-06
Validate image size base integrity on PR build

### LINKED ISSUE #936: Enhance the usability of the image size validation tests
It would be valuable to run the image size validation as part of CI - https://github.com/dotnet/dotnet-docker/tree/nightly/tests/performance.  This will help detect regression upfront as the Dockerfiles are updated.  

In order for this to be really valuable #373 and #792 should be fixed.

mthalman:
* Updates the image size validation pipeline to make use of the new functionality from https://github.com/dotnet/docker-tools/pull/350.
* Includes an additional image size validation pipeline that runs for PRs. This will only validate the integrity of the baseline file and not cause any failure for any size differences.
* Updates the Validate-ImageSize PowerShell script to make use of the new command functionality in Image Builder.  This now allows the script to update the local baseline files.
* Updates the baseline files with the latest data.

Fixes #936


tests/performance/Validate-ImageSize.ps1
```diff
@@ -1,8 +1,21 @@
...
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+[cmdletbinding(
+    DefaultParameterSetName = 'Validate'
+)]
 param(
-    [switch]$UpdateBaselines,
     [switch]$UseLocalImages,
```
MichaelSimons: Should this be part of the `Update` parameter set?  How does it impact the `Validate` behavior?

mthalman: By not specifying a parameter set, it means it applies to all parameter sets.  `UseLocalImages` is relevant for both commands.  It's used by the base class `ImageSizeCommand.GetImageSize` method which is called by the commands here:
https://github.com/dotnet/docker-tools/blob/5e82dc55696e87b6bfa86c6e61e05f78c9b54503/src/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs#L51

and here:
https://github.com/dotnet/docker-tools/blob/5e82dc55696e87b6bfa86c6e61e05f78c9b54503/src/Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageSizeBaselineCommand.cs#L47

MichaelSimons: Of course, I got mixed up and was only thinking of the BaselineIntegrityOnly scenario of Validate.

build-and-test.ps1
```diff
@@ -1,16 +1,41 @@
...
+
+if ($Build) {
+    & ./eng/common/build-and-test.ps1 `
+        -VersionFilter $VersionFilter `
+        -OSFilter $OSFilter `
+        -ArchitectureFilter $ArchitectureFilter `
+        -OptionalImageBuilderArgs $OptionalImageBuilderArgs `
+        -SkipTesting:$Test
+}
+if ($Test) {
```
MichaelSimons: Is there any reason these changes shouldn't be pushed to the common build-and-test script?  It would be valuable for all repos to have the ability to just run tests.  I think it is also reasonable to push the TestCategories notion to the common script.

build-and-test.ps1
```diff
@@ -1,16 +1,41 @@
...
+    $Test = $true
+}
+
+if ($Build) {
+    & ./eng/common/build-and-test.ps1 `
+        -VersionFilter $VersionFilter `
+        -OSFilter $OSFilter `
+        -ArchitectureFilter $ArchitectureFilter `
+        -OptionalImageBuilderArgs $OptionalImageBuilderArgs `
+        -SkipTesting:$Test
```
MichaelSimons: Don't you want to always skip testing here?

build-and-test.ps1
```diff
@@ -1,16 +1,41 @@
...
 param(
     [string]$VersionFilter = "*",
     [string]$OSFilter = "*",
     [string]$ArchitectureFilter = "amd64",
+    [Parameter(ParameterSetName = "Build")]
+    [Parameter(ParameterSetName = "BuildAndTest")]
     [string]$OptionalImageBuilderArgs,
-    [switch]$SkipTesting = $false
+    [Parameter(ParameterSetName = "Build")]
+    [switch]$Build,
```
MichaelSimons: These names feel strange to me - they don't clearly convey the behavior.  For example - `build-and-test -test`.  I think the arguments need to more clearly indicate build only or test only.

build-and-test.ps1
mthalman: I long considered this and don't see much value in it.  It depends on how "correct" you want the common script to be.  The reason I modified this script to use parameter sets was to enforce the use of parameters that are only relevant to the particular scenario (i.e. don't pass TestCategories if you're only doing build).  If we apply that same sort of enforcement to the common script, that means this wrapper script now needs to call that script appropriately, it can't just map all of its parameters to all of the relevant parameters in the common script because that will lead to parameter set violations.  That means this script will need to have conditional logic for determine which mode is being used and call the common script appropriately.  That leads to basically the same logic as I've provided here.

I think the most appropriate thing to do is to keep these changes and apply them to dotnet-framework-docker repo as well.  Then update the common script to only handle build.  The wrapper scripts can then invoke the common build script or their local test script.

MichaelSimons: I'm fine with your proposal on refactoring the common build-and-test script to only handle building.

tests/run-tests.ps1
```diff
@@ -14,8 +14,8 @@ param(
     [switch]$DisableHttpVerification,
     [switch]$PullImages,
     [string]$ImageInfoPath,
-    [ValidateSet("runtime", "runtime-deps", "aspnet", "sdk", "image-size")]
-    [string[]]$TestCategories = @("runtime", "runtime-deps", "aspnet", "sdk")
+    [ValidateSet('runtime', "runtime-deps", 'aspnet', 'sdk', 'sample', "image-size")]
```
MichaelSimons: Curious why the mix of using double and single quotes here?

mthalman: Introduced when merging from nightly.  Fixed

PR #1574, opened 2020-01-13
Document how to retrieve the image artifact details

MichaelSimons:
This PR introduces documentation describing how to retrieve the licensing and source pedigree of the components contained within the .NET Core images.


documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+Navigating the [Package link](https://pkgs.alpinelinux.org/package/v3.11/main/x86/zlib) will show the detailed package information such as the license and originating source.
+
+![zlib package information](./image-artifact-details-alpine-package-details.png)
+
+### debian
+
+Licensing information is present in Debian images, as is demonstrated below.
+
+```console
+$ docker run --rm debian:buster-slim find . | grep copyright
```
mthalman: Instead of looking for copyright files, can't the same package lookup technique be used as Alpine:

`docker run --rm debian:buster-slim apt list -i`

Then you can search for the relevant package at https://packages.debian.org/index.  From the package's page, you can find the copyright file link.

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+<http://www.gnu.org/copyleft/gpl.txt> for the terms of the latest version
+of the GNU General Public License.
+```
+
+### ubuntu
+
+Licensing information is present in Ubuntu images, as is demonstrated below.
+
+```console
+$ docker run --rm ubuntu:bionic find . | grep copyright
```
mthalman: Same feedback on using package listing command as I made for debian.  Ubuntu's package search page is at https://packages.ubuntu.com/.

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+of the GNU General Public License.
+```
+
+### buildpack-deps
+
+.NET Core uses the Debian and Ubuntu variants of [`buildpack-deps`](https://hub.docker.com/_/buildpack-deps), enabling you to use the same patterns described above to retrieve the licensing information for these images.
+
+## Retrieving Base Image Pedigree Details
+
+.NET Core relies on the official Docker image repositories to provide detailed information about the images. This information includes layer information, source code, license information, etc. This information is stored in the [Official Images "Extended Information" repository](https://github.com/docker-library/repo-info) and is split into two types.
```
mthalman:
```suggestion
.NET Core relies on the official Docker image repositories to provide detailed information about the images. This information includes layer information, source code, license information, etc. This information is stored in the [Official Images "Extended Information" repository](https://github.com/docker-library/repo-info) and is split into two types:
```

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+* https://sources.debian.net/src/acl/2.2.53-4/debian/copyright/ (for direct copyright/license information)
+* http://snapshot.debian.org/package/acl/2.2.53-4/ (for access to the source package after it no longer exists in the archive)
+
+##### `dpkg` source package: `adduser=3.118`
+
+...
+
+### Pedigree Details for Base Images used by .NET Core Images
+
+The following links provide the pedigree details for the .NET Core linux base images.
```
mthalman:
```suggestion
The following links provide the pedigree details for the .NET Core Linux base images.
```

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+    # Enable detection of running in a container
+    DOTNET_RUNNING_IN_CONTAINER=true
+```
+
+The pattern for [Retrieving License Information From Base Images](#retrieving-license-information-from-base-images) can be repeated for packages that are added in .NET Core images (above and beyond the base image). The first package listed in the Dockerfile is `ca-certificates`, which will be used to demonstrate the pattern below.
+
+There is no guarantee that the requested package is not present in the base image, but it typically will not be (hence why it is explicitly installed), as is demonstrated below.
+
+```console
+$ docker run --rm debian:buster-slim find . | grep copyright | grep -c ca-certificates
```
mthalman: Again, instead of searching for copyright, I suggest using the distro-specific package search web pages.

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+```
+
+You can retrieve the source for any of these packages, as demonstrated below.  The first step is to find the package version.
+
+```console
+$ docker run -it --rm mcr.microsoft.com/dotnet/core/runtime-deps:3.1-buster-slim apt list | grep ca-certificates
+ca-certificates/now 20190110 all [installed,local]
+```
+
+The source for a particular version of a package can be retrieved with the [`apt-get source`](https://manpages.debian.org/buster/apt/apt-get.8.en.html) command. This example simply prints the source uris.
```
mthalman:
```suggestion
The source for a particular version of a package can be retrieved with the [`apt-get source`](https://manpages.debian.org/buster/apt/apt-get.8.en.html) command. This example simply prints the source URLs.
```

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+% docker run --rm mcr.microsoft.com/dotnet/core/aspnet:3.1-buster-slim find ./usr/share/dotnet | grep -i third
+./usr/share/dotnet/ThirdPartyNotices.txt
+./usr/share/dotnet/shared/Microsoft.AspNetCore.App/3.1.0/THIRD-PARTY-NOTICES.txt
+```
+
+### .NET Core SDK Image
+
+The [SDK image](https://hub.docker.com/_/microsoft-dotnet-core-sdk/) includes the .NET Core SDK, which includes various .NET Core components, with associated licenses and third party notice files.
+
+Note: The SDK image is based on [`buildpack-deps`](https://hub.docker.com/_/buildpack-deps), which includes components that we distribute but not use (like Python).
```
mthalman:
```suggestion
Note: The SDK image is based on [`buildpack-deps`](https://hub.docker.com/_/buildpack-deps), which includes components that we distribute but do not use (like Python).
```

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
+# .NET Core Image Artifact Details
+
+.NET Core container images contain multiple software components, each built from different source and with different licensing. It is important that users of these images can determine the license requirements of these images and have a link back to the source for pedigree purposes.
```
mthalman: Is pedigree a commonly used term in this context?  To me, pedigree (in the context of software) refers to the historical line of a product or company across many releases.  What's really being described here is itemizing and cross-referencing.  It just seems like an odd term.  Perhaps it can be explained further what its meaning is if it's going to be used.

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+
+This document is intended to be informative, primarily defining patterns for discovering license and pedigree information and describing expectations (whether the desired information can be found within the images). It does not list all of the license and pedigree information you might want. This information can change at any time, which also demonstrates why using the documented patterns is important.
+
+The combination of [.NET Core Dockerfiles](https://github.com/dotnet/dotnet-docker) and [.NET Core container images](https://hub.docker.com/_/microsoft-dotnet-core/) are the source and binary artifacts that are used in this document and intended to be used to determine the compliance of .NET Core containers.
+
+This document describes the licensing and source pedigree of the following components:
+
+* Base images (from Docker Hub)
```
mthalman: I feel like things should be laid out a bit differently.  There's a heading for `Base images` but then it immediately gets into talking about package licenses.  To me, that needs to be part of the `Packages` section which could be divided into two parts: base image and .NET.  And the only thing that should be in the `Base images` would be the repo info stuff.

documentation/image-artifact-details.md
MichaelSimons: Ideally, all images should be self-describing in terms of the licensing.  I think the ideal pattern is to retrieve the licence information from the images.  Because it is not possible with Alpine, the lookup method is suggested only for Alpine.

documentation/image-artifact-details.md
MichaelSimons: Changed to `URIs`.

documentation/image-artifact-details.md
mthalman: But this method relies on convention rather than a deterministic approach.  What if a package has a `license` file or some other name instead of `copyright`?

MichaelSimons: It appears debian packages are required to have a machine readable copyright with the name copyright - https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/.  This is what the Docker repo-info logic relies on as well - https://github.com/docker-library/repo-info/blob/master/.local-scripts/gather-dpkg.sh#L66.

As I previously stated, I think it is ideal to retrieve the copyright from the image itself.  I view the pattern described in the document as being sufficient for now with the longer term solution being the `tag-details` feature which will gather this information and present it in a friendly format that is easy to consume.

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+.NET Core images are provided for a set of Linux distros, delivered via [MCR](https://azure.microsoft.com/blog/microsoft-syndicates-container-catalog/) (and advertised on Docker Hub). The Linux distros are acquired from Docker Hub as part of building the .NET Core images. The following base images are delivered as a component of .NET Core images.
+
+* [alpine](https://hub.docker.com/_/alpine)
+* [debian](https://hub.docker.com/_/debian)
+* [ubuntu](https://hub.docker.com/_/ubuntu)
+* [buildpack-deps](https://hub.docker.com/_/buildpack-deps)
+
+.NET Core is dependent on those images, and the individuals/organizations that build them, to satisfactorily describe the licensing and pedigree of these images.
+
+## Retrieving License Information From Base Images
```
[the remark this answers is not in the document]
richlander: Yes and no. We should say something about Windows. However, Windows is proprietary so most of this is out of scope. We should make it clear that this is about open source Linux distros and then explain why Windows is out of scope (because it is a proprietary, licenses, OS).

documentation/image-artifact-details.md
richlander: Ideally, we would offer both. I like the simplicity of the grep command. It removes any fear that people might have in doing this. It is very telling that Docker uses the same pattern.

documentation/image-artifact-details.md
richlander: Pedigree is the term that I've used / heard used as it relates to source code and its associated licensing and origin. In any case, we're not talking about horse breeding.

documentation/image-artifact-details.md
richlander: Everything is a package! There are really two topics:

- How can I interrogate an image to determine licensing and version images of the packages it includes?
- Where can I look to learn about source pedigree information about an image from offline sources (since it is difficult/impossible to learn everything from an image).

It is also important to understand the copyright burden that comes from a base image vs a dev platform (like .NET) vs a higher layer.

richlander: Maybe we could rename "Packages" -> "Additional packages" ... would that help? I think "Base images" is the correct name. It's going to confuse folks calling it anything else.

mthalman: > Maybe we could rename "Packages" -> "Additional packages" ... would that help? I think "Base images" is the correct name. It's going to confuse folks calling it anything else.

Yes, "Additional Packages" would be better I think.  

> There are really two topics:
> 
> * How can I interrogate an image to determine licensing and version images of the packages it includes?
> * Where can I look to learn about source pedigree information about an image from offline sources (since it is difficult/impossible to learn everything from an image).

That's actually a pretty good synopsis and should be included in the doc IMO.  It's just a tad bit obscure on what the larger picture is here.

documentation/image-artifact-details.md
mthalman: Who's the audience here?  Do we expect technically-inclined lawyers to read this?  Are they going to know what it means?  I certainly don't think it would hurt to just have a single sentence that explains it for clarity.  BTW, `"source code pedigree"` yields 85 hits on Bing.

richlander: We could say: "have a link back to the source for copyright and pedigree purposes"

richlander:
Another thing ... I don't want use to boil the ocean here. The presence of this doc will be a MASSIVE step forward. Let's get it published and see what feedback we get.

documentation/image-artifact-details.md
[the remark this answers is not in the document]
MichaelSimons: Added a couple sentences clarifying the scope.

documentation/image-artifact-details.md
```diff
@@ -0,0 +1,476 @@
...
+
+This document describes the licensing and source pedigree of the following components:
+
+* Base images (from Docker Hub)
+* Packages (from Linux package managers)
+* .NET Core and other components (acquired as .tar.gz/.zip files)
+
+## Base Images used by .NET Core Images
+
+.NET Core images are provided for a set of Linux distros, delivered via [MCR](https://azure.microsoft.com/blog/microsoft-syndicates-container-catalog/) (and advertised on Docker Hub). The Linux distros are acquired from Docker Hub as part of building the .NET Core images. The following base images are delivered as a component of .NET Core images.
```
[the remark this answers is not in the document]
MichaelSimons: Not in scope.  This is now called out in the documents opening.

documentation/image-artifact-details.md
MichaelSimons: I just completed a major refactoring of the commands.  I found a lookup technique for each distro including Alpine.  The only external lookup needed is to find the source information for Alpine packages.  Take a look once I submit it and let me know what you think.

documentation/image-artifact-details.md
MichaelSimons: I took Rich's suggestion on a slightly different wording.  Matt if you have other specific suggestions feel free to share them.

documentation/image-artifact-details.md
MichaelSimons: I renamed "Packages" -> "Additional packages".  I also incorporated the synopsis in the document's opening.

MichaelSimons:
I pushed a fairly substantial update.  Please let me know what you think of the proposed patterns for retrieving the license information.

MichaelSimons:
@richlander - Thoughts on linking to this doc from the image repo READMEs?

richlander:
We should probably link to the doc from the official readmes and under the guidance section in the samples readme.

documentation/image-artifact-details.md
```diff
@@ -57,37 +58,54 @@ ca-certificates-cacert-20191127-r0 x86_64 {ca-certificates} (MPL-2.0 GPL-2.0-or-
 libcrypto1.1-1.1.1d-r2 x86_64 {openssl} (OpenSSL) [installed]
 ```
 
-You can lookup the individual package information from the [alpine package website](https://pkgs.alpinelinux.org/packages) by entering the package name, branch (which corresponds to the Alpine version), and architecture. Finding the [information for the `zlib` package](https://pkgs.alpinelinux.org/packages?name=zlib&branch=v3.11&repo=main&arch=x86) is demonstrated below.
+You can retrieve the license for any package, as demonstrated below.
```
mthalman: Call out that the example demonstrates retrieving the license for the `musl` package.  Not sure if you're aware but the list command above shows the same license information as this command, but for all packages obviously.

documentation/image-artifact-details.md
```diff
@@ -57,37 +58,54 @@ ca-certificates-cacert-20191127-r0 x86_64 {ca-certificates} (MPL-2.0 GPL-2.0-or-
...
-You can then print any of these copyright files, as demonstrated below.
+You can retrieve the license for any package, as demonstrated below. This example retrieves the `apt` license from the `debian:buster-slim` image by performing the following steps with a single command:
+
+1. List out the package contents
+1. Find the copyright file
+1. Print out the copyright file contents
 
 ```console
-$ docker run --rm debian:buster-slim cat ./usr/share/doc/apt/copyright
+$ docker run --rm debian:buster-slim sh -c "cat \`dpkg -L apt | grep copyright\`"
```
mthalman: This command doesn't work from PowerShell or Windows CMD.  When run from PowerShell, it looks for more console input.  From CMD, it returns an error saying `cat: invalid option -- 'L'`.  Same comment applies to Ubuntu.

documentation/image-artifact-details.md
```diff
@@ -112,28 +130,49 @@ See /usr/share/common-licenses/GPL-2, or
 of the GNU General Public License.
 ```
 
+You can retrieve the source for package, as demonstrated below. This example retrieves the `apt` source from the `debian:buster-slim` image by performing the following steps with a single command:
```
mthalman: Maybe make this more clear by saying that this returns URLs of archive files that contain the source.  Same comment applies to Ubuntu.

documentation/image-artifact-details.md
```diff
@@ -1,9 +1,16 @@
...
+.NET Core container images contain multiple software components, each built from different source and with different licensing. It is important that users of these images can determine the license requirements of these images and have a link back to the source for copyright and pedigree purposes.
 
 This document is intended to be informative, primarily defining patterns for discovering license and pedigree information and describing expectations (whether the desired information can be found within the images). It does not list all of the license and pedigree information you might want. This information can change at any time, which also demonstrates why using the documented patterns is important.
 
+This document focuses solely on open source Linux distros. Windows is out of scope because it is a proprietary licensed OS.
+
+This document answers two basic questions:
+
+* How can I interrogate an image to determine licensing and version information of the packages and components it includes?
+* Where can I look to learn about source pedigree information for an image from offline sources (since it is difficult/impossible to learn everything from an image)?
```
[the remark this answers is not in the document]
MichaelSimons: fixed

documentation/image-artifact-details.md
MichaelSimons: Fixed.

documentation/image-artifact-details.md
MichaelSimons: Expanded the wording to:

> The package listing above includes the license in the fourth cell. You can also retrieve the license for any single package, as demonstrated below. This example retrieves the `musl` license from the `alpine:3.11` image.

documentation/image-artifact-details.md
MichaelSimons: I was able to easily fix this by switching from double quotes to single quotes.

```
docker run --rm ubuntu:bionic sh -c 'cat `dpkg -L apt | grep copyright`'
```

MichaelSimons:
FYI - I just added links from the READMEs to the image-artifact-details.md.

documentation/image-artifact-details.md
mthalman: That works for PowerShell.  For CMD, I need to use double quotes: 
```
docker run --rm ubuntu:bionic sh -c "cat `dpkg -L apt | grep copyright`"
```

MichaelSimons: I personally don't care to call this out.  If we receive feedback from customers on this then we can refactor.

PR #1600, opened 2020-01-17
Add tests for sample Dockerfiles

### LINKED ISSUE #392: Add unit tests to ensure samples always work
@MichaelSimons commented on [Fri Jun 09 2017](https://github.com/dotnet/dotnet-docker-samples/issues/51)

Unit tests should be created to ensure the samples always build and run

mthalman:
Adds tests for the sample Dockerfiles that are not published as Docker images.  These Dockerfiles are meant for illustration purposes but they should be verified to work correctly.  These changes refactor the existing sample test logic to support this scenario.  Since there are no pre-existing images to pull and these Dockerfiles are not built by our normal infrastructure, the test itself handles the building of the image.

Those people who want to run tests on the samples can run:
```powershell
./tests/run-tests.ps1 -TestCategories @('sample') -PullImages
```

Fixes #392


tests/Microsoft.DotNet.Docker.Tests/Config.cs
```diff
@@ -18,6 +18,7 @@ public static class Config
             Environment.GetEnvironmentVariable("RUNNING_TESTS_IN_CONTAINER") != null;
         public static string RepoPrefix { get; } = Environment.GetEnvironmentVariable("REPO_PREFIX") ?? string.Empty;
         public static string Registry { get; } = Environment.GetEnvironmentVariable("REGISTRY") ?? GetManifestRegistry();
+        public static string SamplesPath { get; } = Environment.GetEnvironmentVariable("SAMPLES_PATH");
```
MichaelSimons: What is the value in this config when it is just hard coded in the run-tests script?  It makes it more difficult to run the tests outside of the script (e.g. VS).  Wouldn't it simplify the system to simply hard-code the value within the tests vs the runner script?

tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
```diff
@@ -8,14 +8,29 @@ namespace Microsoft.DotNet.Docker.Tests
 {
     public class SampleImageData : ImageData
     {
+        /// <summary>
+        /// Gets or sets a value indicating that this represents a local sample only and is not published.
+        /// </summary>
+        public bool LocalOnly { get; set; }
+
+        public string Dockerfile { get; set; }
```
MichaelSimons: Thoughts on making this DockerfileSuffix?  It would be nice to incorporate it in the tag name when specified vs the OS.  It would better differentiate the images.

tests/Microsoft.DotNet.Docker.Tests/SampleImageData.cs
```diff
@@ -8,14 +8,29 @@ namespace Microsoft.DotNet.Docker.Tests
 {
     public class SampleImageData : ImageData
     {
+        /// <summary>
+        /// Gets or sets a value indicating that this represents a local sample only and is not published.
+        /// </summary>
+        public bool LocalOnly { get; set; }
```
MichaelSimons: Given the majority of instances specify this, have you considered inverting the logic.  `IsPublished`

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -31,37 +34,115 @@ public static IEnumerable<object[]> GetImageData()
...
+        {
+            string appTag = ImageData.GetImageName("complexapp-local-app", "sample");
+            string testTag = ImageData.GetImageName("complexapp-local-test", "sample");
+            string sampleFolder = $"{Config.SamplesPath}/complexapp";
+            string dockerfilePath = $"{sampleFolder}/Dockerfile";
+            string containerName = ImageData.GenerateContainerName("sample-complex");
+            string tempDir = null;
+            try
+            {
+                DockerHelper.Build(appTag, dockerfilePath, contextDir: sampleFolder, pull: Config.PullImages);
```
MichaelSimons: It feels like we should also be running the resulting app image and verifying it behaves as expected.

mthalman:
I'm seeing some very strange errors when running the complexapp sample in the Windows ARM leg of the build.  This is going to require some more investigation.  In the meantime, I'm going to pull out my changes to [samples/aspnetapp/Dockerfile.nanoserver-x64-slim](https://github.com/dotnet/dotnet-docker/pull/1600/files#diff-d2ddb543af354a734ed87af97e3fd1ba) so that can get in right away.

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -71,7 +71,9 @@ public string Rid
             }
         }
 
-        public virtual string GetIdentifier(string type) => $"{type}-{DateTime.Now.ToFileTime()}";
+        public virtual string GetIdentifier(string type) => GenerateContainerName(type);
+
+        public static string GenerateContainerName(string prefix) => $"{prefix}-{DateTime.Now.ToFileTime()}";
```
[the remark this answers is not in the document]
mthalman: I'm not following.  That method already makes use of `ImageData.GetIdentifier`.

tests/Microsoft.DotNet.Docker.Tests/Config.cs
mthalman: Perhaps it would simplify things but it also makes it more sensitive to breakage.  The test assembly could only be run from the default location where it gets built, for example.  I figured it'd be best to encapsulate any knowledge of repo locations within the script because scripts are meant to be run from their specific location whereas assemblies are meant to be portable.  And the script already has knowledge about the location of the test project itself.  So it seem a natural spot to define repo paths like this.

That being said, I could go either way on this.  Let me know if you'd still be prefer moving it into the test code.

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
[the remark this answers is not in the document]
MichaelSimons: My mistake.

tests/Microsoft.DotNet.Docker.Tests/Config.cs
MichaelSimons: That's a fair point.  The piece I think needs to be addressed is the experience when someone tries to run the tests and the variable is not set.  The rest of the variables default to a sensible value that allow the tests to run in a meaningful way.  This variable looks to cause a null ref if not set.  

There is one other repo artifact the tests depend on - manifest.json.  The tests handle this differently in that the manifest is treated as content (e.g. CopyToOutputDirectory).  Would it make sense to treat the sample artifacts the same?  This would eliminate the need for the variable.

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -80,20 +80,33 @@ public void VerifyComplexAppSample()
...
             {
+                // Test that the app works
                 DockerHelper.Build(appTag, dockerfilePath, contextDir: sampleFolder, pull: Config.PullImages);
-                DockerHelper.Build(testTag, dockerfilePath, target: "test", contextDir: sampleFolder);
+                string containerName = ImageData.GenerateContainerName("sample-complex");
+                string output = DockerHelper.Run(appTag, containerName);
+                Assert.StartsWith("string: The quick brown fox jumps over the lazy dog", output);
 
-                DockerHelper.Run(testTag, containerName, skipAutoCleanup: true);
+                if (DockerHelper.DockerOS == "windows" &&
```
MichaelSimons: Consider utilizing `!DockerHelper.IsLinuxContainerModeEnabled` to be consistent with the rest of the tests.  It makes it easier to search to find the special cases if the code utilizes the same pattern.

tests/Microsoft.DotNet.Docker.Tests/TestData.cs
```diff
@@ -59,17 +59,33 @@ public static class TestData
...
-            new SampleImageData { OS = OS.NanoServer1809,  Arch = Arch.Arm },
-            new SampleImageData { OS = OS.NanoServer1903,  Arch = Arch.Amd64 },
-            new SampleImageData { OS = OS.NanoServer1909,  Arch = Arch.Amd64 },
+            new SampleImageData { OS = OS.NanoServer1809, Arch = Arch.Amd64, IsPublished = true },
+            new SampleImageData { OS = OS.NanoServer1809, Arch = Arch.Arm,   IsPublished = true },
+            new SampleImageData { OS = OS.NanoServer1903, Arch = Arch.Amd64, IsPublished = true },
+            new SampleImageData { OS = OS.NanoServer1909, Arch = Arch.Amd64, IsPublished = true },
+
+            new SampleImageData { OS = OS.NanoServer1809, Arch = Arch.Arm,   DockerfileSuffix = "nanoserver-arm32" },
+            new SampleImageData { OS = OS.NanoServer1909, Arch = Arch.Amd64, DockerfileSuffix = "nanoserver-x64"  },
```
MichaelSimons:
```suggestion
            new SampleImageData { OS = OS.NanoServer1909, Arch = Arch.Amd64, DockerfileSuffix = "nanoserver-x64" },
```

eng/pipelines/dotnet-core-samples.yml
```diff
@@ -1,5 +1,13 @@
 trigger: none
-pr: none
+pr:
+  branches:
+    include:
+    - master
+    - nightly
+  paths:
+    include:
```
MichaelSimons: Thought on adding a few `exclude:` paths in the primary build pipelines for nightly and master?  As they relate to this PR, `samples/*` feels like a good optimization.  On a larger picture, excluding README.* as well as documentation/* seems like a good idea.

Should `eng/common/*` and `eng/pipelines/*` also be included?  It feels like these tests should run with infra changes.

eng/pipelines/dotnet-core-nightly.yml
```diff
@@ -4,7 +4,14 @@ trigger:
     include:
     - nightly
 pr:
-- nightly
+  branches:
+    include:
+    - nightly
+  paths:
+    include:
+    - '*'
```
MichaelSimons: Is this necessary?  From the documentation, it seems like you only need to specify the exclude.

mthalman: I haven't seen a doc example of an `exclude` without an `include`.  In fact, I see this example: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#wildcards which does exactly what I do in this change.

MichaelSimons: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=example#triggers

eng/pipelines/dotnet-core-samples.yml
mthalman: For the README and doc exclusions, I'll log an issue for that.

mthalman: https://github.com/dotnet/dotnet-docker/issues/1609

MichaelSimons:
Am I correct in that the public dnceng build definition for the samples pr validation will need to be created after this is merged in order to complete the work?

mthalman:
> Am I correct in that the public dnceng build definition for the samples pr validation will need to be created after this is merged in order to complete the work?

Correct

PR #1622, opened 2020-01-24
Rename launch profile

richlander:
The launch profile added in the sample is not container-specific, so changing its name, to "publicdev" (a more accurate description of what the profile is accomplishing). After that, will start a conversation to get this change into the .NET 5.0 templates.

@glennc suggested that we switch to a port like `5050`, for two reasons:

* Maps more closely to the localhost:5000 that is used by default.
* Proxies and other things apparently use the 80xx range, so there could be conflicts. 

If we want to do that, we may want to do that throughout the samples. We could also wait to have that conversation as part of adding this concept to the official ASP.NET Core templates.

### DIFF
--- a/samples/aspnetapp/aspnetapp/Properties/launchSettings.json
+++ b/samples/aspnetapp/aspnetapp/Properties/launchSettings.json
@@ -23,7 +23,7 @@
         "ASPNETCORE_ENVIRONMENT": "Development"
       }
     },
-    "container": {
+    "publicdev": {
       "commandName": "Project",
       "launchBrowser": true,
       "applicationUrl": "http://+:80",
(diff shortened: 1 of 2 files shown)


richlander:
Filed the issue @ https://github.com/dotnet/dotnet-docker/issues/1623. We can then probably merge this one. The only question in this PR then is the name of the profile. You good with "publicdev"?

MichaelSimons:
"publicdev" is palatable but I'm not in love with it. I can't come up with a better suggestion therefore I have no right to comment!

richlander:
I wanted to just call it "public", but that's doesn't account for the environment setting.

Another framing is that we consider launchsettings.json to be only for development scenarios. If that's the case, we could elide the "dev" part.

We could also go with "external" instead of "public".

MichaelSimons:
I think I prefer "external" instead of "public".

PR #1624, opened 2020-01-24
Added 'Image Update Policy' section to the READMEs

MichaelSimons:
Added an 'Image Update Policy' section to the READMEs to call out when .NET Core images will be updated.  The purpose is to communicate our policy so that customers can know what to expect.  The desired outcome is that more customers feel compelled to take a dependency on our images.

I also tweaked one of the licensing URLs which was language specific to be neutral.

/cc @rotata - this is related to the question I asked in regards to changing the MCR readme template to make it more obvious that the full tag listing is the complete list of supported tags.

### DIFF
--- a/README.aspnet.md
+++ b/README.aspnet.md
@@ -117,6 +117,11 @@ You can retrieve a list of all available tags for dotnet/core/aspnet at https://
 
 See [Microsoft Support for .NET Core](https://github.com/dotnet/core/blob/master/microsoft-support.md) for the support lifecycle.
 
+# Image Update Policy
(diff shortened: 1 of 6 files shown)


dleeapho:
The policy looks good. You might want to consider placing it after `Support`, before the `Feedback` section, I just feels it goes together more.

MichaelSimons:
@dleeapho - Thanks for the feedback.  I wrestled on the location - before the `License` or after `Support`.  I moved the `Policy` section after `Support` as suggested.

PR #1638, opened 2020-01-29
Cleanup unused update-dependencies-from-versions build pipeline

MichaelSimons:
We haven't used this update-dependency flow for some time.  The update-dependency flow from drop is the foreseeable way forward.  

- Removed update-dependencies-from-versions build pipeline
- Renamed update-dependencies-from-drop build pipeline
- Removed the unused `--build-info` option from update-dependencies
- Added the scheduled trigger to the update-dependencies build pipeline.  I will removed the modeled trigger once merged.


eng/update-dependencies/Program.cs
```diff
@@ -71,62 +69,20 @@ private static DependencyUpdateResults UpdateFiles(IEnumerable<IDependencyInfo>
             return DependencyUpdateUtils.Update(updaters, buildInfos);
         }
 
-        private static async Task<IEnumerable<IDependencyInfo>> GetBuildInfoAsync()
+        private static IEnumerable<IDependencyInfo> GetBuildInfoAsync()
```
mthalman: Remove `Async` from the method name.

eng/update-dependencies/Program.cs
```diff
@@ -36,7 +34,7 @@ public static async Task Main(string[] args)
 
                 Options.Parse(args);
 
-                IEnumerable<IDependencyInfo> buildInfos = await GetBuildInfoAsync();
+                IEnumerable<IDependencyInfo> buildInfos = GetBuildInfoAsync();
```
mthalman: Capture the results with `ToArray` since it is enumerated multiple times.

eng/update-dependencies/Program.cs
MichaelSimons: Fixed

eng/update-dependencies/Program.cs
MichaelSimons: fixed

Issue #331, opened 2017-11-06
Windows Docker Tag Scheme Changed


richlander:
# Windows Docker Tag Scheme Changed

.NET Core and [.NET Framework](https://github.com/Microsoft/dotnet-framework-docker/issues/40) Windows container images have been updated to use a new tag naming scheme for Windows container images, for both [Windows Server Version 1709](https://docs.microsoft.com/en-us/windows-server/get-started/get-started-with-1709) and [Windows Server 2016](https://docs.microsoft.com/en-us/windows-server/get-started/server-basics).

You will see the following changes in .NET Docker repos:

- Tag names will better describe [Windows Release Channels](https://docs.microsoft.com/en-us/windows-server/get-started/semi-annual-channel-overview).
- Only stable tags will be used.
- You will need to use image digests as the identifier for patch releases, as appropriate.

The following tags are examples of the new tag scheme:

* 2.0.0-runtime-nanoserver-1709
* 2.0.0-runtime-nanoserver-sac2016
* 4.7.1-windowsservercore-ltsc2016
* 4.7.1-windowsservercore-1709

This scheme is in use in the following repos:

* [microsoft/aspnet](https://hub.docker.com/r/microsoft/aspnet/)
* [microsoft/aspnetcore](https://hub.docker.com/r/microsoft/aspnetcore/)
* [microsoft/aspnetcore-build](https://hub.docker.com/r/microsoft/aspnetcore-build/)
* [microsoft/dotnet](https://hub.docker.com/r/microsoft/dotnet/)
* [microsoft/dotnet-framework](https://hub.docker.com/r/microsoft/dotnet-framework/)
* [microsoft/wcf](https://hub.docker.com/r/microsoft/wcf/)

## Details

* [dotnet/dotnet-docker #329](https://github.com/dotnet/dotnet-docker/pull/329)
* [Microsoft/dotnet-framework-docker #39](https://github.com/Microsoft/dotnet-framework-docker/pull/39)
* [dotnet/announcements #37](https://github.com/dotnet/announcements/issues/37)
* [dotnet/announcements #36](https://github.com/dotnet/announcements/issues/36)

With the release of [Windows Server 1709](https://blogs.technet.microsoft.com/windowsserver/2017/10/17/windows-server-version-1709-available-for-download/), Windows Server container images adopted stable tags, in addition to the release-specific tags already in use.

[...]

PR #1643, opened 2020-01-30
Add update-readmes pipeline

MichaelSimons:
Related to #331

I will port the common infra back to docker-tools and dotnet-framework-docker once merged.

### DIFF
--- a/eng/pipelines/update-readmes.yml
+++ b/eng/pipelines/update-readmes.yml
@@ -1,0 +1,14 @@
+trigger: none
+pr: none
+variables:
+- group: DotNet-Maestro
+- template: variables/common.yml
+- name: manifest
+  value: manifest.json
+jobs:
+- job: UpdateReadmes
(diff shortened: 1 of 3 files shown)


eng/pipelines/update-readmes.yml
```diff
@@ -0,0 +1,14 @@
+trigger: none
+pr: none
+variables:
+- group: DotNet-Maestro
+- template: variables/common.yml
+- name: manifest
```
mthalman: Should this be defined as a variable in the pipeline UI so that we can override it for samples perhaps?

MichaelSimons: I was planning on doing that and leaving this here as a default.  Would that be confusing?

mthalman: Does the UI actually take precedence over this value then?

MichaelSimons: yes

PR #1650, opened 2020-02-03
Update path exclusions for pipelines

### LINKED ISSUE #1609: Exclude some paths from PR trigger
Now that [PR triggers for YAML pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#pr-triggers) support path filtering for GitHub repos, the PR pipeline validation for this repo can be updated to exclude some paths that shouldn't trigger a build:
* `README*`
* `documentation/*`

mthalman:
Ensures that PR builds will not trigger for commits that only consist of README or documentation changes.

Fixes #1609

### DIFF
--- a/eng/pipelines/dotnet-core-nightly.yml
+++ b/eng/pipelines/dotnet-core-nightly.yml
@@ -3,6 +3,11 @@ trigger:
   branches:
     include:
     - nightly
+  paths:
+    exclude:
+    - documentation/*
+    - README*
+    - samples/*
 pr:
   branches:
     include:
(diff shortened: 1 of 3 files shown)


eng/pipelines/dotnet-core-nightly.yml
```diff
@@ -9,6 +9,8 @@ pr:
     - nightly
   paths:
     exclude:
```
MichaelSimons: Should the same exclude paths be specified in the official build triggers as well?

mthalman: Yes, good call

PR #1654, opened 2020-02-04
Build samples from build-and-test script

### LINKED ISSUE #1613: Dev build script should build sample images
On a machine in which the sample images have not been built or pulled, attempting to run the `build-and-test.ps1` script will result in errors similar to `'mcr.microsoft.com/dotnet/core/samples:dotnetapp-buster-slim' could not be found on disk.`  This is because when the `run-tests.ps1` script gets invoked, it defaults to not pulling images.  And since the build script didn't end up building the sample images, those images will not exist on disk and result in an error when the sample tests run.

The build script should be updated to include sample images in the set of images that are built.

mthalman:
Fixes #1613


build-and-test.ps1
```diff
@@ -34,6 +34,12 @@ if ($build) {
         -OSFilter $OSFilter `
         -ArchitectureFilter $ArchitectureFilter `
         -OptionalImageBuilderArgs $OptionalImageBuilderArgs
+
```
MichaelSimons: I don't like the UX when utilizing the filters.  I don't think users should have to build the samples each time.

```
build-and-test.ps1 -VersionFilter 5.0
build-and-test.ps1 -OSFilter alpine*
```

mthalman: Should we just skip building samples if `VersionFilter` or `OSFilter` is anything but `*`?

MichaelSimons: I like that but I think you should be able to just build the samples.  This feels like a scenario that would benefit from a single manifest.

mthalman: Then I'm not quite clear on what you're wanting then.  Are you wanting samples to be skipped if you execute `build-and-test.ps1 -VersionFilter 5.0`?  And you want some mechanism to _only_ build samples?  How does a single manifest help with this?  The product and samples have very different paths to their Dockerfiles so the filtering scheme doesn't really align at all.

MichaelSimons: The script parameters should support scoping the build to a subset of the repo.  This should support things like building the images for a particular version, a particular os, or both.  With this PR, I would expect that you could build just the samples or just a subset of product images.  So I want everything 😸.  I see a single manifest helping because we could use the same parameters that exist in the eng/common/build script.

```
    # Version of .NET Core to filter by
    [string]$VersionFilter = "*",

    # Name of OS to filter by
    [string]$OSFilter = "*",

    # Type of architecture to filter by
    [string]$ArchitectureFilter = "amd64",

    # Custom path filters that override the other filter options
    [string]$PathFilters,
```

With this, if you want to build just the samples you would use `-PathFilters samples/*`

richlander: Any reason why we have to use the word "Filter"? It seems to make a simple concept more complicated for no reason. What am I missing?

tests/Microsoft.DotNet.Docker.Tests/ImageVersion.cs
```diff
@@ -11,5 +11,8 @@ public static class ImageVersion
         public static readonly Version V2_1 = new Version(2, 1);
         public static readonly Version V3_0 = new Version(3, 0);
         public static readonly Version V3_1 = new Version(3, 1);
+
+        public static readonly Version SampleVersion = new Version(SampleVersionString);
+        public const string SampleVersionString = "3.1";
```
MichaelSimons: I have mixed feeling on this being a constant.  I am confident this will get overlooked at some point in time.  Thoughts on parsing this out of one of the sample Dockerfiles?

build-and-test.ps1
MichaelSimons: I like rich's suggestion here.  This is going to have a ripple effect though so it may be best to do with a separate PR.

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -32,7 +32,7 @@ public static IEnumerable<object[]> GetImageData()
                 .Select(imageData => new object[] { imageData });
         }
 
-        [Theory]
+        [SkippableTheory(ImageVersion.SampleVersionString)]
```
MichaelSimons: Is this going to cause issues if you run build-and-test with a 3.1 versionfilter?  The published samples won't get built locally but these test will try to test the published samples? Because of this, it feels like it is inappropriate to trick the system here by specifying a version.

build-and-test.ps1
```diff
@@ -1,50 +1,58 @@
...
+        -PathFilters $PathFilters `
+        -OptionalImageBuilderArgs $OptionalImageBuilderArgs `
+        -Manifest manifest.samples.json
 }
-if ($test) {
+if ($Mode -eq "BuildAndTest" -or $Mode -eq "Test") {
+
+    $localTestCategories = $TestCategories
+
+    if ($PathFilters -and $TestCategories.Contains("sample")) {
```
MichaelSimons: I don't think this is desirable in some cases.  This primary one I am thinking of is if you want to explicitly build and test the samples.  `build-and-test.ps1 -PathFilters '--path samples/*'.  I was thinking if a version filter was specified, then we would skip the tests.  If you are using a path filter, then I think it is up to you to customize the TestCategories as appropriate.

build-and-test.ps1
mthalman: Logged https://github.com/dotnet/docker-tools/issues/409 to handle this rename

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
mthalman: The samples would still get built in that case.  They're built if you specify _any_ version filter because the `VersionFilter` parameter is not set when building samples.

MichaelSimons: I see that now.  That isn't desirable IMO.  When I am running this locally, I don't want the samples getting built when I am targetting 5.0 via a version filter.

mthalman: Fixed

build-and-test.ps1
mthalman: Ok, that makes things simpler if we say that it's up to the caller to set `TestCategories` if they use custom paths.  Fixed.

build-and-test.ps1
```diff
@@ -1,44 +1,59 @@
...
+    # Build the product images
     & ./eng/common/build.ps1 `
         -VersionFilter $VersionFilter `
         -OSFilter $OSFilter `
         -ArchitectureFilter $ArchitectureFilter `
+        -PathFilters $PathFilters `
         -OptionalImageBuilderArgs $OptionalImageBuilderArgs
+
+    # Build the sample images
+    & ./eng/common/build.ps1 `
```
MichaelSimons: I pulled these changes and ran through a few scenarios.  This does not work well on Windows and results in the following exception being thrown.

```
System.ArgumentException: An item with the same key has already been added. Key: [samples/dotnetapp/Dockerfile, Microsoft.DotNet.ImageBuilder.Models.Image.ImageData]
   at System.Collections.Generic.TreeSet`1.AddIfNotPresent(T item)
   at System.Collections.Generic.SortedDictionary`2.Add(TKey key, TValue value)
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.BuildImages() in D:\repos\docker-tools\src\Microsoft.DotNet.ImageBuilder\src\Commands\BuildCommand.cs:line 71
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.ExecuteAsync() in D:\repos\docker-tools\src\Microsoft.DotNet.ImageBuilder\src\Commands\BuildCommand.cs:line 38
   at Microsoft.DotNet.ImageBuilder.ImageBuilder.Main(String[] args) in D:\repos\docker-tools\src\Microsoft.DotNet.ImageBuilder\src\ImageBuilder.cs:line 72
```

This is because of the shared Dockerfiles in the manifest (one for each supported Windows version).  The system is trying to rebuild each of them.  This isn't valid because they are designed to be built in a different host environment in order to pickup a different base image.  We should figure out a way to only build the version that matches the users host.

PR #1655, opened 2020-02-04
Fix image size validation pipeline

### LINKED ISSUE #1588: Image size validation pipeline needs to pull images
The logic for [image size validation pipeline](https://github.com/dotnet/dotnet-docker/blob/master/eng/pipelines/dotnet-core-size-validation.yml) invokes the validation [script](https://github.com/dotnet/dotnet-docker/blob/master/tests/performance/Validate-ImageSize.ps1) with default parameters.  The script had been changed to not pull images by default, optimizing for the local dev environment.  Since the images will not exist on the build agent, the logic needs to be updated to pull the images by passing the `-PullImages` option.

A test build that did this, however, ran into a space issue using a Linux agent from the Azure Pipelines pool.  There will need to be changes made to account for this.

Some options:
* Divide up the Linux image validation between AMD64 and ARM so that they are run on separate agents.  This limits the number of images that need to be stored on a given agent.
* Investigate whether it is feasible to use the REST API to retrieve the image size rather than pulling the image.

mthalman:
The image size validation pipeline was not configured to pull images causing it to run into errors when attempting to get the image size.

After fixing that issue, it was discovered that the hosted Linux agents do not have enough disk space to contain all of the Linux images that need to be pulled.  To workaround that, the changes in https://github.com/dotnet/docker-tools/pull/400 are used so that we can split apart Linux image validation between multiple jobs.  One job handles AMD64 images, one handles ARM images, and the other does baseline integrity validation.  Baseline integrity validation cannot be done in the other two jobs because they are only handling a partial set of the Linux images so that would lead to extraneous image errors.

Fixes #1588


eng/common/templates/jobs/validate-image-sizes.yml
```diff
@@ -1,12 +1,34 @@
 jobs:
-- job: LinuxPerfTests
+- ${{ if eq(variables['System.TeamProject'], 'internal') }}:
```
MichaelSimons: Do you think it would be useful to capture in a comment the reasoning behind the job partitioning?  It is not obvious from the yml alone as to why linux legs are split apart by arch but the windows legs aren't/

Issue #1645, opened 2020-01-30
CURL vulnerabilities


lenisha:
## Steps to reproduce the issue
1. Use mcr.microsoft.com/dotnet/core/aspnet images
2. Dotnet images flagged with vulnerabilities in CURL package

Is it possible to remove curl from the image once donet sdk or runtime was installed?

mthalman:
* Which specific image?  Please provide the image tag.
* What vulnerability scanner is being used?
* What are the vulnerabilities that are being flagged?

lenisha:
Tag: mcr.microsoft.com/dotnet/core/aspnet:2.2
AquaSecurity
https://curl.haxx.se/docs/vuln-7.52.0.html

MichaelSimons:
@lenisha, Please upgrade to a supported version of .NET Core.  2.2 went EOL on Dec 23, 2019.  See this [blog post](https://devblogs.microsoft.com/dotnet/net-core-2-2-will-reach-end-of-life-on-december-23-2019/) for details.

MichaelSimons:
Looking outside of the specific scenario in which this issue was reported on, we think it makes sense to consider removing curl as well as ca-certificates from the runtime images.  The motivating reasons are to reduce the attack surface area and to reduce the size of the images to help download and startup perf.  It makes sense to keep them within the sdk images as those are tuned for development scenarios.  This is considered a breaking change and therefore should only be done for the next <major>.<minor> release which would be 5.0.

The original reason why curl was included is because these images were defined prior to Docker supporting multi-staged builds.  Without multi-staged builds, it was necessary to acquire  tools such as curl in order to download assets directly within Dockerfiles while adhering to the [Dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) which recommend not using cmds like ADD which have support for downloading artifacts.

shouaib-ahmed:
The issue exists in all Debian Buster based Docker images regardless of .net version. See https://security-tracker.debian.org/tracker/source-package/curl . What is the best way to address this?

MichaelSimons:
We posted a [Staying up-to-date with .NET Container Images](https://devblogs.microsoft.com/dotnet/staying-up-to-date-with-net-container-images/) blog post a while ago.  In the _Security Updates and Vulnerability Scanning_ section, we recommend using the Alpine images if you are concerned about vulnerabilities as it is the most secure offering.  

[...]

PR #1657, opened 2020-02-04
Remove curl and ca-certificates from 5.0 runtime images

MichaelSimons:
Related to #1645 

The impact to the size of the image on disk is minimal.

image (buster-slim) | previous size (MB) | new size (MB)
runtime-deps | 111 | 109
runtime | 195 | 191 
aspnet | 213 | 209

I'd like feedback on the base image for the installer stage.  I waffled on a couple patterns that can be seen [here](https://github.com/dotnet/dotnet-docker/commit/a66cf88cd1f3a6152dd3cb4696067a3c3d22f6f9)


mthalman:
How is it that runtime and aspnet increased in size?

MichaelSimons:
Because I accidentally inversed the numbers when I typed them - I corrected the table.

tests/performance/ImageSize.nightly.linux.json
```diff
@@ -20,7 +20,15 @@
     "3.0/runtime-deps/bionic/arm64v8": 99094631,
     "3.1/runtime-deps/focal/amd64": 129846045,
     "3.1/runtime-deps/focal/arm32v7": 96976919,
-    "3.1/runtime-deps/focal/arm64v8": 119013389
+    "3.1/runtime-deps/focal/arm64v8": 119013389,
+    "5.0/runtime-deps/alpine3.11/amd64": 9937498,
+    "5.0/runtime-deps/alpine3.11/arm64v8": 9733357,
+    "5.0/runtime-deps/bionic/amd64": 106459634,
```
mthalman: Just seeing that the image size validation failed.  `bionic` should be `buster-slim`.

MichaelSimons: Fixed

richlander:
LGTM

If I understand correctly, you went back and forth between a self-curated image based on the distro (+ packages you curated) or an existing curated buildpack-deps image that is probably bigger than you need but self-curated.

I like the buildpack-deps approach. It's way simpler and more likely to be cached on an arbitrary machine.

MichaelSimons:
@richlander - you are correct.  The main reason I was considering the self-curated image is because it shared the same base image as the final stage.  I decided against this because of the complication it adds to the Dockerfile over utilizing buildpack-deps.

PR #1673, opened 2020-02-07
Add important firewall announcement to readme for docker hub

mbrown555:


eng/common/templates/jobs/validate-image-sizes.yml
```diff
@@ -45,4 +45,4 @@ jobs:
     parameters:
       dockerClientOS: windows
       ${{ if eq(variables['System.TeamProject'], 'public') }}:
-        validationMode: integrity
+        validationMode: integrity
```
mbrown555: this is a mistake, I'm fixing it

PR #1678, opened 2020-02-13
Removed "" around default kestrel password in documentation

thylux:


mthalman:
What's the reason for this change?

thylux:
Sorry, when I was struggling to make HTTPS work on local docker I I saw somewhere that the password shouldn't have "".
After some changes I got it working and thought that it was part of the solution.

I tried again with and without "" and it is working both ways.

By the way, looks like the problem I was having was because I called my executable with --urls http://+;http**s**://+ instead of --urls http**s**://+;http://+.
With the first way https was not responding.

Closing pull request.

Issue #1685, opened 2020-02-14
 .NET Core Servicing Release (2.1.16, 3.0.3, 3.1.2)


mthalman:
# .NET Core Servicing Release

## Release Versions
_The set of .NET Core versions that are being released as a unit._

* 2.1.16/2.1.804
* 3.0.3/3.0.103
* 3.1.2/3.1.102

## 1. Master Branch Tasks
1. - [x] Merge appropriate commits from nightly branch (new/EOL distros, infra changes, etc):
      - [x] Alpine 3.11: https://github.com/dotnet/dotnet-docker/pull/1562
      - [x] PowerShell 7.0.0-rc.2: https://github.com/dotnet/dotnet-docker/pull/1594
1. - [x] Wait for .NET Core archive files (.zip, .tar.gz) to be available at blob storage location
1. - [x] Run `update-dependencies` tool to update all the necessary files to reflect the specified .NET Core versions (run this command for each version being released):
      - [x] `dotnet run --project .\eng\update-dependencies\update-dependencies.csproj --sdk-version 2.1.804 --runtime-version 2.1.16 --aspnet-version 2.1.16`
      - [x] `dotnet run --project .\eng\update-dependencies\update-dependencies.csproj --sdk-version 3.0.103 --runtime-version 3.0.3 --aspnet-version 3.0.3`
      - [x] `dotnet run --project .\eng\update-dependencies\update-dependencies.csproj --sdk-version 3.1.102 --runtime-version 3.1.2 --aspnet-version 3.1.2`
1. - [x] Inspect generated changes for correctness
1. - [x] Commit generated changes
1. - [x] Create PR
1. - [x] Get PR signoff
1. - [x] Merge PR
1. - [x] Wait for changes to be mirrored to internal [dotnet-docker repo](https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet-docker) (internal MSFT link)
1. - [x] Build images - Queue build stage of [dotnet-docker pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=373) (internal MSFT link) with variables:

      Servicing release:

          imageBuilder.pathArgs: --path '2.1*' --path '3.0*' --path '3.1*'
          stages: build
1. - [x] Wait for NuGet packages to be published during release tic-toc
1. - [x] Test and publish images - Queue build of [dotnet-docker pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=373) (internal MSFT link) with variables:

[...]

PR #1686, opened 2020-02-14
Updates for .NET Core servicing release

mthalman:
Merges changes to include Dockerfiles for Alpine 3.11 and the update to PowerShell 7.0.0-rc.2.

Related to https://github.com/dotnet/dotnet-docker/issues/1685


MichaelSimons:
I think it would be useful to add the Feb updates as a separate commit to this PR so that we have one merge commit for the Feb release.  Thoughts?

mthalman:
> I think it would be useful to add the Feb updates as a separate commit to this PR so that we have one merge commit for the Feb release. Thoughts?

Done

PR #1692, opened 2020-02-18
Fix secret name for https certificate password

StevenLiekens:
I couldn't get these samples to work until I changed the name of the variable for the development certificate password.

``` diff
-Kestrel:Certificates:Development:Password
+Kestrel:Certificates:Default:Password
```

I believe this is the only supported way to configure a default certificate (ie. not bound to an endpoint), see: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-3.1#listenoptionsusehttps.

### DIFF
--- a/samples/run-aspnetcore-https-development.md
+++ b/samples/run-aspnetcore-https-development.md
@@ -68,7 +68,7 @@ dotnet dev-certs https --trust
 Configure application secrets, for the certificate:
 
 ```console
-dotnet user-secrets -p aspnetapp\aspnetapp.csproj set "Kestrel:Certificates:Development:Password" "crypticpassword"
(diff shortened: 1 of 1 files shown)


mthalman:
The sample instructions are correct as they are defined using `Kestrel:Certificates:Development:Password`.  Since the instructions are  using dev certs, the certificate lookup logic falls back to the dev cert.  From the [link](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-3.1#listenoptionsusehttps) that you provided (emphasis mine):
> Any HTTPS endpoint that doesn't specify a certificate (HttpsDefaultCert in the example that follows) falls back to the cert defined under Certificates > Default **or the development certificate**.

PR #1693, opened 2020-02-19
Update Dockerfile SHA updater

### LINKED ISSUE #1576: update-dependencies tool not accessing checksum URLs correctly
There is an aspect of how the update-dependencies tool retrieves the checksum for a .NET artifact that needs to be fixed.  It attempts to access checksum files with a URL pattern of `https://dotnetcli.blob.core.windows.net/dotnet/checksums/<version>-<product>-sha.txt` but this results in a 404.  The checksums are now available for an entire release (both runtime and SDK) in a single URL and the text must be parsed to access the necessary value: `https://dotnetcli.blob.core.windows.net/dotnet/checksums/<version>-sha.txt`.

mthalman:
* Fixes the logic for retrieving SHAs from dotnetcli blob storage. The SHAs are contained in one file and need to be parsed out.
* Provides a summary of any errors that were encountered.
* Changed to not compute SHAs by default. A new option is provided that will enable this logic: `--compute-shas`.

https://github.com/dotnet/core/issues/4278 does impact the use of this functionality currently, requiring that the `--compute-shas` option be used.

Fixes #1576

### DIFF
--- a/eng/update-dependencies/DockerfileShaUpdater.cs
+++ b/eng/update-dependencies/DockerfileShaUpdater.cs
@@ -189,46 +189,7 @@ namespace Dotnet.Docker
                 productVersion = _options.RuntimeVersion;
             }
 
-            string uri = $"https://dotnetcli.blob.core.windows.net/dotnet/checksums/{productVersion}-sha.txt";
-            if (!s_releaseChecksumCache.TryGetValue(uri, out Dictionary<string, string> checksumEntries))
-            {
-                checksumEntries = new Dictionary<string, string>();
-                s_releaseChecksumCache.Add(uri, checksumEntries);
-
-                Trace.TraceInformation($"Downloading '{uri}'.");
-                using (HttpClient client = new HttpClient())
-                using (HttpResponseMessage response = await client.GetAsync(uri))
-                {
-                    if (response.IsSuccessStatusCode)
-                    {
-                        string checksums = await response.Content.ReadAsStringAsync();
-                        string[] checksumLines = checksums.Replace("\r\n", "\n").Split("\n");
-                        if (!checksumLines[0].StartsWith("Hash") || !String.IsNullOrEmpty(checksumLines[1]))
-                        {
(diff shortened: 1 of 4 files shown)


eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -228,5 +263,17 @@ private static string SubstituteVariableValue(string input, string name, string
                 .Replace($"$Env:{name}", value)   // Windows PS ENV variable reference format
                 .Replace($"%{name}%", value);     // Windows CMD variable reference format
         }
+
+        private class ReleaseChecksumEntry
```
MichaelSimons: I'm not seeing where this is referenced.  What am I overlooking?

eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -169,37 +178,63 @@ private static async Task<string> GetDotNetCliChecksumsShaAsync(string productDo
...
+                envName.Contains("aspnet", StringComparison.OrdinalIgnoreCase))
+            {
+                productVersion = _options.RuntimeVersion;
+            }
 
-            Trace.TraceInformation($"Downloading '{uri}'.");
-            using (HttpClient client = new HttpClient())
-            using (HttpResponseMessage response = await client.GetAsync(uri))
+            string uri = $"https://dotnetcli.blob.core.windows.net/dotnet/checksums/{productVersion}-sha.txt";
+            if (!s_releaseChecksumCache.TryGetValue(uri, out Dictionary<string, string> checksumEntries))
```
MichaelSimons: Consider pulling out the LoadDotNetReleaseChecksums into it's own method for readability.

eng/update-dependencies/DockerfileShaUpdater.cs
mthalman: Leftover from earlier work that was abandoned.  Removed.

PR #1699, opened 2020-02-22
Update PowerShell to v7.0.0-rc.3

adityapatwardhan:


MichaelSimons:
Can you also update the new [5.0 images](https://github.com/dotnet/dotnet-docker/tree/nightly/5.0)?

adityapatwardhan:
We have not built or tested on 5.0. Is it supposed to just work?

MichaelSimons:
PS will work in the 5.0 images because the roll-forward policy will allow the 5.0 runtime to be used.  The Docker unit test verify this as well.  

The first 5.0 preview is coming up shortly.  It would be inconvenient if PS was not included in the SDK images.

PR #1700, opened 2020-02-22
Fix header tags for Environment and Metrics

DenisBalan:
Invalid HTML, tag h5 is opened but h1 is closed.
Closing h5.


mthalman:
@DenisBalan - Thank you for your contribution.

PR #1701, opened 2020-02-24
Azure Artifacts tweaks

johnterickson:
* Make feed path match in config and env var
* Don't pass secret on the command-line

### DIFF
--- a/samples/snippets/nuget-credentials.md
+++ b/samples/snippets/nuget-credentials.md
@@ -114,7 +114,7 @@ COPY --from=build /app/out ./
 ENTRYPOINT ["dotnet", "dotnetapp.dll"]
 ```
 
-Before running `docker build`, first populate the Nuget_CustomFeedUserName and Nuget_CustomFeedPassword environment variables with appropriate secrets. Then, this Dockerfile would be built using this command:
+Before running `docker build`, first populate the `Nuget_CustomFeedUserName` and `Nuget_CustomFeedPassword` environment variables with appropriate secrets. Then, this Dockerfile would be built using this command:
 
 ```bash
 docker build --build-arg Nuget_CustomFeedUserName --build-arg Nuget_CustomFeedPassword .
--- a/samples/snippets/nuget-credentials.md
+++ b/samples/snippets/nuget-credentials.md
@@ -172,7 +172,7 @@ ENTRYPOINT ["dotnet", "dotnetapp.dll"]
 
 _Note that a script is called to install the Credential Provider. When `dotnet restore` is run, the Credential Provider is invoked to resolve the credentials and it retrieves them from the `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` environment variable._
 
-Before running `docker build`, first populate the FEED_ACCESSTOKEN environment variable with an access token. Then, this Dockerfile would be built using this command:
+Before running `docker build`, first populate the `FEED_ACCESSTOKEN` environment variable with an access token. Then, this Dockerfile would be built using this command:
 
 ```bash
 docker build --build-arg FEED_ACCESSTOKEN .
--- a/samples/snippets/nuget-credentials.md
+++ b/samples/snippets/nuget-credentials.md
@@ -5,7 +5,7 @@ There are often cases when your Dockerfile needs to build a .NET project that ma
 There are two aspects to be aware of when it comes to protecting credentials:
 
 1. _Exposing credentials in an image that is published to a registry._ This is the most obvious scenario. You don't want consumers of your image to have access to data they should not have. The options listed below demonstrate how to prevent this from happening.
-2. _Exposing secrets locally on the Docker host machine._ This isn't necessarily an issue; it depends on your build processes. Just be aware that while your published image may not be exposing sensitive data, there can be sensitive data stored on the Docker host machine that you may want to clean up depending on how the machine is managed in your build workflow. For example, when using Azure DevOps pipelines with a hosted agent, there's no need to clean up data because the virtual machine of the build agent is automatically deleted at the end of the build. But if you manage your own agent, you're in charge of cleaning any leftover state should it be necessary.
+2. _Exposing secrets locally on the Docker host machine._ This isn't necessarily an issue; it depends on your build processes. Just be aware that while your published image may not be exposing sensitive data, there can be sensitive data stored on the Docker host machine that you may want to clean up depending on how the machine is managed in your build workflow. For example, when using Azure DevOps pipelines with a hosted agent, there's no need to clean up data because the virtual machine of the build agent is automatically deleted at the end of the build. But if you manage your own agent, you're in charge of cleaning any leftover state should it be necessary. Because some environments log command lines, it is a security best-practice to avoid putting secrets on the command line.  To avoid this with `docker build`, refer to [just the name of the environment](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) variable without the `=` or the value.
 
 ## Use a multi-stage build to protect nuget.config passed by build context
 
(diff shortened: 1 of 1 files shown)


samples/snippets/nuget-credentials.md
```diff
@@ -175,7 +175,8 @@ _Note that a script is called to install the Credential Provider. When `dotnet r
 This Dockerfile would be built using this command:
 
 ```bash
-docker build --build-arg FEED_ACCESSTOKEN=<access_token> .
+# Populate FEED_ACCESSTOKEN env var with an access token.
```
jmyersmsft: My brain first parsed the comment as an explanation of the next line rather than an instruction, so I suggest:

```suggestion
# First populate the FEED_ACCESSTOKEN environment variable with an access token, then:
```

samples/snippets/nuget-credentials.md
```diff
@@ -137,7 +137,7 @@ The `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` environment variable is a well-known var
 <configuration>
   <packageSources>
     <add key="public" value="https://api.nuget.org/v3/index.json" />
-    <add key="customfeed" value="https://mycustomfeedurl"  />
+    <add key="customfeed" value="https://thalman.pkgs.visualstudio.com/_packaging/test/nuget/v3/index.json"  />
```
mthalman: I actually didn't realize that this file had specified `thalman.pkgs.visualstudio.com`.  That's my own personal Azure DevOps instance that I was using for testing.  So I don't really want that specified here.  Can you keep this as is and also change the value in `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS`?  If you want it to be more representative of the URL format, you can keep that but replace `thalman` and `test` with placeholders.

johnterickson: :) 

I will borrow fabrikam from https://docs.microsoft.com/en-us/azure/devops/pipelines/packages/nuget-restore?view=azure-devops

samples/snippets/nuget-credentials.md
```diff
@@ -175,7 +175,8 @@ _Note that a script is called to install the Credential Provider. When `dotnet r
 This Dockerfile would be built using this command:
 
 ```bash
-docker build --build-arg FEED_ACCESSTOKEN=<access_token> .
+# First populate the FEED_ACCESSTOKEN environment variable with an access token, then:
```
MichaelSimons: I think this needs further explanation.  
1. Reference the [Docker documentation](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) which explains the functionality being used here.
1. I don't think it is clear why the user shouldn't use the `=` syntax.  Setting the env ahead of time will at times be more work.  Call out why using an environment variable should be considered here.  
1. The explanation belongs in the text and not in the code cmd snippet.
1. This pattern should be consistently used throughout the snippet.  For example the following should use this pattern as well:
    > docker build --build-arg Nuget_CustomFeedUserName=<username> --build-arg Nuget_CustomFeedPassword=<password> .

johnterickson: Thanks for the feedback!

samples/snippets/nuget-credentials.md
```diff
@@ -5,7 +5,7 @@ There are often cases when your Dockerfile needs to build a .NET project that ma
...
+2. _Exposing secrets locally on the Docker host machine._ This isn't necessarily an issue; it depends on your build processes. Just be aware that while your published image may not be exposing sensitive data, there can be sensitive data stored on the Docker host machine that you may want to clean up depending on how the machine is managed in your build workflow. For example, when using Azure DevOps pipelines with a hosted agent, there's no need to clean up data because the virtual machine of the build agent is automatically deleted at the end of the build. But if you manage your own agent, you're in charge of cleaning any leftover state should it be necessary. Because some environments log command lines, it is a security best-practice to avoid putting secrets on the command line.  To avoid this with `docker build`, refer to [just the name of the environment](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) variable without the `=` or the value.
```
mthalman: You cut off the the hyperlink at the word "environment" but should probably extend to include "environment variable".

samples/snippets/nuget-credentials.md
```diff
@@ -172,13 +172,13 @@ ENTRYPOINT ["dotnet", "dotnetapp.dll"]
 
 _Note that a script is called to install the Credential Provider. When `dotnet restore` is run, the Credential Provider is invoked to resolve the credentials and it retrieves them from the `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` environment variable._
 
-This Dockerfile would be built using this command:
+Before running `docker build`, first populate the FEED_ACCESSTOKEN environment variable with an access token. Then, this Dockerfile would be built using this command:
```
mthalman:
```suggestion
Before running `docker build`, first populate the `FEED_ACCESSTOKEN` environment variable with an access token. Then, this Dockerfile would be built using this command:
```

samples/snippets/nuget-credentials.md
```diff
@@ -114,10 +114,10 @@ COPY --from=build /app/out ./
 ENTRYPOINT ["dotnet", "dotnetapp.dll"]
 ```
 
-This Dockerfile would be built using this command:
+Before running `docker build`, first populate the Nuget_CustomFeedUserName and Nuget_CustomFeedPassword environment variables with appropriate secrets. Then, this Dockerfile would be built using this command:
```
mthalman:
```suggestion
Before running `docker build`, first populate the `Nuget_CustomFeedUserName` and `Nuget_CustomFeedPassword` environment variables with appropriate secrets. Then, this Dockerfile would be built using this command:
```

samples/snippets/nuget-credentials.md
```diff
@@ -172,13 +172,13 @@ ENTRYPOINT ["dotnet", "dotnetapp.dll"]
...
-This Dockerfile would be built using this command:
+Before running `docker build`, first populate the FEED_ACCESSTOKEN environment variable with an access token. Then, this Dockerfile would be built using this command:
 
 ```bash
-docker build --build-arg FEED_ACCESSTOKEN=<access_token> .
+docker build --build-arg FEED_ACCESSTOKEN .
 ```
 
-Passing the access token to the `docker build` command in this manner can be useful in automated scenarios when that value is stored as an environment variable on the Docker host machine or can be retrieved from an external secrets storage location and passed to the `docker build` command.
+Passing the access token to the `docker build` command in this manner can be useful in automated scenarios when that value is stored as an environment variable on the Docker host machine or can be retrieved from an external secrets storage location and passed to the `docker build` command.
```
mthalman: Nit: Remove trailing space at the end of the paragraph.

samples/snippets/nuget-credentials.md
mthalman: I would include a statement saying that this requires the value to be stored locally in an environment variable whose name matches the name used for the `build-arg`.

johnterickson: Makes sense

johnterickson:
@MichaelSimons or @mthalman, could you merge?

samples/snippets/nuget-credentials.md
```diff
@@ -5,7 +5,7 @@ There are often cases when your Dockerfile needs to build a .NET project that ma
...
+2. _Exposing secrets locally on the Docker host machine._ This isn't necessarily an issue; it depends on your build processes. Just be aware that while your published image may not be exposing sensitive data, there can be sensitive data stored on the Docker host machine that you may want to clean up depending on how the machine is managed in your build workflow. For example, when using Azure DevOps pipelines with a hosted agent, there's no need to clean up data because the virtual machine of the build agent is automatically deleted at the end of the build. But if you manage your own agent, you're in charge of cleaning any leftover state should it be necessary. Because some environments log command lines, it is a security best-practice to avoid putting secrets on the command line.  To avoid this with `docker build`, store the secret in an enviornment variable with the name you want to build argument to have.  Then, on the `docker build` command line, refer to [just the name of the environment variable](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) without the `=` or the value.
```
mthalman:
```suggestion
2. _Exposing secrets locally on the Docker host machine._ This isn't necessarily an issue; it depends on your build processes. Just be aware that while your published image may not be exposing sensitive data, there can be sensitive data stored on the Docker host machine that you may want to clean up depending on how the machine is managed in your build workflow. For example, when using Azure DevOps pipelines with a hosted agent, there's no need to clean up data because the virtual machine of the build agent is automatically deleted at the end of the build. But if you manage your own agent, you're in charge of cleaning any leftover state should it be necessary. Because some environments log command lines, it is a security best-practice to avoid putting secrets on the command line.  To avoid this with `docker build`, store the secret in an enviornment variable with the name you want the build argument to have.  Then, on the `docker build` command line, refer to [just the name of the environment variable](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) without the `=` or the value.
```

MichaelSimons: Sorry Matt I missed your signoff.

PR #1707, opened 2020-02-25
Update PT issue template to check Windows images w/ script

mthalman:
Now that there's a [script](https://github.com/dotnet/docker-tools/pull/411) to check the status of the base Windows images, the Patch Tuesday issue template can make use of it in the workflow.


mthalman:
> Are you planning to mirror this change to dotnet-framework-docker as well?

Yes, but I wanted to confirm the wording here first.

PR #1713, opened 2020-02-28
Trigger nightly build on Dockerfile or manifest change only

### LINKED ISSUE #1708: Change nightly pipeline trigger to only run on Dockerfile updates
Currently the [dotnet-core-nightly](https://github.com/dotnet/dotnet-docker/blob/master/eng/pipelines/dotnet-core-nightly.yml) pipeline is triggered whenever a commit to the nightly branch is made.  This just wastes a lot of compute because there are often changes made to the branch that don't result in a change to the images that get published.

I propose that the pipeline be updated so that it only triggers if a change is made to a Dockerfile.  Admittedly, the nightly branch can be useful to test out infrastructure changes as well.  In those cases, I think it's more appropriate for the dev to manually queue a test build if they want to verify the pipeline.

mthalman:
Fixes #1708

### DIFF
--- a/eng/pipelines/dotnet-core-nightly.yml
+++ b/eng/pipelines/dotnet-core-nightly.yml
@@ -4,11 +4,12 @@ trigger:
     include:
     - nightly
   paths:
-    exclude:
-    - .github/*
-    - documentation/*
-    - README*
-    - samples/*
+    include:
+    - manifest.json
+    - 2.1/*
+    - 3.0/*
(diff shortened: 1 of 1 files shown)


mthalman:
> This change raises a question for me whether or not there is value in defining an issue template for onboarding a new .NET version. There are a few nuances that might be beneficial to capture.

I thought the same.  Would probably be worthwhile to have a location to capture things like this.

PR #1760, opened 2020-03-24
Merge changes from nightly

mthalman:


mthalman:
> IDK if it was possible or not but it would be nice to cherry-pick changes when merging if plausible as it keeps the branch history cleaner.

I've created a new PR that does this.

PR #1773, opened 2020-03-26
Update Image Builder ref that uses new image info schema

mthalman:
References a version of Image Builder that is configured to use the new image info schema as defined by https://github.com/dotnet/docker-tools/issues/438.

### DIFF
--- a/manifest.samples.json
+++ b/manifest.samples.json
@@ -1,8 +1,5 @@
 {
   "registry": "mcr.microsoft.com",
-  "variables": {
-    "SampleProductVersion": "3.1"
-  },
   "repos": [
     {
       "id": "samples",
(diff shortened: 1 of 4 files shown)


mthalman:
Need to make some changes to the manifest still.

mthalman:
Had to make a fix for the product version parsing: https://github.com/dotnet/docker-tools/pull/450

manifest.samples.json
```diff
@@ -8,6 +11,7 @@
       "mcrTagsMetadataTemplatePath": "eng/mcr-tags-metadata-templates/samples-tags.yml",
       "images": [
         {
+          "productVersion": "$(SampleProductVersion)",
```
MichaelSimons: Is this required in the samples?  I'm concerned about the version not getting updated when the samples change.

mthalman: No, not needed.  I'm taking this out with my updates to the version parsing.

PR #1779, opened 2020-03-27
Fix test logic to support new image info schema

mthalman:
The test code reads from the image info file to determine whether it needs to pull any images.  Since the image info schema was changed as part of https://github.com/dotnet/docker-tools/issues/438, the code is being updated to support this new schema.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
@@ -27,8 +27,7 @@ namespace Microsoft.DotNet.Docker.Tests
         {
             ImageInfoData = new Lazy<JObject>(() =>
             {
-                //string imageInfoPath = Environment.GetEnvironmentVariable("IMAGE_INFO_PATH");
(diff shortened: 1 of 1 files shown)


tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -21,17 +21,18 @@ public abstract class ImageData
...
+        private static readonly Lazy<JObject> ImageInfoData;
 
         static ImageData()
         {
-            ImageInfoData = new Lazy<JArray>(() =>
+            ImageInfoData = new Lazy<JObject>(() =>
             {
-                string imageInfoPath = Environment.GetEnvironmentVariable("IMAGE_INFO_PATH");
+                //string imageInfoPath = Environment.GetEnvironmentVariable("IMAGE_INFO_PATH");
+                string imageInfoPath = @"C:\repos\versions\build-info\docker\image-info.dotnet-dotnet-docker-nightly.json";
```
MichaelSimons: It looks like some local changes for testing didn't get removed.

PR #1813, opened 2020-04-06
Create abstract image pedigree discovery guidance doc and along w/.NET specific version

MichaelSimons:
* Created an abstract version of the _Determining License and Source Pedigree for Container Images_ document that is intended to be shared across other Microsoft docker repositories.  This document may eventually get hosted outside of this repo.
* Created a new .NET Core specific version of the _Determining License and Source Pedigree for Container Images_ that includes instructions specific to the .NET Core and other components (beyond base images and packages).

**Note:**  The files should get renamed but I left them as is for now so that you can more easily see what was changed.


MichaelSimons:
Closing as this has changed a lot.  I will create a new PR shortly.

Issue #1743, opened 2020-03-13
Please add back DOTNET_SDK_VERSION environment variable in 3.1 SDK images


LordNibbler274:
It appears that the 3.1 SDK images no longer set the DOTNET_SDK_VERSION environment variable, that is available in the 2.1 images. Can we add it back please as this made it very simple to fully qualify the path to the SDK?

richlander:
Can you describe the underlying scenario where you need access to this path? Thanks.

LordNibbler274:
This is probably quite niche but I would like to be able to call vstest.console.dll in the container:

https://github.com/G-Research/DotNetDockerTest/blob/fe38bc8b0b3d0a0f7f73349d66f71324ae0db5d3/src/DockerVSTestTask.cs#L191

Having this variable set means that I don't need to know the container or the specific SDK version being used in order for this to work.

MichaelSimons:
When fixing this, we want to ensure the ENV is set consistently across all images.  In the 2.1 images it is not set in the images that are built with multi-stage Dockerfiles.

MichaelSimons:
A potential workaround would be to grab the version from dotnet output of `dotnet --info` or `dotnet --list-sdks`

richlander:
Wouldn't `dotnet --version` be better? No parsing.

MichaelSimons:
That works if you want the SDK version within the SDK image but doesn't work if you want the runtime version.

The scenario in which I think this would be useful is a slight permutation of the scenario originally described.  For the crossgen usage scenarios we would want to users to be able to write SDK based dockerfiles the contain the following following:

`dotnet publish -c release -o /app --no-restore --target-version $DOTNET_VERSION`

LordNibbler274:
I think that adding DOTNET_VERSION seems like a good way to go. Ultimately anything that is consistent between the different images is fine for my purposes. For the short term my workaround has simply been to build off the microsoft image and add the environment variable myself but being able to use the official image would be much better, thanks.

PR #1832, opened 2020-04-13
Update base image of 5.0 SDK for Alpine and Nano

### LINKED ISSUE #1351: Change the base image of the Windows and Alpine SDK images to maximize layer sharing
The SDK images do not have a consistent pattern of base images.  The Ubuntu and Debian images are based on the buildpack-deps:scm images which provide a rich experience as they include curl, wget, ca-certificates, git, etc.  

When Windows containers came around, the option to base the SDK images on buildpack-deps:scm was unavailable.  As a result, the Windows SDK images are based on the base NanoServer OS images.  On Windows curl is already included but SCM (e.g. git) is not.  With 3.0 PS is included which provides other options for downloading content.  IIRC no feedback has been provided in regards to not having git in the image.

When Alpine support was added, the buildpack-deps:scm option was again unavailable.  The buildpack-deps repo maintainers believe the Alpine images are not intended for development scenarios (https://github.com/docker-library/buildpack-deps/issues/51#issuecomment-260421322).  As a result the Alpine images were based on the runtime-deps images.  The runtime-deps were chosen over the base Alpine image because of the need for the native dependencies.

I think the correct choice was made to utilize the buildpack-deps:scm images where available.  For the images where buildpack-deps is not available, I think adjustments should be made.  In these scenarios basing the SDK images on the runtime images would support maximize layer re-use.  This will provide customers with a better experience.  This is valuable in particular with the Windows images as they are larger when compared to the Linux images.  The runtime specific layers on disk are ~70 MB and the aspnet layers add another ~20 MB.

mthalman:
In order to allow for more layer sharing amongst images, the 5.0 SDK images for Alpine and Nano Server have been changed to be based on their respective ASP.NET images.  This requires that the runtime and ASP.NET components of the SDK be omitted when extracting its contents since they are derived from base layers.

Fixes #1351

### DIFF
--- a/3.1/sdk/alpine3.10/amd64/Dockerfile
+++ b/3.1/sdk/alpine3.10/amd64/Dockerfile
@@ -24,7 +24,7 @@ RUN dotnet_sdk_version=3.1.201 \
     && dotnet_sha512='9a8f14be881cacb29452300f39ee66f24e253e2df947f388ad2157114cd3f44eeeb88fae4e3dd1f9687ce47f27d43f2805f9f54694b8523dc9f998b59ae79996' \
     && echo "$dotnet_sha512  dotnet.tar.gz" | sha512sum -c - \
     && mkdir -p /usr/share/dotnet \
-    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz ./packs ./sdk ./templates ./*.txt \
+    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz ./packs ./sdk ./templates ./LICENSE.txt ./ThirdPartyNotices.txt \
     && rm dotnet.tar.gz \
     # Trigger first run experience by running arbitrary cmd
     && dotnet help
--- a/3.1/sdk/alpine3.10/amd64/Dockerfile
+++ b/3.1/sdk/alpine3.10/amd64/Dockerfile
@@ -1,4 +1,4 @@
-ARG REPO=mcr.microsoft.com/dotnet/core/runtime-deps
+ARG REPO=mcr.microsoft.com/dotnet/core/aspnet
 FROM $REPO:3.1-alpine3.10
 
 ENV \
@@ -24,8 +24,7 @@ RUN dotnet_sdk_version=3.1.201 \
     && dotnet_sha512='9a8f14be881cacb29452300f39ee66f24e253e2df947f388ad2157114cd3f44eeeb88fae4e3dd1f9687ce47f27d43f2805f9f54694b8523dc9f998b59ae79996' \
     && echo "$dotnet_sha512  dotnet.tar.gz" | sha512sum -c - \
     && mkdir -p /usr/share/dotnet \
-    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz \
-    && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet \
+    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz ./packs ./sdk ./templates ./*.txt \
     && rm dotnet.tar.gz \
     # Trigger first run experience by running arbitrary cmd
     && dotnet help
--- a/3.1/sdk/nanoserver-1809/amd64/Dockerfile
+++ b/3.1/sdk/nanoserver-1809/amd64/Dockerfile
@@ -1,5 +1,7 @@
 # escape=`
 
+ARG REPO=mcr.microsoft.com/dotnet/core/aspnet
+
 # Installer image
 FROM mcr.microsoft.com/windows/servercore:1809 AS installer
 
@@ -31,10 +33,16 @@ RUN $powershell_version = '7.0.0'; `
     Remove-Item -Force PowerShell.Windows.x64.$powershell_version.nupkg; `
     Remove-Item -Path \powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force
 
+# Delete everything in the dotnet folder that's not needed in the SDK layer
+RUN Get-ChildItem -Exclude "*.txt","packs","sdk","templates","shared" -Path dotnet | Remove-Item -Force -Recurse; `
+    Get-ChildItem -Exclude Microsoft.WindowsDesktop.App -Path dotnet\shared | Remove-Item -Force -Recurse
+
 # SDK image
-FROM mcr.microsoft.com/windows/nanoserver:1809
+FROM $REPO:3.1-nanoserver-1809
 
 ENV `
+    # Unset ASPNETCORE_URLS derived from aspnet base image
+    ASPNETCORE_URLS= `
     # Enable detection of running in a container
     DOTNET_RUNNING_IN_CONTAINER=true `
     # Enable correct mode for dotnet watch (only mode supported in a container)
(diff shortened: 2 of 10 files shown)


5.0/sdk/nanoserver-1809/amd64/Dockerfile
```diff
@@ -42,14 +44,19 @@ ENV `
     # Skip extraction of XML docs - generally not useful within an image/container - helps performance
     NUGET_XMLDOC_MODE=skip `
     # PowerShell telemetry for docker image usage
-    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetCoreSDK-NanoServer-1809
+    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetCoreSDK-NanoServer-1809 `
+    # Unset ASPNETCORE_URLS derived from aspnet base image
```
MichaelSimons, line 48: I have been following the practice of sorting our multi-line ENV instructions.  I think reasons stated in the Dockerfile best practices for [sorting multi-line arguments](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#sort-multi-line-arguments) apply here.

5.0/sdk/nanoserver-1809/amd64/Dockerfile
```diff
@@ -42,14 +44,19 @@ ENV `
...
+    ASPNETCORE_URLS=
 
 # In order to set system PATH, ContainerAdministrator must be used
 USER ContainerAdministrator
 RUN setx /M PATH "%PATH%;C:\Program Files\dotnet;C:\Program Files\powershell"
 USER ContainerUser
 
-COPY --from=installer ["/dotnet", "/Program Files/dotnet"]
```
MichaelSimons, line 52: Have you explored to possibility of using a single COPY so that we can maintain a single layer for the SDK in the resulting image?  I think this would be more desirable.  This can be done by creating a src location in the installer image that only contains the components you want to copy into the final image.

5.0/sdk/nanoserver-1809/amd64/Dockerfile
```diff
@@ -42,14 +44,19 @@ ENV `
...
     NUGET_XMLDOC_MODE=skip `
     # PowerShell telemetry for docker image usage
-    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetCoreSDK-NanoServer-1809
+    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetCoreSDK-NanoServer-1809 `
+    # Unset ASPNETCORE_URLS derived from aspnet base image
+    ASPNETCORE_URLS=
 
 # In order to set system PATH, ContainerAdministrator must be used
 USER ContainerAdministrator
 RUN setx /M PATH "%PATH%;C:\Program Files\dotnet;C:\Program Files\powershell"
```
MichaelSimons, line 53: It looks like this needs refactoring because I think dotnet would be in the path multiple times because of the base image change.

mthalman:
I'll log an issue to add a test as a separate task so that it can take advantage of the changes in https://github.com/dotnet/dotnet-docker/issues/1743.

3.1/sdk/nanoserver-1909/amd64/Dockerfile
```diff
@@ -31,10 +33,16 @@ RUN $powershell_version = '7.0.0'; `
     Remove-Item -Force PowerShell.Windows.x64.$powershell_version.nupkg; `
     Remove-Item -Path \powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force
 
+# Delete everything in the dotnet folder that's not needed in the SDK layer
```
MichaelSimons, line 36: I feel like `not needed in the SDK layer` should be clarified here.  I don't want to give the impression the content isn't in the resulting SDK image.  I'm not sure how to word this.

```suggestion
# Delete the dotnet content that is not contained within the aspnet base layers.
```

3.1/sdk/nanoserver-1809/amd64/Dockerfile
```diff
@@ -31,10 +33,16 @@ RUN $powershell_version = '7.0.0'; `
     Remove-Item -Force PowerShell.Windows.x64.$powershell_version.nupkg; `
     Remove-Item -Path \powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force
 
+# Delete everything in the dotnet folder that's not needed in the SDK layer
```
MichaelSimons, line 36: Have you considered consolidating the three RUN instructions in the installer stage?  The advantage is that it would eliminate the writing of the contents being deleted in the final RUN to an image layer.

3.1/sdk/alpine3.10/amd64/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN dotnet_sdk_version=3.1.201 \
     && dotnet_sha512='9a8f14be881cacb29452300f39ee66f24e253e2df947f388ad2157114cd3f44eeeb88fae4e3dd1f9687ce47f27d43f2805f9f54694b8523dc9f998b59ae79996' \
     && echo "$dotnet_sha512  dotnet.tar.gz" | sha512sum -c - \
     && mkdir -p /usr/share/dotnet \
-    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz ./packs ./sdk ./templates ./*.txt \
+    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz ./packs ./sdk ./templates ./LICENSE.txt ./ThirdPartyNotices.txt \
```
MichaelSimons, line 27: I like the explicitness of this, is there value in using a consistent pattern within the Windows Dockerfiles?

mthalman: Yes, might as well make them consistent.

3.1/sdk/nanoserver-1809/amd64/Dockerfile
MichaelSimons: It looks like merging the RUN instructions in the 3.1 1803 Dockerfile was missed.

MichaelSimons: I meant 1809.

3.1/sdk/nanoserver-1809/amd64/Dockerfile
```diff
@@ -1,40 +1,48 @@
...
     }; `
-    `
     \dotnet\dotnet tool install --add-source . --tool-path \powershell --version $powershell_version PowerShell.Windows.x64; `
     \dotnet\dotnet nuget locals all --clear; `
     Remove-Item -Force PowerShell.Windows.x64.$powershell_version.nupkg; `
-    Remove-Item -Path \powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force
+    Remove-Item -Path \powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force; `
+    `
+    # Delete everything in the dotnet folder that's not needed in the SDK layer but will instead be derived from base layers
+    $excludedItems = 'LICENSE.txt','ThirdPartyNotices.txt','packs','sdk','templates','shared'; `
```
MichaelSimons, line 36: The variable name here is a little deceptive IMO because of `Microsoft.WindowsDesktop.App`.  Was done so that the next line doesn't wrap?  If so could you do the following instead?

```
    Get-ChildItem -Exclude 'LICENSE.txt','ThirdPartyNotices.txt','packs','sdk','templates','shared' -Path dotnet `
        | Remove-Item -Force -Recurse; `
    Get-ChildItem -Exclude Microsoft.WindowsDesktop.App -Path dotnet\shared | Remove-Item -Force -Recurse
```

mthalman: Yes, that was the reason.  Your suggestion makes sense.

PR #1848, opened 2020-04-16
Drop use of buildpack-deps as base image

### LINKED ISSUE #1814: Upcoming breaking change: Drop reference to buildpack-deps for SDK images
# Upcoming breaking change: Drop reference to buildpack-deps for SDK images

We intend to change the parent image of [.NET SDK images](https://hub.docker.com/_/microsoft-dotnet-core-sdk), from a [buildpack-deps](https://hub.docker.com/_/buildpack-deps) image to a [dotnet/core/aspnet](https://hub.docker.com/_/microsoft-dotnet-core-aspnet) one. This change will only be made to 5.0 images, and only affects Debian and Ubuntu images, as described in the following table.

|Image|Current parent image| New parent image|
|------|----------------------|-----------------------------|
|sdk:5.0-buster| buildpack-deps:buster-scm| aspnet:5.0-buster-slim|
|sdk:5.0-focal| buildpack-deps:focal-scm| aspnet:5.0-focal|
|sdk:5.0-alpine|runtime-deps:5.0| aspnet:5.0-alpine
|sdk:5.0-nanoserver|nanoserver|aspnet:5.0-nanoserver

Alpine and Nano Server are included in the table above, because a similar (but not the same) change is being made for those images. See: https://github.com/dotnet/dotnet-docker/issues/1351.

The benefits are the following:

* Size savings, on the order of the entirety of the size of the `aspnet` image in the case of Debian Buster, and the entirety of the .NET layer of the `aspnet` image in the case of Ubuntu Focal.
* Reduction in surface area, for potential security vulnerabilities.

We don't have the size saving on-hand currently. When that information is available, it will be added to this issue.

We will install `git`, `curl`, and `wget` packages in the 5.0 SDK images going forward.

We are planning to make this change for .NET 5.0 Preview 3 (April 2020).

## Breaks and Workarounds

If you rely on tools provided by `buildpack-deps:scm` (beyond the three mentioned above), you will be broken by this change. We would like to know about that. Please share your scenario in the issue comments.

The complete list of tools provided by `buildpack-deps:scm` can be easily discovered by looking at the following Dockerfiles, for the `curl` and `scm` layers.

For Debian:

* [buildpack-deps:buster-scm](https://github.com/docker-library/buildpack-deps/blob/99a1c33fda559272e9322b02a5d778bbd04154e7/buster/scm/Dockerfile)
* [buildpack-deps:buster-curl](https://github.com/docker-library/buildpack-deps/blob/b0fc01aa5e3aed6820d8fed6f3301e0542fbeb36/buster/curl/Dockerfile)

For Ubuntu:

* [buildpack-deps:focal-scm](https://github.com/docker-library/buildpack-deps/blob/1bf287b61b2c02d8890f4806a9bfb2c7042b308d/focal/scm/Dockerfile)
* [buildpack-deps:focal-curl](https://github.com/docker-library/buildpack-deps/blob/1bf287b61b2c02d8890f4806a9bfb2c7042b308d/focal/curl/Dockerfile)

If you rely on these tools as part of `docker build`, you are encouraged to install the required tool(s) yourself in your Dockerfile, or rely on multi-stage build, by using a `buildpack-deps` stage, as we do for [.NET runtime images](https://github.com/dotnet/dotnet-docker/blob/master/5.0/runtime/focal/amd64/Dockerfile#L4).

If you rely on these tools as part of `docker run`, you can potentially install them via a script.

Please share feedback if you this change breaks your workflow.

## Context

Since .NET Core 1.0, we've been releasing  SDK images on top of [buildpack-deps](https://hub.docker.com/_/buildpack-deps) images. We made this choice for two reasons:

- We needed access to curl to download .NET Core.
- We saw that other developer platforms (for their SDK) depend on `buildpacks-deps:scm` ([example](https://github.com/docker-library/golang/blob/master/1.14/buster/Dockerfile#L1)), which contains curl.

This made choosing `buildpacks-deps:scm` for the .NET Core SDK images an easy choice. We did that and moved on.

Later, Docker developed the multi-stage-build feature, which opened up many new possibilities. In particular, this feature enables you to download or build assets in one stage, and then copy into another. That's a pattern we've adopted for our [runtime images](https://github.com/dotnet/dotnet-docker/blob/master/5.0/runtime/focal/amd64/Dockerfile#L4).

We realized early on that the `buildpack-deps` dependency was problematic because it meant that the SDK images could not depend on runtime images. This means for multi-stage builds that you download the runtime twice, once with the SDK and another time for the final image. That's very unfortunate.

Both Windows and Alpine don't have a `buildpack-deps` image offered. That means that we have more freedom to layer our images. That's what we're proposing to do @ https://github.com/dotnet/dotnet-docker/issues/1351.

Upon further thought, we realized that we didn't have any actual customer evidence that `buildpack-deps` was needed (as part of `sdk` images). We have plenty of evidence that the layer is problematic, and can deliver a sizable performance improvement to anyone using the `sdk` and a `runtime` or `aspnet` image in a workflow (multi-stage build of a .NET image would be the most obvious example). We believe that this change will have a large positive impact on millions of builds ever day, and have a massive impact on any commercial build host.

We hope that we can deliver this benefit to .NET docker image users. In particular, we hope that we can find satisfactory workarounds for users that are impacted by this change.

### LINKED ISSUE #1843: Need test to validate SDK image contents
As part of the changes in #1832, the SDK images for Alpine and Nano were changed to be based on the ASP.NET image.  The SDK layer explicitly installs only the SDK bits that do not exist in the base layers.  There needs to be a test which validates these SDK images to verify that the dotnet directory's contents match what exists in the SDK .zip/.tar file.  This will ensure that all the necessary bits exist in the image and there wasn't an error in how files are being extracted.

mthalman:
* Updates 5.0 SDK buster and focal images to be based on aspnet base image instead of buildpack-deps.
* Adds ca-certificates, curl, git, and wget to 5.0 SDK images for alpine, buster, and focal.
* Adds test to validate the contents of the SDK.
* Adds tests to validate the derivation of the .NET version environment variables.

Fixes #1814
Fixes #1843


MichaelSimons:
Can you provide a summary of the size impact to this change?  What I am looking for is the base image size, final image size and the original image size.  Having this for the debian and ubuntu amd64 images would be sufficient.

mthalman:
## Image Size Summary

### Focal

| aspnet Base | Original (based on buildpack-deps) | Final |
| - | - | - |
| 233 MB | 738 MB | 630 MB |

### Buster

| aspnet Base | Original (based on buildpack-deps) | Final |
| - | - | - |
| 209 MB | 734 MB | 598 MB |

mthalman:
I just realized that buster SDK is now based on buster-slim.  I suppose we should change its tag to be buster-slim now?

MichaelSimons:
> I just realized that buster SDK is now based on buster-slim. I suppose we should change its tag to be buster-slim now?

Yes, I think so.

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -121,6 +139,97 @@ public void VerifyPowerShellScenario_NonDefaultUser(ProductImageData imageData)
...
+            string architecture = imageData.Arch switch
+            {
+                Arch.Amd64 => "x64",
+                Arch.Arm => "arm",
+                Arch.Arm64 => "arm64",
+                _ => throw new InvalidOperationException($"Unexpected architecture type: '{imageData.Arch}'"),
+            };
+
+            string sdkUrl =
+                $"https://dotnetcli.azureedge.net/dotnet/Sdk/{sdkVersion}/dotnet-sdk-{sdkVersion}-{osType}-{architecture}.{fileType}";
```
MichaelSimons, line 214: Shouldn't this take into account `-musl` for alpine?

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -121,6 +139,97 @@ public void VerifyPowerShellScenario_NonDefaultUser(ProductImageData imageData)
...
+                .Replace("c:\\program files\\dotnet", String.Empty)
+                .Replace("\\", "/")
+                .Replace("\r\n", "\n")
+                .Split("\n")
+                .Select(file => file.TrimStart('/'))
+                .OrderBy(file => file);
+
+            IEnumerable<string> expectedDotnetFiles = await GetSdkContentsAsync(imageData);
+
+            Assert.Equal(expectedDotnetFiles, actualDotnetFiles);
```
MichaelSimons, line 182: I feel like we should be doing more than a file name comparison.  Specifically for the license and TPN.  With linux it would be pretty easy to do an checksum/md5sum comparison of all files.  Perhaps it is equally as easy with PS.  Performance may be a concern though for doing it on all files.

tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
```diff
@@ -33,6 +33,23 @@ public void VerifyEnvironmentVariables(ProductImageData imageData)
...
 
+            EnvironmentVariableInfo aspnetVersionVariableInfo = GetAspnetVersionVariableInfo(imageData, DockerHelper);
+            if (aspnetVersionVariableInfo != null)
+            {
+                variables.Add(aspnetVersionVariableInfo);
+            }
+
+            if (imageData.Version.Major >= 5)
+            {
+                string version = imageData.GetProductVersion(DotNetImageType.Runtime, DockerHelper);
```
MichaelSimons, line 44: Should this make use of the `RuntimeImageTests.GetRuntimeVersionVariableInfo` method?

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -121,6 +136,129 @@ public void VerifyPowerShellScenario_NonDefaultUser(ProductImageData imageData)
...
+                "| select -ExpandProperty Value";
+            string command = $"pwsh -Command \"{powerShellCommand}\"";
+
+            string containerFileList = DockerHelper.Run(
+                image: imageData.GetImage(ImageType, DockerHelper),
+                command: command,
+                name: imageData.GetIdentifier("DotnetFolder"));
+
+            IEnumerable<SdkContentFileInfo> actualDotnetFiles = containerFileList
+                .Replace("\\", "/")
```
MichaelSimons, line 177: I see the `SdkContentFileInfo` ctor does some path normalization as well.  Is the `.Replace("\\", "/")` necessary here or can it just reside in the ctor?

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -121,6 +136,129 @@ public void VerifyPowerShellScenario_NonDefaultUser(ProductImageData imageData)
...
+            }
+
+            string powerShellCommand =
+                $"Get-ChildItem -File -Force -Recurse '{dotnetPath}' " +
+                "| Get-FileHash -Algorithm SHA512 " +
+                "| select @{name='Value'; expression={$_.Hash + '  ' +$_.Path}} " +
+                "| select -ExpandProperty Value";
+            string command = $"pwsh -Command \"{powerShellCommand}\"";
+
+            string containerFileList = DockerHelper.Run(
```
MichaelSimons, line 171: There is a lot going on in this test case.  What do you think about encapsulating the logic that around retrieving the actual SdkContentFileInfo into a separate method?

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -121,6 +136,129 @@ public void VerifyPowerShellScenario_NonDefaultUser(ProductImageData imageData)
...
+            // Skip file comparisons for 3.1 until https://github.com/dotnet/sdk/issues/11327 is fixed.
+            if (imageData.Version.Major == 3)
+            {
+                return;
+            }
+
+            int fileCount = expectedDotnetFiles.Count();
+            for (int i = 0; i < fileCount; i++)
+            {
+                Assert.Equal(expectedDotnetFiles.ElementAt(i), actualDotnetFiles.ElementAt(i));
```
MichaelSimons, line 201: How useful is the output when the contents are different?  It looks as though if the files aren't aligned (e.g. if this isn't comparing files with the same names) or the counts are different, then the output would be fairly useless.  Would it make sense to dump out the two lists when there are differences so that the output can be used to find the actual differences rather than having to debug and manually diff the folders?

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -121,6 +136,129 @@ public void VerifyPowerShellScenario_NonDefaultUser(ProductImageData imageData)
...
+            {
+                using SHA512 sha512 = SHA512.Create();
+                byte[] sha512HashBytes = sha512.ComputeHash(File.ReadAllBytes(file.FullName));
+                string sha512Hash = BitConverter.ToString(sha512HashBytes).Replace("-", String.Empty);
+                yield return new SdkContentFileInfo(
+                    file.FullName.Substring(tempFolderContext.Path.Length), sha512Hash);
+            }
+        }
+
+        private async Task<IEnumerable<SdkContentFileInfo>> GetSdkContentsAsync(ProductImageData imageData)
```
MichaelSimons, line 222: This name is a little ambiguous as it doesn't indicate what sdk contents are retrieved.  In the scope of this test, I think the reader should assume the SDK docker image not the archive.
```suggestion
        private async Task<IEnumerable<SdkContentFileInfo>> GetSdkArchiveContentsAsync(ProductImageData imageData)
```

tests/performance/ImageSize.nightly.linux.json
```diff
@@ -101,9 +101,9 @@
     "3.1/sdk/bionic/arm64v8": 629031352,
     "3.1/sdk/focal/amd64": 656452398,
     "3.1/sdk/focal/arm64v8": 672793717,
-    "5.0/sdk/buster/amd64": 689019701,
-    "5.0/sdk/buster/arm32v7": 644889584,
-    "5.0/sdk/buster/arm64v8": 712029090,
+    "5.0/sdk/buster-slim/amd64": 689019701,
```
MichaelSimons, line 104: I would expect this change would trigger a failure in the size tests because of the reduced image sizes.

tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
mthalman: No, because the runtime needs to validate that the variable exists for 2.1 on Linux.  But 2.1 ASP.NET images depend on runtime-deps, not runtime, so it won't be set in that case.

MichaelSimons: My suggestion was to literally call the `RuntimeImageTests.GetRuntimeVersionVariableInfo` method within the if existing if statement.  I realize the conditions are different but the logic to retrieve the EnvironmentVariableInfo appeared identical and since it was encapsulated within a method it seemed to make sense to utilize it here.

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
mthalman: Yeah, that makes sense.  I've updated it to output the full list of expected and actual results if there's going to be any errors logged.

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
mthalman: Not necessary, removed.

tests/performance/ImageSize.nightly.linux.json
```diff
@@ -101,9 +101,9 @@
     "3.1/sdk/bionic/arm64v8": 629031352,
     "3.1/sdk/focal/amd64": 656452398,
     "3.1/sdk/focal/arm64v8": 672793717,
-    "5.0/sdk/buster/amd64": 689019701,
-    "5.0/sdk/buster/arm32v7": 644889584,
-    "5.0/sdk/buster/arm64v8": 712029090,
+    "5.0/sdk/buster-slim/amd64": 598366352,
```
MichaelSimons, line 104: Do the focal images need to be updated as well or are they still within the threshold?

mthalman: There were a bunch of image size baselines that were out of date.  The nanoserver changes were a result of the Preview 3 update.

mthalman:
Here's a comparison of the benefits these changes provide in terms of the **download cost savings**.  The canonical scenario here is a multi-stage Dockerfile that builds and publishes an ASP.NET app (see the sample [ASP.NET Dockerfile](https://github.com/dotnet/dotnet-docker/blob/master/samples/aspnetapp/Dockerfile)).  The scenario involves first pulling the SDK image in order to build the application and then pulling the ASP.NET image to contain the published application.  Before these changes, there was no layer sharing between the ASP.NET image and the SDK image.  This was wasteful because the SDK image does contain ASP.NET bits; it just wasn't based on the ASP.NET image.  With the changes that were made here, the SDK is now based on the ASP.NET image which means that once you've pulled the SDK image, you now essentially have the ASP.NET image as well.  This provides substantial download savings when both images need to be pulled.

### Focal

| | Before | After |
| - | - | - |
| **1. Pull SDK image** | 268 MB | 232 MB |
| **2. Pull ASP.NET image** | 64 MB | 10 KB (manifest only) |

**Net download savings:** 100 MB (-30%)

### Buster

| | Before | After |
| - | - | - |
| **1. Pull SDK image** | 280 MB | 218 MB |
| **2. Pull ASP.NET image** | 84 MB | 4 KB (manifest only) |

**Net download savings:** 146 MB (-40%)

Issue #1765, opened 2020-03-25
Introduce non-core repositories for .NET 5


MichaelSimons:
The [Introducing .NET 5](https://devblogs.microsoft.com/dotnet/announcing-net-5-preview-4-and-our-journey-to-one-net/) blog announces a single unified .NET platform going forward.  

**.NET 5 = .NET Core vNext**

With the _.NET 5_ release the _core_ terminology is dropped from the naming.  As a result it would make sense to do the same in the .NET docker repositories going forward.

**Current Repositories:**

* [dotnet/core](https://hub.docker.com/_/microsoft-dotnet-core/): .NET Core -> product family repo
* [dotnet/core/sdk](https://hub.docker.com/_/microsoft-dotnet-core-sdk/): .NET Core SDK
* [dotnet/core/aspnet](https://hub.docker.com/_/microsoft-dotnet-core-aspnet/): ASP.NET Core Runtime
* [dotnet/core/runtime-deps](https://hub.docker.com/_/microsoft-dotnet-core-runtime-deps/): .NET Core Runtime Dependencies
* [dotnet/core/samples](https://hub.docker.com/_/microsoft-dotnet-core-samples/): .NET Core Samples
* [dotnet/core-nightly](https://hub.docker.com/_/microsoft-dotnet-core-nightly/): .NET Core (Preview)

**Proposed new repositories for .NET 5 going forward:**

* dotnet
* dotnet/sdk
* dotnet/aspnet
* dotnet/runtime-deps
* dotnet/samples
* dotnet-nightly

**Benefits:**
1. A shorter name is more convenient to type.
2. It will be beneficial to align the Docker repository names with the official product name.  Having _core_ in the name won't necessarily be confusing now but over time it will become more confusing especially for new users.

**Sample Usage**

```Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:5.0
```

```Dockerfile
FROM mcr.microsoft.com/dotnet/runtime:5.0
```

PR #1888, opened 2020-05-04
Update 5.0 repo name to "nightly" instead of "core-nightly"

mthalman:
Related to #1765


manifest.json
```diff
@@ -1787,7 +1811,15 @@
               "variant": "v8"
             }
           ]
-        },
+        }
+      ]
+    },
+    {
+      "id": "sdk-preview",
```
MichaelSimons, line 1818: Is it reasonable to name items like the id, readmePath, and template in durable terms that won't change once 5.0 is RTM?  If we used the terms like `sdk` and `core-sdk` we would have long term stable names that won't change.

README.aspnet.preview.md
```diff
@@ -0,0 +1,126 @@
...
+docker run -it --rm -p 8000:80 --name aspnetcore_sample mcr.microsoft.com/dotnet/core/samples:aspnetapp
+```
+
+After the application starts, navigate to `http://localhost:8000` in your web browser.
+
+See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotnet/dotnet-docker/blob/master/samples/host-aspnetcore-https.md) to use HTTPS with this image.
+
+# Related Repos
+
+.NET Core:
```
MichaelSimons, line 43: Shouldn't the `core` terminology be eliminated in the new .NET readmes except for where 3.1 and older versions are referenced?

README.aspnet.preview.md
```diff
@@ -0,0 +1,126 @@
...
+
+```console
+docker run -it --rm -p 8000:80 --name aspnetcore_sample mcr.microsoft.com/dotnet/core/samples:aspnetapp
+```
+
+After the application starts, navigate to `http://localhost:8000` in your web browser.
+
+See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotnet/dotnet-docker/blob/master/samples/host-aspnetcore-https.md) to use HTTPS with this image.
+
+# Related Repos
```
MichaelSimons, line 41: What are you thinking for the UX for relating/referring users between the .NET and .NET Core repos?  We should have messaging that helps provide clarity between the various repos.

manifest.json
mthalman: Makes senses.  Fixed.

README.aspnet.preview.md
mthalman: Fixed

README.aspnet.preview.md
mthalman: This is definitely tricky.  I've proposed some changes in the README.aspnet.md and README.aspnet.preview.md files.  Let me know what you think and I can roll those out to the rest of the readmes.

README.aspnet.preview.md
```diff
@@ -0,0 +1,138 @@
...
+
+.NET Core:
+
+* [dotnet/core](https://hub.docker.com/_/microsoft-dotnet-core/): .NET Core
+* [dotnet/core/samples](https://hub.docker.com/_/microsoft-dotnet-core-samples/): .NET Core Samples
+
+.NET Core (Preview):
```
MichaelSimons, line 50: I lean towards it is only necessary to link to the product family here.  The pattern is to only link to the individual repos that are part of the same product family.

README.aspnet.preview.md
MichaelSimons: I agree it is tricky.  I think this is looking good.  It will help once 5.0 is shipped and not in "preview".

mthalman:
I'll address the remaining review comments in a separate PR.  We're wanting to get this in now to jumpstart the process of onboarding to Docker Hub.

PR #1894, opened 2020-05-07
Fix pipeline schedules to always run

mthalman:
There are a couple pipelines that run on a schedule but are configured incorrectly such that they will only run if there had been a new commit since the last run.  Updated the configuration to set them to always run, regardless of commits.


MichaelSimons:
Thanks for getting to the bottom of this issue!

PR #1908, opened 2020-05-12
Include ca-certificates pkg in 5.0 runtime-deps

mthalman:
Adds the `ca-certificates` package to the runtime-deps images to allow HTTPS connections to succeed by default.  Alpine doesn't require this package in order for this scenario to work since it has the `ca-certificates-cacert` package by default.

Updates the tests to inject custom logic to the `Main` method which verifies the scenario.

Related to #1897

### DIFF
--- a/5.0/runtime-deps/buster-slim/amd64/Dockerfile
+++ b/5.0/runtime-deps/buster-slim/amd64/Dockerfile
@@ -1,9 +1,10 @@
 FROM amd64/debian:buster-slim
 
-# .NET Core dependencies
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         ca-certificates \
+        \
+# .NET Core dependencies
         libc6 \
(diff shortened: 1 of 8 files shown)


5.0/runtime-deps/buster-slim/amd64/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM amd64/debian:buster-slim
 # .NET Core dependencies
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
+        ca-certificates \
```
MichaelSimons, line 6: Thoughts on using the existing format used in the previous versions?  This would make the formatting consistent and help make it clear that ca-certs is not a dependency of .NET Core itself.

mthalman: Fixed

PR #1919, opened 2020-05-14
Include 5.0 in core repos

mthalman:
Updates the manifest so that 5.0 images are published to the `dotnet/core` repo in addition to the `dotnet` repo.  For example, the 5.0 SDK image for Ubuntu Focal will have a tag like `mcr.microsoft.com/dotnet/core/sdk:5.0-focal` in addition to `mcr.microsoft.com/dotnet/sdk:5.0-focal`.  The 5.0 tags using the `dotnet/core` repo will not be published in readme tags table.

This ensures that consumers of the old repo, `dotnet/core`, will continue to get the latest 5.0 images.  This is intended to be a temporary measure for the transition to the new repo name.  The intent is to no longer tag the images with the `dotnet/core` repo in .NET 5 Preview 5.

Related to #1765

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -389,6 +389,150 @@
               "variant": "v8"
             }
           ]
+        },
+        {
+          "productVersion": "$(5.0-RuntimeVersion)",
+          "sharedTags": {
+            "$(5.0-RuntimeVersion)": {
+              "isUndocumented": true
+            },
+            "5.0": {
+              "isUndocumented": true
+            }
+          },
+          "platforms": [
+            {
+              "dockerfile": "5.0/runtime-deps/buster-slim/amd64",
+              "os": "linux",
+              "osVersion": "buster",
+              "tags": {
+                "$(5.0-RuntimeVersion)-buster-slim": {
+                  "isUndocumented": true
+                },
+                "5.0-buster-slim": {
+                  "isUndocumented": true
+                }
+              }
+            },
+            {
+              "architecture": "arm",
+              "dockerfile": "5.0/runtime-deps/buster-slim/arm32v7",
+              "os": "linux",
+              "osVersion": "buster",
+              "tags": {
+                "$(5.0-RuntimeVersion)-buster-slim-arm32v7": {
+                  "isUndocumented": true
+                },
+                "5.0-buster-slim-arm32v7": {
+                  "isUndocumented": true
+                }
+              },
+              "variant": "v7"
+            },
+            {
+              "architecture": "arm64",
+              "dockerfile": "5.0/runtime-deps/buster-slim/arm64v8",
+              "os": "linux",
+              "osVersion": "buster",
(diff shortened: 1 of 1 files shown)


mthalman:
Actually, I'm going to remove the "new" buster-slim tags from the "old" `dotnet/core` repo.  The intent is to not introduce any new tags in the old repo.  Stand by.

manifest.json
```diff
@@ -2612,6 +2612,9 @@
                 },
                 "5.0-buster-slim": {
```
MichaelSimons, line 2613: I don't think we should be including the `buster-slim` tags at all in this repo.  The only reason we are continuing to push to the old repo is for backwards compatibility.

mthalman: Yep, fixing this up.

manifest.json
```diff
@@ -2026,6 +2587,193 @@
...
+            {
+              "buildArgs": {
+                "REPO": "$(Repo:core-aspnet)"
+              },
+              "architecture": "arm",
+              "dockerfile": "5.0/sdk/buster-slim/arm32v7",
+              "os": "linux",
+              "osVersion": "buster",
+              "tags": {
+                "$(5.0-SdkVersion)-buster-slim-arm32v7": {
```
MichaelSimons, line 2624: The arm32v7 and arm64v8 images should have `buster` tags instead of `buster-slim` tags just like the amd64 images.

manifest.json
```diff
@@ -389,6 +389,150 @@
               "variant": "v8"
             }
           ]
+        },
+        {
+          "productVersion": "$(5.0-RuntimeVersion)",
+          "sharedTags": {
+            "$(5.0-RuntimeVersion)": {
```
MichaelSimons, line 396: Should we be pushing any of the 3 part version tags?  These would be new tags with preview 4 therefore it would not be breaking anyone if we didn't push them.

mthalman: I don't think so.  I think it's actually preferable if some tried to pull `mcr.microsoft.com/dotnet/core/sdk:5.0.100-preview.4-focal` and it failed.  That would force them to figure out why it's not there, a push to get people to migrate to the new repo name.

MichaelSimons: Agreed.

manifest.json
```diff
@@ -2026,6 +2479,175 @@
...
+            {
+              "buildArgs": {
+                "REPO": "$(Repo:core-aspnet)"
+              },
+              "architecture": "arm64",
+              "dockerfile": "5.0/sdk/buster-slim/arm64v8",
+              "os": "linux",
+              "osVersion": "buster",
+              "tags": {
+                "$(5.0-SdkVersion)-buster-slim-arm64v8": {
```
MichaelSimons, line 2540: I don't think we want to push any of the 3 part version tags.

PR #1923, opened 2020-05-15
Update package dependencies in 5.0 images

mthalman:
Because of the changes in https://github.com/dotnet/dotnet-docker/pull/1908, we no longer need to be installing `ca-certificates` in the SDK images.  These changes remove that package from the Dockerfiles.

Since `ca-certificates` is missing from the Alpine `runtime-deps` image, that package is being moved from `sdk` to `runtime-deps`.  This enables parity across the distros.

It was also found that .NET dependencies were still being installed in the `sdk` images which no longer needs to be done because those `sdk` images have been switched from `buildpack-deps` to `aspnet` as the base image.

### DIFF
--- a/5.0/runtime-deps/alpine3.11/amd64/Dockerfile
+++ b/5.0/runtime-deps/alpine3.11/amd64/Dockerfile
@@ -1,7 +1,9 @@
 FROM amd64/alpine:3.11
 
-# .NET Core dependencies
 RUN apk add --no-cache \
+    ca-certificates \
+    \
+# .NET Core dependencies
     krb5-libs \
     libgcc \
     libintl \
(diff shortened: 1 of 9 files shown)


MichaelSimons:
> The one sorta-gray area is with Alpine. Alpine has the ca-certificates-cacert package installed by default which should be sufficient for most scenarios. So ca-certificates is being removed from the Alpine SDK image as well.

How confident are you that `ca-certificates-cacert` is sufficient?  Adding ca-certificates appears to be 553kB.

mthalman:
> To have parity across our images, it would be nice to add ca-certificates in Alpine.

Good point on the parity aspect.  That's important.  I suppose I should move it up to Alpine runtime-deps then?

MichaelSimons:
>I suppose I should move it up to Alpine runtime-deps then?

Yes, that is what I was thinking.

mthalman:
I realized that it's more than just `ca-certificates`.  As a result of the transition from buildpack-deps to aspnet for the base image, we no longer need to install the .NET prereqs.

MichaelSimons:
> I realized that it's more than just ca-certificates. As a result of the transition from buildpack-deps to aspnet for the base image, we no longer need to install the .NET prereqs.

Good catch, I completely missed that in the code review.

Issue #1906, opened 2020-05-12
The client firewall rules update notice in the readmes is poorly formatted


MichaelSimons:
Example 1:
![image](https://user-images.githubusercontent.com/8290530/81632773-b822f200-93d0-11ea-90a4-cf905dbf0d15.png)

Example 2:
![image](https://user-images.githubusercontent.com/8290530/81632912-059f5f00-93d1-11ea-8db3-f66319169c69.png)

The `For more info` line should not be formatted as a header.  The formatting of the .NET readme's is different than the other Microsoft repos.

![image](https://user-images.githubusercontent.com/8290530/81632872-ec96ae00-93d0-11ea-9ca5-5c788219e509.png)

MichaelSimons:
@mbrown555, can you correct this formatting issue?

PR #1926, opened 2020-05-15
Adjust formatting of the MCR firewall change notification

MichaelSimons:
Related #1906

Here is a [sample repo](https://hub.docker.com/repository/docker/msimons/test2) I published the readme to so that you can see the rendered description.

The formatting of the nightly repos are slightly different as compared to the [official repos](https://github.com/dotnet/dotnet-docker/pull/1925) because of the existing prelude text.

cc @mbrown555

### DIFF
--- a/README.samples.md
+++ b/README.samples.md
@@ -3,7 +3,6 @@
 To provide a consistent FQDNs, the data endpoint will be changing from *.cdn.mscr.io to *.data.mcr.microsoft.com
 
 For more info, see [MCR Client Firewall Rules](https://aka.ms/mcr/firewallrules).
----------------------------------------------------------------------------------
 
 # Featured Tags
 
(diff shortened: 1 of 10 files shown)


README.samples.md
```diff
@@ -3,7 +3,6 @@
 To provide a consistent FQDNs, the data endpoint will be changing from *.cdn.mscr.io to *.data.mcr.microsoft.com
 
 For more info, see [MCR Client Firewall Rules](https://aka.ms/mcr/firewallrules).
----------------------------------------------------------------------------------
```
mthalman, line 6: Add the line here too?

MichaelSimons: No because this isn't a "nightly" readme and therefore doesn't have the same prelude comment as the other readmes which is the reason for the separator.  This format is the same as all of the readmes in master.

PR #1950, opened 2020-05-27
Initial dotnet-monitor Docker image and documentation

jander-msft:
Creates a single Docker image consisting of the dotnet-monitor tool that can be used as a sidecar container to monitor diagnostic information about .NET processes.
- The image is intended to be an appliance rather than a base image to build on top.
- The image is based on ASP.NET Core 3.1 on Alpine 3.11 distro at this time. In the near future, I will look into moving to a .NET 5.0 base image, depending on the optimizations we want to make for the image and container.
- The image is only tagged as latest, 5.0, and 5.0.0 at this time as we are currently determining the tagging scheme. We plan to increment the version for each new release of dotnet-monitor e.g. the next non-servicing release after 5.0.0 would be 5.1.0.


manifest.monitor.json
```diff
@@ -0,0 +1,34 @@
+{
+  "readmePath": "README.md",
+  "registry": "mcr.microsoft.com",
+  "variables": {
+    "MonitorVersion": "5.0-preview.6"
```
wiktork, line 5: Did we want to start with version 1 and disconnect the versioning from .net versioning?

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG BASEREPO=mcr.microsoft.com/dotnet/core/aspnet
+ARG STAGEREPO=mcr.microsoft.com/dotnet/core/sdk
+
+# Install dotnet-monitor tool into SDK image. For .NET Core 3.1, only the SDK is able to install dotnet tools.
+FROM $STAGEREPO:3.1-alpine3.11 AS stage
+ENV DOTNET_MONITOR_VERSION=5.0.0-preview.20275.1
```
wiktork, line 6: Does this dockerfile get auto updated? Do we have to keep changing the source version?

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# To reduce image size, remove the copy of the nupkg under the tools.
+RUN find /root/.dotnet/tools -print | grep -i '.*[.]nupkg$' | xargs rm
+
+# Create tools image by copying tools from stage container
+FROM $BASEREPO:3.1-alpine3.11
+ENV PATH="${PATH}:/root/.dotnet/tools"
+COPY --from=stage /root/.dotnet/tools /root/.dotnet/tools
+
+ENTRYPOINT ["dotnet-monitor", "collect"]
```
wiktork, line 22: Would it be clearer if we just copied the payload to top level folder (similar to /app) and launched it directly?

monitor/alpine3.11/amd64/Dockerfile
jander-msft: With the current changes, this does not get automatically updated. We will have to manually update the package version and the hash value by hand for now.

Additional work will need to be done in the dotnet/diagnostics repository to push the hash value to some public location (e.g. public blob storage) and then additional work in dotnet/dotnet-docker to update the update-dependencies project to understand from where to read that information and how to update the dockerfiles.

manifest.monitor.json
jander-msft: This is actually the version of the dotnet-monitor package, which is 5.0.0-preview.20275.1 in the dockerfile, but we want to not have to update the manifest for every incremental change in the package version. I believe that dotnet/diagnostics repository already forked for .NET 5 preview 5 and the master branch currently tracks with .NET 5.0 preview 6+ at this time.

I think we'd want to keep the product version and version tag (when I add that into the manifest) to be the same as whatever dotnet-monitor uses so that the docker image has a version that is the same as the dotnet-monitor tool within that image.

With that, if we want the version decoupled, then dotnet-monitor should decouple its versioning from .NET in general within the dotnet/diagnostics repository.

monitor/alpine3.11/amd64/Dockerfile
jander-msft: I did it this way because this is the standard global location for installing .NET tools into Linux and it allows us to add more tools in the future without having to move things around (although, what would be very easy to do).

In any case, I'm fine with any location for this.

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG BASEREPO=mcr.microsoft.com/dotnet/core/aspnet
+ARG STAGEREPO=mcr.microsoft.com/dotnet/core/sdk
+
+# Install dotnet-monitor tool into SDK image. For .NET Core 3.1, only the SDK is able to install dotnet tools.
+FROM $STAGEREPO:3.1-alpine3.11 AS stage
+ENV DOTNET_MONITOR_VERSION=5.0.0-preview.20275.1
+ENV PATH="${PATH}:/root/.dotnet/tools"
```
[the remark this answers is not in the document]
jander-msft: Not needed

eng/mcr-tags-metadata-templates/monitor-tags.yml
```diff
@@ -0,0 +1,2 @@
+$(McrTagsYmlRepo:monitor)
+$(McrTagsYmlTagGroup:latest)
```
mthalman, line 2: Missing EOF new line.

manifest.json
```diff
@@ -1962,6 +1964,35 @@
           ]
         }
       ]
+    },
+    {
+      "id": "monitor",
+      "name": "dotnet/nightly/monitor",
+      "readmePath": "README.monitor.preview.md",
```
mthalman, line 1971: There's no need to use `preview` in the name here.  The only reason `preview` exists in the other readme filenames is because we needed distinct readmes as part of the transition from `dotnet/core` repo names to just `dotnet` going forward for 5.0+.  Once 2.1 and 3.1 go out of support, everything will move into the non-preview readme files.

manifest.json
```diff
@@ -1962,6 +1964,35 @@
...
+            {
+              "buildArgs": {
+                "BASEREPO": "$(Repo:core-aspnet)",
+                "INSTALLERREPO": "$(Repo:core-sdk)"
+              },
+              "dockerfile": "monitor/alpine3.11/amd64",
+              "os": "linux",
+              "osVersion": "alpine3.11",
+              "tags": {
+                "latest": {}
```
mthalman, line 1990: Technically, `latest` should be moved up to `sharedTags`.  The tags specified here should be platform-specific.  Since you only have one platform, all the shared tags could just as very well be moved here too.  I'm not sure if our infrastructure would support not having any platform-specific tags, particularly with the readme tags generation.

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,19 @@
+ARG BASEREPO=mcr.microsoft.com/dotnet/core/aspnet
+ARG INSTALLERREPO=mcr.microsoft.com/dotnet/core/sdk
+
+# Install dotnet-monitor tool into SDK image. For .NET Core 3.1, only the SDK is able to install dotnet tools.
+FROM $INSTALLERREPO:3.1-alpine3.11 AS installer
+ENV DOTNET_MONITOR_VERSION=5.0.0-preview.20278.1
+RUN wget -O dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg https://pkgs.dev.azure.com/dnceng/public/_apis/packaging/feeds/dotnet5-transport%40Local/nuget/packages/dotnet-monitor/versions/$DOTNET_MONITOR_VERSION/content?api-version=5.1-preview.1 \
```
mthalman, line 7: To simplify the content here, you could just omit the version value for the output nupkg file name.

jander-msft: 'dotnet tool install' requires the package version to be in the file name. I'll add a comment to that effect.

manifest.json
jander-msft: The tag generation seems to be okay with it, but the image size validation falls over because it requires at least one platform specific tag. I've moved the 'latest' tag back to the platform.tags object for now and will move it to sharedTags once we agree on platform specific tags.

jander-msft:
@mthalman and @MichaelSimons, are there any other changes you would like me to make before completing these changes? I'll make issues for the other items that Michael highlight in the review summary.

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    && dotnetmonitor_sha512='26d5b43b5bf9ea8dbdd810e6892481a23774490f23bc3e9ea08b9b50c5e5b66111210b90794a846a0ca696d1ed6835976cf453c71928129851aa51149103568e' \
+    && echo "$dotnetmonitor_sha512  dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg" | sha512sum -c - \
+    && dotnet tool install dotnet-monitor --tool-path /app --add-source / --version $DOTNET_MONITOR_VERSION --no-cache \    
+    # To reduce image size, remove the copy of the nupkg under the tools.
+    && find /app -print | grep -i '.*[.]nupkg$' | xargs rm \
+    && rm dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg
+
+# Create tools image by copying tools from installer container.
+FROM $BASEREPO:3.1-alpine3.11
+COPY --from=installer /app /app
```
MichaelSimons, line 18: Have you considered setting the WORKDIR to `/app`?  It may help set the context if anyone interactively runs the image or wants to tweak the entrypoint.

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,20 @@
+ARG BASEREPO=mcr.microsoft.com/dotnet/core/aspnet
```
MichaelSimons, line 1: `BASEREPO` is ambiguous because of the multiple stages.  I lean towards ASPNET_REPO and SDK_REPO as they make the FROM statements more readable.

Regardless of the name, it should use snake casing (`_`) to be consistent with the existing convention.

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ARG INSTALLERREPO=mcr.microsoft.com/dotnet/core/sdk
+
+# Install dotnet-monitor tool into SDK image. For .NET Core 3.1, only the SDK is able to install dotnet tools.
+FROM $INSTALLERREPO:3.1-alpine3.11 AS installer
+ENV DOTNET_MONITOR_VERSION=5.0.0-preview.20278.1
+# 'dotnet tool install' requires package versions to be in the file name.
+RUN wget -O dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg https://pkgs.dev.azure.com/dnceng/public/_apis/packaging/feeds/dotnet5-transport%40Local/nuget/packages/dotnet-monitor/versions/$DOTNET_MONITOR_VERSION/content?api-version=5.1-preview.1 \
+    && dotnetmonitor_sha512='26d5b43b5bf9ea8dbdd810e6892481a23774490f23bc3e9ea08b9b50c5e5b66111210b90794a846a0ca696d1ed6835976cf453c71928129851aa51149103568e' \
+    && echo "$dotnetmonitor_sha512  dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg" | sha512sum -c - \
+    && dotnet tool install dotnet-monitor --tool-path /app --add-source / --version $DOTNET_MONITOR_VERSION --no-cache \
```
MichaelSimons, line 11:
```suggestion
    && dotnet tool install dotnet-monitor --tool-path /app --add-source / --version $DOTNET_MONITOR_VERSION --no-cache \
```

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,20 @@
+ARG BASEREPO=mcr.microsoft.com/dotnet/core/aspnet
+ARG INSTALLERREPO=mcr.microsoft.com/dotnet/core/sdk
+
+# Install dotnet-monitor tool into SDK image. For .NET Core 3.1, only the SDK is able to install dotnet tools.
```
MichaelSimons, line 4: I see value in simplifying some of the comments.  In particular I see `For .NET Core 3.1, only the SDK is able to install dotnet tools.` as adding confusion.  This Dockerfile is using the recommended pattern for 3.1.  There are no other supported options for installing a global tool without the SDK.  I also see value in using a consistent comments indicating the various stages across the .NET Dockerfiles - e.g `# Installer image`.  Lastly we have found it helps the readability to have double blank lines between the stages.

```
ARG ASPNET_REPO=mcr.microsoft.com/dotnet/core/aspnet
ARG SDK_REPO=mcr.microsoft.com/dotnet/core/sdk

# Installer image
FROM $SDK_REPO:3.1-alpine3.11 AS installer

# Install .NET Monitor
ENV DOTNET_MONITOR_VERSION=5.0.0-preview.20278.1
RUN wget -O dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg https://pkgs.dev.azure.com/dnceng/public/_apis/packaging/feeds/dotnet5-transport%40Local/nuget/packages/dotnet-monitor/versions/$DOTNET_MONITOR_VERSION/content?api-version=5.1-preview.1 \
    && dotnetmonitor_sha512='26d5b43b5bf9ea8dbdd810e6892481a23774490f23bc3e9ea08b9b50c5e5b66111210b90794a846a0ca696d1ed6835976cf453c71928129851aa51149103568e' \
    && echo "$dotnetmonitor_sha512  dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg" | sha512sum -c - \
    && dotnet tool install dotnet-monitor --tool-path /app --add-source / --version $DOTNET_MONITOR_VERSION --no-cache \
    # To reduce image size, remove the copy of the nupkg under the tools.
    && find /app -print | grep -i '.*[.]nupkg$' | xargs rm \
    && rm dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg


# Monitor image
FROM $ASPNET_REPO:3.1-alpine3.11

COPY --from=installer /app /app

ENTRYPOINT ["/app/dotnet-monitor", "collect"]
```

monitor/alpine3.11/amd64/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+
+# Monitor image
+FROM $ASPNET_REPO:3.1-alpine3.11
+
+WORKDIR /app
+COPY --from=installer /app .
+ENV PATH="${PATH}:/app"
+
+ENTRYPOINT ["dotnet-monitor", "collect"]
```
jander-msft, line 25: @wiktork , @shirhatti: Should we enable metrics in the container by default (adding '--metrics' switch) so that kubernetes deployments don't have to adjust the command entrypoint for this image. Maybe this is something that we can also enable via environment variable so that deployments can just specify an environment variable without having to replicate the existing command entrypoint.

wiktork: It is enabled by default. I added this switch for an easy disablement.

jander-msft: Thanks for the correction.

eng/Get-TagsDocumentation.ps1
```diff
@@ -40,3 +40,5 @@ $gitRepo = "https://github.com/dotnet/dotnet-docker"
     dotnet/$($repoName)sdk README.sdk.preview.md manifest.json $gitRepo $Branch -ReuseImageBuilderImage
 & $PSScriptRoot/common/Invoke-ReadmeGeneration.ps1 `
     dotnet/core/samples README.samples.md manifest.samples.json $gitRepo -ReuseImageBuilderImage
+& $PSScriptRoot/common/Invoke-ReadmeGeneration.ps1 `
+    dotnet/$($repoName)monitor README.monitor.md manifest.json $gitRepo $Branch -ReuseImageBuilderImage
```
mthalman, line 44: Missing EOF new line.

Issue #1966, opened 2020-06-01
Add support for Alpine 3.12


MichaelSimons:
# New Linux Distro Version

Distro: [Alpine 3.12](https://alpinelinux.org/posts/Alpine-3.12.0-released.html)

## ImageBuilder Tasks

- [x] Ensure that the ImageBuilder supports the new distro version in the [code](https://github.com/dotnet/docker-tools/blob/master/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs) to generate the correct README display name from the version specified in the manifest

## Nightly Branch Tasks

1. - [x] Copy the Dockerfiles of the most recent version of that distro (or author new ones for an entirely new distro) and place them in a version-specific folder under their respective variants (runtime-deps, runtime, aspnet, sdk). Note: not all variants have a corresponding runtime-deps image.
1. - [x] Modify the Dockerfiles as appropriate for any specific changes related to the new distro version
1. - [x] Update [manifest.json](https://github.com/dotnet/dotnet-docker/blob/nightly/manifest.json) to reference the new set of Dockerfiles with the appropriate tags
      - [x] Move any distro-specific floating tags to the newer version (e.g. `3.1-alpine`)
1. - [x] Update the [test data](https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/TestData.cs) to include the new distro version
1. - [x] Update the [tags metadata templates](https://github.com/dotnet/dotnet-docker/tree/master/eng/mcr-tags-metadata-templates) to include the new distro version
1. - [x] Run the command to update the READMEs: `.\eng\Get-TagsDocumentation.ps1`
1. - [x] Inspect generated changes for correctness
1. - [x] Consider whether sample Dockerfiles should be authored if this is a new distro and them to the [samples](https://github.com/dotnet/dotnet-docker/tree/master/samples)
1. - [x] Run the command to build and test your changes: `.build-and-test.ps1 -OS <os>`
   1. - [x] Run the command to update the image size baseline file: `.\tests\performance\Validate-ImageSize.ps1 -UpdateBaselines`
1. - [x] Commit generated changes
1. - [x] Create PR
1. - [x] Get PR signoff
1. - [x] Merge PR to nightly branch
1. - [x] Wait for automatically queued CI build to finish on [dotnet-docker-nightly pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=359) (internal MSFT link)
[...]

PR #1971, opened 2020-06-02
Add Dockerfiles for Alpine 3.12

mthalman:
I also decided to just go ahead and update the .NET Monitor Dockerfile to use Alpine 3.12.

Note that the 2.1 images have significantly increased due to the installation of lttng-ust which has a dependency on python3 in the Alpine 3.12 package feed.  I'm following up with the maintainer to see whether this was intentional.

Related to #1966


manifest.json
```diff
@@ -2065,7 +2412,7 @@
       "mcrTagsMetadataTemplatePath": "eng/mcr-tags-metadata-templates/monitor-tags.yml",
       "images": [
         {
-          "productVersion": "$(MonitorFullVersion)",
+          "productVersion": "$(3.1-RuntimeVersion)",
```
jander-msft, line 2415: .NET Monitor is supposed to be versioned separately from .NET runtime/sdk/etc. It's not tied to any specific version of .NET Core.

mthalman: I know but `productVersion` basically represents .NET version and this image is based on 3.1, not 5.0.  My changes in this PR actually exposed an existing issue that we didn't catch.  For our PR builds, we build the entire graph of dependent images in a single job.  This `productVersion` value is used as part of that grouping logic.  What was actually happening is that the monitor image was being built in the job for the 5.0 image instead of 3.1.  This ended up automatically pulling the image it needed from MCR which is why we never caught it.  But with these new changes, there isn't an existing 3.12 image to pull down so it causes an error.

@MichaelSimons - What are your thoughts on this?  My proposed changes fix the problem but then since it's also used for reporting purposes, it doesn't actually present the results with the correct value (the .NET Monitor version).

MichaelSimons: I agree this needs to be the .NET Version for the way it is used currently.  The name is a little confusing in this context but that is acceptable.

Regarding the specific build issue, I've always wanted a way to disable the Docker auto pull functionality so that it could be disabled in the builds in order to catch scenarios like this.  I poked around a long time ago and never found support.  Maybe it is worth logging an issue in Moby for this.

tests/performance/ImageSize.nightly.linux.json
```diff
@@ -3,6 +3,7 @@
     "2.1/runtime-deps/stretch-slim/amd64": 99052362,
     "2.1/runtime-deps/stretch-slim/arm32v7": 79329954,
     "2.1/runtime-deps/alpine3.11/amd64": 12333463,
+    "2.1/runtime-deps/alpine3.12/amd64": 55514811,
```
MichaelSimons, line 6: Please log an issue to track this size regression if it is not resolved at the time this is merged.

manifest.json
mthalman: https://github.com/moby/moby/issues/41064

tests/performance/ImageSize.nightly.linux.json
mthalman: https://github.com/dotnet/dotnet-docker/issues/1973

PR #1974, opened 2020-06-03
Update monitor tags, folder structure, and readme.

jander-msft:
Add more tags for .NET Monitor 5.0.0 preview release.
Nest .NET Monitor 5.0.X dockerfiles under 5.0 folder.
Update readme with minor changes.


manifest.json
```diff
@@ -8,8 +8,7 @@
     "3.1-SdkVersion": "3.1.300",
     "5.0-RuntimeVersion": "5.0.0-preview.5",
     "5.0-SdkVersion": "5.0.100-preview.5",
-    "MonitorFullVersion": "5.0.0",
-    "MonitorMajorMinorVersion": "5.0"
+    "Monitor-5.0-Version": "5.0.0-preview.1"
```
jander-msft, line 11: @shirhatti, please confirm that the first preview of .NET Monitor will be Preview 1.

manifest.json
```diff
@@ -2414,20 +2413,22 @@
...
+            "latest": {}
           },
           "platforms": [
             {
               "buildArgs": {
                 "ASPNET_REPO": "$(Repo:core-aspnet)",
                 "SDK_REPO": "$(Repo:core-sdk)"
               },
-              "dockerfile": "monitor/alpine3.12/amd64",
+              "dockerfile": "monitor/5.0/alpine3.12/amd64",
```
MichaelSimons, line 2426: Since the tag is distro version agnostic, the path should reflect that.  In other words if you never intend to support multiple alpine versions concurrently, then including the alpine version in the path is unnecessary/undesirable.

manifest.json
shirhatti: Ack. First version will be preview1

PR #1975, opened 2020-06-04
Exclude diagnostics on monitor image to prevent self-diagnosis.

jander-msft:
With dotnet-monitor running in ASP.NET Core 3.1, it establishes its own diagnostics pipe server, which is visible through the REST API that dotnet-monitor provides. We should set the COMPlus_EnableDiagnostics environment variable to 0 in order to disable the diagnostics pipe server within dotnet-monitor. If absolutely necessary, users of the image can override this by setting the value  to 1 within their orchestration environment.

### DIFF
--- a/monitor/5.0/alpine/amd64/Dockerfile
+++ b/monitor/5.0/alpine/amd64/Dockerfile
@@ -20,7 +20,11 @@ FROM $ASPNET_REPO:3.1-alpine3.12
 
 WORKDIR /app
 COPY --from=installer /app .
-ENV COMPlus_EnableDiagnostics=0
-ENV PATH="${PATH}:/app"
+
+ENV \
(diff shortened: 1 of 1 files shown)


monitor/5.0/alpine/amd64/Dockerfile
```diff
@@ -20,6 +20,7 @@ FROM $ASPNET_REPO:3.1-alpine3.12
 
 WORKDIR /app
 COPY --from=installer /app .
+ENV COMPlus_EnableDiagnostics=0
```
MichaelSimons, line 23: Suggestion: use the multiple variables ENV syntax as demonstrated [here](https://github.com/dotnet/dotnet-docker/blob/master/5.0/sdk/alpine3.11/amd64/Dockerfile#L4)

A comment may be useful on why this is set.

Issue #1973, opened 2020-06-03
Image size regression for 2.1 runtime-deps on Alpine 3.12


mthalman:
The changes to add support for Alpine 3.12 introduced a sizeable increase to the 2.1 runtime-deps image.  It went from 12.3 MB in Alpine 3.11 to 55.5 MB in Alpine 3.12.  This is because of the addition of a python3 dependency from the lttng-ust package.  This issue tracks the Alpine package issue that was logged as a result of this dependency change: https://gitlab.alpinelinux.org/alpine/aports/-/issues/11610.

mthalman:
This has been fixed in the Alpine package feed.  The python3 package is no longer a dependency.  It's been made a dependency of the sub-package lttng-ust-tools.

MichaelSimons:
Won't the image size baseline need to be updated?

PR #1979, opened 2020-06-04
Update image size baselines

mthalman:
Updating 2.1 images for Alpine 3.12 now that #1973 is fixed.

Also updated Nano Server 2004 SDK image.  Not sure why the original value was incorrect.  I've checked the original 2004 SDK image that was published and it's size is relatively was same as the updated value.


mthalman:
Merging since the image size validation pipeline succeeded.

Issue #1972, opened 2020-06-03
Dockerfile directory structure should be refactored.


MichaelSimons:
Currently the Dockerfile directory structure follows the following pattern.

`<.net version>/<repo>/<distro>/<arch>`

This pattern made sense when all of the images were in a single repo within Docker hub - microsoft/dotnet.  The directory structure matched the tag pattern - `<.net version>-<repo>-<distro>-<arch>`.  With the move to MCR, the image variants were split into their own individual repos but the directory structure was not refactored to align with the new tagging pattern  

There is active work to onboard additional image variants.  The first one is [dotnet/monitor](https://github.com/dotnet/dotnet-docker/pull/1950).  The dotnet-monitor tool is going to version independently of the .NET Core product.  Because of this, the top level version number is going to cause confusion because not all repos will have the same version numbers.

The Dockerfile directory structure should be consistent across image variants and should support independent versioning.  Based on this, the following directory structure is recommended:

`<repo>/<product version>/<distro>/<arch>`

PR #1999, opened 2020-06-11
Refactor Dockerfile directories to align with tags

MichaelSimons:
Related #1972


eng/update-dependencies/Program.cs
```diff
@@ -293,10 +293,7 @@ private static string GetBuildVersion(this IEnumerable<IDependencyInfo> buildInf
         private static IEnumerable<IDependencyUpdater> GetUpdaters(
             string dockerfileVersion, IEnumerable<IDependencyInfo> buildInfos)
         {
-            string[] dockerfiles = Directory.GetFiles(
-                Path.Combine(RepoRoot, dockerfileVersion),
-                "Dockerfile",
-                SearchOption.AllDirectories);
+            string[] dockerfiles = Directory.GetFiles(RepoRoot, "Dockerfile", SearchOption.AllDirectories);
```
mthalman, line 296: This is no longer filtering by `dockerfileVersion`.  Won't this attempt to update all the Dockerfiles instead of just the version being targeted when update-dependencies is invoked?

MichaelSimons: Yes I'm still working through these changes.  The PR is marked as draft.

MichaelSimons: This is now fixed.

MichaelSimons:
> Once this gets merged and built, it's going to cause a bunch of additional entries made into https://github.com/dotnet/versions/blob/master/build-info/docker/image-info.dotnet-dotnet-docker-nightly.json. So that file can either be prepped to reflect the new paths prior to building these changes or it can be fixed up after the fact.

Per an offline conversation, we think the tool will handle removing the obsolete entries.

eng/common/build.ps1
```diff
@@ -61,7 +61,7 @@ try {
         $args += " --path " + ($Paths -join " --path ")
     }
     else {
-        $args += " --path '$Version/*'"
+        $args += " --path '*/$Version/*'"
```
MichaelSimons, line 64: I will be back porting this to https://github.com/dotnet/docker-tools/blob/master/eng/common/build.ps1

MichaelSimons:
This is ready for final review.

mthalman:
Need to update the trigger paths for the nightly build: https://github.com/dotnet/dotnet-docker/blob/nightly/eng/pipelines/dotnet-core-nightly.yml#L6

MichaelSimons:
> Need to update the trigger paths for the nightly build: https://github.com/dotnet/dotnet-docker/blob/nightly/eng/pipelines/dotnet-core-nightly.yml#L6

Thanks for the catch - fixed.

PR #2009, opened 2020-06-16
Add src directory for all product Dockerfiles

MichaelSimons:
Related to #1972


eng/pipelines/dotnet-core-nightly.yml
```diff
@@ -6,11 +6,8 @@ trigger:
   paths:
     include:
     - manifest.json
-    - aspnet/*
-    - monitor/*
-    - runtime/*
-    - runtime-deps/*
-    - sdk/*
+    - src/*
```
mthalman, line 9: ❤

Issue #1819, opened 2020-04-08
PowerShell telemetry variable not set correctly for Alpine 3.11


mthalman:
The Alpine 3.11 SDK Dockers set the `POWERSHELL_DISTRIBUTION_CHANNEL` to `PSDocker-DotnetCoreSDK-Alpine-3.10` when they should set it to `PSDocker-DotnetCoreSDK-Alpine-3.11`.

https://github.com/dotnet/dotnet-docker/blob/efb5ca7568387a2ab5cb8cdb96d0b8456ccd7448/3.1/sdk/alpine3.11/amd64/Dockerfile#L16

https://github.com/dotnet/dotnet-docker/blob/efb5ca7568387a2ab5cb8cdb96d0b8456ccd7448/5.0/sdk/alpine3.11/amd64/Dockerfile#L16

PR #2014, opened 2020-06-18
Fix incorrect POWERSHELL_DISTRIBUTION_CHANNEL value

MichaelSimons:
Another occurrence of https://github.com/dotnet/dotnet-docker/issues/1819


MichaelSimons:
Merging because the validation leg for the changed Dockerfile is green.

PR #2019, opened 2020-06-19
Modify update-dependencies to handle dotnet-monitor.

jander-msft:
Modify the update-dependencies tool to be able to check for the latest dotnet-monitor 5.0 tool and update its version and checksum in the Dockerfiles as well as update the manifest with corresponding version tag. I will update the update-dependencies pipeline to provide the dotnet-monitor channel, which is currently "net5/dev", after merging in these changes.

### DIFF
--- a/eng/update-dependencies/Program.cs
+++ b/eng/update-dependencies/Program.cs
@@ -20,6 +20,7 @@ namespace Dotnet.Docker
     public static class Program
     {
         private const string AspNetCoreBuildInfoName = "aspnet";
+        private const string MonitorBuildInfoName = "monitor";
         private const string RuntimeBuildInfoName = "core-setup";
         private const string SdkBuildInfoName = "cli";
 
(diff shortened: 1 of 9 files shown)


jander-msft:
CC @wiktork @hoyosjs

eng/update-dependencies/Program.cs
```diff
@@ -71,9 +72,12 @@ public static async Task Main(string[] args)
 
         private static DependencyUpdateResults UpdateFiles(IEnumerable<IDependencyInfo> buildInfos)
         {
+            // NOTE: dotnet-monitor will diverge its versioning in the future; this will need to be refactored in
```
jander-msft, line 75: I can take try to take care of this now or work on it in the next few days instead of adding more debt to the repo.

jander-msft: I'll make this change separately since it looks to be an extensive refactor.

PR #2024, opened 2020-06-22
Create updaters for each product independently.

jander-msft:
While dotnet-monitor does share the same version with the runtime/aspnet/sdk, that will not always be the case. With these changes, I've refactored the update-dependencies tool to update each product independently so that each product may be versioned independent of the other products.

### DIFF
--- a/eng/update-dependencies/DockerfileShaUpdater.cs
+++ b/eng/update-dependencies/DockerfileShaUpdater.cs
@@ -36,12 +36,6 @@ namespace Dotnet.Docker
         private static readonly Regex s_lzmaDownloadUrlRegex = new Regex(s_lzmaUrlPattern);
         private static readonly Regex s_productShaRegex = new Regex(s_productShaPattern);
         private static readonly Regex s_lzmaShaRegex = new Regex(s_lzmaShaPattern);
-        private static readonly Regex s_versionRegex = VariableHelper.GetValueRegex(
-            VariableHelper.AspNetVersionName,
-            VariableHelper.AspNetCoreVersionName,
-            VariableHelper.DotnetSdkVersionName,
-            VariableHelper.DotnetVersionName,
-            VariableHelper.MonitorVersionName);
 
         private static readonly Dictionary<string, string> s_shaCache = new Dictionary<string, string>();
         private static readonly Dictionary<string, Dictionary<string, string>> s_releaseChecksumCache =
@@ -49,11 +43,14 @@ namespace Dotnet.Docker
 
         private Regex _downloadUrlRegex;
         private readonly Options _options;
+        private readonly Regex _versionRegex;
 
-        private DockerfileShaUpdater(string dockerfilePath, Regex regex, Regex downloadUrlRegex, Options options) : base()
+        private DockerfileShaUpdater(string dockerfilePath, Regex regex, Regex downloadUrlRegex, Options options, string[] versionVariableNames) : base()
         {
             _downloadUrlRegex = downloadUrlRegex;
             _options = options;
+            _versionRegex = VariableHelper.GetValueRegex(versionVariableNames);
+
             Path = dockerfilePath;
             Regex = regex;
             VersionGroupName = ValueGroupName;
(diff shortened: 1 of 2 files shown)


mthalman:
What is functionally incorrect with how things work right now?  The tool can already handle the products being independently versioned so it's not clear to me what this is changing.

jander-msft:
> What is functionally incorrect with how things work right now? The tool can already handle the products being independently versioned so it's not clear to me what this is changing.

If you look at what was formerly UpdateFiles:
```
            // NOTE: dotnet-monitor will diverge its versioning in the future; this will need to be refactored in	            List<IDependencyInfo> buildInfos = new List<IDependencyInfo>();
            // some way to allow updaters to be collected for each product's individual version.	            List<IDependencyUpdater> updaters = new List<IDependencyUpdater>();
            string buildVersion = buildInfos.GetBuildVersion(RuntimeBuildInfoName) ??	
                buildInfos.GetBuildVersion(SdkBuildInfoName) ??	
                buildInfos.GetBuildVersion(AspNetCoreBuildInfoName) ??	
                buildInfos.GetBuildVersion(MonitorBuildInfoName);	
            string productVersion = buildVersion.Split('-')[0];	
            string dockerfileVersion = productVersion.Substring(0, productVersion.LastIndexOf('.'));	
            IEnumerable<IDependencyUpdater> updaters = GetUpdaters(dockerfileVersion, buildInfos);
```

It would only take the version of the first product that it could find (runtime, then sdk, then aspnetcore, then dotnet-monitor), create the Dockerfile version, and only search for Dockerfiles that had that version. So when .NET Core goes to 6.0 or if dotnet-monitor moves to 5.1.X, then dotnet-montior would not get updated anymore if any of runtime, sdk, or aspnetcore versions were specified.

jander-msft:
> > What is functionally incorrect with how things work right now? The tool can already handle the products being independently versioned so it's not clear to me what this is changing.
> 
> If you look at what was formerly UpdateFiles:
> 
> ```
>             // NOTE: dotnet-monitor will diverge its versioning in the future; this will need to be refactored in
>             // some way to allow updaters to be collected for each product's individual version.
>             string buildVersion = buildInfos.GetBuildVersion(RuntimeBuildInfoName) ??	
>                 buildInfos.GetBuildVersion(SdkBuildInfoName) ??	
>                 buildInfos.GetBuildVersion(AspNetCoreBuildInfoName) ??	
>                 buildInfos.GetBuildVersion(MonitorBuildInfoName);	
>             string productVersion = buildVersion.Split('-')[0];	
>             string dockerfileVersion = productVersion.Substring(0, productVersion.LastIndexOf('.'));	
>             IEnumerable<IDependencyUpdater> updaters = GetUpdaters(dockerfileVersion, buildInfos);
> ```
> 
> It would only take the version of the first product that it could find (runtime, then sdk, then aspnetcore, then dotnet-monitor), create the Dockerfile version, and only search for Dockerfiles that had that version. So when .NET Core goes to 6.0 or if dotnet-monitor moves to 5.1.X, then dotnet-montior would not get updated anymore if any of runtime, sdk, or aspnetcore versions were specified.

Technically, we could run the update-dependencies tool for each product independently so that only one product is updated for each run of the tool, but that seemed like that was not the intended use based on how the pipeline already invokes the tool with all version information passed at once.

MichaelSimons:
We have had several issues in the past in aspnet/sdk/runtime that blocked uptake of the latest builds until product changes were made.  In these type of situations, those issues shouldn't block uptake of the latest monitor/diagnostics versions.  If we onboard more image variants that are disparate, the pattern of a single PR to update all may not scale well in the aforementioned scenario.  I therefore lean towards having separate PRs.

jander-msft:
> We have had several issues in the past in aspnet/sdk/runtime that blocked uptake of the latest builds until product changes were made. In these type of situations, those issues shouldn't block uptake of the latest monitor/diagnostics versions. If we onboard more image variants that are disparate, the pattern of a single PR to update all may not scale well in the aforementioned scenario. I therefore lean towards having separate PRs.

I've created a separate PR (#2027) that should run the update-dependencies tool just for dotnet-monitor so that it creates a separate update PR. If this is more of what you had in mind, I'll close out the refactoring PR.

MichaelSimons:
I like your new proposal in #2027.

PR #2027, opened 2020-06-22
Create separate update PR for dotnet-monitor

jander-msft:


eng/pipelines/update-dependencies.yml
```diff
@@ -25,6 +25,15 @@ jobs:
...
+      --compute-shas
+    displayName: Run Update Dependencies for SDK
+  - script: >
+      docker run --rm -v /var/run/docker.sock:/var/run/docker.sock update-dependencies
+      --user $(dotnetDockerBot.userName)
+      --email $(dotnetDockerBot.email)
+      --password $(BotAccount-dotnet-docker-bot-PAT)
       --monitor-version $(monitorVer)
+      --version-source-name dotnet/diagnostics
       --compute-shas
```
mthalman, line 38: This option probably shouldn't be included for .NET Monitor.  It will cause the tool to fallback to manually computing the checksum if the checksum file isn't available.  This logic only exists for legacy, 2.1, purposes.  Without providing this option, an error will be produced if the checksum file cannot be found, which is what should happen for .NET monitor.

jander-msft:
I'm not certain that when the update-dependencies tool runs and creates a PR for the core-sdk updates that it (1) locally commits the changes, and (2) reverts the changes after it creates the PR. If the changes are committed and not reverted, then the run of the update-dependencies tool for dotnet-monitor will likely create a PR that contains both the core-sdk updates commit as well as dotnet-monitor updates commit.

So I'm thinking that I can split the UpdateDependencies pipeline job into two, one for core-sdk and one for dotnet-monitor so that there is not possibility for including each other's changes, but keep both jobs in the same yml file. Additionally, I can then split the get-drop-versions.sh script into the original one (pre-dotnet-monitor) and a new one for dotnet-monitor.

@MichaelSimons, @mthalman, what do you think?

mthalman:
> So I'm thinking that I can split the UpdateDependencies pipeline job into two, one for core-sdk and one for dotnet-monitor so that there is not possibility for including each other's changes, but keep both jobs in the same yml file.

Good point. To keep things simple, that seems like the way to go.

Issue #1933, opened 2020-05-18
Templating system for Dockerfiles


MichaelSimons:
The .NET Docker repos prioritize Dockerfile readability.  By this I mean using a single Dockerfile to build arm, amd64, multiple distros, etc image variants is avoided because it would required Dockerfiles ARGs to specify version numbers, artifact shas, base images, etc.  Having individual Dockerfiles per platform makes it is easy for users to copy a Dockerfile and adjust for their own specific purposes.  It also makes it easy for users to understand what is in a particular image.

The drawback to this approach is that maintaining the Dockerfiles is more work.  If a change is needed for a particular image variant, it typically needs to be copied across several Dockerfiles.  Likewise, when support is added for a new distro, several Dockerfiles need to be added.  Ensuring all Dockerfiles are identical/symmetric is left to the human eye and there have been cases in the past where the Dockerfiles have gotten out of sync.  Some of these cases were cosmetic but others have been functional (e.g. [Consistent environment variable names for DOTNET_RUNNING_IN_CONTAINER](https://github.com/dotnet/dotnet-docker/issues/677))

There are upcoming features ([Approach for using Crossgen2](https://github.com/dotnet/dotnet-docker/issues/1791) and [Add Server Core Images](https://github.com/dotnet/dotnet-docker/issues/1852)) where symmetry is critical.  The reason is that we would like the first installer stage to be identical in order to gain build optimizations (e.g. build cache).

The official Docker image repositories make use of a template system (e.g. [golang](https://github.com/search?q=repo%3Adocker-library%2Fgolang+extension%3Atemplate&type=Code&ref=advsearch&l=&l=)).  A similar mechanism could be used for .NET Dockerfiles.

**Work Breakdown**

- [x] Ensure consistency across Dockerfiles per version (e.g. ensure Dockerfiles that will be generated from a single template are the same)
- [x] Add ImageBuilder support for breaking manifest.json into separate files
- [x] Introduce manifest.versions.json
- [x] Add ImageBuilder generateDockerfiles command
- [x] Define templates and manifest metadata
- [x] Refactor UpdateDependencies in response to templates
- [x] Add unit test to ensure Dockerfiles are in sync with templates
- [ ] Add documentation for template syntax and workflow for making Dockerfile changes
- [ ] Notify contributing partners of template changes - .NET diagnostics and PowerShell

PR #2066, opened 2020-07-09
Add manifest.versions.json file to encapsulate version variables

MichaelSimons:
This is being added to support the [Dockerfile templates feature](#1933).  Eventually the versions file will contain the product build versions along with the product checksums.

### DIFF
--- a/eng/update-dependencies/ManifestUpdater.cs
+++ b/eng/update-dependencies/ManifestUpdater.cs
@@ -40,11 +40,11 @@ namespace Dotnet.Docker
 
             string productVersion = version.Split('-')[0];
             string dockerfileVersion = productVersion.Substring(0, productVersion.LastIndexOf('.'));
-            string versionVariableName = $"{imageVariantName}\\|{dockerfileVersion}\\|product-version";
+            string versionVariableName = $"{imageVariantName}|{dockerfileVersion}|product-version";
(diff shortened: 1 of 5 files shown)


MichaelSimons:
Doh - I meant to update update-dependencies.  It is a simple fix.

MichaelSimons:
I have updated update-dependencies appropriately.

eng/update-dependencies/ManifestUpdater.cs
```diff
@@ -40,10 +40,10 @@ public ManifestUpdater(string imageVariantName, string version, string repoRoot)
 
             string productVersion = version.Split('-')[0];
             string dockerfileVersion = productVersion.Substring(0, productVersion.LastIndexOf('.'));
-            string versionVariableName = $"{dockerfileVersion}-{imageVariantName}Version";
+            string versionVariableName = $"{imageVariantName}\\|{dockerfileVersion}\\|product-version";
```
mthalman, line 43: Can you add a comment stating that the use of two backslashes is because a backslash needs to be preserved in the resulting string for use in a regex?  That had me confused at first.

MichaelSimons: I tweaked the pattern to make use of Regex.Escape.  This makes the code more robust and readable.

mthalman: Nice.  Didn't know about that method.

Issue #1892, opened 2020-05-07
Update Images references to 3.1


richlander:
Our docs should always point to supported images.

https://github.com/dotnet/dotnet-docker/blob/master/documentation/scenarios/nuget-credentials.md#using-the-azure-artifact-credential-provider

PR #2073, opened 2020-07-10
Update nuget credentials doc

### LINKED ISSUE #1895: NuGet credential docs should include Windows instructions or use x-plat PS scripts
A user reached out to me that they got confused by the following line @ https://github.com/dotnet/dotnet-docker/blob/master/documentation/scenarios/nuget-credentials.md#using-the-azure-artifact-credential-provider

``` Dockerfile
RUN curl -L https://raw.githubusercontent.com/Microsoft/artifacts-credprovider/master/helpers/installcredprovider.sh  | bash
```

They were using a Windows Server Core image. Can we use [this script](https://github.com/Microsoft/artifacts-credprovider#installation-on-windows)? Does it work cross-platform?

We should also discuss (again) if the NuGet credential provider should be included in the .NET SDK images.

mthalman:
Updates the NuGet credential instructions to have a Windows-specific example for the cred provider since there are differences in how the Dockerfile needs to be defined compared to Linux.  Specifically, a workaround is needed for https://github.com/microsoft/artifacts-credprovider/issues/201.

Fixes #1895
Related to #1892


documentation/scenarios/nuget-credentials.md
```diff
@@ -35,7 +35,7 @@ Here's an example `nuget.config` file containing the credentials:
 In the Dockerfile, the `nuget.config` file is copied via the build context and stored in the file system only for the `build` stage. In the `runtime` stage, only the binaries of the application are copied, not the `nuget.config` file so the credentials are not exposed in the final image.
 
 ```Dockerfile
-FROM mcr.microsoft.com/dotnet/core/sdk:3.0 AS build
+FROM mcr.microsoft.com/dotnet/core/sdk AS build
```
MichaelSimons, line 38: I'm not a fan of this change.  Our recommendation is to not use the latest tags for production scenarios.  We know we have lots of users who copy our samples.  This is a pattern that we would not want copied without changes.  I realize this change is convenient for us in maintaining this document but I think we should be showing what is best for our customers.

mthalman: Fixed

mthalman:
@richlander - I've updated this to include a sample project that can be used to test a feed with.

MichaelSimons (approved):

PR #2083, opened 2020-07-14
Add Dockerfile templates

MichaelSimons:
Related to #1933 

This change introduces the set of templates to generate the complete set of Dockerfiles from.  From now on, the Dockerfiles should not be directly edited, rather the Dockerfile templates should be updated and then the Dockerfiles should be regenerated.  The update-dependency tool is being updated here to reflect this new process.  As part of #1933 a unit test will be added to ensure the Dockerfiles are always in sync with the templates.


MichaelSimons:
This change requires https://github.com/dotnet/docker-tools/pull/581, as well as the subsequent change to pickup the version of ImageBuilder with it before it can be merged.

mthalman:
cc @adityapatwardhan

eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -21,100 +21,112 @@ namespace Dotnet.Docker
...
+                        VersionGroupName = ShaValueGroupName
+                    };
+
+                    string archPattern = updater._arch == string.Empty ? string.Empty : "|" + updater._arch;
+                    string osPattern = updater._os == string.Empty ? string.Empty : "|" + updater._os;
+                    updater.Regex = new Regex(
+                        $"{shaVariablePattern.Replace(".*", Regex.Escape(osPattern + archPattern))} \"(?<{ShaValueGroupName}>.*)\"");
+
+                    return updater;
+                });
```
mthalman, line 90: Add a ToArray to this Enumerable, especially since it's outputting tracing information as part of enumeration.

eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -21,100 +21,112 @@ namespace Dotnet.Docker
...
-            Trace.TraceInformation($"DockerfileShaUpdater is processing '{Path}'.");
-            string dockerfile = File.ReadAllText(Path);
-            return GetArtifactShaAsync(dockerfile).Result;
+            usedBuildInfos = new IDependencyInfo[] { productInfo };
+
+            string downloadUrl = urls[_productName]
+                .Replace("$VERSION", _buildVersion)
+                .Replace("$OS", _os)
+                .Replace("$ARCH", _arch)
+                .Replace("$COMPRESS_EXT", _os.Contains("win") ? "zip" : "tar.gz")
```
mthalman, line 104: I feel like `ARCHIVE_EXT` would be a better name.

eng/update-dependencies/Options.cs
```diff
@@ -1,60 +1,42 @@
...
+using System;
+using System.Collections.Generic;
 using System.CommandLine;
+using System.Linq;
 
 namespace Dotnet.Docker
 {
     public class Options
     {
-        public string AspnetVersion { get; private set; }
```
mthalman, line 10: Since the options for this tool have changed, can you also update the issue templates in https://github.com/dotnet/dotnet-docker/tree/master/.github/ISSUE_TEMPLATE/releases accordingly?

eng/update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: Good suggestion - changed.

eng/update-dependencies/Options.cs
MichaelSimons: Done

eng/update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: fixed

PR #2097, opened 2020-07-17
remove announcement and add license notice

mbrown555:
we want a standard link to license info in all readmes


mbrown555:
@mthalman @MichaelSimons please review this one as well, I don't seem to have the option to add reviewers.

PR #2114, opened 2020-07-21
Create tests that verifies the Dockerfiles and templates are in sync

MichaelSimons:
related #1933

These changes introduce a pre-build validation step in the builds.  This works nicely for validating the templates and Dockerfiles are in-sync.  It will also provide a nice place to validate the readme tag listing is up-to-date as well.  I will back port the common eng infrastructure changes to docker-tools once merged.

### DIFF
--- a/eng/dockerfile-templates/Get-GeneratedDockerfiles.ps1
+++ b/eng/dockerfile-templates/Get-GeneratedDockerfiles.ps1
@@ -1,20 +1,22 @@
 #!/usr/bin/env pwsh
 param(
-    # Paths to the Dockerfiles to generate.
-    [string[]]$Paths
+    [switch]$Validate
 )
 
-if ($Paths) {
-    $pathArgs = " --path " + ($Paths -join " --path ")
+if ($Validate) {
+    $customImageBuilderArgs = " --validate"
 }
 
 $repoRoot = (Get-Item "$PSScriptRoot").Parent.Parent.FullName
 
 $onDockerfilesGenerated = {
     param($ContainerName)
-    Exec "docker cp ${ContainerName}:/repo/src $repoRoot"
+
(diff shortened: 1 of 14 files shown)


eng/dockerfile-templates/Get-GeneratedDockerfiles.ps1
```diff
@@ -1,20 +1,22 @@
...
-    Exec "docker cp ${ContainerName}:/repo/src $repoRoot"
+
+    if (-Not $Validate) {
+        Exec "docker cp ${ContainerName}:/repo/src $repoRoot"
+    }
 }
 
 & $PSScriptRoot/../common/Invoke-ImageBuilder.ps1 `
-    -ImageBuilderArgs "generateDockerfiles --architecture * --os-type *$pathArgs" `
+    -ImageBuilderArgs "generateDockerfiles --architecture '*' --os-type '*'$customImageBuilderArgs" `
```
MichaelSimons, line 21: The quoted asterisks are required when running in a linux environment.

tests/Microsoft.DotNet.Docker.Tests/Microsoft.DotNet.Docker.Tests.csproj
```diff
@@ -18,13 +18,6 @@
     <Content Include="TestAppArtifacts/*">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="../../manifest.json">
```
MichaelSimons, line 21: Continuing this pattern didn't seem like a good choice as we would need to copy pretty much the entire repo as content in order to validate the Dockerfile templates.  Instead an ENV is now used to specify the source repo directory.

Issue #514, opened 2018-04-30
Create a developer-guide.md to provide assistance to those wanting to contribute.


MichaelSimons:
There should be a developer-guide.me that describe the following:

1. Build images locally
1. Run tests locally
1. Modify existing images
1. Create new images
1. Add/Modify/Obsolete image tags

Our repo should be encouraging external contributions.

See https://github.com/dotnet/dotnet-docker/pull/506#issuecomment-384729639 for a concrete case on why this is needed.

jsacapdev:
@MichaelSimons can i help with this? even if it means running through and validating the PR from an external contributor? If you think I can, let me know how you think I might be able to help.

MichaelSimons:
@jsacapdev - A major component of this documentation includes describing the [manifest.json](https://github.com/dotnet/dotnet-docker/blob/master/manifest.json) - what it contains and how it is used.  Other important aspects are to describe the build infra as well as the tests.  I would welcome your help on validating the PR from an external contributor standpoint.  I will keep you posted.  Thanks

Issue #2015, opened 2020-06-18
Alpine 3.12 SDK image contains old PS version


MichaelSimons:
Alpine 3.11 contains 7.0.1 - https://github.com/dotnet/dotnet-docker/blob/nightly/src/sdk/3.1/alpine3.11/amd64/Dockerfile#L33

Alpine 3.12 contains 7.0.1 - https://github.com/dotnet/dotnet-docker/blob/nightly/src/sdk/3.1/alpine3.12/amd64/Dockerfile#L33

MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/pull/2001

MichaelSimons:
This issue did not get merged to master during the July servicing.  Re-opening and setting to Rollout.

cc @mthalman

MichaelSimons:
When the fix is merged into master, the dockerfileTemplate should be defined in the manifest.

PR #2123, opened 2020-07-23
Merging Dockerfile templates feature work from nightly to master

### LINKED ISSUE #1933: Templating system for Dockerfiles

MichaelSimons:
Fixes #1933 and #514


MichaelSimons:
Because of #2015, the template for the 3.1 sdk alpine 3.12 dockerfileTemplate is not specified in the manifest.  When the fix for #2015 is merged into master, the template should be defined in the manifest.

PR #2124, opened 2020-07-23
Remove 'core' terminology from 5.0 Dockerfiles

### LINKED ISSUE #2084: 5.0 Dockerfiles still use "core" in the default repo name
As part of the changes for https://github.com/dotnet/dotnet-docker/issues/1765, the repo name for the 5.0 images dropped "core" from the name.  But the 5.0 Dockerfiles also have a `REPO` ARG that defines the default repo name to use.  The build infrastructure overrides this based on what's defined in the manifest so the official images end up being tagged correctly.  But if you were to build one of those Dockerfiles without overriding `REPO`, the wrong repo would be used, resulting in an out-of-date base image since that repo is no longer being updated.

Example:
https://github.com/dotnet/dotnet-docker/blob/5731be852a857b3b49acacb75d881f6d516c562c/src/aspnet/5.0/alpine3.12/amd64/Dockerfile#L1

MichaelSimons:
Fixes #2084 

Remove 'core' terminology from:

- REPO ARG
- Comments
- POWERSHELL_DISTRIBUTION_CHANNEL ENV 

The 'core' terminology was left for places referencing ASP.NET Core.

### DIFF
--- a/eng/dockerfile-templates/sdk/5.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/sdk/5.0/Dockerfile.alpine
@@ -1,4 +1,4 @@
-ARG REPO=mcr.microsoft.com/dotnet/core/aspnet
+ARG REPO=mcr.microsoft.com/dotnet/aspnet
 FROM $REPO:5.0-{{OS_VERSION}}
 
 ENV \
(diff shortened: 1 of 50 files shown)


eng/dockerfile-templates/sdk/5.0/Dockerfile.alpine
```diff
@@ -12,14 +12,14 @@ ENV \
     # Skip extraction of XML docs - generally not useful within an image/container - helps performance
     NUGET_XMLDOC_MODE=skip \
     # PowerShell telemetry for docker image usage
-    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetCoreSDK-{{OS_ARCH_HYPHENATED}}
+    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetSDK-{{OS_ARCH_HYPHENATED}}
```
mthalman, line 15: This change will need to be coordinated with PowerShell.

MichaelSimons: @adityapatwardhan - Do you have any objection to this change?  For .NET 5.0 onward, the 'core' terminology is being dropped.

adityapatwardhan: I will add our PM here who works on telemetry dashboards which use this value.

SydneyhSmith: Thanks @adityapatwardhan this looks good to me, I'll make note of this change for our dashboards

PR #2128, opened 2020-07-24
Enable optional jobs for update-dependencies pipeline

### LINKED ISSUE #2125: update-dependencies pipeline needs to allow control over which jobs execute
There needs to be independent control over which of the update-dependencies jobs run.  This allows for scenarios where we need to independently disable updates from dotnet/installer or dotnet/diagnostics.

mthalman:
There's a need to be able to optionally define which of the products gets updated by the update-dependencies pipeline.  These changes define two queue-time variables for controlling this behavior: `update-dotnet-enabled` and `update-diagnostics-enabled`. It turns out that the easiest way to consume these was through the use of stages since a stage can have a condition.  Using `${{ if }}` expressions isn't possible because they can't reference custom variables.  I also looked at templatizing the UpdateDependencies jobs but it turned out there's just enough variation between them that it made it pointless.

Fixes #2125

### DIFF
--- a/eng/pipelines/update-dependencies.yml
+++ b/eng/pipelines/update-dependencies.yml
@@ -9,41 +9,52 @@ schedules:
   always: true
 variables:
 - template: variables/common.yml
-jobs:
-- job: UpdateDependenciesCoreSdk
-  displayName: Update Dependencies (dotnet/core-sdk)
-  pool: Hosted Ubuntu 1604
-  steps:
-  - script: $(engPath)/get-drop-versions.sh $(channel)
-    displayName: Get Versions
-  - script: docker build -t update-dependencies -f $(engPath)/update-dependencies/Dockerfile --pull .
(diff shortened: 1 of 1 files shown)


eng/pipelines/update-dependencies.yml
```diff
@@ -9,41 +9,52 @@ schedules:
...
+        --compute-shas
+      displayName: Run Update Dependencies
+- stage: Diagnostics
+  condition: eq(variables['update-diagnostics-enabled'], 'true')
+  dependsOn: [] # Allows it to run in parallel
+  jobs:
+  - job: UpdateDependencies
+    displayName: Update Dependencies (dotnet/diagnostics)
+    variables:
+      isEnabled: $(update-diagnostics-enabled)
```
MichaelSimons, line 44: Is this necessary?

mthalman: Nope, removed

PR #2222, opened 2020-09-08
Update release templates to handle workflow for image caching

mthalman:


mthalman:
> Should the official builds be updating the internal repos instead? Would those changes automatically get mirrored to the public repos? I ask this because it seems like there are other scenarios where we could get cache misses if the mirror is slow. The most likely is when a commit is made in nightly when a build it already happening. This commit will trigger a build immediately after the one in progress completes.

I'm thinking maybe we should be committing to the internal repo.  I've logged https://github.com/dotnet/docker-tools/issues/642 for this.  Once we triage that, I'll address this PR appropriately.

mthalman:
Closing this since we'll be addressing the issue via infrastructure.

Issue #1852, opened 2020-04-20
Request for feedback: Should we produce .NET Core + Server Core container images?


richlander:
# Request for feedback: Should we produce .NET Core + Server Core container images?

We are *exploring* the possibility of producing .NET Core + Server Core container images. This is due to hearing feedback from .NET Core users that the Nano Server lifecycle relative to a Windows Server host doesn't work well. See https://github.com/dotnet/dotnet-docker/issues/1842 for a recent report on and explanation of this topic. We have also seen users needing to use Server Core images because Nano Server is not compatible with their workload.

Some context: We have long believed (on the .NET team) that customer demand for the combination of .NET Core and Server Core would be low and that producing .NET Core images for both Server Core and Nano Server would be confusing and send a mixed message. In short, we thought that nearly all .NET Core users would want Nano Server images. With that in mind, we decided to let customers produce Server Core container images themselves (but, of course, still support .NET Core with Server Core, generally). We are re-assessing this viewpoint and stance, but need your feedback to do so.

It would be very useful to know how many people/companies would use .NET Core + Server Core images if we made them available. Please tell us if you'd use these images, if you would use them instead of the .NET Core + Nano Server images, and what type of app you would host, preferably in this issue. If you are willing (or would prefer a private conversation), we will want to talk to you directly to learn more about your scenarios. You can reach out to me at rlander@ms.

## Details

We have discussed a few different variants of what .NET Core + Server Core images could look like. There are two (meta) use cases we are considering, in terms of the types of apps that you host across your containerized .NET app portfolio:

1. All apps are .NET Core.
2. Mixture of .NET Core and .NET Framework apps, sometimes hosted in the same image.

Some context: In order to get the best performance for .NET Framework apps, we need to NGEN the framework. In fact, without NGEN, the performance is not acceptable. See https://github.com/microsoft/dotnet-framework-docker/issues/29 for past experience on this topic. NGEN is even more important starting with the Server Core 2004 container image release. See: [We made Windows Server Core container images 40% smaller](https://devblogs.microsoft.com/dotnet/we-made-windows-server-core-container-images-40-smaller/). 

[...]

PR #2274, opened 2020-09-28
Windows Server Core 2019 Dockerfiles for 5.0

mthalman:
New Dockerfile templates are defined for Windows Server Core since the Nano Server templates are multi-stage while the Server Core templates can be single stage.

Related to #1852

### DIFF
--- a/eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
+++ b/eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
@@ -3,8 +3,7 @@
 ARG REPO=mcr.microsoft.com/dotnet/runtime
 FROM $REPO:5.0-{{OS_VERSION}}
 
-ARG ASPNET_VERSION={{VARIABLES["aspnet|5.0|build-version"]}}
-ENV ASPNET_VERSION $ASPNET_VERSION
+ENV ASPNET_VERSION {{VARIABLES["aspnet|5.0|build-version"]}}
 
 RUN powershell -Command `
         $ErrorActionPreference = 'Stop'; `
--- a/eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
+++ b/eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
@@ -1,0 +1,22 @@
+# escape=`
+
+ARG REPO=mcr.microsoft.com/dotnet/runtime
+FROM $REPO:5.0-{{OS_VERSION}}
+
+ARG ASPNET_VERSION={{VARIABLES["aspnet|5.0|build-version"]}}
+ENV ASPNET_VERSION $ASPNET_VERSION
+
+RUN powershell -Command `
+        $ErrorActionPreference = 'Stop'; `
+        $ProgressPreference = 'SilentlyContinue'; `
+        `
+        # Install ASP.NET Core Runtime
+        Invoke-WebRequest -OutFile aspnetcore.zip https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/aspnetcore-runtime-$Env:ASPNET_VERSION-win-x64.zip; `
+        $aspnetcore_sha512 = '{{VARIABLES["aspnet|5.0|win|x64|sha"]}}'; `
+        if ((Get-FileHash aspnetcore.zip -Algorithm sha512).Hash -ne $aspnetcore_sha512) { `
+            Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
+            exit 1; `
+        }; `
+        `
+        tar -C $Env:ProgramFiles\dotnet -oxzf aspnetcore.zip ./shared/Microsoft.AspNetCore.App; `
+        Remove-Item -Force aspnetcore.zip
--- a/eng/dockerfile-templates/sdk/5.0/Dockerfile.windowsservercore
+++ b/eng/dockerfile-templates/sdk/5.0/Dockerfile.windowsservercore
@@ -1,0 +1,53 @@
+# escape=`
+
+ARG REPO=mcr.microsoft.com/dotnet/aspnet
+FROM $REPO:5.0-{{OS_VERSION}}
+
+ARG DOTNET_SDK_VERSION={{VARIABLES["sdk|5.0|build-version"]}}
+
+ENV `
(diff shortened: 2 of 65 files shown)


eng/dockerfile-templates/sdk/5.0/Dockerfile.windowsservercore
```diff
@@ -0,0 +1,53 @@
...
+    ASPNETCORE_URLS= `
+    DOTNET_SDK_VERSION=$DOTNET_SDK_VERSION `
+    # Enable correct mode for dotnet watch (only mode supported in a container)
+    DOTNET_USE_POLLING_FILE_WATCHER=true `
+    # Skip extraction of XML docs - generally not useful within an image/container - helps performance
+    NUGET_XMLDOC_MODE=skip `
+    # PowerShell telemetry for docker image usage
+    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetSDK-{{OS_ARCH_HYPHENATED}}
+
+RUN `
```
MichaelSimons, line 19: What is the reason for the `RUN` instruction being on a separate line by itself?

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -34,7 +34,7 @@ public static IEnumerable<object[]> GetImageData()
                 .Select(imageData => new object[] { imageData });
         }
 
-        [Theory]
+        [SkippableTheory("windowsservercore-ltsc2019")]
```
MichaelSimons, line 37: Am I correct in my understanding that this is temporary?  This is needed because no `s_windowsSampleTestData` entries exists for Server Core scenarios.  Because of that this would fail in those legs.  Once 5.0 releases, we will want to add Server Core samples?

eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
```diff
@@ -0,0 +1,22 @@
...
+        `
+        # Install ASP.NET Core Runtime
+        Invoke-WebRequest -OutFile aspnetcore.zip https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/aspnetcore-runtime-$Env:ASPNET_VERSION-win-x64.zip; `
+        $aspnetcore_sha512 = '{{VARIABLES["aspnet|5.0|win|x64|sha"]}}'; `
+        if ((Get-FileHash aspnetcore.zip -Algorithm sha512).Hash -ne $aspnetcore_sha512) { `
+            Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
+            exit 1; `
+        }; `
+        `
+        tar -C $Env:ProgramFiles\dotnet -oxzf aspnetcore.zip ./shared/Microsoft.AspNetCore.App; `
```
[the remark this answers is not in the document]
mthalman: Yes, PowerShell doesn't have a way to extract specific files/dirs.

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
mthalman: Yes, you're reasoning is correct.  I was originally thinking that we wouldn't have Server Core samples because we'd want the multi-arch tag of samples to be consistent with the product repos in resolving to Nano Server.  But that doesn't mean we can't have non-published samples for Server Core.  I'll add those and then I can get rid of this `SkippableTheory`.

mthalman: Ah yes, I guess I can't add them until we actually have some published images that they can reference.  Ok then.

eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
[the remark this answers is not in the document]
MichaelSimons: Should we using `tar` in all Windows Dockerfiles?  The benefit is consistency over the using all PS constructs.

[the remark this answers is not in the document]
mthalman: Yes, I've updated the other templates to do this.

eng/dockerfile-templates/sdk/5.0/Dockerfile.windowsservercore
```diff
@@ -0,0 +1,53 @@
...
+            Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
+            exit 1; `
+        }; `
+        & $Env:ProgramFiles\dotnet\dotnet tool install --add-source . --tool-path $Env:ProgramFiles\powershell --version $powershell_version PowerShell.Windows.x64; `
+        & $Env:ProgramFiles\dotnet\dotnet nuget locals all --clear; `
+        Remove-Item -Force PowerShell.Windows.x64.$powershell_version.nupkg; `
+        Remove-Item -Path $Env:ProgramFiles\powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force;"
+
+# In order to set system PATH, ContainerAdministrator must be used
+USER ContainerAdministrator
```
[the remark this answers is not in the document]
mthalman: Fixed

eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
```diff
@@ -0,0 +1,22 @@
+# escape=`
+
+ARG REPO=mcr.microsoft.com/dotnet/runtime
+FROM $REPO:5.0-{{OS_VERSION}}
+
+ARG ASPNET_VERSION={{VARIABLES["aspnet|5.0|build-version"]}}
```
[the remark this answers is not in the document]
mthalman: Fixed

eng/dockerfile-templates/sdk/5.0/Dockerfile.windowsservercore
mthalman: Fixed

MichaelSimons (approved):

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
mthalman: https://github.com/dotnet/dotnet-docker/issues/2275

PR #2277, opened 2020-09-29
Update Server Core Dockerfiles to reference latest 5.0 build

mthalman:
An update was made in https://github.com/dotnet/dotnet-docker/pull/2267 without updating the new Server Core Dockerfiles.


MichaelSimons (approved):
Am I correct in my understanding that these weren't updated because the PR to add them was merged after the update-dependencies tool created #2267?  In the future they will get updated automatically?

mthalman:
> Am I correct in my understanding that these weren't updated because the PR to add them was merged after the update-dependencies tool created #2267? In the future they will get updated automatically?

Yes, this was simply because the tool hadn't been run after the Server Core Dockerfiles were added.

Issue #2182, opened 2020-08-19
Consider manifest tag for Alpine


richlander:
We are (hopefully) in the process of creating ARM32 Alpine image. This puts us in a position where we will have support for the complete architecture matrix that we support (x64, ARM64, ARM32). That means that we have the option of creating a manifest tag for Alpine. We have that for Debian but haven't for Alpine.

Today, we have:

- `5.0-alpine`
- `5.0-alpine-arm64v8`

Going forward, we could move to the following (including for Ubuntu and Debian):

- `5.0-alpine` (manifest tag)
- `5.0-alpine-amd64`
- `5.0-alpine-arm64v8`
- `5.0-alpine-arm32v7`

Our non-manifest tags have always offered x64 tags as special (no arch specified). This approach would move to a new model:

- Manifest tags have no arch in them and support all arches.
- Concrete tags are fully specified, including arches.

This would be very nice for using .NET container images on Apple Silicon. It's the first mainline developer scenario for ARM64.

Related: https://github.com/dotnet/runtime/issues/2021

mthalman:
These tags were published as part of the 5.0 RC2 release.

PR #2281, opened 2020-09-30
Ensure all concrete tags have arch in name

mthalman:
The first phase of #2182 is to update all existing concrete 5.0 tags so that they explicitly include an architecture suffix.


mthalman:
Closing to target feature branch.

PR #2282, opened 2020-09-30
Ensure all concrete tags have arch in name

mthalman:
The first phase of #2182 is to update all existing concrete 5.0 tags so that they explicitly include an architecture suffix.


MichaelSimons (approved):
Can you create an issue that tracks doing this for the samples when we upgrade them to 5.0?

mthalman:
> Can you create an issue that tracks doing this for the samples when we upgrade them to 5.0?

https://github.com/dotnet/dotnet-docker/issues/2284

PR #2288, opened 2020-10-02
Define multi-arch tags for each Linux distro

mthalman:
Defines a multi-arch tag for each Linux distro for 5.0.  Because of limitations in the manifest schema, this required duplicating the platforms in order to assign them a distinct set of shared tags.

Related to #2182

### DIFF
--- a/README.aspnet.preview.md
+++ b/README.aspnet.preview.md
@@ -63,7 +63,7 @@ See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotn
 Tags | Dockerfile | OS Version
 -----------| -------------| -------------
 5.0.0-rc.2-buster-slim-amd64, 5.0-buster-slim-amd64, 5.0.0-rc.2-buster-slim, 5.0-buster-slim, 5.0.0-rc.2, 5.0, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/5.0/buster-slim/amd64/Dockerfile) | Debian 10
-5.0.0-rc.2-alpine3.12-amd64, 5.0-alpine3.12-amd64, 5.0-alpine-amd64, 5.0.0-rc.2-alpine3.12, 5.0-alpine | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/5.0/alpine3.12/amd64/Dockerfile) | Alpine 3.12
+5.0.0-rc.2-alpine3.12-amd64, 5.0-alpine3.12-amd64, 5.0-alpine-amd64, 5.0.0-rc.2-alpine3.12, 5.0-alpine3.12, 5.0-alpine | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/5.0/alpine3.12/amd64/Dockerfile) | Alpine 3.12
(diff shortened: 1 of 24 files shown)


mthalman:
Found an issue in the matrix generation for PR builds that needs to be addressed.

manifest.json
```diff
@@ -1774,7 +1878,135 @@
...
+              "tags": {},
+              "variant": "v8"
+            }
+          ]
+        },
+        {
+          "id": "nanoserver-1809",
           "productVersion": "$(dotnet|5.0|product-version)",
+          "sharedTags": {
+            "$(dotnet|5.0|product-version)-nanoserver-1809": {},
```
MichaelSimons, line 1928: Given we only support amd64 for Windows, I wonder if these should be define as concrete tags instead of manifest tags?  Are we adding _overhead_ for no value?

README.aspnet.preview.md
```diff
@@ -62,49 +62,49 @@ See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotn
 ##### .NET 5.0 Preview Tags
 Tags | Dockerfile | OS Version
 -----------| -------------| -------------
-5.0.0-rc.2-buster-slim-amd64, 5.0-buster-slim-amd64, 5.0.0-rc.2, 5.0, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/5.0/buster-slim/amd64/Dockerfile) | Debian 10
-5.0.0-rc.2-alpine3.12-amd64, 5.0-alpine3.12-amd64, 5.0-alpine-amd64 | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/5.0/alpine3.12/amd64/Dockerfile) | Alpine 3.12
-5.0.0-rc.2-focal-amd64, 5.0-focal-amd64 | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/5.0/focal/amd64/Dockerfile) | Ubuntu 20.04
+5.0.0-rc.2-buster-slim-amd64, 5.0-buster-slim-amd64, 5.0.0-rc.2-buster-slim, 5.0-buster-slim, 5.0.0-rc.2, 5.0, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/5.0/buster-slim/amd64/Dockerfile) | Debian 10
+5.0.0-rc.2-alpine3.12-amd64, 5.0-alpine3.12-amd64, 5.0-alpine-amd64, 5.0.0-rc.2-alpine3.12, 5.0-alpine | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/5.0/alpine3.12/amd64/Dockerfile) | Alpine 3.12
```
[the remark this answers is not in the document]
mthalman: Added

manifest.json
mthalman: Based on discussion, I've removed these tags.  Their only purpose was to have consistency with the Linux tagging pattern.  Functionally, they offered no value since Windows has only one architecture that we support.  So the value in having them just isn't there since it just adds unnecessary tags for people to have to sort through.  We can always add them if we get feedback that having consistency is desired.

MichaelSimons: I think we want to remove the arch tagsfor Windows and leave the simpler, non-arch tags.

mthalman: > I think we want to remove the arch tagsfor Windows and leave the simpler, non-arch tags.

I'd much prefer having consistency in the tagging scheme when it comes essentially for free, without any extra complexity.  That is, all concrete tags are arch-specific, period.  Any arch-less tags are shared tags.  What would be the benefit of not including the arch in the concrete tags?

MichaelSimons (approved):

PR #2296, opened 2020-10-06
Pipeline to validate PR without caching

### LINKED ISSUE #2247: PR validation with caching disabled
By default, the PR builds have image caching enabled.  This is useful because it avoids build and testing images that haven't changed.  But this is problematic when we want to use the PR build to validate infrastructure changes.  Not building any images or only building a subset may mean that code paths in the infrastructure do not get exercised and leads to a lack of test coverage for the change.

We should consider having a special pipeline defined for PR builds that executes with caching disabled in order to test infrastructure changes.  The trigger paths should be set to appropriate values in order to limit when this pipeline runs.  It's also not necessary to build the entire portfolio of images but rather just a representative subset that best exercises code paths that should be tested.

mthalman:
Defines a new pipeline YAML file that is intended to be used to validate PRs without caching enabled.  The pipeline should only build/test a subset of the images.  Rather than defining that subset within the YAML, I felt it more appropriate to define it in the variables of the build definition in AzDO where it can be more easily adjusted as necessary.  See https://github.com/dotnet/dotnet-docker/issues/2247#issuecomment-704325852 for the suggested set of paths to use.

This pipeline is configured to be triggered only by changes to the eng and tests folders.  A change to the src folder would likely end up causing a cache miss in the existing PR pipeline so this one wouldn't need to run in that case.  The primary use case for this pipeline is to validate infrastructure changes in the eng folder.

Fixes #2247

### DIFF
--- a/eng/pipelines/dotnet-core-nightly-pr-no-cache.yml
+++ b/eng/pipelines/dotnet-core-nightly-pr-no-cache.yml
@@ -23,3 +23,4 @@ stages:
   parameters:
     buildMatrixType: platformVersionedOs
     buildMatrixCustomBuildLegGroupArgs: --custom-build-leg-group pr-build --custom-build-leg-group test-dependencies
+    noCache: true
--- a/eng/pipelines/dotnet-core-nightly-pr-no-cache.yml
+++ b/eng/pipelines/dotnet-core-nightly-pr-no-cache.yml
@@ -1,0 +1,25 @@
+trigger: none
+pr:
+  branches:
+    include:
+    - nightly
+  paths:
+    include:
+    - eng/*
+    - tests/*
+
+resources:
+  repositories:
+  - repository: VersionsRepo
+    type: github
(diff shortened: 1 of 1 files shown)


eng/pipelines/dotnet-core-nightly-pr-no-cache.yml
```diff
@@ -0,0 +1,25 @@
...
+
+resources:
+  repositories:
+  - repository: VersionsRepo
+    type: github
+    endpoint: dotnet
+    name: dotnet/versions
+
+variables:
+- template: variables/core.yml
```
MichaelSimons, line 19: Does it make sense for this to set the `NoCache` build option by default?

mthalman: Yes, I meant to do that.  Easy to do with a parameter.

MichaelSimons (approved):

PR #2301, opened 2020-10-08
Define more granular PR pipelines

### LINKED ISSUE #2161: Stop building/testing the full repo during PR validation.
Currently the full set of Dockerfiles are built and tested whenever a src, test, eng, or manifest file is changed in the repo.  Most often this is not necessary and is therefore wasting a fair amount of compute.  This is more prevalent in the _nightly_ branch than the _master_ branch.  For example when an auto-update PR is made to pull in the most recent product versions, the Dockerfiles for the product versions that are in-service are unaffected and do not need to be built/tested.

A simple change could be made to include version triggers and only build/test the Dockerfiles for the affected version.  This approach is similar to what is used in the [dotnet/dotnet-buildtools-prereqs-docker](https://github.com/dotnet/dotnet-buildtools-prereqs-docker) repo today.  The difference is the buildtools repo triggers on the OS variant.

The downside to this approach is a separate build definition must exist for each of the product versions in support.  This adds small amount of maintenance overhead but the savings in compute overshadows this maintenance cost.

mthalman:
Breaks up the single PR pipeline that we had into more granular pipelines that are triggered by changes to specific folders in the repo.  This allows us to only run builds for .NET Core 2.1, 3.1, 5.0, or the monitor images based on which Dockerfiles had changed.

Fixes #2161

### DIFF
--- a/eng/pipelines/dotnet-core-nightly-pr-2.1.yml
+++ b/eng/pipelines/dotnet-core-nightly-pr-2.1.yml
@@ -7,7 +7,6 @@ pr:
   paths:
     include:
     - manifest.json
-    - manifest.versions.json
     - eng/*
     - src/aspnet/2.1/*
     - src/runtime/2.1/*
--- a/eng/pipelines/dotnet-core-nightly-pr-2.1.yml
+++ b/eng/pipelines/dotnet-core-nightly-pr-2.1.yml
@@ -1,0 +1,34 @@
+trigger: none
+pr:
+  branches:
+    include:
+    - nightly
+    - feature/*
+  paths:
+    include:
+    - manifest.json
+    - manifest.versions.json
+    - eng/*
+    - src/aspnet/2.1/*
+    - src/runtime/2.1/*
+    - src/runtime-deps/2.1/*
+    - src/sdk/2.1/*
+    - tests/*
+
+resources:
+  repositories:
+  - repository: VersionsRepo
+    type: github
+    endpoint: dotnet
+    name: dotnet/versions
+
+variables:
+- template: variables/core.yml
+- name: imageBuilder.pathArgs
+  value: --path 'src/*/2.1/*'
+
+stages:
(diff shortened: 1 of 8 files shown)


eng/pipelines/dotnet-core-nightly-pr-2.1.yml
```diff
@@ -0,0 +1,34 @@
+trigger: none
+pr:
+  branches:
+    include:
+    - nightly
+    - feature/*
+  paths:
+    include:
+    - manifest.json
+    - manifest.versions.json
```
MichaelSimons, line 10: Is going to work as desired because of this include.  It looks like any time a version # is changed, it will trigger all PR legs to run.  When the auto update PRs are made to pull in a new product versions, only the affected Dockerfiles should be built.  If we can't solve this use case I question the value of this change.

mthalman: Here are the options I see:

1. Remove the trigger entirely for this file from all the PR pipelines.  It would be very unusual to update that file without a Dockerfile change and technically incorrect since it should fail the pre-build validation.  This would be a compromise to keep things simple since a change to that file without also changing a Dockerfile wouldn't trigger a build that would validate this.
2. Define a pipeline that would be configured to just run pre-build validation and have it trigger from any manifest change, including this file.  Downside is that it adds yet another pipeline.  The other PR pipelines would still continue to run pre-build tests to ensure they don't unnecessarily run a bunch of jobs in case those tests fail.
3. Split out the manifest.versions.json into separate files: 2.1, 3.1, 5.0, monitor that can each be individually triggered upon.

MichaelSimons: My preference is #1.

mthalman: That works for me.  It was close between 2 and 1 for me.

MichaelSimons (approved):

PR #2318, opened 2020-10-19
Update version resolution logic

mthalman:
When nearing release, .NET switches to stable branding for its versioning scheme.  This means that the versions of shared frameworks contained in the SDK zip file are not build-specific.  In the case of the upcoming 5.0 release, the versions of runtime and ASP.NET Core being referenced are just labeled as 5.0.0.  This breaks the existing logic in the `get-drop-versions.sh` script because it relies on this value being build-specific in order to generate the correct input to the update-dependencies tool.

I've updated the logic to account for this scenario by not relying on the .version files for the shared frameworks and instead lookup up the build-specific version numbers associated with the SDK build by using the Version.Details.xml file.

This also required changes to the update-dependencies tool because the URL path to the files is different.  The folder name contains the build version but the filename does not.  The simplest way I could think of this was just to use a heuristic that accounts for this URL format in the case when the version string contains `-rtm.`.

### DIFF
--- a/eng/get-drop-versions.sh
+++ b/eng/get-drop-versions.sh
@@ -12,15 +12,19 @@ sudo apt-get update && \
 
 curl -SLo sdk.zip https://aka.ms/dotnet/$channel/Sdk/dotnet-sdk-win-x64.zip
 
-commitSha=$(unzip -p sdk.zip "sdk/*/.version" | cat | head -1)
-# Remove last character (end-of-line)
-commitSha=${commitSha%?}
-sdkVer=$(unzip -p sdk.zip "sdk/*/.version" | cat | head -2 | tail -1)
+unzip -p sdk.zip "sdk/*/.version" > sdkversion
+
+commitSha=$(cat sdkversion | head -1)
+commitSha=${commitSha%?} # Remove last character (newline)
+
+sdkVer=$(cat sdkversion | head -2 | tail -1)
+
+rm sdkversion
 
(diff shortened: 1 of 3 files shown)


eng/get-drop-versions.sh
```diff
@@ -7,13 +7,23 @@ set -u
 
 channel=$1
 
+sudo apt-get update && \
+    sudo apt-get install -y --no-install-recommends libxml2-utils
+
 curl -SLo sdk.zip https://aka.ms/dotnet/$channel/Sdk/dotnet-sdk-win-x64.zip
 
+commitSha=$(unzip -p sdk.zip "sdk/*/.version" | cat | head -1)
```
MichaelSimons, line 15: It's a little inefficient that the script unpacks the sdk `.version` file multiple times.

eng/pipelines/update-dependencies.yml
```diff
@@ -25,9 +25,6 @@ stages:
     - script: >
         docker run --rm -v /var/run/docker.sock:/var/run/docker.sock update-dependencies
         $(dockerfileVersion)
-        --user $(dotnetDockerBot.userName)
```
MichaelSimons, line 28: I'm not understanding why these are being removed.  Can you explain this change?

mthalman: That was for testing purposes.  Added this back.

eng/get-drop-versions.sh
mthalman: Fixed

MichaelSimons (approved):

PR #2326, opened 2020-10-21
Update 5.0 to latest RTM build

mthalman:
Because of the stable branding for the version, the NuGet config file needed to be updated to reference the build-specific feed location.


README.aspnet.preview.md
```diff
@@ -62,49 +62,49 @@ See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotn
 ##### .NET 5.0 Preview Tags
```
MichaelSimons, line 62: In preparation for RTM, I propose we remove the preview sections in the readmes and get the 5.0 tags in the proper location.  It's not necessary with this PR but should get tracked with an issue.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config.nightly
```diff
@@ -4,7 +4,7 @@
     <clear />
     <add key="dotnet2.1" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet2.1/nuget/v3/index.json" />
     <add key="dotnet3.1" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet3.1/nuget/v3/index.json" />
-    <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
+    <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/5.0.100-rtm.20518.2/nuget/v3/index.json" />
```
MichaelSimons, line 7: Is the plan to manually update this each time?

mthalman: Yes. Given the relatively short timespan that this needs to be maintained and the infrequency of actually needing to have a build-specific feed URL, I think that's a reasonable approach to take.

README.aspnet.preview.md
mthalman: https://github.com/dotnet/dotnet-docker/issues/2327

eng/dockerfile-templates/aspnet/5.0/Dockerfile.alpine
```diff
@@ -3,8 +3,9 @@ FROM $REPO:5.0-{{OS_VERSION}}{{ARCH_TAG_SUFFIX}}
 
 # Install ASP.NET Core
 ENV ASPNET_VERSION {{VARIABLES["aspnet|5.0|build-version"]}}
+ARG PRODUCT_VERSION={{VARIABLES["dotnet|5.0|product-version"]}}
```
MichaelSimons, line 6: Is this a permanent change or just a temporary change until 5.0 releases?

mthalman: This would be temporary until release because 5.0 build versions include that "-rtm.XXXXX.X" suffix.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config.nightly
MichaelSimons: I don't object but if an auto PR is created everyday for a couple weeks, I think it would be worth the effort to update the ImageScenarioVerifier logic to read the version from the versions manifest and replace the value.

eng/dockerfile-templates/aspnet/5.0/Dockerfile.alpine
MichaelSimons: Consider removing the ARG and just hard coding the value in the URL.  Having the two version numbers, ASPNET_VERSION and PRODUCT_VERSION, hurt the readability IMO.  Also please log an issue to restore the original pattern after 5.0 releases.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config.nightly
mthalman: The auto PR stuff will need to be disabled because dotnet/installer doesn't point to the latest RTM build.  This will all need to be done manually as new coherent builds are made available.

MichaelSimons (approved):

eng/dockerfile-templates/aspnet/5.0/Dockerfile.alpine
mthalman: https://github.com/dotnet/dotnet-docker/issues/2331

PR #2328, opened 2020-10-21
Fix path variables in 5.0 PR pipelines

mthalman:
The PR build for 5.0 leads to build failures because it doesn't end up building the required runtime-deps that it shares with 3.1.


MichaelSimons (approved):

mthalman:
Unfortunately, this requires a path filter that includes all of 3.1 and 5.0 for both the 3.1 and 5.0 pipelines.  Because of the dependencies, we can't just build a partial set of the graph, otherwise testing will fail.

mthalman:
Closing this in favor of reverting the splitting of the PR pipelines.

Issue #2317, opened 2020-10-19
Publish all supported images to the non-core branded repositories


MichaelSimons:
For .NET 5.0 a new set of non-core branded repositories were created.  See this [announcement](https://github.com/dotnet/dotnet-docker/issues/1939) for details.  The original plan was to publish all 5.0 and higher images to the new `mcr.microsoft.com/dotnet` repos and continue publishing the 3.1 and 2.1 images to the existing `mcr.microsoft.com/dotnet/core` repos.  

This presents some challenges for our customers in certain scenarios.  If they have automation/tooling/infra around the .NET repositories, they may have to add conditional logic that inspects version numbers to know which .NET repos to target.  If all of the .NET images were published to the new non-core repos, it would eliminate the need for this conditional logic.  Customers could make a complete switch to use the new repos.  This would be similar to what was done during the transition from publishing the .NET images on DockerHub to MCR.

MichaelSimons:
To further improve the UX, we should redirect users from the "core" branded repositories on Docker Hub to the new repositories.  The "core" branded repositories should no longer get indexed and appear in the Docker Hub search results.  This will simplify the number of choices our customers have to pick from when searching for .NET images.

Issue #2307, opened 2020-10-13
`latest` tag is not getting updated in mcr.microsoft.com/dotnet/core-nightly/runtime-deps


MichaelSimons:
The `mcr.microsoft.com/dotnet/core-nightly/runtime-deps:latest` tag has not been updated since 2020-05-07.

``` console
> docker pull mcr.microsoft.com/dotnet/core-nightly/runtime-deps:latest
latest: Pulling from dotnet/core-nightly/runtime-deps
54fec2fa59d0: Pull complete
f685ad152116: Pull complete
Digest: sha256:d2c84b0b046a813a8813bef034e5646ecf1d715514bbf6cad3e815899d45bda7
Status: Downloaded newer image for mcr.microsoft.com/dotnet/core-nightly/runtime-deps:latest
mcr.microsoft.com/dotnet/core-nightly/runtime-deps:latest
> docker inspect mcr.microsoft.com/dotnet/core-nightly/runtime-deps:latest -f "{{ .Created }}"
2020-05-07T00:24:17.813296738Z
```

This is only an issue in the _nightly_ repo

FYI, I discovered this while reviewing https://github.com/dotnet/dotnet-docker/pull/2306.  I diffed `master` and `nightly` to see the differences between the two branches.

mthalman:
Fixed by https://github.com/dotnet/dotnet-docker/pull/2340.

PR #2340, opened 2020-10-26
Move 2.1/3.1 to new repo names

mthalman:
Related to #2317

Moves the 2.1 and 3.1 images from the `mcr.microsoft.com/dotnet/core/*` repos to the `mcr.microsoft.com/dotnet/*` repos.  But it also syndicates those images to the original `mcr.microsoft.com/dotnet/core/*` repos for backwards compatibility.

Since everything is now in one set of repos, the readmes are combined.  However, the readmes aren't fully updated to reflect the repo transition.  This will be covered in https://github.com/dotnet/dotnet-docker/issues/2339.


manifest.json
```diff
@@ -2223,9 +2585,12 @@
         {
           "productVersion": "$(sdk|3.1|product-version)",
           "sharedTags": {
-            "$(sdk|3.1|product-version)": {},
-            "3.1": {},
-            "latest": {}
+            "$(sdk|3.1|product-version)": {
```
MichaelSimons, line 2588: This made me realize there is a problem.  The latest tags in the `core` repos are not going to get updated/syndicated.

MichaelSimons:
It looks like this PR will address #2307 as well.

manifest.json
mthalman: A couple options I can think of:

1. Add special-case logic in Image Builder to handle syndication of the latest tag for 3.1 images.  This wouldn't apply to `runtime-deps` since `latest` is shared by both 3.1 and 5.0.  I should then update `runtime-deps:latest` to also be syndicated in the manifest.

1. Add another field to tags that indicates whether to only publish it to the syndicated repository and not the repository it's contained within.  The downside is that this becomes pervasive throughout much of Image Builder, requiring that this flag be checked when consuming tags.  So `aspnet` could look like this:
```
{
  "productVersion": "$(dotnet|3.1|product-version)",
  "sharedTags": {
    "$(dotnet|3.1|product-version)": {
      "syndicatedRepo": "$(syndicatedAspnetRepo)"
     },
     "3.1": {
       "syndicatedRepo": "$(syndicatedAspnetRepo)"
     },
     "latest": {
       "syndicatedRepo": "$(syndicatedAspnetRepo)",
       "publishToSyndicateOnly": true,
       "docType": "Undocumented"
     }
  }
}
```

I prefer option 1.

MichaelSimons: Can you explain more about what are thinking for #1?  What would you use to trigger the special logic?  Is #2 the same as #1 but just adding the metadata to trigger the logic versus hard coding a condition?

mthalman: For option 1, I was thinking of just hardcoding the relevant publish commands to include `latest` as a syndicate-only tag if it ever encounters syndication for `3.1`.  Like, literally hardcoding "latest" and "3.1" in the Image Builder code.

A third option that avoids hardcoding and avoids the introduction of a tag that needs to be checked through Image Builder code is adding a `syndicatedTags` field:

```
{
  "productVersion": "$(dotnet|3.1|product-version)",
  "sharedTags": {
    "$(dotnet|3.1|product-version)": {
      "syndicatedRepo": "$(syndicatedAspnetRepo)"
     },
     "3.1": {
       "syndicatedRepo": "$(syndicatedAspnetRepo)"
       "syndicatedTags": [
         "3.1",
         "latest"
       ]
     }
  }
}
```

If `syndicatedTags` is not set or empty, it will just syndicate the image with the name of the tag that's marked to be syndicated.  Otherwise, it will syndicate the image with the tags specified in the array.

MichaelSimons: I am opposed to hard coding this logic within Image Builder.  This behavior should be based on declared metadata.  This syndication is going to be around for a long time.  Option #3 seems to give the most flexibility.

mthalman: I've made changes to use the new option for destinationTags from https://github.com/dotnet/docker-tools/pull/683 which solves the issue of keeping the `latest` tag updated in the original repos.

manifest.json
```diff
@@ -127,8 +197,36 @@
...
+            "$(dotnet|5.0|product-version)-buster-slim": {
+              "documentationGroup": "5.0"
+            },
+            "5.0": {
+              "documentationGroup": "5.0"
+            },
+            "5.0-buster-slim": {
+              "documentationGroup": "5.0"
+            },
+            "latest": {
```
MichaelSimons, line 224: I think it would be better to declare the syndication of the latest tag on the 3.1 tag.  I think this expresses the intent more clearly.  When 3.1 reaches EOL, we will want to stop syndication.  Also when latest is moved to 6.0 and it is not a shared image, we don't want to syndicate the 6.0 image.

MichaelSimons (approved):

manifest.json
mthalman: Good suggestion.

MichaelSimons (approved):

PR #1584, opened 2020-01-15
Adopt PowerShell syntax on Windows

richlander:

PR #2341, opened 2020-10-27
ssl dependencies for asp.net

plooploops:
Notes for using aspnetapp sample with ssl dependencies (.pfx with password).

### DIFF
--- a/samples/run-aspnetcore-https-development.md
+++ b/samples/run-aspnetcore-https-development.md
@@ -46,7 +46,11 @@ Use the following instructions, for your operating system configuration. The com
 
 ![Developer Tools -- Delete cookie](https://user-images.githubusercontent.com/2608468/40246148-875fee5a-5a7c-11e8-9728-7da89a491014.png)
 
+Further, if you're loading SSL certificates and trimming assemblies as part of the publish, you'll also need to update the project file for the sample.  Please see details for how you can [support SSL certificates](https://docs.microsoft.com/en-us/dotnet/core/deploying/trim-self-contained#support-for-ssl-certificates).
+
 ### Windows using Linux containers
+
+The following example uses PowerShell.
 
 Navigate to sample:
 
@@ -57,7 +61,7 @@ cd samples\aspnetapp
 Generate cert and configure local machine:
 
 ```console
-dotnet dev-certs https -ep %USERPROFILE%\.aspnet\https\aspnetapp.pfx -p crypticpassword
+dotnet dev-certs https -ep $env:USERPROFILE\.aspnet\https\aspnetapp.pfx -p crypticpassword
 dotnet dev-certs https --trust
 ```
 
@@ -84,7 +88,7 @@ docker build --pull -t aspnetapp .
 Run the container image with ASP.NET Core configured for HTTPS:
 
 ```console
-docker run --rm -it -p 8000:80 -p 8001:443 -e ASPNETCORE_URLS="https://+;http://+" -e ASPNETCORE_HTTPS_PORT=8001 -e ASPNETCORE_ENVIRONMENT=Development -v %APPDATA%\microsoft\UserSecrets\:/root/.microsoft/usersecrets -v %USERPROFILE%\.aspnet\https:/root/.aspnet/https/ aspnetapp
+docker run --rm -it -p 8000:80 -p 8001:443 -e ASPNETCORE_URLS="https://+;http://+" -e ASPNETCORE_HTTPS_PORT=8001 -e ASPNETCORE_ENVIRONMENT=Development -v $env:APPDATA\microsoft\UserSecrets\:/root/.microsoft/usersecrets -v $env:USERPROFILE\.aspnet\https:/root/.aspnet/https/ aspnetapp
 ```
 
 After the application starts, navigate to `http://localhost:8000` in your web browser.
@@ -168,6 +172,8 @@ After the application starts, navigate to `http://localhost:8000` in your web br
 
 ### Windows using Windows containers
 
+The following example uses PowerShell.
+
 Navigate to sample:
 
 ```console
@@ -177,7 +183,7 @@ cd samples\aspnetapp
 Generate cert and configure local machine:
 
 ```console
-dotnet dev-certs https -ep %USERPROFILE%\.aspnet\https\aspnetapp.pfx -p crypticpassword
+dotnet dev-certs https -ep $env:USERPROFILE\.aspnet\https\aspnetapp.pfx -p crypticpassword
 dotnet dev-certs https --trust
 ```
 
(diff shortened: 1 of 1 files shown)


plooploops:
@MichaelSimons @mthalman broken out notes on using SSL dependencies with aspnetapp sample.

mthalman:
After thinking about this, I'm realizing that there's not anything about this problem of SSL dependencies that is specific to Docker containers, right?  You should be able to encounter this same problem when publishing a .NET project outside the context of a Docker build.  In that case, it seems more appropriate that this scenario be described in the [official deployment docs](https://github.com/dotnet/docs/tree/master/docs/core/deploying) instead.

plooploops:
From the context of including the dependencies, that makes sense.  That's broken out to a separate PR: https://github.com/dotnet/docs/pull/21262

If the core docs are updated, perhaps we can add a link inside this one when covering the samples? 

https://github.com/dotnet/dotnet-docker/blob/master/samples/run-aspnetcore-https-development.md#building-and-running-the-sample-with-https

mthalman:
Yes, it seems reasonable to add a link to the official docs as you suggest.

samples/run-aspnetcore-https-development.md
```diff
@@ -46,6 +46,8 @@ Use the following instructions, for your operating system configuration. The com
 
 ![Developer Tools -- Delete cookie](https://user-images.githubusercontent.com/2608468/40246148-875fee5a-5a7c-11e8-9728-7da89a491014.png)
 
+Further, if we're loading SSL certificates and trimming assemblies as part of the build, we'll also need to update the project file for the sample.  Please see details for how we can [support SSL certificates](https://docs.microsoft.com/en-us/dotnet/core/deploying/trim-self-contained#support-for-ssl-certificates).
```
mthalman, line 49:
```suggestion
Further, if we're loading SSL certificates and trimming assemblies as part of the publish, we'll also need to update the project file for the sample.  Please see details for how we can [support SSL certificates](https://docs.microsoft.com/en-us/dotnet/core/deploying/trim-self-contained#support-for-ssl-certificates).
```

samples/run-aspnetcore-https-development.md
```diff
@@ -56,11 +58,21 @@ cd samples\aspnetapp
 
 Generate cert and configure local machine:
 
-```console
+# [Command](#tab/cmd)
```
mthalman, line 61: What is the intent of having a header and a hyperlink for `Command` and why is it referencing `#tab/cmd`?  The link doesn't go anywhere.  Take a look at the preview of the page on GitHub for how this looks because it doesn't look right.

plooploops: The intent was to have multiple tabs to show the difference between using Command Line vs. Powershell. Given that we're looking to have the doc in Github, please see the updated version which is simplified.

samples/run-aspnetcore-https-development.md
plooploops: Thank you for the clarification! Updated.

samples/run-aspnetcore-https-development.md
```diff
@@ -56,11 +58,21 @@ cd samples\aspnetapp
 
 Generate cert and configure local machine:
 
-```console
+In a Command Line console, we can use:
+
+```cmd
```
mthalman, line 63: `cmd` is not a supported language for GitHub Markdown.  See the full list at https://github.com/github/linguist/blob/master/lib/linguist/languages.yml.  You'll want something like `bat`.

samples/run-aspnetcore-https-development.md
```diff
@@ -46,6 +46,8 @@ Use the following instructions, for your operating system configuration. The com
 
 ![Developer Tools -- Delete cookie](https://user-images.githubusercontent.com/2608468/40246148-875fee5a-5a7c-11e8-9728-7da89a491014.png)
 
+Further, if we're loading SSL certificates and trimming assemblies as part of the publish, we'll also need to update the project file for the sample.  Please see details for how we can [support SSL certificates](https://docs.microsoft.com/en-us/dotnet/core/deploying/trim-self-contained#support-for-ssl-certificates).
```
MichaelSimons, line 49: The use of `we`/`we're` feels inconsistent with the existing documentation which uses `you`.

MichaelSimons (commented):
I don't think we should be including both the PS and CMD commands.  We have already made the decision to only include PS snippets (see https://github.com/dotnet/dotnet-docker/pull/1584).  We believe this give the best UX.  Providing both detracts from the readability.  It is left up to the reader to translate if they want to use CMD.  We should be migrating any lingering sample documentation to be based only on PS and call it out as done in the referenced PR.

samples/run-aspnetcore-https-development.md
plooploops: Updated, let me know if there's other notes!

samples/run-aspnetcore-https-development.md
plooploops: Updated per @MichaelSimons note on https://github.com/dotnet/dotnet-docker/pull/1584

samples/run-aspnetcore-https-development.md
```diff
@@ -83,9 +88,12 @@ docker build --pull -t aspnetapp .
 
 Run the container image with ASP.NET Core configured for HTTPS:
 
-```console
-docker run --rm -it -p 8000:80 -p 8001:443 -e ASPNETCORE_URLS="https://+;http://+" -e ASPNETCORE_HTTPS_PORT=8001 -e ASPNETCORE_ENVIRONMENT=Development -v %APPDATA%\microsoft\UserSecrets\:/root/.microsoft/usersecrets -v %USERPROFILE%\.aspnet\https:/root/.aspnet/https/ aspnetapp
+In a Powershell console, you can use: 
+
+```powershell
```
MichaelSimons, line 93: I'm not convinced the use of `powershell` here is appropriate at least in the current form.  It introduces an inconsistency in two forms.  First it is inconsistent with the other samples documentation that show Powershell snippets.  Second the rendering of the various snippets in this section appear very different in the rendered output as shown below.

image

My recommendation is to use `console` and if you feel that using `powershell` would provide value, then log a separate issue for that change.

samples/run-aspnetcore-https-development.md
```diff
@@ -56,10 +58,13 @@ cd samples\aspnetapp
...
 
-```console
-dotnet dev-certs https -ep %USERPROFILE%\.aspnet\https\aspnetapp.pfx -p crypticpassword
+In a Powershell console, we can use:
+
+```powershell
+dotnet dev-certs https -ep $env:USERPROFILE\.aspnet\https\aspnetapp.pfx -p crypticpassword
 dotnet dev-certs https --trust
 ```
+---
```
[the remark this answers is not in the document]
plooploops: Cleaning up, they were part of the tab approach (that will work in md for docs but isn't part of the approach here with GH).

samples/run-aspnetcore-https-development.md
plooploops: Cleaned up for now.  We can log a separate issue for using `powershell`.

samples/run-aspnetcore-https-development.md
```diff
@@ -56,10 +58,13 @@ cd samples\aspnetapp
 
 Generate cert and configure local machine:
 
-```console
-dotnet dev-certs https -ep %USERPROFILE%\.aspnet\https\aspnetapp.pfx -p crypticpassword
+In a Powershell console, we can use:
```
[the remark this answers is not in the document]
plooploops: Cleaned up!

MichaelSimons:
@plooploops, It looks like this PR needs to get rebased?

mthalman (approved):

plooploops:
@MichaelSimons any other updates required here before a merge?  Looks like @mthalman has this approved.

MichaelSimons (approved):

MichaelSimons:
Thanks for fixing this Andy.

PR #2351, opened 2020-10-29
Change permissions on dotnet-monitor

shirhatti:
Currently `dotnet-monitor` has permissions of 744. When attempting run as a non-root UID (as is recommended for containers) dotnet-monitor fails to start with the familiar `standard_init_linux.go:211: exec user process caused "exec format error"`.

Given that there are there are no security implications here, I think this change makes sense.

The only limiting thing should be whether dotnet-monitor has access to the AF_UNIX socket created by the target process' runtime which is not impacted by this change

### DIFF
--- a/src/monitor/5.0/alpine/amd64/Dockerfile
+++ b/src/monitor/5.0/alpine/amd64/Dockerfile
@@ -27,6 +27,6 @@ ENV \
     # Add dotnet-monitor path to front of PATH for easier, prioritized execution
     PATH="/app:${PATH}"
 
-RUN ["chmod", "755", "dotnet-monitor"]
+RUN chmod 755 dotnet-monitor
 
 ENTRYPOINT ["dotnet-monitor", "collect"]
--- a/src/monitor/5.0/alpine/amd64/Dockerfile
+++ b/src/monitor/5.0/alpine/amd64/Dockerfile
@@ -27,4 +27,6 @@ ENV \
     # Add dotnet-monitor path to front of PATH for easier, prioritized execution
     PATH="/app:${PATH}"
 
+RUN ["chmod", "755", "dotnet-monitor"]
+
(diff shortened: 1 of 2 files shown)


shirhatti:
cc @jander-msft

src/monitor/5.0/alpine/amd64/Dockerfile
```diff
@@ -27,4 +27,6 @@ ENV \
     # Add dotnet-monitor path to front of PATH for easier, prioritized execution
     PATH="/app:${PATH}"
 
+RUN ["chmod", "755", "dotnet-monitor"]
```
mthalman, line 30:
```suggestion
RUN chmod 755 dotnet-monitor
```

mthalman:
Ideally, there would be some tests added to https://github.com/dotnet/dotnet-docker/tree/master/tests/Microsoft.DotNet.Docker.Tests which verify dotnet-monitor is executable for a non-root user.  Currently there are no tests for dotnet-monitor.

jander-msft (approved):

src/monitor/5.0/alpine/amd64/Dockerfile
jander-msft: +1, at the very least its more readable

MichaelSimons (commented):
All of the Dockerfiles in this repo are generated from templates.  While this doesn't add value for the monitor image at this time, it will down the road as multiple versions and OSes are supported.  Please update the template at https://github.com/dotnet/dotnet-docker/blob/nightly/eng/dockerfile-templates/monitor/5.0/Dockerfile.alpine.  The unit test that is failing is because the Dockerfiles and templates are not in sync.  See https://github.com/dotnet/dotnet-docker/blob/master/CONTRIBUTING.md#editing-dockerfiles for more details.

mthalman (approved):

PR #2354, opened 2020-10-30
Add required env vars to samples

### LINKED ISSUE #2319: DOTNET_RUNNING_IN_CONTAINER variable isn't set in sample app
The sample Dockerfile for SCD on Windows, https://github.com/dotnet/dotnet-docker/blob/master/samples/aspnetapp/Dockerfile.nanoserver-x64-slim, produces a container that doesn't have the `DOTNET_RUNNING_IN_CONTAINER` environment variable set.  Normally this variable would be provided by the runtime image but since this is a SCD app, its base image is the base Nano Server image.  The web page of the app outputs "Containerized: false" which is incorrect.  We should consider updating the Dockerfile to set that variable.

Other Dockerfile that is also affected:
https://github.com/dotnet/dotnet-docker/blob/master/samples/dotnetapp/Dockerfile.nanoserver-x64-slim

mthalman:
Explicitly adds the `DOTNET_RUNNING_IN_CONTAINER` environment variable to the "slim" variant sample Dockerfiles because it doesn't get the benefit of deriving from the `runtime` image which defines that variable.

Add unit tests to ensure sample Dockerfiles have the required environment variables in the same way as is done for the product.

Fixes #2319

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
@@ -42,6 +42,9 @@ namespace Microsoft.DotNet.Docker.Tests
             {
                 string output = DockerHelper.Run(image, containerName);
                 Assert.StartsWith("Hello from .NET Core!", output);
+
+                ValidateEnvironmentVariables(imageData, image, SampleImageType.Dotnetapp);
+
                 return Task.CompletedTask;
             });
         }
@@ -69,6 +72,8 @@ namespace Microsoft.DotNet.Docker.Tests
                     {
                         await ImageScenarioVerifier.VerifyHttpResponseFromContainerAsync(containerName, DockerHelper, OutputHelper);
                     }
+
+                    ValidateEnvironmentVariables(imageData, image, SampleImageType.Aspnetapp);
                 }
                 finally
                 {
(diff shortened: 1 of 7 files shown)


tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -132,6 +132,45 @@ public void VerifyComplexAppSample()
             }
         }
 
+        [SkippableTheory("windowsservercore-ltsc2019")]
+        [MemberData(nameof(GetImageData))]
+        public async Task VerifyAspnetEnvironmentVariables(SampleImageData imageData)
+        {
+            await VerifySampleAsync(imageData, SampleImageType.Aspnetapp, (image, containerName) =>
+            {
+                List<EnvironmentVariableInfo> variables = new List<EnvironmentVariableInfo>();
```
MichaelSimons, line 141: I'm curious what your reasoning is for creating separate test cases to validate the variables versus adding the validation to the existing methods?  I can see some pros and cons to both approaches but as currently implemented I see two detractors.  

1. Both test cases are building the same Dockerfiles which seems like a waste and adds overhead.
1. The existing test case names now a have some ambiguity on how on what they cover versus these new methods.

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
```diff
@@ -132,6 +132,45 @@ public void VerifyComplexAppSample()
...
+            });
+        }
+
+        [SkippableTheory("windowsservercore-ltsc2019")]
+        [MemberData(nameof(GetImageData))]
+        public async Task VerifyDotnetEnvironmentVariables(SampleImageData imageData)
+        {
+            await VerifySampleAsync(imageData, SampleImageType.Dotnetapp, (image, containerName) =>
+            {
+                List<EnvironmentVariableInfo> variables = new List<EnvironmentVariableInfo>();
```
MichaelSimons, line 161: Consider a sharing the common logic between the two new methods.

tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs
mthalman: Makes sense.  I've removed the separate test cases.

MichaelSimons (approved):

PR #2355, opened 2020-10-30
Define arch-specific tags for samples

### LINKED ISSUE #2284: Define arch-specific tags for samples
When the sample images get upgraded to 5.0, the tags should be updated to use arch-specific suffixes (e.g. `dotnetapp-buster-slim` -> `dotnetapp-buster-slim-amd64`).  This continues on the work of #2182.

mthalman:
Maintains the tagging pattern used in https://github.com/dotnet/dotnet-docker/issues/2182 by ensuring that all concrete Linux tags for the samples are arch-specific.

Fixes #2284

### DIFF
--- a/README.samples.md
+++ b/README.samples.md
@@ -62,8 +62,8 @@ See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotn
 ## Linux amd64 Tags
 Tags | Dockerfile | OS Version
 -----------| -------------| -------------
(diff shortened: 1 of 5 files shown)


MichaelSimons (commented):
Will the test code need to account for this change at all?  It seems like this in particular will be an issue - https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/ImageData.cs#L104

MichaelSimons (approved):

Issue #2348, opened 2020-10-28
Update documentation/snippets to use 5.0 instead of 3.1


mthalman:
There is various documentation that has code samples and snippets that reference 3.1 tags.  As part of the 5.0 release, these references should be updated to 5.0.

Some examples:
* https://github.com/dotnet/dotnet-docker/blob/master/documentation/image-artifact-details.md
* https://github.com/dotnet/dotnet-docker/blob/master/documentation/scenarios/installing-dotnet.md
* https://github.com/dotnet/dotnet-docker/blob/master/documentation/scenarios/nuget-credentials.md

PR #2356, opened 2020-11-02
Update docs to use 5.0 instead of 3.1

mthalman:
Fixes https://github.com/dotnet/dotnet-docker/issues/2348

### DIFF
--- a/documentation/tagging-guidelines.md
+++ b/documentation/tagging-guidelines.md
@@ -16,7 +16,7 @@ The .NET Core image tags strive to align with the tagging practices utilized by 
     * `5.0.0-focal-arm64v8`
     * `5.0.0-focal-amd64`
 
-> NOTE: In versions 2.1 and 3.1 of .NET Core, tags for the AMD64 architecture do not include the "-amd64" suffix (e.g. `2.1.23-alpine3.12`). In .NET 5.0 and higher, it is explicitly included (e.g. `5.0.0-focal-amd64`).
+> NOTE: In versions 2.1 and 3.1 of .NET Core, tags for the AMD64 architecture do not include the "-amd64" suffix (e.g. `2.1.23-alpine3.12`). In .NET 5.0 and higher, it is explicitly included for Linux distros (e.g. `5.0.0-focal-amd64`).
 
 1. `<Major.Minor Version>-<OS>-<Architecture>`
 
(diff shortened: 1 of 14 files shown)


MichaelSimons (commented):
1. Should https://github.com/dotnet/dotnet-docker/blob/nightly/documentation/scenarios/using-system-drawing-common.md get updated as well?  There are some 3.1 references.
1. The architecture section of https://github.com/dotnet/dotnet-docker/blob/nightly/documentation/tagging-guidelines.md looks like it needs updating.  It should call out that in pre-5.0, not architecture includes amd64.

mthalman:
> 1. Should https://github.com/dotnet/dotnet-docker/blob/nightly/documentation/scenarios/using-system-drawing-common.md get updated as well?  There are some 3.1 references.

I explicitly left that as 3.1 because the scenario is intended to get .NET Core on Windows Server Core.  For 5.0, we already provide that image so it's not relevant.

documentation/tagging-guidelines.md
```diff
@@ -10,39 +10,43 @@ The .NET Core image tags strive to align with the tagging practices utilized by
...
     * `2.1-stretch-slim-arm32v7`
-    * `2.2-alpine3.9`
-    * `3.0-nanoserver-1809`
-    * `3.1-disco-arm64v8`
+    * `3.1-alpine3.12`
+    * `3.1-nanoserver-1809`
+    * `5.0-focal-amd64`
+    * `5.0-focal-arm64v8`
+
+> NOTE: In versions 2.1 and 3.1 of .NET Core, tags for the AMD64 architecture do not include the "-amd64" suffix (e.g. `3.1-alpine3.12`). In .NET 5.0 and higher, it is explicitly included (e.g. `5.0-focal-amd64`).
```
MichaelSimons, line 31: I feel like this should also get called out in the `Architecture` line of the `Tag Parts` section

samples/selecting-tags.md
```diff
@@ -70,36 +67,32 @@ The following tags demonstrate the pattern used to describe each processor, usin
...
-* `3.1-buster`
-* `3.1-buster-slim`
-* `3.1-nanoserver-2004`
-* `3.1-nanoserver-1909`
-* `3.1-nanoserver-1903`
-* `3.1-nanoserver-1809`
+* `5.0-alpine`
+* `5.0-bionic`
+* `5.0-buster-slim`
+* `5.0-nanoserver-2004`
```
MichaelSimons, line 73: Shouldn't these include the amd64 suffix?  Should 2009 be included? What about Windows Server Core?

documentation/tagging-guidelines.md
mthalman: Fixed

samples/selecting-tags.md
mthalman: Yes on all of these.

MichaelSimons (approved):

Issue #2396, opened 2020-11-13
[mcr.microsoft.com/dotnet/sdk:5.0] "dotnet watch run" doesn't reload the project when files are changed 


elafarge:
### Describe the Bug

Our local `docker-compose` development stack consists in a set of containers running our ASP.NET Core APIs where we mount our code and run it with `dotnet watch run`. 

We upgraded our services to .NET 5 yesterday, and therefore, our base images to `mcr.microsoft.com/dotnet/sdk:5.0` and noticed a bug: files aren't reloaded when changed on the host.

A similar issue seems to have been reported [here](https://github.com/dotnet/sdk/issues/14081) around a month ago.

### Steps to Reproduce

**Environment:**
I'm running Docker on a Linux host but colleagues had the same issue on OSX. I have no information about docker on Windows or in the WSL (but I suspect it impacts at least docker running in the WSL).

**Requirements:** 
- bash, docker and the .NET 5 CLI

1. Create a new simple "Hello world" webapp
```
mkdir /tmp/test && cd /tmp/test
dotnet new web
```

2. Run it with `dotnet watch run` and the code mounted in a container
```
docker run -it --rm --workdir /app -v ${PWD}/:/app mcr.microsoft.com/dotnet/sdk:5.0 dotnet watch run
```
You'll notice that when opening, altering and saving the `./Startup.cs` file, the service isn't reloaded.

### Other Information

The issue doesn't occur on the Ubuntu Focal image (`5.0-focal`). Neither does it on the `alpine`-based image but for some unknown reason, (re)build times are significantly slower on Alpine :thinking: .

If that can help, I had the same issue when building a custom .NET Core SDK image with the `./dotnet-install.sh` scripts on `debian:buster-slim`, `debian:slim` and `debian:bullseye`, but not when basing the same custom image off `ubuntu:focal`. Oh and one last thing, it used to work well with the `3.1` image. Maybe a missing dependency in the base `debian` images ?

We just switched our local development environment to the official focal image, which completely solved our issue (so we basically don't care if and when this issue gets fixed :angel: ). 
Since the `5.0` (the "main" image tag line) image is impacted, I thought this bug was worth reporting though.

### Output of `docker version`

```
Client:
 Version:           19.03.13-ce
 API version:       1.40
 Go version:        go1.15.2
 Git commit:        4484c46d9d
 Built:             Sat Sep 26 12:04:46 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
[...]

PR #2403, opened 2020-11-18
Add presence check for procps on Debian-based images

elafarge:
On Debian-based distributions, the procps package is required for
the "dotnet watch" command to work accordingly.

This has introduced a regression in the official sdk:5.0 image
recently.

This commit adds a test that ensures this package is present on
Debian-based distribution (we're assuming an image is Debian based
if the "dpkg" command is present).

Issue: https://github.com/dotnet/dotnet-docker/issues/2396

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
@@ -121,6 +121,33 @@ namespace Microsoft.DotNet.Docker.Tests
 
         [Theory]
         [MemberData(nameof(GetImageData))]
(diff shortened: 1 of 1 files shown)


elafarge:
I didn't expect any impact on the ARM64 Alpine build (it seems that it has `dpkg` installed on it, maybe I'll rely on something else to figure out which OS I'm running on). Will work on that in a few hours.

MichaelSimons:
@elafarge, Thanks for your willingness to help out here.  Per the discussion in #2396, we are going remove this dependency from `dotnet-watch` therefore no changes are needed to the SDK images.  https://github.com/dotnet/aspnetcore/issues/27950 is tracking this work.

PR #2428, opened 2020-12-02
Sample Dockerfiles for Windows Server Core

### LINKED ISSUE #2275: Include sample Dockerfiles for Windows Server Core
With the introduction of Windows Server Core images with https://github.com/dotnet/dotnet-docker/issues/1852, there should be sample Dockerfiles based on Server Core once there are official images so that the community can see examples of apps based on Server Core in addition to Nano Server.

Currently, the sample tests explicitly exclude Server Core which will need to be enabled once these images are made available: see https://github.com/dotnet/dotnet-docker/pull/2274#discussion_r496784026.

mthalman:
Fixes #2275

### DIFF
--- a/samples/aspnetapp/Dockerfile.windowsservercore-x64-slim
+++ b/samples/aspnetapp/Dockerfile.windowsservercore-x64-slim
@@ -1,7 +1,7 @@
 # escape=`
 
 # https://hub.docker.com/_/microsoft-dotnet
-FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build
+FROM mcr.microsoft.com/dotnet/sdk:5.0-windowsservercore-ltsc2019 AS build
 WORKDIR /source
 
 # copy csproj and restore as distinct layers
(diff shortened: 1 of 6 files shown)


samples/aspnetapp/Dockerfile.windowsservercore-x64-slim
```diff
@@ -0,0 +1,29 @@
+# escape=`
+
+# https://hub.docker.com/_/microsoft-dotnet
+FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build
```
MichaelSimons, line 4: Thoughts on using the WSC based sdk here?  I feel like it would provide a slightly better experience in the case you don't have the base nano and WSC images pulled.  Having the build and final layers based on the same OS would reduce the number of unique layers needed to build the sample.

mthalman: Makes total sense to do that.  Fixed

MichaelSimons (approved):

PR #2484, opened 2021-01-07
Fix update-dependencies for .NET 6.0

mthalman:
The content of the `Version.Details.xml` file has changed for .NET 6.0 which requires some updates to the target dependency names in order to gather the correct version numbers to pass to the update-dependencies tool.

Example of `Version.Details.xml` that is being used: https://raw.githubusercontent.com/dotnet/installer/8efbfffe79d15d851d2a34ae502fa53f9430a708/eng/Version.Details.xml

### DIFF
--- a/eng/get-drop-versions.sh
+++ b/eng/get-drop-versions.sh
@@ -23,8 +23,8 @@ rm sdkversion
 
 curl -SLo versionDetails.xml https://raw.githubusercontent.com/dotnet/installer/$commitSha/eng/Version.Details.xml
 
-runtimeVer=$(xmllint --xpath string\(//ProductDependencies/Dependency[@Name=\'Microsoft.NETCore.App.Internal\']/@Version\) versionDetails.xml)
-aspnetVer=$(xmllint --xpath string\(//ProductDependencies/Dependency[@Name=\'Microsoft.AspNetCore.App.Ref.Internal\']/@Version\) versionDetails.xml)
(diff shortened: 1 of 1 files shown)


eng/get-drop-versions.sh
```diff
@@ -23,8 +23,8 @@ rm sdkversion
 
 curl -SLo versionDetails.xml https://raw.githubusercontent.com/dotnet/installer/$commitSha/eng/Version.Details.xml
 
-runtimeVer=$(xmllint --xpath string\(//ProductDependencies/Dependency[@Name=\'Microsoft.NETCore.App.Internal\']/@Version\) versionDetails.xml)
-aspnetVer=$(xmllint --xpath string\(//ProductDependencies/Dependency[@Name=\'Microsoft.AspNetCore.App.Ref.Internal\']/@Version\) versionDetails.xml)
+runtimeVer=$(xmllint --xpath string\(//ProductDependencies/Dependency[@Name=\'VS.Redist.Common.NetCore.SharedFramework.x64.6.0\']/@Version\) versionDetails.xml)
```
MichaelSimons, line 26: This appears to add a coupling to `6.0` within the code versus having it specified as config (e.g. the channel command line option).  Would it be acceptable to retrieve the version from the `Microsoft.NETCore.App.Ref` and `Microsoft.AspNetCore.App.Ref` dependencies?  Alternatively we could ignore the version within the dependency name of the SharedFrameworks.

mthalman: We can't use the `.Ref` dependencies because they don't necessarily change with each build.  I'll look into generalizing the string matching to not hardcode the version.

mthalman: Ok, I made use of the `starts-with` function in order to not hardcode the version number.

MichaelSimons (approved):

PR #2489, opened 2021-01-08
Disable PowerShell tests for 6.0 sdk until it's compatible

mthalman:
PowerShell was mistakenly included in the 6.0 sdk even though that version of PowerShell is based on 5.0.  There is not yet a version based on 6.0 so I'm removing it until a version is available.


MichaelSimons (commented):
I lean towards leaving PS in the images and simply disable the tests.  I realize PS won't work in the current images.  I prefer to keep the structure of the Dockerfiles and images what they should be.  Thoughts?

mthalman:
> I lean towards leaving PS in the images and simply disable the tests. I realize PS won't work in the current images. I prefer to keep the structure of the Dockerfiles and images what they should be. Thoughts?

Sure, I can get behind that.

MichaelSimons (approved):

PR #2490, opened 2021-01-08
Remove usage of deprecated syntax for ENV instruction

### LINKED ISSUE #2449: Remove usage of legacy ENV syntax
The syntax for `ENV` instructions that doesn't use an assignment operator is being phased out.  All Dockerfiles that use this syntax should be updated.

https://github.com/docker/cli/pull/2743

Legacy syntax:
```Dockerfile
ENV var value
```
Current syntax:
```Dockerfile
ENV var=value
```

mthalman:
Updates the usage of `ENV name value` syntax to `ENV name=value`.

Fixes #2449


mthalman:
I need to make some updates on other instances I need to fix.

MichaelSimons (approved):

PR #2498, opened 2021-01-11
Upgrade System.CommandLine version

### LINKED ISSUE #2493: Replace dependency on System.CommandLine
The update-dependencies tool has a dependency on an older, unsupported version of the `System.CommandLine` NuGet package which had existed in a `dotnet.myget.org` feed but that feed has now been deprecated and no longer available.  For that reason, restoring packages for the project will fail with an error similar to the following:

```
/update-dependencies/update-dependencies.csproj : warning NU1603: update-dependencies depends on System.CommandLine (>= 0.1.0-e170603-2) but System.CommandLine 0.1.0-e170603-2 was not found. An approximate best match of System.CommandLine 0.3.0-alpha.20054.1 was resolved.
/update-dependencies/Options.cs(27,13): error CS0246: The type or namespace name 'ArgumentSyntax' could not be found (are you missing a using directive or an assembly reference?) [/update-dependencies/update-dependencies.csproj]
/update-dependencies/Options.cs(27,40): error CS0103: The name 'ArgumentSyntax' does not exist in the current context [/update-dependencies/update-dependencies.csproj]
```

The project needs to take a new dependency on a supported version of `System.CommandLine`.

mthalman:
Removes the use of an older, unsupported version of `System.CommandLine` that is no longer available from the deprecated `dotnet.myget.org` feed.  Replaces it with the new version of `System.CommandLine` from https://github.com/dotnet/command-line-api. This has a new API model that the code has been refactored to consume.

Fixes #2493

### DIFF
--- a/eng/update-dependencies/Options.cs
+++ b/eng/update-dependencies/Options.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections.Generic;
 using System.CommandLine;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


eng/update-dependencies/Options.cs
```diff
@@ -22,61 +21,29 @@ public class Options
...
+            GitHubEmail = email;
+            GitHubPassword = password;
+            GitHubUser = user;
+            ComputeChecksums = computeShas;
         }
+
+        public static IEnumerable<Symbol> GetCliSymbols() =>
+            new Symbol[]
+            {
+                new Argument<string>("dockerfile-version", "Version to the Dockerfiles to update"),
```
MichaelSimons, line 40: I realize this was the previous wording...
```suggestion
                new Argument<string>("dockerfile-version", "Version of the Dockerfiles to update"),
```

eng/update-dependencies/Program.cs
```diff
@@ -68,9 +83,9 @@ public static async Task Main(string[] args)
             Environment.Exit(0);
         }
 
-        private static void ParseOptions(string[] args)
+        private static void InitOptions(Options options)
         {
-            Options.Parse(args);
+            Options = options;
```
MichaelSimons, line 89: Thoughts on encapsulating this within the Options class?  Arguably this should have been done before.  It could even be invoked from the ctor which wasn't possible before.

MichaelSimons (approved):
Nice to see the relatively easy migration.

PR #2501, opened 2021-01-14
Update dotnet-monitor to output JSON logging by default.

jander-msft:
This change has the dotnet-monitor tool output its logging in the JSON format, which is a better format for ingesting into analytic platforms, such as Container Insights, because it allows for discrete field selection with minimal parsing.

### DIFF
--- a/eng/dockerfile-templates/monitor/5.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/monitor/5.0/Dockerfile.alpine
@@ -25,7 +25,13 @@ ENV \
     # Disable debugger and profiler diagnostics to avoid diagnosing self.
     COMPlus_EnableDiagnostics=0 \
     # Add dotnet-monitor path to front of PATH for easier, prioritized execution
-    PATH="/app:${PATH}"
+    PATH="/app:${PATH}" \
+    # Logging: JSON format so that analytic platforms can get discrete entry information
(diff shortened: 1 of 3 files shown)


eng/dockerfile-templates/monitor/5.0/Dockerfile.alpine
```diff
@@ -25,7 +25,13 @@ ENV \
     # Disable debugger and profiler diagnostics to avoid diagnosing self.
     COMPlus_EnableDiagnostics=0 \
     # Add dotnet-monitor path to front of PATH for easier, prioritized execution
-    PATH="/app:${PATH}"
+    PATH="/app:${PATH}" \
```
mthalman, line 28: Could you move `PATH` to be the last one in the list of ENVs?  That way they'll be alphabetical.  Same comment applies to the other Dockerfile too.

MichaelSimons (approved):
I had the same comment as Matt on alphabetizing.  The rest LGTM.

wiktork (approved):

shirhatti (approved):

eng/dockerfile-templates/monitor/5.0/Dockerfile.alpine
jander-msft: Done

PR #2516, opened 2021-01-19
Fix the version of PowerShell in 5.0 sdk

### LINKED ISSUE #2507: 5.0 sdk images should be updated to contain PowerShell 7.1.1
The 5.0 sdk images were incorrectly updated to the PowerShell 7.2 (preview) channel in Dec via https://github.com/dotnet/dotnet-docker/pull/2401.  The 5.0 sdk images should be to the PowerShell 7.1 channel.

mthalman:
This is a cherry-pick of dedafa9390ecb40de483e370b11750ba40032889 and updates the tags for the stable .NET version of 5.0 sdk.  This updates the version of PowerShell installed in the images because they were mistakenly configured to use 7.2 preview when they should be using 7.1.

This changes updates all floating tags to use the correct version of PowerShell.  The stable tags (e.g. `5.0.102-buster-slim-amd64`) remain unchanged and will continue to have the incorrect version of PowerShell.  For those users that wish to use a stable tag and target the correct version of PowerShell, they will need to update the tag with the newly added suffix to the version number (e.g. `5.0.102-1-buster-slim-amd64`).  This was done because users that make use of stable version tags generally want to avoid any potential destabilization.  So we're not going to update the image associated with a stable tag; it will require an explicit update of the tag to move to a different version of PowerShell.

Fixes #2507


MichaelSimons (approved):
LGTM.  It's nice to see the infrastructure made it so easy to produce the `-1` tag.

Issue #2504, opened 2021-01-14
Add 6.0 SDK Alpine Arm images 


MichaelSimons:
In 6.0 the SDK has added support for Alpine Arm

* https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-alpha.1.21054.2/dotnet-sdk-6.0.100-alpha.1.21054.2-linux-musl-arm64.tar.gz
* https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-alpha.1.21054.2/dotnet-sdk-6.0.100-alpha.1.21054.2-linux-musl-arm.tar.gz

As a result it is now possible to create `dotnet/sdk:6.0-alpine3.12-arm32v7` and `dotnet/sdk:6.0-alpine3.12-arm64v8` images.  This is great as is completes the image variant matrix for all os/architecture variants.

PR #2543, opened 2021-01-26
Add SDK Dockerfiles for Arm arch on Alpine

mthalman:
Things of note:
* Buster SDK is now no longer needed in order to test the runtime and aspnet images for Arm.
* PowerShell doesn't provide an Arm version of Alpine (see https://github.com/PowerShell/PowerShell/issues/12937 and https://github.com/PowerShell/PowerShell/issues/14667) so PowerShell needed to be removed from the sdk for Arm.

Related to #2504


MichaelSimons (approved):
It's great to see that we now have parity across all image and OS variants!

PR #2556, opened 2021-01-29
Add dotnet-monitor variable and metrics endpoint tests.

jander-msft:
These changes add two unit tests for dotnet-monitor:
- Basic environment variable test
- Basic metrics endpoint availability test (tests that the container does run and that the metrics endpoint is accessible).

I plan to add additional tests in a future change for the other endpoints once https://github.com/dotnet/diagnostics/pull/1820 is completed.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/MonitorImageData.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/MonitorImageData.cs
@@ -1,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class MonitorImageData : ImageData
+    {
+        public Version Version { get; set; }
+        public string VersionString => Version.ToString(2);
+
+        public string GetImage(DockerHelper dockerHelper)
+        {
+            string tagVersion = Version.ToString(2);
+
+            string tag = GetTagName(tagVersion, OS);
+
+            string imageName = GetImageName(tag, "monitor");
+
+            PullImageIfNecessary(imageName, dockerHelper);
+
+            return imageName;
+        }
+
+        protected override string GetArchTagSuffix() => string.Empty;
+    }
+}
--- a/tests/Microsoft.DotNet.Docker.Tests/MonitorImageData.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/MonitorImageData.cs
@@ -12,12 +12,11 @@ namespace Microsoft.DotNet.Docker.Tests
         public string RuntimeVersionString => RuntimeVersion.ToString(2);
         public Version Version { get; set; }
         public string VersionString => Version.ToString(2);
+        public string OSTag { get; set; }
 
         public string GetImage(DockerHelper dockerHelper)
         {
-            string tagVersion = Version.ToString(2);
-
-            string tag = GetTagName(tagVersion, OS);
+            string tag = GetTagName(VersionString, OSTag);
 
             string imageName = GetImageName(tag, "monitor");
 
--- a/tests/Microsoft.DotNet.Docker.Tests/MonitorImageData.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/MonitorImageData.cs
@@ -8,6 +8,8 @@ namespace Microsoft.DotNet.Docker.Tests
 {
     public class MonitorImageData : ImageData
     {
+        public Version RuntimeVersion { get; set; }
+        public string RuntimeVersionString => RuntimeVersion.ToString(2);
         public Version Version { get; set; }
         public string VersionString => Version.ToString(2);
 
--- a/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
@@ -15,8 +15,6 @@ namespace Microsoft.DotNet.Docker.Tests
     [Trait("Category", "monitor")]
     public class MonitorImageTests
     {
-        private static readonly string s_samplesPath = Path.Combine(Config.SourceRepoRoot, "samples");
-
         public MonitorImageTests(ITestOutputHelper outputHelper)
         {
             OutputHelper = outputHelper;
@@ -31,13 +29,6 @@ namespace Microsoft.DotNet.Docker.Tests
         {
             return TestData.GetMonitorImageData()
                 .Select(imageData => new object[] { imageData });
-        }
-
-        [Theory]
-        [MemberData(nameof(GetImageData))]
-        public void VerifyEnvironmentVariables(MonitorImageData imageData)
-        {
-            ValidateEnvironmentVariables(imageData, imageData.GetImage(DockerHelper));
         }
 
         [Theory]
--- a/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
@@ -33,6 +33,25 @@ namespace Microsoft.DotNet.Docker.Tests
 
         [Theory]
         [MemberData(nameof(GetImageData))]
+        public void VerifyEnvironmentVariables(MonitorImageData imageData)
+        {
+            List<EnvironmentVariableInfo> variables = new List<EnvironmentVariableInfo>();
+            variables.AddRange(ProductImageTests.GetCommonEnvironmentVariables());
+
+            // ASPNETCORE_URLS has been unset to allow the default URL binding to occur.
+            variables.Add(new EnvironmentVariableInfo("ASPNETCORE_URLS", string.Empty));
+            // Diagnostics should be disabled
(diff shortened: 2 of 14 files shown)


tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            await verifyImageAsync(image, containerName);
+        }
+
+        private void ValidateEnvironmentVariables(MonitorImageData imageData, string image)
+        {
+            List<EnvironmentVariableInfo> variables = new List<EnvironmentVariableInfo>();
+            variables.AddRange(ProductImageTests.GetCommonEnvironmentVariables());
+
+            // ASPNETCORE_URLS has been unset to allow the default URL binding to occur.
+            variables.Add(new EnvironmentVariableInfo("ASPNETCORE_URLS", string.Empty));
```
wiktork, line 94: This has been added to the image, is it still required here?

wiktork (approved):

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -0,0 +1,105 @@
...
+                        // Verify metrics endpoint is accessible
+                        await ImageScenarioVerifier.VerifyHttpResponseFromContainerAsync(
+                            containerName,
+                            DockerHelper,
+                            OutputHelper,
+                            52325,
+                            "metrics");
+                    }
+
+                    ValidateEnvironmentVariables(imageData, image);
```
MichaelSimons, line 68: How is this different that what is done in VerifyEnvironmentVariables?

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    [Trait("Category", "monitor")]
+    public class MonitorImageTests
+    {
+        private static readonly string s_samplesPath = Path.Combine(Config.SourceRepoRoot, "samples");
```
mthalman, line 18: Unused

MichaelSimons, line 18: I'm not spotting a reference to this.  What is this needed for?

tests/run-tests.ps1
```diff
@@ -14,7 +14,7 @@ param(
     [switch]$DisableHttpVerification,
     [switch]$PullImages,
     [string]$ImageInfoPath,
-    [ValidateSet("runtime", "runtime-deps", "aspnet", "sdk", "pre-build", "sample", "image-size")]
+    [ValidateSet("runtime", "runtime-deps", "aspnet", "sdk", "pre-build", "sample", "image-size", "monitor")]
     [string[]]$TestCategories = @("runtime", "runtime-deps", "aspnet", "sdk")
```
mthalman, line 18: This will need to include `monitor` by default in order for these tests to be run by the pipeline.

```suggestion
    [string[]]$TestCategories = @("runtime", "runtime-deps", "aspnet", "sdk", "monitor")
```

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            await verifyImageAsync(image, containerName);
+        }
+
+        private void ValidateEnvironmentVariables(MonitorImageData imageData, string image)
+        {
+            List<EnvironmentVariableInfo> variables = new List<EnvironmentVariableInfo>();
+            variables.AddRange(ProductImageTests.GetCommonEnvironmentVariables());
+
+            // ASPNETCORE_URLS has been unset to allow the default URL binding to occur.
```
MichaelSimons, line 93: Do you see value in validating the various logging variables defined in the image?

mthalman (commented):
~~This will also require pipeline changes~~  Disregard

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
jander-msft: Removed

jander-msft: Unused and removed.

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
jander-msft: I've removed the VerifyEnvironmentVariable test in favor of incorporating it into other tests.

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
jander-msft: At this time, I don't think its valuable to check the logging variables as they are not crucial to the functioning of the tool or the image.

tests/run-tests.ps1
jander-msft: Added

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
MichaelSimons: From a unit testing perspective and consistency with the other image types, I prefer the separate test case.

jander-msft: I based these tests on the sample tests, which does it the way that I have it here. However, those tests are testing environment variables in different images whereas these tests are testing the same image. I'll add back the VerifyEnvironmentVariables unit test and remove the call from this test. But, if HTTP verification is disabled, then the VerifyMetricsEndpoint test becomes nearly pointless, but that's probably okay since the test is mainly for verifying the endpoint.

MichaelSimons: Thanks.  FYI, DisableHttpVerification is only used in development scenarios so I don't see this as a concern.

tests/run-tests.ps1
mthalman: Ah, this causes issues with the other tests.  I'll have to think about this.

mthalman:
@jander-msft - I'm working on a proposal for fixing the test issue that we have.  I'll share it with you when it's ready.

mthalman:
@jander-msft:
The problem we have with the tests is that pipeline doesn't know which test categories should be executed for a given test job. So it doesn't provide test categories and leaves it up to the defaults defined in the `run-tests.ps1` script.  Now that `monitor` has been added to the mix, it attempts to run the monitor tests in all the test jobs.  Because of the filtering logic in the tests, it determines that certain test jobs aren't associated with the same OS, .NET version, etc. as the monitor.  So the test data gets filtered out, leaving no input to the Theory test case.  XUnit doesn't like having empty input data for theories and so it throws.

Here's my proposal of the fix for the testing issue: https://gist.github.com/mthalman/2de08d4c56ddec13091eeffb438689e5.  If you add those files to the test project and replace all usages of `Theory` with `DotNetTheory`, things should work.  This is basically an extension of the default behavior in XUnit's `TheoryDiscoverer` class.  If the base implementation has determined that no input data exists for the theory, this custom implementation will catch that and replace the result with data that tells XUnit to skip the test instead of marking it as a failed test.

If you agree with that approach, please add it to this PR and make the necessary Theory replacements to validate that this work with your changes.

tests/Microsoft.DotNet.Docker.Tests/MonitorImageData.cs
```diff
@@ -0,0 +1,31 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class MonitorImageData : ImageData
+    {
+        public Version RuntimeVersion { get; set; }
+        public string RuntimeVersionString => RuntimeVersion.ToString(2);
```
jander-msft, line 12: @mthalman I've differentiated runtime version vs image version so that when running tests the build machines can test the images that it built e.g. a machine builds the dotnet-monitor image only if building .NET Core 2.1 Linux x64 images.

I think it would also be useful to add a PRODUCT_VERSION environment variable so that we can filter the dotnet-monitor images on product version (rather than on runtime version) and add this as an option to the run-tests.ps1 script. Maybe I'll create an issue so that we can discuss further (and I won't add that to this change at this time).

tests/Microsoft.DotNet.Docker.Tests/TestData.cs
```diff
@@ -136,6 +154,14 @@ public static IEnumerable<ImageData> FilterImagesByVersion(this IEnumerable<Prod
                     || Regex.IsMatch(imageData.VersionString, versionFilterPattern, RegexOptions.IgnoreCase));
         }
 
+        public static IEnumerable<ImageData> FilterImagesByRuntimeVersion(this IEnumerable<MonitorImageData> imageData)
```
mthalman, line 157: Suggest renaming this to FilterMonitorImagesByRuntimeVersion just to make its purpose even more clear.

mthalman (approved):

MichaelSimons (approved):

Issue #2530, opened 2021-01-21
Add support for Alpine 3.13


mthalman:
# New Linux Distro Version

Distro: [Alpine 3.13](https://alpinelinux.org/posts/Alpine-3.13.0-released.html)

## ImageBuilder Tasks

- [x] Ensure that the ImageBuilder supports the new distro version in the [code](https://github.com/dotnet/docker-tools/blob/master/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs) to generate the correct README display name from the version specified in the manifest

## Nightly Branch Tasks

1. - [x] Copy the Dockerfiles of the most recent version of that distro (or author new ones for an entirely new distro) and place them in a version-specific folder under their respective variants (runtime-deps, runtime, aspnet, sdk). Note: not all variants have a corresponding runtime-deps image.
1. - [x] Modify the Dockerfiles as appropriate for any specific changes related to the new distro version
1. - [x] Update [manifest.json](https://github.com/dotnet/dotnet-docker/blob/nightly/manifest.json) to reference the new set of Dockerfiles with the appropriate tags
      - [x] Move any distro-specific floating tags to the newer version (e.g. `3.1-alpine`)
1. - [x] Update the [test data](https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/TestData.cs) to include the new distro version
1. - [x] Update the [tags metadata templates](https://github.com/dotnet/dotnet-docker/tree/master/eng/mcr-tags-metadata-templates) to include the new distro version
1. - [x] Run the command to update the READMEs: `.\eng\readme-templates\Get-GeneratedReadmes.ps1`
1. - [x] Run the command to update the image size baseline file: `.\tests\performance\Validate-ImageSize.ps1 -UpdateBaselines`
1. - [x] Inspect generated changes for correctness
1. - [x] Consider whether sample Dockerfiles should be authored if this is a new distro and them to the [samples](https://github.com/dotnet/dotnet-docker/tree/master/samples)
1. - [x] Run the command to build and test your changes: `.build-and-test.ps1 -OS <os>`
1. - [x] Commit generated changes
1. - [x] Create PR
1. - [x] Get PR signoff
1. - [x] Merge PR to nightly branch
[...]

PR #2566, opened 2021-02-01
Define Alpine 3.13 runtime-deps in 5.0 folder

mthalman:
Because we will not be publishing Alpine 3.13 images for the servicing release of .NET Core 2.1 and 3.1 next week, I'm restructuring the files here in order to prep for that release next week.  We will be publishing Alpine 3.13 images for 5.0 and 6.0 Preview 1.  So I'm making a copy of the runtime-deps Dockerfile to be contained in the 5.0 folder and shared between 5.0 and 6.0 instead of before where it was shared by 3.1, 5.0, and 6.0.  It wouldn't have made sense to have the 3.1 folder contain a Alpine 3.13 Dockerfile when aren't going to be release that Alpine version for 3.1.

It still will use the `runtime-deps/3.1/Dockerfile.alpine` Dockerfile template because the content we need is still exactly the same; it's just the placement of the generated Dockerfile that's being changed.

Once we're ready to ship Alpine 3.13 images for .NET Core 3.1, we can have them share a single Dockerfile again.  I'll log an issue after merging this PR to ensure that work gets done.

Related to https://github.com/dotnet/dotnet-docker/issues/2530


MichaelSimons (commented):
LGTM - Please add a note in https://github.com/dotnet/dotnet-docker/issues/2530, to track reverting these changes once 2.1 and 3.1 support Alpine 3.13

PR #2582, opened 2021-02-05
Update 6.0 Debian from Buster to Bullseye

### LINKED ISSUE #2576: Update 6.0 images to bullseye
Debian 11 will likely release before 6.0. We should just start the release targeting bullseye.

This is similar to what we did with .NET Core 2.0 and stretch, if I recall correctly.

mthalman:
Fixes #2576


MichaelSimons (approved):
LGTM

Issue #2544, opened 2021-01-26
Remove test workaround once PowerShell has Alpine support for Arm


mthalman:
The changes in https://github.com/dotnet/dotnet-docker/pull/2543 included a change to disable a test due to PowerShell not being available in the image because of its lack of support for Arm on Alpine (https://github.com/PowerShell/PowerShell/issues/14667, https://github.com/PowerShell/PowerShell/issues/12937).  Once PowerShell does provide support and it is incorporated into the sdk images, the test workaround should be removed.

https://github.com/dotnet/dotnet-docker/blob/426ed5f048c16cd675b72c30263e45be392215c3/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs#L168-L172

lbussell:
This will be addressed by https://github.com/dotnet/dotnet-docker/pull/4926 (not by installing PowerShell on arm Alpine images)

PR #2599, opened 2021-02-12
Update PowerShell for 3.1, 5.0 and 6.0 SDK images

adityapatwardhan:
Update all the corresponding dokerfiles for 3.1, 5.0 and 6.0 SDK images


mthalman:
@adityapatwardhan - Can you revert the changes that were made in https://github.com/dotnet/dotnet-docker/pull/2489?  Please include that in this PR so that the tests can validate things.

adityapatwardhan:
@mthalman Done.

mthalman:
@adityapatwardhan - Sorry for the churn on this.  The tests are failing due to no Alpine support on Arm for PowerShell.  So the test changes you made need to be refined to only skip the tests for this skip condition rather than for all of 6.0.  See https://github.com/dotnet/dotnet-docker/issues/2544 for example code.

adityapatwardhan:
Updated

PR #2603, opened 2021-02-16
Add Dockerfiles for 6.0 Preview 1

mthalman:
Related to https://github.com/dotnet/dotnet-docker/issues/2602


MichaelSimons (approved):
:shipit:

eng/dockerfile-templates/sdk/6.0/Dockerfile.alpine
```diff
@@ -0,0 +1,46 @@
+ARG REPO=mcr.microsoft.com/dotnet/aspnet
+FROM $REPO:6.0-{{OS_VERSION}}{{ARCH_TAG_SUFFIX}}
+
+ENV \
+    # Unset ASPNETCORE_URLS from aspnet base image
+    ASPNETCORE_URLS= \
```
alrz, line 6: Why? It took me a few days to figure out what's going on. @mthalman

mthalman: This has been the case since .NET Core 3.0. See https://github.com/dotnet/dotnet-docker/issues/1199 and https://github.com/dotnet/dotnet-docker/pull/1361.

alrz: > This has been the case since .NET Core 3.0.

Okay, but why bullseye is different from the default image?

PR #2605, opened 2021-02-18
Update dotnet-monitor version source and manually update to latest version

jander-msft:
Arcade made a change recently that prevented the ability to create aka.ms latest links for repositories that do not have aka.ms channel names. Thus dotnet-monitor has not been able to provide updated versions to dotnet-docker. This change temporarily fixes the version of dotnet-monitor while Arcade resolves the issue and pushes out a new SDK.

Additionally, dotnet-monitor has moved to its own repository (https://github.com/dotnet/dotnet-monitor) so I've updated the version source to reflect that.

I'll create an issue in dotnet-docker and assign it to myself to undo the temporary version fixing.

### DIFF
--- a/eng/get-drop-versions-monitor.sh
+++ b/eng/get-drop-versions-monitor.sh
@@ -12,6 +12,4 @@ monitorVer=$(tr -d '\r\n' < dotnet-monitor.nupkg.version)
 
 rm dotnet-monitor.nupkg.version
 
-#echo "##vso[task.setvariable variable=monitorVer]$monitorVer"
-# Temporarily fix the version since Arcade cannot properly create aka.ms for assets without an aka.ms channel name.
-echo "##vso[task.setvariable variable=monitorVer]5.0.0-preview.4.21117.6"
+echo "##vso[task.setvariable variable=monitorVer]$monitorVer"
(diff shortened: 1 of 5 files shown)


wiktork (approved):

mthalman (approved):

eng/get-drop-versions-monitor.sh
```diff
@@ -12,4 +12,6 @@ monitorVer=$(tr -d '\r\n' < dotnet-monitor.nupkg.version)
 
 rm dotnet-monitor.nupkg.version
 
-echo "##vso[task.setvariable variable=monitorVer]$monitorVer"
+#echo "##vso[task.setvariable variable=monitorVer]$monitorVer"
+# Temporarily fix the version since Arcade cannot properly create aka.ms for assets without an aka.ms channel name.
+echo "##vso[task.setvariable variable=monitorVer]5.0.0-preview.4.21117.6"
```
mthalman, line 17: It would seem better to leave this script as it was and just disable the automation.  There's no sense in running schedule automation for something that's just returning a fixed value.  Just run the update-dependencies script locally when you want to update it and submit a PR that way.

mthalman (approved):

PR #2608, opened 2021-02-19
Additional dotnet-monitor tests

### LINKED ISSUE #2117: Create tests for the monitor images
A set of unit tests should be created for the `monitor` images.  To conform with the existing tests, two basic tests come to mind.

1. Validate the static state such as the `ENVs` defined in the image.
1. A basic scenario test that ensures dotnet-monitor is installed correctly and is usable.

I consider these tests to be a requirement before merging the images to the master branch.

jander-msft:
These changes refactor the existing tests and add a few more.

Basic container tests:
- VerifyMonitorDefault: Run image without modification and test that it has a viable metrics endpoint.
- VerifyMonitorNoHttps: Disables HTTPS on the artifacts port and checks that /processes returns 401 due to lack of authentication.
- VerifyMonitorNoHttpsNoAuth: Disables HTTPS and authentication and checks that /processes returns empty list of processes.

Basic scenario tests:
- VerifyConnectModeNoAuth: Verifies that dotnet-monitor can discover the samples app running in another container when both have a mounted volumne at /tmp.
- VerifyListenModeNoAuth: Verifies that dotnet-monitor can be set to listening mode and the samples app in the other container can connect to dotnet-monitor and dotnet-monitor is able to see that process.

closes #2117

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
@@ -216,7 +216,18 @@ namespace Microsoft.DotNet.Docker.Tests
             }
         }
 
-        public static async Task VerifyHttpResponseFromContainerAsync(string containerName, DockerHelper dockerHelper, ITestOutputHelper outputHelper, int containerPort = 80, string pathAndQuery = null)
+        public static Task<HttpResponseMessage> GetHttpResponseFromContainerAsync(string containerName, DockerHelper dockerHelper, ITestOutputHelper outputHelper, int containerPort = 80, string pathAndQuery = null)
+        {
+            return GetHttpResponseFromContainerAsync(
+                containerName,
+                dockerHelper,
+                outputHelper,
+                containerPort,
+                pathAndQuery,
+                m => m.EnsureSuccessStatusCode());
+        }
+
+        public static async Task<HttpResponseMessage> GetHttpResponseFromContainerAsync(string containerName, DockerHelper dockerHelper, ITestOutputHelper outputHelper, int containerPort = 80, string pathAndQuery = null, Action<HttpResponseMessage> validateCallback = null)
         {
             int retries = 30;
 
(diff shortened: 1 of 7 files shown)


jander-msft:
@mthalman, seems the new tests that depend on using the sample images are failing because PULL_IMAGES is not set and the samples are not built in these pipelines. Do you think it would be appropriate to force pull the sample images from these two tests?

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -216,7 +216,18 @@ private async Task RunTestAppImage(string image, bool runAsAdmin = false, string
...
+        public static Task<HttpResponseMessage> GetHttpResponseFromContainerAsync(string containerName, DockerHelper dockerHelper, ITestOutputHelper outputHelper, int containerPort = 80, string pathAndQuery = null)
+        {
+            return GetHttpResponseFromContainerAsync(
+                containerName,
+                dockerHelper,
+                outputHelper,
+                containerPort,
+                pathAndQuery,
+                m => m.EnsureSuccessStatusCode());
+        }
```
mthalman, line 228: This overload seems unnecessary since the other overload on line 230 has an optional parameter.  You could just modify that method to default `validateCallback` to `m => m.EnsureSuccessStatusCode()` if the value is null.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -232,24 +243,44 @@ public static async Task VerifyHttpResponseFromContainerAsync(string containerNa
...
+                    HttpResponseMessage result = null;
                     try
                     {
-                        using (HttpResponseMessage result = await client.GetAsync(url))
+                        result = await client.GetAsync(url);
+                        outputHelper.WriteLine($"HTTP {result.StatusCode}\n{(await result.Content.ReadAsStringAsync())}");
+
+                        if (null != validateCallback)
                         {
-                            outputHelper.WriteLine($"HTTP {result.StatusCode}\n{(await result.Content.ReadAsStringAsync())}");
```
mthalman, line 239: I think this logging should be left intact regardless of what the callback does with the result.

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -50,47 +98,420 @@ public void VerifyEnvironmentVariables(MonitorImageData imageData)
...
         {
-            string image = imageData.GetImage(DockerHelper);
+            imageName = imageData.GetImage(DockerHelper);
+            containerName = imageData.GetIdentifier("monitortest");
+        }
 
-            string containerName = imageData.GetIdentifier("monitor");
+        private void GetNames(SampleImageData imageData, out string imageName, out string containerName)
+        {
+            imageName = imageData.GetImage(SampleImageType.Aspnetapp, DockerHelper);
```
mthalman, line 451: You'll need to explicitly pull the sample image before this line.  This is currently causing a test failure in the build because the tests are enabled to not pull images by default.  This is a safe-guard to prevent the accidentally pulling of an image during a `docker build` when we only want to test images that have been locally built.  In this case, you do want images that have not been built locally.  To do that, call `DockerHelper.Pull`.  That would require a little bit of refactoring of `SampleImageData.GetImage` cuz you'll obviously need to know the name of the image in order to pull it but the `GetImage` method that gives you the name also validates that the image exists.  So maybe add a `forcePull` optional parameter to that method?

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -50,47 +98,420 @@ public void VerifyEnvironmentVariables(MonitorImageData imageData)
...
+                    {
+                        using HttpResponseMessage responseMessage =
+                            await ImageScenarioVerifier.GetHttpResponseFromContainerAsync(
+                                monitorName,
+                                DockerHelper,
+                                OutputHelper,
+                                DefaultArtifactsPort,
+                                UrlPath_Processes);
+
+                        JsonDocument document = JsonDocument.Parse(responseMessage.Content.ReadAsStream());
```
wiktork, line 247: nit: Did you want to create a helper around the root element extraction?

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
jander-msft: It is left intact. It immediately follows the HttpClient.GetAsync call unconditionally.

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
jander-msft: Done

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
jander-msft: Added allowPull parameter with default value of false.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
jander-msft: I've made the change as you suggested.

mthalman (approved):

PR #2613, opened 2021-02-22
Update image size baselines

mthalman:
Some size changes to the SDK were introduced from the transition of 6.0 Preview 1 to Preview 2.  See https://github.com/dotnet/sdk/issues/15996.

### DIFF
--- a/tests/performance/ImageSize.nightly.linux.json
+++ b/tests/performance/ImageSize.nightly.linux.json
@@ -145,15 +145,15 @@
     "src/sdk/5.0/focal/amd64": 615235315,
     "src/sdk/5.0/focal/arm32v7": 564440607,
     "src/sdk/5.0/focal/arm64v8": 637793431,
-    "src/sdk/6.0/bullseye-slim/amd64": 636616449,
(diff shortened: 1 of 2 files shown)


MichaelSimons (approved):
Thoughts on logging a tracking issue on our side for this?

mthalman:
> Thoughts on logging a tracking issue on our side for this?

What benefit do you see in doing that?  I see tracking issues as being useful in cases where we are blocked or require a workaround.  But that's not the case for this.

MichaelSimons:
> What benefit do you see in doing that? I see tracking issues as being useful in cases where we are blocked or require a workaround. But that's not the case for this.

To raise visibility of the issue to the docker triage/tactics and track the progress of the issue.  If the issue you logged doesn't get acted upon, docker triage/tactics may decide to push back.

mthalman:
> To raise visibility of the issue to the docker triage/tactics and track the progress of the issue. If the issue you logged doesn't get acted upon, docker triage/tactics may decide to push back.

Sounds fair.  Tracking issue: https://github.com/dotnet/dotnet-docker/issues/2616

PR #2617, opened 2021-02-23
Consume mirrored tags in build pipeline

mthalman:
Updates the build pipeline so that base images will be pulled from the mirror location instead of their base origin.  This is done by ensuring that the mirrored tags are up-to-date by executing the `copyBaseImages` command as a pre-build step.  And then providing the `source-repo-prefix` option to the `build` command to specify the location of the mirrored tags.

Related to https://github.com/dotnet/docker-tools/issues/613


eng/common/templates/jobs/copy-base-images.yml
```diff
@@ -0,0 +1,28 @@
+parameters:
+  name: null
+  additionalOptions: null
+  
+jobs:
```
MichaelSimons, line 5: Should this be skipped for PR validation because it won't have access to the necessary credentials?

eng/common/templates/stages/build-test-publish-repo.yml
```diff
@@ -32,13 +32,28 @@ stages:
...
+      pool:
+        vmImage: $(defaultLinuxAmd64PoolImage)
+      additionalOptions: "--manifest $(manifest)"
   - template: ../jobs/generate-matrix.yml
     parameters:
       matrixType: ${{ parameters.buildMatrixType }}
       name: GenerateBuildMatrix
       customBuildLegGroupArgs: ${{ parameters.buildMatrixCustomBuildLegGroupArgs }}
 
+  # This is just a helper job with no tasks that will complete once all pre-build-related jobs have finished
```
MichaelSimons, line 47: The intent of this is nice but given there is only one dependent consumer and because it still needs to reference GenerateBuildMatrix the value is questionable IMO.

eng/common/templates/jobs/copy-base-images.yml
mthalman: I see that the build is attempting to copy the images and fails but the failure is ignored.  I think it would be worthwhile to still execute this in a PR but with a dry-run option instead.  This would be the same approach that is done in the publish stage.  It helps to identify potential issues that could be caught in a PR even if it is a dry run.

MichaelSimons: That makes sense.

eng/common/templates/stages/build-test-publish-repo.yml
mthalman: Yeah, I debated with myself on that.  I'm fine with omitting it.

eng/common/templates/steps/set-dry-run.yml
```diff
@@ -0,0 +1,9 @@
+steps:
+- script: |
+      # Use dry-run option for certain publish operations if this is not a production build
+      dryRunArg=""
+      if [[ "$PUBLISHREPOPREFIX" != "public/" || "$SYSTEM_TEAMPROJECT" == "public" ]]; then
+        dryRunArg=" --dry-run"
+      fi
+      
+      echo "##vso[task.setvariable variable=dryRunArg]$dryRunArg"
```
MichaelSimons, line 9: Missing EOF

MichaelSimons (approved):

PR #2627, opened 2021-02-26
Update aka.ms path to sdk zip file

mthalman:
There have been some recent changes to the generated aka.ms links that point to the blob locations for .NET bits.  The `Sdk` portion of the path is no longer needed since the link implicitly points to that location.

### DIFF
--- a/eng/get-drop-versions.sh
+++ b/eng/get-drop-versions.sh
@@ -10,7 +10,7 @@ channel=$1
 sudo apt-get update && \
     sudo apt-get install -y --no-install-recommends libxml2-utils
 
-curl -SLo sdk.zip https://aka.ms/dotnet/$channel/Sdk/dotnet-sdk-win-x64.zip
+curl -SLo sdk.zip https://aka.ms/dotnet/$channel/dotnet-sdk-win-x64.zip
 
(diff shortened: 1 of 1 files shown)


eng/get-drop-versions.sh
```diff
@@ -10,7 +10,7 @@ channel=$1
 sudo apt-get update && \
     sudo apt-get install -y --no-install-recommends libxml2-utils
 
-curl -SLo sdk.zip https://aka.ms/dotnet/$channel/Sdk/dotnet-sdk-win-x64.zip
+curl -SLo sdk.zip https://aka.ms/dotnet/$channel/dotnet-sdk-win-x64.zip
```
MichaelSimons, line 13: I saw the notification from engineering services mentioning the new format which includes the build quality.  Is that going to affect this URL as well?  e.g. `https://aka.ms/dotnet/5.0.3xx/daily/`

mthalman: Are you specifically asking whether non-6.0 versions also follow this pattern?  The answer is yes.

And for 6.0, the channel variable will be defined as `6.0.1xx-preview2/daily`

MichaelSimons (approved):

PR #2694, opened 2021-03-23
Update samples

richlander:
- Updated dotnetapp to be more directly helpful in testing .NET in containers, based on https://github.com/dotnet/core/blob/main/samples/dotnet-runtimeinfo/Program.cs.
- Enabled nulllable for the console samples (for good measure). The ASP.NET app didn't just work for nullable, so will wait until we upgrade to the .NET 6.
- This replaces `dotnetapp` @ https://hub.docker.com/_/microsoft-dotnet-samples

New experience:

image
 
And `dotnet run` on Windows:

image

Current experience:

image

### DIFF
--- a/samples/dotnetapp/Program.cs
+++ b/samples/dotnetapp/Program.cs
@@ -66,8 +66,11 @@ if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) &&
     string usageBytes = File.ReadAllLines("/sys/fs/cgroup/memory/memory.usage_in_bytes")[0];
     string limitBytes = File.ReadAllLines("/sys/fs/cgroup/memory/memory.limit_in_bytes")[0];
 
-    long.TryParse(usageBytes, out long usage);
-    long.TryParse(limitBytes, out long limit);
+    if (!long.TryParse(usageBytes, out long usage) ||
+        !long.TryParse(limitBytes, out long limit))
+    {
+        return;
+    }
 
     // above this size is unlikely to be an intentionally constrained cgroup
     if (limit < 10 * gibi)
--- a/samples/dotnetapp/Program.cs
+++ b/samples/dotnetapp/Program.cs
@@ -1,64 +1,58 @@
 using System;
+using System.IO;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using static System.Console;
+using static System.IO.File;
 
-public static class Program
+// Variant of https://github.com/dotnet/core/tree/main/samples/dotnet-runtimeinfo
+
+WriteLine("\r\n" +
+"         42\r\n" +
+"         42              ,d                             ,d\r\n" +  
+"         42              42                             42\r\n" +
+" ,adPPYb,42  ,adPPYba, MM42MMM 8b,dPPYba,   ,adPPYba, MM42MMM\r\n" +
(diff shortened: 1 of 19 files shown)


samples/dotnetapp/Program.cs
```diff
@@ -1,64 +1,95 @@
 using System;
+using System.IO;
+using System.Reflection;
```
MichaelSimons, line 3: Is `System.Reflection` used?

samples/dotnetapp/Program.cs
```diff
@@ -1,64 +1,95 @@
...
+    if (int.TryParse(cpuquota, out int quota) && quota > 0)
+    {
+        WriteLine($"cfs_quota_us: {quota}");
     }
 
-    private static string GetBot() 
+    string usageBytes = File.ReadAllLines("/sys/fs/cgroup/memory/memory.usage_in_bytes")[0];
+    string limitBytes = File.ReadAllLines("/sys/fs/cgroup/memory/memory.limit_in_bytes")[0];
+
+    long.TryParse(usageBytes, out long usage);
```
MichaelSimons, line 68: If the parsing fails, wouldn't it be better to shortcut the subsequent display logic?

richlander: It kinda already does that. The longs will all be 0 and will no-op. Seemed reasonable to me.

richlander: I fixed it.

MichaelSimons: Isn't that assuming if one fails, both fail?  Also note this introduces a CA warning.  It would be nice to have our samples be clean.

MichaelSimons (approved):

samples/dotnetapp/Program.cs
```diff
@@ -1,64 +1,97 @@
...
+WriteLine($"{nameof(Environment.ProcessorCount)}: {Environment.ProcessorCount}");
 
-          WriteLine();
-          WriteLine($"      {message}{GetBot()}");
-          
-          WriteLine("Environment:");
-          WriteLine($"{RuntimeInformation.FrameworkDescription} ({RuntimeInformation.ProcessArchitecture})");
-          WriteLine(RuntimeInformation.OSDescription);
+long mebi = 1048576;
+long gibi = mebi * 1024;
```
mthalman, line 53:
```suggestion
const long Gibi = mebi * 1024;
```

samples/dotnetapp/Program.cs
```diff
@@ -1,64 +1,97 @@
...
-    {
-          var defaultMessage = "Hello from .NET!";
-          var message = args is object && args.Length > 0 ? string.Join(' ', args): defaultMessage;
+  string prettyname = "PRETTY_NAME";
+  foreach(string line in File.ReadAllLines(osrel))
+  {
+      if (line.StartsWith(prettyname))
+      {
+          ReadOnlySpan<char> value = line.AsSpan()[(prettyname.Length + 2)..^1];
+          Console.WriteLine(value.ToString());
```
mthalman, line 36:
```suggestion
          WriteLine(value.ToString());
```

mthalman (approved):

mthalman (commented):
This will fail compilation.  Need to update the references to those new variable names too.

richlander:
Yup. I pulled locally immediately and noticed that. I also wanted to make another change (remove magic numbers).

PR #2705, opened 2021-03-30
Update complexapp with multi-targeting

richlander:
Managed to mess up the TFMs (multiple times). As an aside, I tried to use this and it didn't work: https://github.com/dotnet/designs/blob/main/accepted/2020/or-greater-defines/or-greater-defines.md.


samples/complexapp/README.md
```diff
@@ -45,107 +35,95 @@ There are two primary ways to test within the workflow of an application contain
...
 
-> Note: See [Establishing docker environment](../establishing-docker-environment.md) for more information on correctly configuring Dockerfiles and `docker build` commands.
-
 ## Running tests as an opt-in stage
 
-There are multiple approaches for testing with containers, such as the `ENTRYPOINT` of an opt-in stage (covered in this section) or as part of `docker build` (covered in the next section). The opt-in test stage approach covered in this section is the recommended approach because it is more flexible.
+The benefit of using an opt-in stage for testing is that it enables using the same environment as the build, allows volume mounting (which isn't possible with `docker build`) to collect test logs, and is opt-in so you to skip running tests if you don't want to pay that cost. The downside is that you need to orchestrate building and testing with some form of script. One `docker` command cannot build and test with this model.
```
mthalman, line 40:
```suggestion
The benefit of using an opt-in stage for testing is that it enables using the same environment as the build, allows volume mounting (which isn't possible with `docker build`) to collect test logs, and is opt-in so you can skip running tests if you don't want to pay that cost. The downside is that you need to orchestrate building and testing with some form of script. One `docker` command cannot build and test with this model.
```

samples/complexapp/README.md
```diff
@@ -45,107 +35,95 @@ There are two primary ways to test within the workflow of an application contain
...
 
-The primary benefit of using an opt-in stage for testing is that it enables using the same environment as the build as an opt-in scenario and allows volume mounting (which isn't possible with `docker build`) to collect test logs.
-
-The [Dockerfile](Dockerfile) includes a `test` stage that demonstrates running via its `ENTRYPOINT`, as follows.
+The [Dockerfile](Dockerfile) includes two `test` stages that demonstrates running tests via an`ENTRYPOINT`, as follows.
```
mthalman, line 42:
```suggestion
The [Dockerfile](Dockerfile) includes two `test` stages that demonstrates running tests via an `ENTRYPOINT`, as follows.
```

samples/complexapp/README.md
```diff
@@ -45,107 +35,95 @@ There are two primary ways to test within the workflow of an application contain
...
+Test Run Successful.
+Total tests: 2
+     Passed: 2
+ Total time: 0.7896 Seconds
+$ ls TestResults/
+_512cd4682573_2021-03-30_18_47_24.trx  _eb8b8c6b50ca_2021-03-30_18_48_30.trx
 ```
 
-### Windows using Windows containers
+To sum up this option, if you wanted to build an image, and then test if for multiple target frameworks, your script would look the following, presumably including error checking after each command. If your tests were not multi-targeted, then the last two commands could be removed.
```
mthalman, line 104:
```suggestion
To sum up this option, if you wanted to build an image, and then test it for multiple target frameworks, your script would look the following, presumably including error checking after each command. If your tests were not multi-targeted, then the last two commands could be removed.
```

samples/complexapp/README.md
```diff
@@ -45,107 +35,95 @@ There are two primary ways to test within the workflow of an application contain
...
 
-The following example uses PowerShell.
-
-```console
-docker build --pull --target test -t complexapp-test .
-docker run --rm -v ${pwd}\TestResults:c:\source\tests\TestResults complexapp-test
+```bash
+docker build --pull -t complexapp .
+docker build --pull -t complexapp:test --target test .
```
mthalman, line 108: You probably wouldn't want to use the `--pull` option for the two test runs.  You'd want to use the cache to ensure they are testing from the same base image.

samples/complexapp/libfoo/libfoo.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
```
mthalman, line 4: Use `netstandard2.1` instead of `net5.0`?

samples/complexapp/README.md
richlander: Good call.

samples/complexapp/libfoo/libfoo.csproj
richlander: Fair enough. Will fix that.

samples/complexapp/README.md
richlander: Actually, it is needed for two of three, so it probably makes sense as-is.

samples/complexapp/README.md
```diff
@@ -45,107 +35,95 @@ There are two primary ways to test within the workflow of an application contain
...
+It is possible to run tests as part of `docker build`. This approach can be useful if you want `docker build` to fail if your tests fail.
 
 ```Dockerfile
 WORKDIR /source/tests
 RUN dotnet restore
 RUN dotnet test --no-restore --logger:trx
 ```
 
-The following example demonstrates this pattern, as if it was part of [Dockerfile](Dockerfile). You would then be able run tests using the normal `docker build` pattern, as follows:
+Note: Multi-targeted tests present a problem with this scenario since it isn't possibly to use two .NET SDK images in the same `Dockerfiile`, as was done in the previous section. With this model, you either need to not multi-target tests, run the tests with the matching target framework as the .NET SDK image supports, or install a second runtime manually.
```
MichaelSimons, line 124: > since it isn't possibly to use two .NET SDK images in the same `Dockerfiile`

As worded, I don't agree with this statement.  You can use an arbitrary number of SDK images if you utilize a multi-stage Dockerfile.  You can't use multiple SDK images in the same Dockerfile stage.

samples/complexapp/Dockerfile
```diff
@@ -15,15 +15,25 @@ COPY libbar/ libbar/
 WORKDIR /source/complexapp
 RUN dotnet build -c release --no-restore
 
-# test stage -- exposes optional entrypoint
+# test stage for .NET 5.0 -- exposes optional entrypoint
 # target entrypoint with: docker build --target test
```
MichaelSimons, line 19: To give the Dockerfile symmetry, What do you think about naming the first test stage `test50`?
```suggestion
# target entrypoint with: docker build --target test50
```

MichaelSimons (commented):
I like the proof that the scenario works but I have mixed feelings about incorporating this into the sample for the following reasons.

1. It adds complexity to the samples for a non-mainline scenario.  I wonder if a single purpose illustration wouldn't serve the reader better - this may be in the form of a blog, one of our [scenarios](https://github.com/dotnet/dotnet-docker/tree/main/documentation/scenarios), or a more standalone sample.
2. The scenario feels like it tries to be realistic yet doesn't feel realistic.  If you have a Dockerfile that produces an application for a specific target framework, what is the value in running the tests for a different target framework?

samples/complexapp/README.md
richlander: It is poorly worded. The text `as was done in the previous section` is the key point. My assumption was that it wasn't possibly to having multiple test stage in a row because stages only get run/used if they are referenced from (I assume) a single inheritance chain of stages, come from the last stage. I suppose you could also do this, but it wouldn't necessarily be convenient/optimal. Fair?

richlander:
I agree with you that this feels very odd. I was struck with that at the start. I'm happy to create a new sample that targets just this use case (one MT'd library with tests). Sounds good? What I did would be easy to re-apply.

MichaelSimons:
> I agree with you that this feels very odd. I was struck with that at the start. I'm happy to create a new sample that targets just this use case (one MT'd library with tests). Sounds good? What I did would be easy to re-apply.

I prefer that approach.

PR #2710, opened 2021-03-31
Update SDK image first run handling

richlander:
We reviewed the first run experience of our SDK Dockerfiles. Each one of them has a call to `dotnet help` to ensure that the first run experience is run in the Dockerfile instead of on first (which is every) run of the container.

We decided to take a two step approach:

1. Use SDK ENVs to describe the intent of the state we want.
1. Ask the SDK team for a more explicit `dotnet init` command to better support this scenario.

This PR is step 1.


MichaelSimons (commented):
The change look good.  I have a couple requests.

1. Can you make this PR against nightly?  We should not be making this change in main without a tag change even though 6.0 is in preview.
2. Can update the SDK ENV tests to ensure the variables and values to not regress - https://github.com/dotnet/dotnet-docker/blob/main/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs#L50

mthalman (approved):
@richlander - Can you update the description of this PR so that we have a record of the background and reasons for these changes?

richlander:
Good call. Done.

PR #2713, opened 2021-04-01
Add multi-targeted library testing sample

richlander:
This sample demonstrates one way of using containers to test a multi-targeted library with multiple .NET versions. I'm pretty happy with the outcome.

This sample was built as one possible answer to this request: https://github.com/microsoft/dotnet-framework-docker/issues/731

Few bits of context:
- I found opportunities to use [OR_GREATER preprocessor symbols](https://github.com/dotnet/designs/blob/main/accepted/2020/or-greater-defines/or-greater-defines.md)
- The library I wrote is a bit more real world than my last attempts.
- The code I made conditional is less realistic since it is supported since .NET Core 3.0. I tried to make the two targets .NET Core 2.1 and .NET 5.0. I ran into two technical and one policy problem with that approach.
- I couldn't build the test (even with `-f netcoreapp2.1`) in the .NET Core 2.1 SDK because it rejected building a library that targeted .NET Standard 2.1. That was annoying.
- I also found that the .NET 5 SDK doesn't support the `_OR_GREATER` directives for .NET Standards. I had to target the nightly .NET 6 SDK to get that to work. https://github.com/dotnet/sdk/issues/16699
- It doesn't make sense to target .NET Core 2.1 since we've already [announced it is going out of support](https://devblogs.microsoft.com/dotnet/net-core-2-1-will-reach-end-of-support-on-august-21-2021/).

In short, I think this is a good model. We can update the actual sample if we'd like.

I didn't add any tests for this sample. I wanted to get validation on the approach before investing any more time.

More scenarios to consider:

- Testing on multiple distros (for the same .NET version).
- Testing a trimmed library in the context of a sample app.


samples/multitargetedlibrary/README.md
```diff
@@ -0,0 +1,73 @@
+# Testing a multi-targeted library in containers
+
+Containers are a convenient system for testing. They enable reliably and cheaply establishing the desired test environment. [Building and testing multiple projects with Docker](../complexapp/README.md) describes how to run unit tests in the same environment as the overall application build. This document describes how to establish multiple different .NET environments in containers to validate the various targets of a multi-targeted library.
+
+The primary mechanism for testing a [multi-targeted library](MovingAverage) is creating, targeting, and running multiple test stages that rely on different versions of the .NET SDK.
+
+## Copying source
+
+The source is copied into a `scratch` based stage. The advantage of this approach is that it doesn't depend on a specific image that may not be the right one.
```
mthalman, line 9: This requires more explanation because I don't understand it.  Is the reason for doing this to encapsulate the WORKDIR and COPY instructions in a common stage that can be reused by the other test stages?  If so, I'm not sure this stage really buys you much. You end up needing to use the two instructions (WORKDIR and COPY) in the test stages anyway. I would suggest removing this stage to not overly complicate things.

samples/multitargetedlibrary/MovingAverage/MovingAverage.cs
```diff
@@ -0,0 +1,54 @@
...
+            _sum += value;
+            _values[_index++] = value;
+            _average = _sum / _windowSize;
+            return _average;
+        }
+
+        public int GetAverage() => _average;
+
+#if NET5_0_OR_GREATER
+        public async IAsyncEnumerable<int> GetAveragesAsync(IAsyncEnumerable<int> source)
```
mthalman, line 45: Having a `Get` method that has side-effects seems strange to me, especially considering that `GetAverage` has no side-effects.  This could maybe just be resolved by naming it something like `AddRangeAsync`.

samples/multitargetedlibrary/README.md
richlander: Try removing it and see what happens. You'll realize that the value is exactly what that sentence says. Your comments on WORKDIR and COPY are a red-herring.

samples/multitargetedlibrary/MovingAverage/MovingAverage.cs
richlander: Is there a design guideline against this? That's only for a "getter", which this is not.

mthalman: There is no design guideline for it but I feel it's internally inconsistent within the sample.  There's an `Add` method that changes the state and returns the average result.  To me, it would make sense to translate that to the async enumerable case as well with a method named `AddRangeAsync` that changes the state and returns the average enumerable result.

samples/multitargetedlibrary/README.md
richlander: I'm wrong. Will fix this.

MichaelSimons: I don't mind this approach, it is especially convenient if you are coping in an explicit set of projects and dependencies.  That being said, this approach does not work on Windows - `scratch` images are not supported.

MichaelSimons:
I'm not seeing that building and running unit tests in a Dockerfile as a common practice.  It may be useful for a simple one or two project app but not in a larger multi-project system.  Rather as mentioned in the linked [issue](https://github.com/microsoft/dotnet-framework-docker/issues/731), devs are utilizing the container constructs CI systems like CircleCI, GH Actions, etc have for running steps or  their entire pipelines within a container.  

We have talked about this in the past, I think we should consider updating our samples to showcase some of the container flows within pipelines - e.g. GH Actions.  In this case showing how to build and test a multi-target library in a single pipeline would arguably be more valuable.

richlander:
We decided we're going to do this differently.

PR #2731, opened 2021-04-09
Update README.md to match the actual Dockerfile

CholoTook:
This is pedantic, but the actual dockerfile has this ENTRYPOINT. I know they are equivalent... perhaps it's better to keep them different for educational purposes, but I figure they should be the same?

Thanks

### DIFF
--- a/samples/dotnetapp/README.md
+++ b/samples/dotnetapp/README.md
@@ -49,7 +49,7 @@ RUN dotnet publish -c release -o /app --no-restore
 FROM mcr.microsoft.com/dotnet/runtime:5.0
 WORKDIR /app
 COPY --from=build /app .
-ENTRYPOINT ["./dotnetapp"]
+ENTRYPOINT ["dotnet", "dotnetapp.dll"]
 ```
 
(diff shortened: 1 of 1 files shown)


CholoTook:
Hmm.. I just noticed some of your OS-specific Dockerfiles use different ENTRYPOINTS too...

```
grep -h ENTRYPOINT Dockerfile* | sort | uniq -c
     11 ENTRYPOINT ["./dotnetapp"]
      4 ENTRYPOINT ["dotnetapp"]
      1 ENTRYPOINT ["dotnet", "dotnetapp.dll"]
```

I know this is pedantic, but as a noob  I was trying to understand things by looking at differences, and the difference in entrypoints was initially confusing.

mthalman:
Thanks for making this correction to the documentation.  You're right that this should be consistent with the actual Dockerfile.

As far as the other ENTRYPOINT usages you've found, those differences are intentional.

* `ENTRYPOINT ["./dotnetapp"]`: Will only work on Linux because the actual executable's file name is literally `dotnetapp` whereas in Windows it's `dotnetapp.exe`.  Because the path is specified, Windows won't do the implicit execution like it does for `dotnetapp` => `dotnetapp.exe`.
* `ENTRYPOINT ["dotnetapp"]`: Will only work on Windows because Windows doesn't require the path syntax; it'll just default to the current directory.
* `ENTRYPOINT ["dotnet", "dotnetapp.dll"]`: Works on either Windows or Linux but only for framework-dependent deployments where the dotnet CLI is known to exist.

CholoTook:
Ah, thanks, I should have guessed there was a good reason.

PR #2754, opened 2021-04-20
Add missing parts of dotnet sdk decompress step

gregoral:
I noticed issues when I tried to use the provided docker file with dotnet 6 preview.
Some parts of the .NET SDK were not installed.

After the change the install completes without issues and docker image is built as expected.

I tested with:
DOTNET_SDK_VERSION=6.0.100-preview.3.21202.5

The dotnet SDK is unpacked, but cannot be executed because the correct version of dotnet does not get installed.

This error message is shown instead:
```
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '6.0.0-preview.3.21201.4' was not found.
  - The following frameworks were found:
        5.0.5 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.
```
This message is of course not the best, since we are trying to install the SDK.

### DIFF
--- a/eng/dockerfile-templates/sdk/6.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/sdk/6.0/Dockerfile.alpine
@@ -31,7 +31,7 @@ RUN wget -O dotnet.tar.gz https://dotnetcli.azureedge.net/dotnet/Sdk/$DOTNET_SDK
     && dotnet_sha512='{{VARIABLES[cat("sdk|6.0|linux-musl|", ARCH_SHORT, "|sha")]}}' \
     && echo "$dotnet_sha512  dotnet.tar.gz" | sha512sum -c - \
     && mkdir -p /usr/share/dotnet \
-    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz ./packs ./sdk ./sdk-manifests ./templates ./LICENSE.txt ./ThirdPartyNotices.txt \
+    && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz ./dotnet ./host ./packs ./sdk ./sdk-manifests ./shared ./templates ./LICENSE.txt ./ThirdPartyNotices.txt \
     && rm dotnet.tar.gz \
     # Trigger first run experience by running arbitrary cmd
     && dotnet help{{if ARCH_VERSIONED = "amd64":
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/dotnet-docker?pullRequest=2754) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/dotnet-docker?pullRequest=2754) before we can accept your contribution.<br/><br/>:x: gregoral [sign now](https://cla.dotnetfoundation.org/dotnet/dotnet-docker?pullRequest=2754)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/dotnet-docker?pullRequest=2754) it.</sub>

mthalman:
This change should not be necessary.  Which Dockerfile did you attempt to build and what command did you use to build it?

The reason it's not necessary is because the `sdk` image is based on the `aspnet` image and the `aspnet` image provides these runtime-related folders that you've explicitly added.  So what is actually happening with your change is that the `sdk` Docker build is overwriting these folders that already existed.  

```console
$ docker run --rm mcr.microsoft.com/dotnet/aspnet:6.0-alpine3.13-amd64 ls /usr/share/dotnet
LICENSE.txt
ThirdPartyNotices.txt
dotnet
host
shared
```

This is why it's important to know which Dockerfile you're targeting.  Perhaps you're attempting to use the Dockerfile from the `nightly` branch?  In that case, the `REPO` arg would need to be overridden to point to the `nightly` repo:

```
docker build --build-arg REPO=mcr.microsoft.com/dotnet/nightly/aspnet .
```

gregoral:
Hi, sorry if I misunderstood how the whole stack is layered.
I was using: [alpine3.13 / amd64 / Dockerfile](https://github.com/dotnet/dotnet-docker/blob/main/src/sdk/5.0/alpine3.13/amd64/Dockerfile)
And changed the .net sdk to the latest preview, which is: 6.0.100-preview.3.21202.5
I guess the proper way would be to change the base image too.

I have already tested it and you are indeed correct.
With mcr.microsoft.com/dotnet/aspnet:6.0.0-preview.3-alpine3.13-amd64 it works as expected.

Sorry for the confusion.
This pull request can be cancelled.

I did also try it out with the "vanilla" alpine linux docker image: alpine:3.13.4 from the official docker repository.
In that case the base image will not have the runtime-related folders and these folders have to be added.

Maybe there could be a check that verifies that the base image has the correct version of the required runtime-related folders.

mthalman:
> Maybe there could be a check that verifies that the base image has the correct version of the required runtime-related folders.

That's what this error is saying:

```
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '6.0.0-preview.3.21201.4' was not found.
  - The following frameworks were found:
        5.0.5 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```

PR #2760, opened 2021-04-22
Update issue templates for .NET Monitor release

mthalman:


.github/ISSUE_TEMPLATE/releases/dotnet-monitor-release.md
```diff
@@ -0,0 +1,26 @@
...
+      - [ ] `dotnet run --project .\eng\update-dependencies\update-dependencies.csproj <major/minor version> --product-version monitor=<version> --channel-name <channel>`
+1. - [ ] Inspect generated changes for correctness
+1. - [ ] Commit generated changes
+1. - [ ] Create PR
+1. - [ ] Get PR signoff
+1. - [ ] Merge PR
+1. - [ ] Wait for changes to be mirrored to internal [dotnet-docker repo](https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet-docker) (internal MSFT link)
+1.  - [ ] Build images - Queue build stage of [dotnet-docker pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=373) (internal MSFT link) with variables:
+
+      `imageBuilder.pathArgs: --path 'src/*/3.1/alpine3.13/amd64' --path 'src/*/5.0/alpine3.13/amd64' --path 'src/*/6.0/alpine3.13/amd64' --path 'src/monitor/*'`
```
jander-msft, line 24: Some of these paths are not necessary for dotnet-monitor, namely the 5.0 and 6.0 paths. The tool only uses the 3.1/alpine3.13/amd64 image.

jander-msft (approved):

.github/ISSUE_TEMPLATE/releases/dotnet-monitor-release.md
mthalman: Understood.  The reason 5.0 and 6.0 need to be included is because of the layer sharing.  The `runtime-deps` image for 3.1 is the same image as 5.0 and 6.0 `runtime-deps`.  So if we end up needing to rebuild that image, then we have to rebuild the whole hierarchy (`runtime`, `aspnet`, `sdk`).  It's this big chain of dependencies that we need to account for.  Whether anything actually _does_ get rebuilt is based on whether a Dockerfile had changed or a newer base image is available.

.github/ISSUE_TEMPLATE/releases/dotnet-monitor-release.md
```diff
@@ -0,0 +1,26 @@
...
+1. - [ ] Confirm images have been ingested by MCR
```
MichaelSimons, line 25: Is this necessary given the build step that blocks on ingestion?

MichaelSimons (approved):

.github/ISSUE_TEMPLATE/releases/dotnet-monitor-release.md
mthalman: True, it's implicit in the build succeeding.  I'll remove it here and from the others.

MichaelSimons (approved):

PR #2810, opened 2021-05-13
Use scale set agent pools for internal builds

mthalman:
Rolling out the changes to consume scale set agent pools for internal builds in the nightly branch for testing.

Related to https://github.com/dotnet/docker-tools/issues/532

### DIFF
--- a/eng/common/templates/stages/build-test-publish-repo.yml
+++ b/eng/common/templates/stages/build-test-publish-repo.yml
@@ -91,8 +91,7 @@ stages:
         ${{ if eq(variables['System.TeamProject'], 'public') }}:
           name: Docker-1809-Public
         ${{ if eq(variables['System.TeamProject'], 'internal') }}:
-          name: DotNetCore-Docker
-          demands: VSTS_OS -equals Windows_Server_2019_Data_Center_RS5
+          name: Docker-1809-Internal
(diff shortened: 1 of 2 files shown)


eng/common/templates/stages/build-test-publish-repo.yml
```diff
@@ -105,8 +104,7 @@ stages:
         ${{ if eq(variables['System.TeamProject'], 'public') }}:
           name: Docker-2004-Public
         ${{ if eq(variables['System.TeamProject'], 'internal') }}:
-          name: DotNetCore-Docker
-          demands: VSTS_OS -equals Windows-ServerDatacenter-2004
+          name: Docker-2004-Internal
```
MichaelSimons, line 107: Instead of utilizing two conditions here would the following work?

`name: ${{ join('Docker-2004-', variables['System.TeamProject']) }}`

MichaelSimons (approved):
LGTM, I left one suggestion that might simplify the logic.

eng/common/templates/stages/build-test-publish-repo.yml
mthalman: Great suggestion.  It's even simpler than that.  Updated.

MichaelSimons: Nice - looks great.

Issue #2830, opened 2021-05-25
.NET Release - 5.0 SDK


mthalman:
# .NET Release

## Release Versions

_The set of .NET versions that are being released as a unit._

* 5.0.300

## 1. Main Branch Tasks

1. - [x] Merge appropriate commits from nightly branch.  Things to check:
      - [ ] New/EOL distro
      - [ ] Infrastructure changes
      - [ ] PowerShell version
      - [ ] Check for additional changes by diffing the main and nightly branches
      - [ ] &lt;add link to PR/commit&gt;
1. - [x] Check whether a change has been made to a Dockerfile that is shared by multiple .NET versions. If a change has been made and the .NET versions that share that file are not being released at the same time, define a separate Dockerfile to isolate the change to the .NET version that's being released. Conversely, after a shared Dockerfile has diverged in such a way, it should be combined again into a shared Dockerfile when the other other .NET version is released. Shared Dockerfiles to check:
      - [ ] 3.1 runtime-deps shared with 5.0 and 6.0
1. - [x] Wait for .NET archive files (.zip, .tar.gz) to be available at blob storage location
1. - [x] Run `update-dependencies` tool to update all the necessary files to reflect the specified .NET versions (run this command for each version being released):
      - [ ] `dotnet run --project .\eng\update-dependencies\update-dependencies.csproj 5.0 --product-version sdk=5.0.300`
1. - [x] Inspect generated changes for correctness
1. - [x] Commit generated changes
1. - [x] Create PR
1. - [x] Get PR signoff
1. - [x] Merge PR
1. - [x] Wait for changes to be mirrored to internal [dotnet-docker repo](https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet-docker) (internal MSFT link)
1. - [x] Build images - Queue build stage of [dotnet-docker pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=373) (internal MSFT link) with variables:

      All releases:

          stages: build

      Servicing release (exclude 5.0/6.0 if separate runtime-deps Dockerfiles were defined):

          noCache: true
          imageBuilder.pathArgs: --path 'src/*/3.1/*' --path 'src/*/5.0/*' --path 'src/*/6.0/*' --path 'src/monitor/*'

1. - [x] Wait for NuGet packages to be published during release tic-toc
[...]

PR #2832, opened 2021-05-25
Merge main to nightly

mthalman:
Related to https://github.com/dotnet/dotnet-docker/issues/2830


src/runtime-deps/6.0/alpine3.13/arm32v7/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM arm32v7/alpine:3.13
```
MichaelSimons, line 1: Is this intentional?  It seems like we should be sharing the 5.0 Dockerfile.

mthalman: Good catch.  Fixed

MichaelSimons (approved):

Issue #2814, opened 2021-05-18
Please provide an image for ASP.NET Core with IIS and ANCM installed


richlander:
I was talking to a customer today. They wanted an image with with ASP.NET Core ready for IIS hosting. They tried to install the ASP.NET Core 3.1 hosting bundle (MSI) installed and that failed.

Proposal:

- For 6.0, enable IIS and install ANCM for the ServerCore image.
- For 3.1, ensure that IIS, ANCM and .NET Core 3.1 can be installed on a Windows Server Core 2019 image. If not, fix whatever the problem is and share the working Dockerfile.

MichaelSimons:
[Triage] Assigning to @richlander, to investigate the 3.1 issue.

richlander:
Filed an issue https://github.com/dotnet/aspnetcore/issues/32955.

mthalman:
Added a sample Dockerfile that demonstrates installing IIS and ANCM: https://github.com/dotnet/dotnet-docker/pull/2908. This is all we're planning to do for now.  Providing this in the aspnet image is undesirable due to the size concerns with IIS (see https://github.com/dotnet/dotnet-docker/pull/2908#issuecomment-871740780).

PR #2868, opened 2021-06-08
Add prototype of ANCM

richlander:
Context: https://github.com/dotnet/dotnet-docker/issues/2814

This is a WIP to determine a plan. This model works.

Open issues:

- Should the `aspnet` Windows Server Core image be configured to use IIS, just like the .NET Framework ASP.NET one is? That's the model I started with, not a suggestion I prefer it.
- Perhaps we should add IIS and `servicemonitor` to the `aspnet` image and leave the IIS configuration to the app? That puts IIS and Kestrel on equal footing. There's asymmetry with the .NET Framework image but that's probably a fine tradeoff. That seems like the right model to me.
- We'll need to ensure that we can download ANCM by version like we do ASP.NET Core.
- Are we getting the latest `servicemonitor`? It's a hard-coded value (both here and in other Dockerfiles.
- The original plan was to only enable this for 6.0. That still seems like the right model. People can copy what we've done for 3.1.

Notes:
- I realize that these changes should be made in a template (I assume). I made these changes to get feedback.
- I assumed that the two RUN statements should be coalesced. That can be fixed later.
- The mis-match between 5.0 and 6.0 and the use of `DOTNET_ROLL_FORWARD` is just to make the sample work w/o changing it.
- I just noticed now that the file names I used are 2.1.x while the ANCM is 6.0. I just copied [this code](https://github.com/microsoft/iis-docker/blob/master/samples/aspnetcoremodule/Dockerfile) and made the minimum changes. That needs to be changed.

Relevant .NET Framework assets:

- https://github.com/microsoft/dotnet-framework-docker/blob/main/src/aspnet/4.8/windowsservercore-ltsc2019/Dockerfile
- https://github.com/microsoft/dotnet-framework-docker/blob/main/samples/aspnetapp/Dockerfile

/cc @shirhatti @BrennanConroy


src/aspnet/6.0/windowsservercore-ltsc2019/amd64/Dockerfile
```diff
@@ -10,16 +10,37 @@ ENV `
...
+    }; `
+    `
+    tar -C $Env:ProgramFiles\dotnet -oxzf aspnetcore.zip ./shared/Microsoft.AspNetCore.App; `
+    Remove-Item -Force aspnetcore.zip
+
+RUN powershell -Command `
+    $ErrorActionPreference = 'Stop'; `
+    $ProgressPreference = 'SilentlyContinue'; `
+    New-Item -Type Directory C:\Setup | powershell -Command Out-Null ; `
+    Invoke-WebRequest 'https://download.visualstudio.microsoft.com/download/pr/2728666c-860b-4a78-ba42-8ec7b2167d42/72b9c82be95a013e2c15a1fc182e2fc4/dotnet-hosting-6.0.0-preview.4.21253.5-win.exe' -UserAgent '' -OutFile C:/setup/dotnet-hosting-2.1.3-win.exe ; `
```
mthalman, line 36: This would instead be accessed from a URL like https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/dotnet-hosting-$Env:ASPNET_VERSION-win.exe

src/aspnet/6.0/windowsservercore-ltsc2019/amd64/Dockerfile
```diff
@@ -10,16 +10,37 @@ ENV `
...
+    $process = start-process -Filepath C:/setup/dotnet-hosting-2.1.3-win.exe -ArgumentList  @('/install', '/q', '/norestart', 'OPT_NO_RUNTIME=1', 'OPT_NO_SHAREDFX=1') -Wait -PassThru ; `
+    `
+    if ($process.ExitCode -ne 0) { `
+      exit $process.ExitCode ; `
+    } `
+    Remove-Item -Force C:/setup/dotnet-hosting-2.1.3-win.exe
+
+EXPOSE 80
+
+ENTRYPOINT ["C:\\ServiceMonitor.exe", "w3svc"]
```
mthalman, line 46: This is what I'm most unsure about.  I understand that in non-IIS cases, the app's Dockerfile would override this ENTRYPOINT.  I would consider that to be the typical case.  And it just feels weird for the typical case to be overriding something.  If someone attempts to run the aspnet image directly, what is the experience?  Are they going to be guided to thinking that IIS is the way to go?

mthalman:
> Are we getting the latest servicemonitor? It's a hard-coded value (both here and in other Dockerfiles.

Yes.  There hasn't exactly been a lot of churn on that front: https://github.com/microsoft/IIS.ServiceMonitor/releases

src/aspnet/6.0/windowsservercore-ltsc2019/amd64/Dockerfile
richlander: This is a specific design question called out up top. I come to the same conclusion as you and am asking if people agree.

PR #2903, opened 2021-06-17
Updated 6.0 images to use latest PowerShell Preview build 7.2.0-preview.7

anmenaga:
This PR is a result of running `dotnet run --project .\eng\update-dependencies\ -- 6.0 --product-version powershell=7.2.0-preview.7 --compute-shas` with Docker in Linux container mode.


mthalman:
@anmenaga - Can you manually update the checksum value for PowerShell.Linux.Alpine?  Looks like there's a bug in update-dependencies that caused it to not get updated.

mthalman (approved):

PR #2908, opened 2021-06-19
Add IIS Support to Windows Server Core images for .NET 6

richlander:
This PR replaces https://github.com/dotnet/dotnet-docker/pull/2868.

- This one adds support via Docker templates.
- I didn't know how to make the SHA checks work (since this change switches files) so I just deleted that. I'll need help on that.
- I added a new IIS sample and added a Windows Server Core entry to the ASP.NET Core README.

I wanted to add a new row in table the ASP.NET Core sample that demonstrated which server was being used. From what I can tell, that's not possible. Maybe there is an ASP.NET Core API to do that but I didn't spend any time looking for that.


eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
```diff
@@ -13,13 +13,16 @@ RUN powershell -Command `
...
+        Invoke-WebRequest -OutFile C:\ServiceMonitor.exe -Uri https://dotnetbinaries.blob.core.windows.net/servicemonitor/2.0.1.10/ServiceMonitor.exe; `
         # Install ASP.NET Core Runtime
-        Invoke-WebRequest -OutFile aspnetcore.zip https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/aspnetcore-runtime-$Env:ASPNET_VERSION-win-x64.zip; `
-        $aspnetcore_sha512 = '{{VARIABLES["aspnet|6.0|win|x64|sha"]}}'; `
-        if ((Get-FileHash aspnetcore.zip -Algorithm sha512).Hash -ne $aspnetcore_sha512) { `
-            Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
-            exit 1; `
-        }; `
+        Invoke-WebRequest -OutFile c:\aspnetcore-hosting.exe https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/dotnet-hosting-$Env:ASPNET_VERSION-win.exe; `
+        $process = start-process -Filepath C:\aspnetcore-hosting.exe -ArgumentList  @('/install', '/q', '/norestart', 'OPT_NO_RUNTIME=1') -Wait -PassThru ; `
```
mthalman, line 23:
```suggestion
        $process = Start-Process -Filepath C:\aspnetcore-hosting.exe -ArgumentList  @('/install', '/q', '/norestart', 'OPT_NO_RUNTIME=1') -Wait -PassThru ; `
```

mthalman, line 23: Do we really not need to explicitly install ASP.NET Core runtime?  Is this being implicitly done by the the dotnet hosting exe?  What about the `OPT_NO_RUNTIME` flag?

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
```diff
@@ -13,13 +13,16 @@ RUN powershell -Command `
...
+        # Acquire ServiceMonitor
+        Invoke-WebRequest -OutFile C:\ServiceMonitor.exe -Uri https://dotnetbinaries.blob.core.windows.net/servicemonitor/2.0.1.10/ServiceMonitor.exe; `
         # Install ASP.NET Core Runtime
-        Invoke-WebRequest -OutFile aspnetcore.zip https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/aspnetcore-runtime-$Env:ASPNET_VERSION-win-x64.zip; `
-        $aspnetcore_sha512 = '{{VARIABLES["aspnet|6.0|win|x64|sha"]}}'; `
-        if ((Get-FileHash aspnetcore.zip -Algorithm sha512).Hash -ne $aspnetcore_sha512) { `
-            Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
-            exit 1; `
-        }; `
+        Invoke-WebRequest -OutFile c:\aspnetcore-hosting.exe https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/dotnet-hosting-$Env:ASPNET_VERSION-win.exe; `
```
mthalman, line 22: We're going to need to do checksum validation here on the downloaded exe.  This requires a bunch of changes.

## Update manifest.versions.json file

We'll need a new entry defined here:

https://github.com/dotnet/dotnet-docker/blob/524a594332b60b88d2736077afeb152cfab0dde2/manifest.versions.json#L26-L35

I would propose something like:

```json
"aspnet-dotnet-hosting|6.0|windows|x64|sha": "<checksum>"
```

## Update update-dependencies tool

That will then require a change to the update-dependencies tool to populate the checksum value at https://github.com/dotnet/dotnet-docker/blob/524a594332b60b88d2736077afeb152cfab0dde2/eng/update-dependencies/DockerfileShaUpdater.cs#L39-L67

Proposed value for that is:

```csharp
{"aspnet-dotnet-hosting", new string[] { $"{DotnetBaseUrl}/aspnetcore/Runtime/$VERSION_DIR/dotnet-hosting-$VERSION_FILE-$OS.exe" }
```

## Update script

And then the script which invokes update-dependencies needs to be updated to include this new entry:

https://github.com/dotnet/dotnet-docker/blob/524a594332b60b88d2736077afeb152cfab0dde2/eng/Set-DotnetVersions.ps1#L42

It should be changed to 

```
$updateDepsArgs += @("--product-version", "aspnet=$AspnetVersion", "--product-version", "aspnet-runtime-targeting-pack=$AspnetVersion", "--product-version", "aspnet-dotnet-hosting=$AspnetVersion")
```

## Consume checksum in template

Finally the Dockerfile needs to validate the checksum:

```Dockerfile
        $dotnet_hosting_sha512 = '{{VARIABLES["aspnet-dotnet-hosting|6.0|win|x64|sha"]}}'; `
        if ((Get-FileHash aspnetcore-hosting.exe -Algorithm sha512).Hash -ne $dotnet_hosting_sha512) { `
            Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
            exit 1; `
        }; `
```

mthalman (commented):
I also see that the tests are failing for this change.  The test fails when attempting to connect to an ASP.NET container that uses this base image.  It runs the command `docker run --name 6.0-app-run-132687101791254395 --rm -d -t -p 80 6.0-build-132687101214184088 dotnet run --urls http://0.0.0.0:80` and fails with `403 - Forbidden: Access is denied`.

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
richlander: This configuration installs ASP.NET Core. I validated that. By default, the installer installs ACM, .NET Runtime and ASP.NET Core Runtime. All other configurations are opt-out.

I removed `OPT_NO_RUNTIME` and then the .NET Core Runtime was installed.

richlander:
Is it possible for you to finish this off @mthalman, pretty please? The rest of the PR is all stuff that I don't know anything about and will be very slow at addressing. The main thing I wanted to do was prove out the model, and that's now been accomplished.

mthalman:
@richlander - Ok, because you said pretty please, I guess I can do it.

richlander:
Thanks, Matt. You are the best!

mthalman:
The test failure is very interesting.  It appears that the container is crashing as a result of the test logic making a request before the service is ready for it.  It first yields the 403 response. A second attempt yields a 503 response and then the container crashes.  By putting in a 10 second delay (for good measure) before attempting the first request, the test passes.  What I don't understand is why these changes cause this difference in behavior.  In theory, the app configuration used for the test is no different than prior to these changes.

richlander:
@mthalman -- That's probably doing to servicemonitor.exe. Do you not see that with .NET Framework images?

mthalman:
But servicemonitor.exe isn't even being executed in this context.  It exists in the image, but that's it.  And now I can't even repro it anymore on my machine any more.  It was reproing and now it's not.  🤦‍♂️

richlander:
OK. I was wondering if you were just talking about the base image vs the sample. Makes sense. Very odd!

mthalman:
I made the following updates: 
* Added checksum validation
* When installing dotnet-hosting, opted out of x86 installation
* Need to disable w3svc service. Otherwise IIS will start when container starts, causing race condition between accepting of requests between IIS and Kestrel.  Plus, it's unneeded by default.

richlander:
Do you think that this will make the July release?

mthalman:
> Do you think that this will make the July release?

I would think so.  I realized that @MichaelSimons wasn't declared as a review so I'll want to get his review on this before merging.

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
```diff
@@ -13,13 +13,28 @@ RUN powershell -Command `
...
+        # Install .NET Core Hosting bundle and ASP.NET Core Runtime
+        Invoke-WebRequest -OutFile c:\aspnetcore-hosting.exe https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/dotnet-hosting-$Env:ASPNET_VERSION-win.exe; `
+        $dotnet_hosting_sha512 = '{{VARIABLES["aspnet-dotnet-hosting|6.0|win|x64|sha"]}}'; `
+        if ((Get-FileHash aspnetcore-hosting.exe -Algorithm sha512).Hash -ne $dotnet_hosting_sha512) { `
+            Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
+            exit 1; `
+        }; `
+        $process = Start-Process -Filepath C:\aspnetcore-hosting.exe -ArgumentList  @('/install', '/q', '/norestart', 'OPT_NO_RUNTIME=1', 'OPT_NO_X86=1') -Wait -PassThru ; `
+        if ($process.ExitCode -ne 0) { `
+            exit $process.ExitCode ; `
```
MichaelSimons, line 38:
```suggestion
            exit $process.ExitCode; `
```

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
```diff
@@ -13,13 +13,28 @@ RUN powershell -Command `
...
-        tar -C $Env:ProgramFiles\dotnet -oxzf aspnetcore.zip ./shared/Microsoft.AspNetCore.App; `
-        Remove-Item -Force aspnetcore.zip
+        # Install .NET Core Hosting bundle and ASP.NET Core Runtime
+        Invoke-WebRequest -OutFile c:\aspnetcore-hosting.exe https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/dotnet-hosting-$Env:ASPNET_VERSION-win.exe; `
+        $dotnet_hosting_sha512 = '{{VARIABLES["aspnet-dotnet-hosting|6.0|win|x64|sha"]}}'; `
+        if ((Get-FileHash aspnetcore-hosting.exe -Algorithm sha512).Hash -ne $dotnet_hosting_sha512) { `
+            Write-Host 'CHECKSUM VERIFICATION FAILED!'; `
+            exit 1; `
+        }; `
+        $process = Start-Process -Filepath C:\aspnetcore-hosting.exe -ArgumentList  @('/install', '/q', '/norestart', 'OPT_NO_RUNTIME=1', 'OPT_NO_X86=1') -Wait -PassThru ; `
```
MichaelSimons, line 36: Does this leave any log files that should be cleaned up?

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
```diff
@@ -13,13 +13,28 @@ RUN powershell -Command `
         $ErrorActionPreference = 'Stop'; `
         $ProgressPreference = 'SilentlyContinue'; `
         `
-        # Install ASP.NET Core Runtime
-        Invoke-WebRequest -OutFile aspnetcore.zip https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/aspnetcore-runtime-$Env:ASPNET_VERSION-win-x64.zip; `
-        $aspnetcore_sha512 = '{{VARIABLES["aspnet|6.0|win|x64|sha"]}}'; `
-        if ((Get-FileHash aspnetcore.zip -Algorithm sha512).Hash -ne $aspnetcore_sha512) { `
+         # Install IIS and disable it for startup
```
MichaelSimons, line 16: What is the size impact of these changes?  How much bigger is the image?

src/aspnet/6.0/windowsservercore-ltsc2019/amd64/Dockerfile
```diff
@@ -13,13 +13,28 @@ RUN powershell -Command `
         $ErrorActionPreference = 'Stop'; `
         $ProgressPreference = 'SilentlyContinue'; `
         `
-        # Install ASP.NET Core Runtime
-        Invoke-WebRequest -OutFile aspnetcore.zip https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$Env:ASPNET_VERSION/aspnetcore-runtime-$Env:ASPNET_VERSION-win-x64.zip; `
-        $aspnetcore_sha512 = 'aa10762145e5db33db667d4c563b7a0098ebe95b744add06686879172d7e2e3796fad0fbae9d631fed2019e882798bc6f5ab064ab6ba3342082ca4eafc2256c4'; `
-        if ((Get-FileHash aspnetcore.zip -Algorithm sha512).Hash -ne $aspnetcore_sha512) { `
+         # Install IIS and disable it for startup
```
MichaelSimons, line 16: I see value in adding a unit test that uses IIS to validate the installation is correct.

samples/aspnetapp/Dockerfile.windowsservercore-iis-x64
```diff
@@ -0,0 +1,30 @@
+# https://hub.docker.com/_/microsoft-dotnet
+FROM mcr.microsoft.com/dotnet/sdk:6.0-windowsservercore-ltsc2019 AS build
```
MichaelSimons, line 2: It's strange IMO that this sample is using 6.0 images when the rest don't.

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
mthalman: As a whole, these changes add about 217 MB.  The vast majority of that, ~200 MB, is from IIS.  This changes the image size from roughly 5.7 GB to 5.9 GB.

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
mthalman: Yes, just a few in the temp folder.  I'll get those cleaned up.

samples/aspnetapp/Dockerfile.windowsservercore-iis-x64
mthalman: Agreed.  But it's only temporary until 6.0 GAs, just a few months.

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
MichaelSimons: Granted the base image is huge, but that aside, this is a non-trivial addition.  How many users will want to use IIS over Kestrel?  Is this something better provided by an image variant or a scenario doc?

samples/aspnetapp/Dockerfile.windowsservercore-iis-x64
MichaelSimons: IMO the samples should be on the same version.  Given this isn't supported in the production images, my preference would be to track this with an issue and added at RTM.

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
richlander: We could go that path. Certainly, we've had very little call for this. That said, it's a relatively small % change for one of our key scenarios on Windows. It's also not an easy/intuitive component to install. If we were going to go that direction, I'd request that the ASP.NET team create an distribution with only ANCM to make it easier.

samples/aspnetapp/Dockerfile.windowsservercore-iis-x64
richlander: How about we target moving all the samples to .NET 6 for Preview 6.

MichaelSimons: I can get onboard with that.

eng/dockerfile-templates/aspnet/6.0/Dockerfile.windowsservercore
MichaelSimons: The runtime and aspnetcore .NET layers combined are ~90 MB.  I lean towards not adding to the size problem ServerCore based images have if this is not considered the mainline use case.

richlander: Sure, but what's the alternative. My issue is that the ANCM installer is difficult to use. It is easy to make your images worse if done incorrectly. See this conversation for evidence of that: https://github.com/dotnet/dotnet-docker/pull/2908#discussion_r655655667.

MichaelSimons: Add [scenario documentation](https://github.com/dotnet/dotnet-docker/tree/main/documentation/scenarios) for this.  Concurrently get the ASP.NET team create an distribution with only ANCM to make it easier.

mthalman:
Because of the [amount of size](https://github.com/dotnet/dotnet-docker/pull/2908#discussion_r660972228) that IIS adds to the image, we've decided to not add IIS to the aspnet image at this time.  As a consequence, ServiceMonitor and the ASP.NET Core Module won't be added either.  Instead, this pattern of installation will be demonstrated and tested in the sample Dockerfile.

eng/Set-DotnetVersions.ps1
```diff
@@ -39,7 +39,7 @@ if ($SdkVersion) {
 }
 
 if ($AspnetVersion) {
-    $updateDepsArgs += @("--product-version", "aspnet=$AspnetVersion", "--product-version", "aspnet-runtime-targeting-pack=$AspnetVersion")
+    $updateDepsArgs += @("--product-version", "aspnet=$AspnetVersion", "--product-version", "aspnet-runtime-targeting-pack=$AspnetVersion", "--product-version")
```
MichaelSimons, line 42: What is the trailing `--product-version` needed for?

samples/aspnetapp/Dockerfile.windowsservercore-iis-x64
```diff
@@ -0,0 +1,49 @@
...
+
+# copy everything else and build app
+COPY aspnetapp/. ./aspnetapp/
+WORKDIR /source/aspnetapp
+RUN dotnet publish -c release -o /app --no-restore
+
+# final stage/image
+FROM mcr.microsoft.com/dotnet/aspnet:6.0-windowsservercore-ltsc2019
+
+WORKDIR /inetpub/wwwroot
```
MichaelSimons, line 20: Thoughts on moving this after the IIS setup just before the COPY?  I think it would help may the COPY more clear.

MichaelSimons (approved):
LGTM

eng/Set-DotnetVersions.ps1
mthalman: Just a mistake.  Deleted.

PR #2944, opened 2021-07-02
Add full version number to base tags

### LINKED ISSUE #2885: Use 3-part version tag in FROM instruction
For all Dockerfiles which have a .NET image tag referenced in a `FROM` instruction, the referenced tag should use the 3-part version number instead of the 2-part number.  This maintains coherency when attempting to build older Dockerfiles whose versions are not currently the latest anymore.

For example, the current version for .NET 5 is 5.0.7.  If I grab the [aspnet Dockerfile for 5.0.6](https://github.com/dotnet/dotnet-docker/blob/5d9e2099db7d73b97a9008ec948dcade7b34904b/src/aspnet/5.0/buster-slim/amd64/Dockerfile) and attempt to build it with nothing cached on my machine, it will pull the latest runtime version, 5.0.7, because it's referencing the tag with the 2-part version number tag, a floating tag which gets updated with each servicing release to point to the latest version:

https://github.com/dotnet/dotnet-docker/blob/5d9e2099db7d73b97a9008ec948dcade7b34904b/src/aspnet/5.0/buster-slim/amd64/Dockerfile#L17

This causes an image to be built that contains incompatible runtimes, 5.0.6 .NET runtime and 5.0.7 ASP.NET Core.

Instead, to always guarantee coherency, such Dockerfiles should reference the matching 3-part version for which the Dockerfile represents.  In this case, the 5.0.6 aspnet Dockerfile should reference the 5.0.6 runtime tag: `FROM $REPO:5.0.6-buster-slim-amd64`.

Inspired by https://github.com/dotnet/dotnet-docker/issues/2843

mthalman:
Fixes #2885


MichaelSimons (commented):
Why aren't the 6.0 Dockerfiles being updated to use this new scheme?

It's unfortunate the infrastructure is limiting the way the Dockerfiles are authored.  Given the choice to inline version within the `FROM` tag reference, it makes me question continuing to use the `DOTNET_VERSION` `ARG` given it is not consistency used throughout the Dockerfile.  Inlining the version makes the Dockerfile easier to read in a way but it makes it less re-useable for the cases where someone wants to copy the Dockerfile and customize it including changing the version.  But these changes erode that value hence the question.

mthalman:
> Why aren't the 6.0 Dockerfiles being updated to use this new scheme?

I was going to say because 6.0 has a version number that includes the build number (e.g. `6.0.0-preview.5.21301.5`) which there is no tag for. But then I realized that the tags are based off the product-version variable, not the build-version variable.  So I should really be referencing the product-version variable in these FROM instructions.

mthalman:
> It's unfortunate the infrastructure is limiting the way the Dockerfiles are authored. Given the choice to inline version within the `FROM` tag reference, it makes me question continuing to use the `DOTNET_VERSION` `ARG` given it is not consistency used throughout the Dockerfile. Inlining the version makes the Dockerfile easier to read in a way but it makes it less re-useable for the cases where someone wants to copy the Dockerfile and customize it including changing the version. But these changes erode that value hence the question.

My [previous comment](https://github.com/dotnet/dotnet-docker/pull/2944#issuecomment-874972346) basically answers this. They don't technically have to be the same value so it wouldn't be correct to define as single ARG.

MichaelSimons:
> > It's unfortunate the infrastructure is limiting the way the Dockerfiles are authored. Given the choice to inline version within the `FROM` tag reference, it makes me question continuing to use the `DOTNET_VERSION` `ARG` given it is not consistency used throughout the Dockerfile. Inlining the version makes the Dockerfile easier to read in a way but it makes it less re-useable for the cases where someone wants to copy the Dockerfile and customize it including changing the version. But these changes erode that value hence the question.
> 
> My [previous comment](https://github.com/dotnet/dotnet-docker/pull/2944#issuecomment-874972346) basically answers this. They don't technically have to be the same value so it wouldn't be correct to define as single ARG.

For release versions won't they always be the same?  If so this doesn't change my mind.  When customers look at these Dockerfiles, they will not understand why the version ARG is not referenced in all places.  Because of this, it question if the ARG should be used everywhere or nowhere.

mthalman:
> For release versions won't they always be the same? If so this doesn't change my mind. When customers look at these Dockerfiles, they will not understand why the version ARG is not referenced in all places. Because of this, it question if the ARG should be used everywhere or nowhere.

I've updated things to conform with this.

eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
```diff
@@ -1,7 +1,7 @@
 # escape=`
 
 ARG REPO=mcr.microsoft.com/dotnet/runtime
-FROM $REPO:5.0-{{OS_VERSION}}
+FROM $REPO:{{VARIABLES["dotnet|5.0|product-version"]}}-{{OS_VERSION}}
```
MichaelSimons, line 4: It feels a bit inconsistent in a way that some Dockerfiles are still using the $ENV in the download urls.  These Dockerfiles still have the version declared multiple places.

MichaelSimons (commented):
In the event that there needs to be a one-off servicing event such as the NuGet patch issue earlier this year, will the templates support a `-1` tagging scheme?  Maybe it's not necessary in that type of event.  We could disable the templates and manually update the necessary Dockerfiles.  It feels like a worthwhile thought exercise however as when these events happen, we need to be able to react quickly.

eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
mthalman: I don't think it's fair to say that this usage is inconsistent with other Dockerfiles that have a fundamentally different pattern for producing the image (i.e. multi-stage vs single stage).  The fact that they're different only stems from that fact.  I think it's perfectly reasonable, and expected, to take advantage of the assets available within that pattern.  Making use of an ENV, that is required to exist, is different IMO than making use of an ARG solely for the purposes of reusability.

mthalman:
> In the event that there needs to be a one-off servicing event such as the NuGet patch issue earlier this year, will the templates support a `-1` tagging scheme? Maybe it's not necessary in that type of event. We could disable the templates and manually update the necessary Dockerfiles. It feels like a worthwhile thought exercise however as when these events happen, we need to be able to react quickly.

Yes, I think things would work very well in that case.  I think the [PowerShell update](https://github.com/dotnet/dotnet-docker/pull/2516) is a better example of this scenario.  We would just update the `product-version` variable in the manifest and that would flow out to all the necessary Dockerfiles.

eng/dockerfile-templates/aspnet/5.0/Dockerfile.windowsservercore
MichaelSimons: >I don't think it's fair to say that this usage is inconsistent with other Dockerfiles

I understand how the Dockerfile pattern is different in respect to mulit-stage vs single stage.  The part that felt inconsistent to me was that the single stage is making use of the ENV where possible, yet it is not the single place where the version is defined.  I thought this was the pattern was either the version should be defined in a single place or it should be inlined everyplace where used.  If I understand, your preference is to utilize the ENV if defined, else inline the version, correct?

mthalman: > If I understand, your preference is to utilize the ENV if defined, else inline the version, correct?

Yes, I feel like we should take advantage of the ENV when it exists.

src/runtime/6.0/focal/amd64/Dockerfile
```diff
@@ -1,12 +1,11 @@
 ARG REPO=mcr.microsoft.com/dotnet/runtime-deps
-ARG DOTNET_VERSION=6.0.0-preview.6.21352.12
 
 # Installer image
 FROM amd64/buildpack-deps:focal-curl as installer
 ARG DOTNET_VERSION
```
MichaelSimons, line 5: This should be removed.

MichaelSimons (approved):

PR #2997, opened 2021-07-28
Consume RPM files for 3.1 and 5.0 on CBL-Mariner

### LINKED ISSUE #2943: Consume RPM packages for CBL-Mariner in 3.1 and 5.0
Currently the 3.1 and 5.0 Dockerfiles for [CBL-Mariner](https://github.com/microsoft/CBL-Mariner) do not install .NET bits from RPM packages but rather directly from tarballs. This was due to the unavailability of the runtime-deps RPM package for CBL-Mariner.  These packages will be available for the next servicing release of .NET.  At that point, they can be installed in the Dockerfiles instead of using the tarballs.

Related to https://github.com/dotnet/dotnet-docker/issues/2837

mthalman:
* Updates Dockerfiles to install RPM package files rather than extract from tarball
* Remove the workaround for https://github.com/dotnet/runtime/issues/53543 since it's been fixed
* Update tests to reflect that 3.1 and 5.0 images for CBL-Mariner now install from package files

Fixes #2943


eng/dockerfile-templates/runtime-deps/5.0/Dockerfile.mariner
```diff
@@ -0,0 +1,28 @@
+FROM cblmariner.azurecr.io/base/core:{{OS_VERSION_NUMBER}}
+
+RUN tdnf install -y \
```
MichaelSimons, line 3: Am I remembering correctly that mariner already had an issue to support file based installation with tdnf that would be able to automatically install dependencies within the native package feed?  We should open an issue in our repo tracking this so that we can cleanup/simplify the pattern here.

MichaelSimons (approved):
LGTM

eng/dockerfile-templates/runtime-deps/5.0/Dockerfile.mariner
mthalman: Yes, I've logged a tracking issue for it: https://github.com/dotnet/dotnet-docker/issues/3003

PR #3074, opened 2021-08-25
Ensure Alpine images have their packages upgraded

mthalman:
Ensures that packages from the base alpine:3.13 image, including apk-tools and libssl1.1, have been upgraded to latest available version.  This addresses the following CVEs:

* CVE-2021-36159
* CVE-2021-3711
* CVE-2021-3712


MichaelSimons (approved):

jackni:
I'm not sure why the commit has been reverted. as I can see the upgrade will address missing future package upgrade, the only issue from the old commit is missing -U param at the end, which will make sure all packages in the package list got updated before upgrade. 

Given this is a public repo, would that be safer or more defensive position then just rely on upstream?

PR #3088, opened 2021-08-30
Generalize naming of common pipeline-related artifacts

mthalman:
The common infrastructure located in the eng/common folder had names of projects, repositories, variables, branches, and paths that were specific to the .NET implementation.  In order to make the infrastructure suitable for general purpose, these names need to be generalized.  

* Defined parameters for internal/public AzDO project names.
* Defined parameters for a reference to the internal/public pipeline repository resource.  This allows the pipeline implementer to be able to declare their own repository resource that represents a versions repo and pass in its identifying name as a parameter to the build-test-publish-repo.yml template.
* Replace all usages of variables derived from secrets of the .NET key vault with generalized names (e.g. `app-dotnetdockerbuild-client-secret` => `acr.servicePrincipalPassword`).
* Defined a new `officialRepoPrefix` so that the pipeline logic can know whether a given `publishRepoPrefix` value represents an official repo location. 
* Defined a new `azdoVersionsRepoInfo.path` and `gitHubVersionsRepoInfo.path` variable to provide a way to specify the directory path where image info files are located in those respective repos.
* Added parameters allowing pipeline implementers to inject custom init steps and custom publish variables in order to customize the logic in the build and publish stages. Specifically, this allows .NET to customize the `publicSourceBranch` and `additionalPublishMcrDogsArgs` variables and reference a .NET-specific variable group in the publish stage.
* Create a new build-test-publish-repo.yml in the eng/pipelines/stages folder (outside of common) to act as a wrapper around the same-named template in the common folder. This allows for a shared implementation of many of the common settings that are used amongst the .NET pipelines.

Related to dotnet/docker-tools#843


dagood:
> Related to #843

Meant to be a link to https://github.com/dotnet/docker-tools/issues/843 (vs. `dotnet/dotnet-docker#843`)?

Is this PR in the right repo to change the common files? It does make sense to test the results in this repo though.

mthalman:
> > Related to #843
> 
> Meant to be a link to [dotnet/docker-tools#843](https://github.com/dotnet/docker-tools/issues/843) (vs. `dotnet/dotnet-docker#843`)?
> 
> Is this PR in the right repo to change the common files? It does make sense to test the results in this repo though.

Yes, thanks.  I've fixed the link.  I'm making these changes in the nightly branch here because I feel it gives better context to the changes to see what changes are necessary to the consuming pipelines which aren't in the docker-tools repo.

dagood (approved):
I haven't dug through the entire thing, but the bits being pulled out make sense to me. A few suggestions.

eng/pipelines/stages/build-test-publish-repo.yml
```diff
@@ -0,0 +1,30 @@
...
+      buildMatrixCustomBuildLegGroupArgs: --custom-build-leg-group pr-build --custom-build-leg-group test-dependencies
+    ${{ if not(parameters.isPullRequestPipeline) }}:
+      buildMatrixCustomBuildLegGroupArgs: --custom-build-leg-group test-dependencies
+    ${{ if eq(variables['System.TeamProject'], parameters.internalProjectName) }}:
+      customPublishVariables:
+      - group: DotNet-AllOrgs-Darc-Pats
+
+    # Template paths must be relative to the YAML job that executes them
+    customBuildInitSteps:
+    - template: ../../../pipelines/steps/set-public-source-branch-var.yml
```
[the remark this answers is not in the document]
mthalman: In normal situations, an absolute path would work.  The problem is in the context of which pipelines use this file and their use of repositories.  In the pipelines for internal builds, two repositories get checked out: the primary repo and the versions repo. For public builds, only the primary repo gets checked out.  (To keep this short, I won't go into the reason for that.)

Having multiple repositories get checked out versus only one impacts the value that is returned for the pre-defined `Build.SourcesDirectory` variable. That's the variable that would be used to establish the root of the absolute path. When just a single repository has been checked out by the pipeline, that variable gives you the path to the root of the repo. When two or more repositories are checked out, that variable gives you the base agent sources directory, the parent directory of all the repositories.

So depending on whether it's an internal or public build, the path to use is different. Using a relative path allows for a value that will work in either context.

In theory, you could use `if` statement to conditionally set the path depending on whether it was public or internal.  But that just adds further complexity to the YAML as well.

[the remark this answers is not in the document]
dagood: Wow, that's a bizarre dependency... thanks for the explanation.

mthalman:
One thing not being addressed in this PR is the usage of `DotNet` in the names for agent pools.  Fixing that will require more changes than I wanted to mix in here.

eng/pipelines/dotnet-core-size-validation.yml
```diff
@@ -13,3 +13,6 @@ variables:
 
 jobs:
 - template: ../common/templates/jobs/validate-image-sizes.yml
+  parameters:
+    internalProjectName:
+    publicProjectName:
```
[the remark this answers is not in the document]
mthalman: Nope, just an oversight.

eng/common/templates/jobs/build-images.yml
```diff
@@ -61,10 +65,12 @@ jobs:
...
   - ${{ parameters.customInitSteps }}
   - template: ../steps/set-image-info-path-var.yml
+    parameters:
+      publicSourceBranch: $(publicSourceBranch)
   - powershell: |
       $imageBuilderBuildArgs = "$(imageBuilder.queueArgs) --image-info-output-path $(artifactsPath)/$(legName)-image-info.json"
-      if ($env:SYSTEM_TEAMPROJECT -eq "internal") {
-        $imageBuilderBuildArgs = "$imageBuilderBuildArgs --registry-override $(acr.server) --repo-prefix $(stagingRepoPrefix) --source-repo-prefix $(mirrorRepoPrefix) --push --registry-creds ""$(acr.server)=$(acr.userName);$(BotAccount-dotnet-docker-acr-bot-password)"""
+      if ($env:SYSTEM_TEAMPROJECT -eq "${{ parameters.internalProjectName }}") {
+        $imageBuilderBuildArgs = "$imageBuilderBuildArgs --registry-override $(acr.server) --repo-prefix $(stagingRepoPrefix) --source-repo-prefix $(mirrorRepoPrefix) --push --registry-creds ""$(acr.server)=$(acr.userName);$(acr.password)"""
```
dagood, line 73: I think this fixes https://github.com/dotnet/docker-tools/issues/835 too. (Well, sort of: it opens up a workaround-seeming fix that looks good enough to me!) For an internal build that I want to prevent from pushing tags to my ACR (like a PR validation build), I can make the pipeline file leave `internalProjectName` as `null`, so this never evaluates to true. :+1:

dagood (approved):

MichaelSimons (approved):

eng/common/templates/steps/set-image-info-path-var.yml
```diff
@@ -1,19 +1,25 @@
...
-        $publicSourceBranch = "nightly"
+steps:
+- powershell: |
+    if ("$(System.TeamProject)" -eq "$(publicProjectName)") {
+        $basePath = "$(gitHubVersionsRepoInfo.path)"
     }
     else {
-        $publicSourceBranch = "main"
+        $basePath= "$(azdoVersionsRepoInfo.path)"
+    }
```
dagood, line 11: Looking at the file again, is this saying that internal builds can only publish image info to the AzDO/internal repo? Don't we want to publish it to the public dotnet/versions repo, as long as we aren't doing a truly private build? (We can't publish anywhere at all in public-project builds, right?)

mthalman: This template sets the variable for the purposes of _reading_ the image info file.  It's not relevant to the publishing of the image info file.  For publish, it publishes to both the public and internal repos. It pushes the commit to both to avoid latency issues (see https://github.com/dotnet/docker-tools/issues/642).  For public project builds, it runs with the dry-run option, so nothing actually gets published.

dagood: Ah, right... and I guess having them different is just an error right now (https://github.com/dotnet/docker-tools/blob/2dddfe26ef6faca5f110fe1f6c179e47a51f3433/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs#L35-L39) but this could be used in the future.

PR #3103, opened 2021-09-02
Update samples to .NET 6

### LINKED ISSUE #3076: Update samples to .NET 6
The next release of 6.0, RC1, is a go-live release.  Our samples should demonstrate 6.0 in action.

One of new additions to 6.0 is the availability of an Arm64 Alpine image for the SDK. This improves the UX for Arm64 users when using multi-arch tags.  By updating the samples, it resolves an issue with the Dockerfile.alpine-arm64 sample that causes it to fail on an Arm64 machine because it references the `sdk:5.0-alpine` tag but there is no Arm64-based image for that tag; but there will be one for 6.0.  Related to https://github.com/dotnet/dotnet-docker/issues/2139.

mthalman:
This also updates the ASP.NET Core sample to use the web minimal API. This was copied from the files that get generated by the web template in VS.

Fixes #3076


samples/aspnetapp/Dockerfile.alpine-x64-slim
```diff
@@ -5,7 +5,7 @@ WORKDIR /source
 # copy csproj and restore as distinct layers
 COPY *.sln .
 COPY aspnetapp/*.csproj ./aspnetapp/
-RUN dotnet restore -r linux-musl-x64
+RUN dotnet restore -r linux-musl-x64 /p:PublishReadyToRun=true
```
mthalman, line 8: This is needed because the PublishReadyToRun option now impacts which packages are used for 6.0 (see https://github.com/dotnet/sdk/issues/20701).

MichaelSimons (approved):

samples/aspnetapp/aspnetapp/Program.cs
```diff
@@ -1,26 +1,30 @@
...
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Hosting;
-using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.Hosting;
-using Microsoft.Extensions.Logging;
-
-namespace aspnetapp
+
```
WeihanLi, line 1: Maybe it's better to remove this empty line

Issue #3110, opened 2021-09-07
New Linux Distro Version - Debian 11 (Bullseye)


mthalman:
# New Linux Distro Version

Distro: Debian 11 (Bullseye)

## ImageBuilder Tasks

- [x] Ensure that the ImageBuilder supports the new distro version in the [code](https://github.com/dotnet/docker-tools/blob/main/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs) to generate the correct README display name from the version specified in the manifest

## Nightly Branch Tasks

1. - [x] Copy the Dockerfiles of the most recent version of that distro (or author new ones for an entirely new distro) and place them in a version-specific folder under their respective variants (runtime-deps, runtime, aspnet, sdk). Note: not all variants have a corresponding runtime-deps image.
1. - [x] Modify the Dockerfiles as appropriate for any specific changes related to the new distro version
1. - [x] Update [manifest.json](https://github.com/dotnet/dotnet-docker/blob/nightly/manifest.json) to reference the new set of Dockerfiles with the appropriate tags
      - [ ] Move any distro-specific floating tags to the newer version (e.g. `3.1-alpine`)
1. - [x] Update the [test data](https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/TestData.cs) to include the new distro version
1. - [x] Update the [tags metadata templates](https://github.com/dotnet/dotnet-docker/tree/main/eng/mcr-tags-metadata-templates) to include the new distro version
1. - [x] Run the command to update the READMEs: `.\eng\readme-templates\Get-GeneratedReadmes.ps1`
1. - [x] Run the command to update the image size baseline file: `.\tests\performance\Validate-ImageSize.ps1 -UpdateBaselines`
1. - [x] Inspect generated changes for correctness
1. - [x] Consider whether sample Dockerfiles should be authored if this is a new distro and them to the [samples](https://github.com/dotnet/dotnet-docker/tree/main/samples)
1. - [x] Run the command to build and test your changes: `.build-and-test.ps1 -OS <os>`
1. - [x] Commit generated changes
1. - [x] Create PR
1. - [x] Get PR signoff
1. - [x] Merge PR to nightly branch
1. - [x] Wait for automatically queued CI build to finish on [dotnet-docker-nightly pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=359) (internal MSFT link)
[...]

PR #3113, opened 2021-09-08
Add Debian Bullseye Dockerfiles for 3.1 and 5.0

mthalman:
Related to #3110


MichaelSimons (approved):
LGTM - One thing I noticed is that our documentation (https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-platforms.md and https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-tags.md) doesn't capture the behavior of the floating tags.  For example the behavior of the 5.0 and 5.0-alpine behave different as new versions of debian and alpine are supported.  It would be good to clarify this.

mthalman:
> LGTM - One thing I noticed is that our documentation (https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-platforms.md and https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-tags.md) doesn't capture the behavior of the floating tags. For example the behavior of the 5.0 and 5.0-alpine behave different as new versions of debian and alpine are supported. It would be good to clarify this.

I've logged https://github.com/dotnet/dotnet-docker/issues/3118 for this.

PR #3139, opened 2021-09-16
Default to non-root user and port 8080 on .NET 6.0 runtimes

### LINKED ISSUE #2249: Improve running .NET in containers as a non-root user
# Improve running .NET in containers as a non-root user
We see an increasing number of requests from users seeking guidance on running containers as non-root users.  A primary driver for running as non-root is related to reducing vulnerabilities.  [This issue](https://github.com/dotnet/dotnet-docker/issues/1757#issuecomment-604497687) discusses why we don't define a non-root user within the .NET images.  This issues explores and tracks ways we can enhance .NET to better support running in containers as a non-root user.

## .NET Docker
* [Include documentation and samples that demonstrate USER best practices](https://github.com/dotnet/dotnet-docker/issues/1772)

## Runtime
* [Kestrel does not support running as non-root on Alpine](https://github.com/dotnet/aspnetcore/issues/4699)

olljanat:
This PR includes hardenings instructed on https://techcommunity.microsoft.com/t5/azure-developer-community-blog/hardening-an-asp-net-container-running-on-kubernetes/ba-p/2542224 to .NET 6.0.0 runtimes and runtime-deps with one minor difference:
* I used UID 1000 and GID 3000 to match with Kyverno [Add Default securityContext](https://kyverno.io/policies/other/add-default-securitycontext/) policy.
~~* Using `DOTNET_` prefix on EnableDiagnostics environment variable like described on https://docs.microsoft.com/en-us/dotnet/core/run-time-config/debugging-profiling~~

~~File system permission changed should be also done to samples but that must be done after #3103 is merged.~~

Closes #2249

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/6.0/Dockerfile.mariner
+++ b/eng/dockerfile-templates/runtime-deps/6.0/Dockerfile.mariner
@@ -19,10 +19,25 @@ RUN dotnet_version={{VARIABLES["runtime|6.0|build-version"]}} \
     && dotnet_sha512='{{VARIABLES[cat("runtime-deps-cm.1|6.0|linux-rpm|", ARCH_SHORT, "|sha")]}}' \
     && echo "$dotnet_sha512  dotnet-runtime-deps.rpm" | sha512sum -c - \
     && rpm --install dotnet-runtime-deps.rpm \
-    && rm dotnet-runtime-deps.rpm
+    && rm dotnet-runtime-deps.rpm \
+    \
+    # Non-root user and group
+    && groupadd \
+        --gid 3000 \
+        dotnet \
+    && adduser \
+        --uid 1000 \
+        --gid 3000 \
+        --shell /usr/sbin/nologin \
+        --no-create-home \
+        --system \
+        "dotnet"
 
 ENV \
-    # Configure web servers to bind to port 80 when present
-    ASPNETCORE_URLS=http://+:80 \
+    # Configure web servers to bind to port 8080 when present
+    ASPNETCORE_URLS=http://+:8080 \
     # Enable detection of running in a container
     DOTNET_RUNNING_IN_CONTAINER=true
+
+# Run as non-root
+USER 1000:3000
--- a/eng/dockerfile-templates/aspnet/6.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/aspnet/6.0/Dockerfile.alpine
@@ -1,5 +1,6 @@
 ARG REPO=mcr.microsoft.com/dotnet/runtime
 FROM $REPO:{{VARIABLES["dotnet|6.0|product-version"]}}-{{OS_VERSION}}{{ARCH_TAG_SUFFIX}}
+USER root
 
 # .NET globalization APIs will use invariant mode by default because DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=true is set
 # by the base runtime-deps image. See https://aka.ms/dotnet-globalization-alpine-containers for more information.
@@ -16,3 +17,6 @@ RUN wget -O aspnetcore.tar.gz https://dotnetcli.azureedge.net/dotnet/aspnetcore/
     && echo "$aspnetcore_sha512  aspnetcore.tar.gz" | sha512sum -c - \
     && tar -ozxf aspnetcore.tar.gz -C /usr/share/dotnet ./shared/Microsoft.AspNetCore.App \
     && rm aspnetcore.tar.gz
+
+# Run as non-root
+USER 1000:3000
--- a/eng/dockerfile-templates/aspnet/6.0/Dockerfile.linux
+++ b/eng/dockerfile-templates/aspnet/6.0/Dockerfile.linux
@@ -1,4 +1,5 @@
 ARG REPO=mcr.microsoft.com/dotnet/runtime
+USER root
 
 # Installer image
 FROM {{ARCH_VERSIONED}}/buildpack-deps:{{OS_VERSION_BASE}}-curl as installer
@@ -21,3 +22,6 @@ ENV \
     Logging__Console__FormatterName=Json
 
 COPY --from=installer ["/shared/Microsoft.AspNetCore.App", "/usr/share/dotnet/shared/Microsoft.AspNetCore.App"]
+
+# Run as non-root
+USER 1000:3000
--- a/eng/dockerfile-templates/aspnet/6.0/Dockerfile.mariner
+++ b/eng/dockerfile-templates/aspnet/6.0/Dockerfile.mariner
@@ -1,5 +1,6 @@
 ARG REPO=mcr.microsoft.com/dotnet/runtime
 FROM $REPO:{{VARIABLES["dotnet|6.0|product-version"]}}-{{OS_VERSION}}{{ARCH_TAG_SUFFIX}}
+USER root
 
 ENV \
     # ASP.NET Core version
@@ -13,3 +14,6 @@ RUN curl -SL --output aspnetcore.rpm https://dotnetcli.azureedge.net/dotnet/aspn
     && echo "$dotnet_sha512  aspnetcore.rpm" | sha512sum -c - \
     && rpm --install aspnetcore.rpm \
     && rm aspnetcore.rpm
+
+# Run as non-root
+USER 1000:3000
--- a/eng/dockerfile-templates/runtime/6.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/runtime/6.0/Dockerfile.alpine
@@ -14,3 +14,6 @@ RUN wget -O dotnet.tar.gz https://dotnetcli.azureedge.net/dotnet/Runtime/$DOTNET
     && tar -C /usr/share/dotnet -oxzf dotnet.tar.gz \
     && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet \
     && rm dotnet.tar.gz
+
+# Run as non-root
+USER 1000:3000
--- a/eng/dockerfile-templates/runtime/6.0/Dockerfile.linux
+++ b/eng/dockerfile-templates/runtime/6.0/Dockerfile.linux
@@ -20,3 +20,6 @@ ENV DOTNET_VERSION={{VARIABLES["runtime|6.0|build-version"]}}
 COPY --from=installer ["/dotnet", "/usr/share/dotnet"]
 
 RUN ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+
+# Run as non-root
+USER 1000:3000
--- a/eng/dockerfile-templates/runtime/6.0/Dockerfile.mariner
+++ b/eng/dockerfile-templates/runtime/6.0/Dockerfile.mariner
@@ -18,3 +18,6 @@ RUN curl -SL --output dotnet-host.rpm https://dotnetcli.azureedge.net/dotnet/Run
     \
     && rpm --install dotnet-host.rpm dotnet-hostfxr.rpm dotnet-runtime.rpm \
     && rm dotnet-host.rpm dotnet-hostfxr.rpm dotnet-runtime.rpm
+
+# Run as non-root
+USER 1000:3000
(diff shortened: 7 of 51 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/dotnet-docker?pullRequest=3139) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/dotnet-docker?pullRequest=3139) before we can accept your contribution.<br/><br/>:x: olljanat [sign now](https://cla.dotnetfoundation.org/dotnet/dotnet-docker?pullRequest=3139)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/dotnet-docker?pullRequest=3139) it.</sub>

olljanat:
@mthalman well we have need for this on our env anyway and there looks to be very nice testing pattern here so I would like to see that those tests can pass without root user.

olljanat:
Ok. Now it passed on "Linux_amd64 6.0-alpine3.14" so I will leave this PR to this state waiting for result from that discussion.

@mthalman as addon to that discussion. Microsoft own best practice guide for AKS on https://docs.microsoft.com/en-us/azure/aks/developer-best-practices-pod-security very clearly says:
> Best practice guidance - To run as a different user or group and limit access to the underlying node processes and services, define pod security context settings. **Assign the least number of privileges required.**
>
> For your applications to run correctly, pods should run as a defined user or group and **not as root**. The securityContext for a pod or container lets you define settings such as runAsUser or fsGroup to assume the appropriate permissions. Only assign the required user or group permissions, and don't use the security context as a means to assume additional permissions. The runAsUser, privilege escalation, and other Linux capabilities settings are only available on Linux nodes and pods.

I see if very confusing and even indifference of customers environments security if Microsoft will decide that this cannot be fixed now to .NET 6.0 when it is still on release candidate phase.

After all this is not even breaking change. Users can anytime switch back to old behavior by instructing their containers orchestration platform to start those containers as a `root`

Also the way how change is done now on this PR is that it only updates `mcr.microsoft.com/dotnet/aspnet` and `mcr.microsoft.com/dotnet/runtime` images. Development image `mcr.microsoft.com/dotnet/sdk` will still run as a `root`

eng/dockerfile-templates/runtime-deps/6.0/Dockerfile.mariner
```diff
@@ -19,10 +19,24 @@ RUN dotnet_version={{VARIABLES["runtime|6.0|build-version"]}} \
...
 
 ENV \
-    # Configure web servers to bind to port 80 when present
-    ASPNETCORE_URLS=http://+:80 \
+    # Configure web servers to bind to port 8080 when present
+    ASPNETCORE_URLS=http://+:8080 \
     # Enable detection of running in a container
     DOTNET_RUNNING_IN_CONTAINER=true
+    # Disable diagnostics
+    ENV DOTNET_EnableDiagnostics=0
```
shirhatti, line 42: 👀 @dotnet/dotnet-diag

jander-msft: 👀 @dotnet/dotnet-monitor

jander-msft, line 42: If setting `DOTNET_EnableDiagnostics=0` is accepted, this will come at a great detriment to the debuggability and diagnosability of containers by default. Tools such as debuggers and tools from dotnet/diagnostics and dotnet/dotnet-monitor will effectively be unusable on containers running these images.

I would propose to not accept this change, especially without clear documentation and remediation steps.

hoyosjs: Why are we trying to disable the diagnostic experience by default on our containers. This image affects SDK images which are often used in developer scenarios, where you want to be able to attach a debugger as needed. Also this means disabling eventing and profiling mechanisms that are widely used and designed for production scenarios. Not running as root is something I can get behind - it's detrimental to security in the event the container is not started with namespace isolation and some other vulnerability-prone configurations. However, disabling diagnostics is and should not be our default experience - if our customers decide to do so, let them set the variable on container startup or create a container based on this and know the tradeoff they are deciding on.

hoyosjs (changes requested):
See comment on diagnostic pipeline disable.

eng/dockerfile-templates/runtime-deps/6.0/Dockerfile.mariner
olljanat: It was picked from that linked blog article. I can drop that change from this PR but probably it should be investigated if diagnostics can be modified on way that it works with `readOnlyRootFilesystem: true` ?

richlander:
I'm not in favor of these changes. The issue is that they are biased to benefiting one group of users and (as a result) harm another set. These images are relatively low-level and are intended to work equally well for a variety of scenarios.

The best example is package installation. Most users probably don't need to install packages (which is great), however, a bunch do. It would be unfortunate if those users needed to change the user back to root in order to install packages. There are other root scenarios, but installing packages (or otherwise installing software to a root location) is the main one. For users like this that want a non-root user, they'd need to change the user twice instead of once.

There is an easy workaround to this. You can change the user as the first line after your FROM statements. That seems to be a great and workable solution that avoids the challenge I just raised.

> I see if very confusing and even indifference of customers environments security if Microsoft will decide that this cannot be fixed now to .NET 6.0 when it is still on release candidate phase.

The nuance that is missing is that this guidance is for apps, less about platforms. Users have 100% control to control various settings for apps. I'm unaware of any other developer platforms making this change. Are you aware of any?

We care deeply about security-oriented topics. That said, we also care deeply about defining where a given change should be made, in our images or in user images. We've discussed this topic many times and continue to believe that it should be resolved in user images. Do you have any thoughts on why that model would be a problem?

olljanat:
> The best example is package installation. Most users probably don't need to install packages (which is great), however, a bunch do. It would be unfortunate if those users needed to change the user back to root in order to install packages. There are other root scenarios, but installing packages (or otherwise installing software to a root location) is the main one. For users like this that want a non-root user, they'd need to change the user twice instead of once.

Reason why I needed to add those switch back to root user statements to current draft is that these images are made on way that SDK images (`mcr.microsoft.com/dotnet/sdk`) are actually build to top of runtime images (`mcr.microsoft.com/dotnet/runtime`) and runtime images are build to top of runtime-deps (`mcr.microsoft.com/dotnet/runtime-deps`) images. If we want to avoid that then we probably should be refactor these on way that there would be one more layer like build-deps or common-deps and runtime/sdk images would be build based on those.

Alternatively adding this feature to buildkit https://github.com/moby/buildkit/issues/2263 would fix it as it is actually `docker build` command/Dockerfile limitation.

> The nuance that is missing is that this guidance is for apps, less about platforms. Users have 100% control to control various settings for apps. I'm unaware of any other developer platforms making this change. Are you aware of any?

Depends on what you mean with "developer platform"? There is very big bunch of ready made Docker images for different tools available on https://bitnami.com which are running with non-root user by default.

> That said, we also care deeply about defining where a given change should be made, in our images or in user images. We've discussed this topic many times and continue to believe that it should be resolved in user images. Do you have any thoughts on why that model would be a problem?

On year 2021 one of the big things which everyone is talking about is Zero Trust Architecture. Running *runtime* version of ASP.NET Core containers with non-root priciples is of course only small part of it but IMO important still. Also people are lazy, if everything works without problems how many really will do security hardening for their images?

Same time there is guidances like this https://docs.microsoft.com/en-us/dotnet/architecture/containerized-lifecycle/design-develop-containerized-apps/build-aspnet-core-applications-linux-containers-aks-kubernetes about how you can easily deploy application which are available from internet to top of AKS.

@richlander also it sounds that we understand these terms differently. For me *runtime* means something which is supposed to be production ready and *sdk* the stuff what developers are using but it sounds that you are thinking both of them as developer tools?

richlander:
> it sounds that you are thinking both of them as developer tools?

No. Here's an example Dockerfile: https://github.com/jellyfin/jellyfin/blob/488e276f01b34c9aa82b53c969c56ccc9647cfbe/Dockerfile#L33. It installs packages in its "runtime layer". This particular app is self-contained, but it is an example of the scenario I was talking about. These APT commands require root permissions.

> There is very big bunch of ready made Docker images for different tools available on https://bitnami.com which are running with non-root user by default.

That's great. That's a curated and opinionated experience. I was asking if any actual dev-platform provider, like .NET, was providing images as non-root. I'm assuming that none of them are. We see our images as unopinionated.

We could in theory provide root and non-root variants for the runtime images, but that would double the number of runtime images we produce.  We don't want to do that, particularly when users can just change the user as the first line of their Dockerfile or maintain a layer (that simply changes that user) for all their apps.

The problem is that this change makes these images both better and worse. We all understand the better part. I'd like to see you address the challenges that come with this change. For example, what's the guidance you would write for those scenarios?

eng/dockerfile-templates/runtime-deps/6.0/Dockerfile.mariner
hoyosjs: I hadn't seen the article, but there's several misconceptions in it:

> `And that is because you should disable some debugging telemetry setting in the Dockerfile by adding an extra ENV statement: ENV COMPlus_EnableDiagnostics=0`

We don't send telemetry as suggested - rather what this is doing is disabling all communication mechanisms for diagnostic purposes. Without this, a large portion of things like profilers, crash dump collection on demand, debugging, and eventing won't work at all. They use the file system because they expose a Unix Domain Socket/pipes for tools to communicate with the app. `readOnlyRootFilesystem` is good for some things, overly zealous for others. One thing that might be worth looking at is making some of these things configurable in finer grain to be able to turn on things like `readOnlyRootFilesystem` if we enable a properly configured volume.

Also, some of the recommendations there are great - like probing capabilities. All these are things that make sense at the application layer - not at the the runtime layer - as you know what features make sense to turn off. Languages and platforms tend to be generic tools, and being opinionated can be harmful for a wide range of people who still use the technology. That's why this is normally done on the leaf layers where you've installed all you need and then you can say this is it, buckle the belt.

For example, dropping all capabilities will probably have the issue of making crash dumps to stop working if you are using anything but Moby/Docker. Some air gapped scenarios might want this - but not all scenarios want to forego having a way of detecting crashing code.

olljanat: Ah so it just need write permissions to `/tmp` folder. Mounting tmpfs to there looks to be doing the trick. This one starts just fine:
```bash
docker run -it --rm --read-only --mount type=tmpfs,destination=/tmp mcr.microsoft.com/dotnet/samples:aspnetapp-buster-slim-amd64
```

olljanat:
> > it sounds that you are thinking both of them as developer tools?
> 
> No. Here's an example Dockerfile: https://github.com/jellyfin/jellyfin/blob/488e276f01b34c9aa82b53c969c56ccc9647cfbe/Dockerfile#L33. It installs packages in its "runtime layer". This particular app is self-contained, but it is an example of the scenario I was talking about. These APT commands require root permissions.

Btw. That example actually also show one way to solve those special use cases as it actually uses standard `debian:bullseye-slim` image instead of `mcr.microsoft.com/dotnet/runtime-deps:5.0-bullseye-slim` on runtime. But I will discuss with those who are more familiar with rootless containers in-case they have some other proposals too.

richlander:
That solution (using bullseye) is a significant step-function in complexity to solve this problem. It means you have to switch from framework-dependent to self-contained deployment just because you want to add packages.

For distroless, on the other hand, it would make a lot more sense to go with rootless. Maybe there is a solution in that space. That feels pretty compelling to me. Any takers?

olljanat:
> For distroless, on the other hand, it would make a lot more sense to go with rootless. Maybe there is a solution in that space. That feels pretty compelling to me. Any takers?

Yes, distroless + rooless runtime image would fit very nicely to my use cases. Afaiu distroless would need https://github.com/dotnet/runtime/issues/43113 to be solved first so marked this PR to draft now.

richlander:
Let's see how the following goes. https://github.com/dotnet/dotnet-docker/issues/3141

olljanat:
#3141 was closed so will close this one too.

PR #3160, opened 2021-09-21
Add GC hard limit % to sample

richlander:


mthalman (approved):
Out of curiosity, at what point does this cease to be a sample and become more of a diagnostic tool?  I'm just thinking that all of the metrics provided here would be nice to have as a .NET global tool.

richlander:
You are on the money, Matt.

This is the tool: https://github.com/dotnet/core/tree/main/samples/dotnet-runtimeinfo

I got into a weird spot where I now have two copies and the tool is now much worse. As part of this effort, I was going to update the same with my latest changes.

PR #3172, opened 2021-09-24
Infra to build distroless tarball image with CBL-Mariner toolkit

mthalman:
This defines a set of scripts and configuration files that provides a way to build a distroless tarball of CBL-Mariner using its toolkit. This tarball will eventually be added to the Docker image once the Dockerfile gets authored.

Related to #3146


mthalman:
Converted to a draft since this needs to be re-worked.

eng/cbl-mariner/toolkit-build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+marinerRepoTag="$1"
+outputPath="$2"
+script=$(readlink -f "$0")
+scriptDir=$(dirname "$script")
+
+# Necessary to avoid having the toolkit think it's running in a CDPx build environment which would cause this build to not work
+rm /.dockerenv
+
+git clone --depth 1 --branch 1.0-stable https://github.com/microsoft/CBL-Mariner.git /CBL-Mariner
```
MichaelSimons, line 13: Any reason this is not included in a pre-made buildtools-prereqs image?  Advantages are eliminate external dependencies and speed of building.  Additionally it may be advantages down the road if when we need to build multiple versions of the runtime-deps tarball.

eng/cbl-mariner/build-distroless.sh
```diff
@@ -0,0 +1,17 @@
...
+
+marinerRepoTag="$1"
+outputPath="$2"
+script=$(readlink -f "$0")
+scriptDir=$(dirname "$script")
+
+# Run container to build distroless tarball
+docker build --pull -t mariner-prereqs $scriptDir
+containerOutputPath="/dotnet-runtime-deps.tar.gz"
+docker run --name mariner-build --privileged mariner-prereqs ./toolkit-build.sh $marinerRepoTag $containerOutputPath
```
MichaelSimons, line 13: It seems like the workflow here may be a little clunky upon errors.  It looks like it requires the user to cleanup the mariner-build container before this script can be run again.  Utilizing a unique container name seems useful.

eng/cbl-mariner/packagelists/dotnet-runtime-deps.json
```diff
@@ -0,0 +1,12 @@
+{
```
MichaelSimons, line 1: It feels strange that this is considered `eng` content because this is going to be version specific.

eng/cbl-mariner/build-distroless.sh
mthalman: Fixed

eng/cbl-mariner/packagelists/dotnet-runtime-deps.json
mthalman: I've updated things so that the path to a version-specific runtime-deps package list file can be passed in when invoking the `build-distroless.sh` script. This makes the implementation in the `eng/cbl-mariner` completely generic wrt to the runtime-deps. The runtime-deps package list file will instead be located next to (or near) the Dockerfile.

mthalman: Note that this will require adding the `jq` package to the buildtools-prereqs image.

PR #3183, opened 2021-09-29
Allow arch to be included in monitor tag test data.

jander-msft:
The test code is attempting to use the "6.0-alpine" tag for the dotnet-monitor image. However, this is not a simple tag. The simple tags for dotnet-monitor are "6.0-alpine-amd64" and "6.0.0-preview.8-alpine-amd64".

The problem is that the MonitorImageData class is overriding the arch tag suffix to be empty. The fix is to allow the default arch tag determination behavior.


jander-msft:
@mthalman, feel free to merge this PR, if you approve of it, to unblock the nightly build.

mthalman (approved):

PR #3188, opened 2021-10-01
Support parameterized agent pools

mthalman:
This removes hard-coded pools in the common build-test-publish-repo template in order for it to be easily reusable by other products. Pools can be defined using template parameters but also have defaults configured.

In order for the pools, and other settings, to be used by the other .NET team-owned repositories, such as dotnet-framework-docker, I've created a wrapper template at `eng/common/stages/dotnet/build-test-publish-repo.yml` which consolidates these settings.

Related to dotnet/docker-tools#843

### DIFF
--- a/eng/common/templates/variables/common.yml
+++ b/eng/common/templates/variables/common.yml
@@ -19,8 +19,6 @@ variables:
   value: true
 - name: manifestVariables
   value: ""
-- name: defaultLinuxAmd64PoolImage
-  value: ubuntu-latest
 - name: mcrImageIngestionTimeout
   value: "00:20:00"
 - name: mcrDocIngestionTimeout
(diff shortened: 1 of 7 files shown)


eng/common/templates/variables/common.yml
```diff
@@ -32,3 +30,20 @@ variables:
...
   value: "''"
+
+- name: defaultLinuxAmd64PoolImage
+  value: ubuntu-latest
+- name: defaultLinuxArm32PoolImage
+  value: null
+- name: defaultLinuxArm64PoolImage
+  value: null
+- name: defaultWindows2016PoolImage
+  value: vs2017-win2016
```
dagood, line 41: This seems like a new value--that previously, the pool was always `Docker-2016-${{ variables['System.TeamProject'] }}`, but now by default you get a hosted pool unless you use `eng/common/templates/stages/dotnet/build-test-publish-repo.yml` or specify it manually. Is this the intent, or am I missing something?

dagood (approved):
Seems like a reasonable extraction to me.

eng/common/templates/variables/common.yml
mthalman: That's correct.  All the non-.NET-specific defaults are set to Hosted Pools since that's the common thing that would be usable by all teams using this infra.

MichaelSimons (approved):

PR #3202, opened 2021-10-12
Use any host for dotnet-monitor URLs.

jander-msft:
In this [PR](https://github.com/dotnet/dotnet-monitor/pull/990), dotnet-monitor is removing the heuristic for binding the metrics URL to "any address" from the tool and just binding both sets of URLs to localhost. However, binding to "any address" is desirable in containers since they are frequently run in bridge networking and dotnet-monitor is specifically designed to allow for Prometheus metrics scraping (which scrapes from a separate pod).

This change sets both set of URLs to bind to "any address" to allow access from outside of the pod.

I've tested that:
- The dotnet-monitor URLs are accessible from other pods using the pod IP address.
- Setting the URLs back to localhost using environment variables is respected and prevents access from other hosts.
- Using "kubectl port-forward" works in both "any address" and localhost configurations.

cc @dotnet/dotnet-monitor

### DIFF
--- a/eng/dockerfile-templates/monitor/6.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/monitor/6.0/Dockerfile.alpine
@@ -40,4 +40,4 @@ ENV \
 
 RUN chmod 755 dotnet-monitor
 
-ENTRYPOINT ["dotnet-monitor", "collect"]
+ENTRYPOINT [ "dotnet-monitor", "collect", "-urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
--- a/src/monitor/6.0/alpine/amd64/Dockerfile
+++ b/src/monitor/6.0/alpine/amd64/Dockerfile
@@ -40,4 +40,4 @@ ENV \
 
 RUN chmod 755 dotnet-monitor
 
-ENTRYPOINT ["dotnet-monitor", "collect"]
+ENTRYPOINT [ "dotnet-monitor", "collect", "-urls", "https://+:52323", "--metricUrls", "http://+:52325" ]


jander-msft:
> > Is there any reasonable test that could be added to verify this functionality?
> 
> The tests already change the donet-monitor container to use "any host" binding (see [here](https://github.com/dotnet/dotnet-docker/blob/2ec15b5593457593f2ca68611b710409c06a7a4f/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs#L515)) using environment variables since the containers are running on a bridge network; otherwise, the tests wouldn't be able to talk to the dotnet-monitor process (it would have failed to bind to localhost) and the reverse lookup of the container port to the host port (see [here](https://github.com/dotnet/dotnet-docker/blob/2ec15b5593457593f2ca68611b710409c06a7a4f/tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs#L246)) would not be necessary if they were running in host mode networking.
> 
> With this change, the tests are now only disabling HTTPS rather than changing how it binds to the host (e.g. the default url of `https://*:52323` is changed to `http://*:52323` using configuration).
> 
> I can look into adding an authentication test, which we should probably have regardless of this change. Me validating authentication was working correctly wasn't so much of a question of whether it would work when changing to "any host" but more of a sanity check. Our recommendation is that authentication is enabled, especially when using "any host" URLs (because now anyone with the container IP address can access the endpoint whereas previously only connections from the same host were allowed on the 52323 port by default).

To further elaborate:
- the existing tests already test that the metrics URL (`http://+:52324` as specified by the container) without modifying that URL; only the port is exposed from the container: https://github.com/dotnet/dotnet-docker/blob/2ec15b5593457593f2ca68611b710409c06a7a4f/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs#L279
- the existing tests change the default URL (`https://+:52323` as specified by the container) to `http://+:52323`, the difference being that HTTP is used instead of HTTPS. What I can do is try to add the ability to test the default HTTPS setting such that the URLs do not need to be overridden at all, but that will require certificate generation, mounting the certificate, configuring the container to use the certificate and password, and then finally updating the tests to not override the URL. The lack of testing the default URL as-is is an already existing issue.

So what I would like to proposed is that, if there are no issues with this current change, is to merge in this change and then I can work on two subsequent changes:
1. Add SSL cert generation and ingestion to the dotnet-monitor tests in order to test HTTPS, which allows the tests to use the image without necessarily modifying the command line.
1. Add API key generation and ingestion to the dotnet-monitor tests in order to test authentication is enabled or disabled as expected in the default usage of the image.

I would argue that these two changes are separate from the changes in this PR. These two in combination would satisfy the ability of testing the image with minimal changes at container startup. Ideally, these would have been done already with the existing settings of the container. I can log separate issues for each of these and work on them in the coming weeks.

@mthalman, is this a reasonable proposal?

mthalman (approved):

PR #3207, opened 2021-10-13
Support optional stable branding in update-deps

mthalman:
The update-dependencies tool will by default trim any version string containing `-rtm.` to a stable branding version (e.g. `6.0.0-rtm.21512.5` -> `6.0.0`) when computing the blob URL.  For the current 6.0 RTM files in the blob container, they don't currently have this stable branding for filenames (i.e. the filename still contains the build number).  Because of the version trimming, this causes a failure when attempting to run update-deps because it's trying to access a filename that doesn't exist.

I've updated the code to allow this version trimming for stable branding to be optional.  We can enable it from the pipeline whenever necessary.  Note that support for stable branding in the Dockerfiles themselves still requires a change to how the Dockerfile is authored but at least the update-dependencies tool can now support either scenario.

### DIFF
--- a/eng/update-dependencies/DockerfileShaUpdater.cs
+++ b/eng/update-dependencies/DockerfileShaUpdater.cs
@@ -91,7 +91,7 @@ namespace Dotnet.Docker
         public static IEnumerable<IDependencyUpdater> CreateUpdaters(
             string productName, string dockerfileVersion, string repoRoot, Options options)
         {
-            string versionsPath = System.IO.Path.Combine(repoRoot, Program.VersionsFilename);
+            string versionsPath = System.IO.Path.Combine(repoRoot, UpdateDependencies.VersionsFilename);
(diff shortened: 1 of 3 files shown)


eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -437,12 +431,22 @@ private static bool TryParsePreviewVersion(string version, out string? previewVe
...
+                buildVersion = version;
+            }
+            else
+            {
+                buildVersion = VersionUpdater.GetBuildVersion(productName, dockerfileVersion, variables);
+            }
+
+            if (buildVersion is not null)
+            {
+                buildVersion = Program.ResolveProductVersion(buildVersion, options);
```
MichaelSimons, line 446: <Nit> We should consider renaming `Program`.  The readability of this is poor.

MichaelSimons (approved):

Issue #3222, opened 2021-10-15
Tracking: icu and krb5 cause unnecessary bash dependency to be included in Mariner distroless derivative


mthalman:
https://microsoft.visualstudio.com/OS/_workitems/edit/36653173

PR #3223, opened 2021-10-15
Add distroless Dockerfiles for CBL-Mariner

mthalman:
Defines distroless Dockerfiles for CBL-Mariner.

The installation of .NET is fairly straightforward, just using a multi-stage build to copy the files extracted from a tarball into the distroless container. The tricky part is the installation of the runtime-deps packages. By using dnf's `installroot` option (which uses chroot under the covers), it provides the ability to install the entire graph of .NET dependencies in an isolated staging location. That staging location can then be copied to the distroless container as part of the build.  This same pattern is used by the tests in order to add some necessary packages to a distroless container to make it testable.  Note that due to two dependent packages having a dependency on bash, bash is ending up installed in the built image.  This is tracked by https://github.com/dotnet/dotnet-docker/issues/3222.

These changes also define a non-root user by default for distroless containers.

Related to #3147, #3148


johnazariah:
Congratulations!

Thank you, @mthalman! :)

src/runtime/6.0/cbl-mariner1.0-distroless/amd64/Dockerfile
```diff
@@ -0,0 +1,26 @@
+ARG REPO=mcr.microsoft.com/dotnet/runtime-deps
+
+# Installer image
+FROM amd64/alpine:latest AS installer
```
MichaelSimons, line 4: Using alpine in the mariner image doesn't seem optimal.  Is the reason because it's small and therefore is quicker to pull in build environments?  I see benefits to using a mariner image here.  Since the runtime-deps is already using a mariner base, it feels like that would be a better choice here.

src/aspnet/6.0/cbl-mariner1.0-distroless/amd64/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN aspnet_version='6.0.0-rtm.21514.23' \
+    && wget -O aspnetcore.tar.gz https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$aspnet_version/aspnetcore-runtime-$aspnet_version-linux-x64.tar.gz \
+    && aspnetcore_sha512='8b19890a158f32edeb109de49f522197a91c921963e4e3e27e26a03c1556f37044a39cde4a9e52281aefdb58e339f0ff8db6b96d57386322cd8f96ebf4fa0f26' \
+    && echo "$aspnetcore_sha512  aspnetcore.tar.gz" | sha512sum -c - \
+    && mkdir -p /dotnet \
+    && tar -ozxf aspnetcore.tar.gz -C /dotnet ./shared/Microsoft.AspNetCore.App \
+    && rm aspnetcore.tar.gz
+
+
+# # ASP.NET Core image
```
MichaelSimons, line 16:
```suggestion
# ASP.NET Core image
```

PR #3268, opened 2021-11-05
Add component governance pipeline for container images

### LINKED ISSUE #3001: Scan images with component governance
Currently the repo only scans .NET projects in the repo with component governance. The built Docker images are not being scanned by default because CG will only scan them automatically if using the Docker build AzDO task.  There is an option to explicitly configure a pipeline to scan a known set of images.  I'm proposing that we use this option to ensure that all of the Docker images built by the repo get scanned.

mthalman:
This pipeline will pull all supported .NET Linux container images and have them scanned by component governance.

Fixes #3001


eng/pipelines/cg-images.yml
```diff
@@ -0,0 +1,32 @@
...
+  steps:
+  - template: ../common/templates/steps/init-docker-linux.yml
+  - script: >
+      $(runImageBuilderCmd) pullImages
+      --architecture '*'
+      --manifest 'manifest.json'
+      --output-var 'pulledImages'
+    displayName: Pull Images
+    name: PullImages
+  - task: ComponentGovernanceComponentDetection@0
```
MichaelSimons, line 28: If I understand correctly, this appears to be scanning each tag.  In this context, it doesn't seem necessary and is potentially wasting compute resources.  It really only needs to scan the unique set of images.  If this is trivially cheap it is not a problem but otherwise this seems like something to consider resolving.

MichaelSimons (commented):
Does CG only support Linux images?  Why are Windows images not being scanned?

eng/pipelines/cg-images.yml
mthalman: Yes, it's scanning each tag. But each tag is for a unique image. The output of the `pullImages` command is a list of tags representing the _unique_ images that were pulled:

https://github.com/dotnet/docker-tools/blob/9f589069309993e387b5b5f05f1375cc068ae1ae/src/Microsoft.DotNet.ImageBuilder/src/Commands/PullImagesCommand.cs#L30-L44

Does this address your concern?

mthalman:
> Does CG only support Linux images? Why are Windows images not being scanned?

Correct, CG only supports Linux containers.

eng/pipelines/cg-images.yml
MichaelSimons: Yes, thanks

MichaelSimons (approved):

PR #3269, opened 2021-11-05
Clean up test code

mthalman:
* Re-enables tests for scenarios that are working.
* Remove tests and checks for EOL versions.


MichaelSimons (approved):
Nice cleanup.

Issue #1455, opened 2019-11-08
Detect package updates and rebuild images


mthalman:
When updates are made available in a dependent package that is installed by one of the .NET Core Dockerfiles, this update should be detected and the image rebuilt and published to provide consumers with the latest version of the package.  There is currently a system for detecting changes to base images and rebuilding dependent images so, in theory, this would be an extension of that to include the detection of package updates.

MichaelSimons:
Recently there was an instance in which `curl` had a security vulnerability which is included in the .NET images.  The ARM32 images didn't pick this up because the base image never changed.  Vulnerability scanning caught this and a build was manually triggered to get the patched curl version that addressed the vulnerability.

Rebuilding any time that any of the packages that are included in the .NET images may cause a lot of churn.  We may want to consider batching these to reduce churn if they are not addressing vulnerabilities.

richlander:
This makes good sense, but is more complicated than the base image updates.

- IMO, we have no good choice but to accept base image updates and rebuild / republish.
- We have a lot more flexibility with packages. We'll get the new package the next time we rebuild .NET Core images, either due to a base image update or .NET Core servicing update. Typically, (together) those happen multiple times per month. So, it only makes to force a package update for CVEs.
- The big question is finding a way to know which of the package updates are functional vs a critical CVE. Anyone know of a magic API that offers that info?

NasAmin:
Can something like Trivy be used?
https://github.com/aquasecurity/trivy

mthalman:
Another example where this would have been useful: https://github.com/dotnet/dotnet-docker/issues/2417.  The `krb5` package had been updated with a security fix but the parent Alpine image never changed.  This meant we never automatically triggered a build to rebuild the image.

mthalman:
As stated in https://github.com/dotnet/docker-tools/pull/824#issuecomment-900506374, the plan is stored the package graph as metadata in the image info file. In addition to actually storing this metadata, we should also provide notifications whenever the package graph changes for a given Dockerfile (see https://github.com/dotnet/dotnet-docker/pull/3271#issuecomment-963167406). This allows us to be aware of any unexpected changes.

mthalman:
Now that the infrastructure supports the generation of SBOMs, it seems more appropriate to consume the SBOM's description of the packages that are installed in an image rather than duplicating that information in the image info file. This would clean up code and provide a single source of truth.

mthalman:
[...]

PR #3271, opened 2021-11-05
Update distroless Mariner to reference prebuilt-ca-certificates package

### LINKED ISSUE #3236: Use prebuilt-ca-certificates package for distroless version of CBL-Mariner
The distroless Dockerfile for CBL-Mariner currently references `ca-certificates-microsoft`:
https://github.com/dotnet/dotnet-docker/blob/58b2670c6fc4bd431ce22f58a9f3d76ee09f1d33/src/runtime-deps/6.0/cbl-mariner1.0-distroless/amd64/Dockerfile#L9

This package has a dependency on bash which causes bash to be included in the built image (not what we want). A pre-built version of this package, named `prebuilt-ca-certificates`, will be available in the next CBL-Mariner update which does not have a bash dependency. Once the pre-built package is available, we should update the Dockerfile to reference that package instead.

This won't completely remove the dependency on bash for the Dockerfile as a whole. https://github.com/dotnet/dotnet-docker/issues/3222 would need to be resolved for that.

mthalman:
Fixes #3236

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/6.0/Dockerfile.mariner
+++ b/eng/dockerfile-templates/runtime-deps/6.0/Dockerfile.mariner
@@ -1,7 +1,7 @@
 FROM cblmariner.azurecr.io/base/core:{{OS_VERSION_NUMBER}}
 
 RUN tdnf install -y \
-        ca-certificates-microsoft \
+        prebuilt-ca-certificates \
         \
         # .NET dependencies
         glibc \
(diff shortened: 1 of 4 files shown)


MichaelSimons (commented):
What do you think about having some type of test for monitoring the package package graph in our distroless images?  It feels like we would want to know when there are changes.  For example a package adds a dependency of bash.  Size detection doesn't seem sufficient here.  The test may be better implemented via the staging phase vs the final image.  I'm not suggesting this is something to include here but this change prompted the thought.

MichaelSimons (approved):

mthalman:
> What do you think about having some type of test for monitoring the package package graph in our distroless images? It feels like we would want to know when there are changes. For example a package adds a dependency of bash. Size detection doesn't seem sufficient here. The test may be better implemented via the staging phase vs the final image. I'm not suggesting this is something to include here but this change prompted the thought.

That would be related to https://github.com/dotnet/dotnet-docker/issues/1455. The work for that would involve storing the package graph in the image info file. In order to be made aware of when graph changes occur, I think it would be reasonable to post an issue notification whenever that computed graph has changed from it's previous result for a given Dockerfile.

PR #3282, opened 2021-11-11
Add dotnet-monitor 7.0 image

### LINKED ISSUE #3277: Add Dockerfile for dotnet-monitor 7.0.0
Now that dotnet-monitor 6.0.0 GA has shipped, the nightly branch can be updated to start building the dotnet-monitor 7.0.0 image.

This does not depend on the existence of the .NET 7 images (yet) since early builds of dotnet-monitor 7.0.0 run on .NET 6.

jander-msft:
The 7.0 builds of the tool are based on .NET 6 at this time, so the Dockerfile is effectively the same as the 6.0 version but with 7.0 branding.

closes #3277

### DIFF
--- a/eng/mcr-tags-metadata-templates/monitor-tags.yml
+++ b/eng/mcr-tags-metadata-templates/monitor-tags.yml
@@ -1,2 +1,4 @@
 $(McrTagsYmlRepo:monitor)
 $(McrTagsYmlTagGroup:6.0-alpine-amd64)
+$(McrTagsYmlTagGroup:7.0-alpine-amd64)
+    customSubTableTitle: .NET Monitor 7.0 Preview Tags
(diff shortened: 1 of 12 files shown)


eng/mcr-tags-metadata-templates/monitor-tags.yml
```diff
@@ -1,2 +1,4 @@
 $(McrTagsYmlRepo:monitor)
 $(McrTagsYmlTagGroup:6.0-alpine-amd64)
+$(McrTagsYmlTagGroup:7.0-alpine-amd64)
```
mthalman, line 3: 7.0 should be listed above 6.0

mthalman (approved):
You'll also need to add a size baseline entry. https://github.com/dotnet/dotnet-docker/blob/337eeb48bbfd7674427ca4018826a17bd17c6c81/tests/performance/ImageSize.nightly.linux.json#L158

PR #3287, opened 2021-11-12
Curl download should fail on 404

### LINKED ISSUE #2998: Downloads with curl should fail if file doesn't exist
All usage of curl in the Dockerfiles use the following options: `-SL --output`.  If the target URL doesn't exist, it will return a 404 but, because of the way curl is configured, it will just write that 404 message to the output file.  The command itself succeeds.  In this case, we don't actually know that something is wrong until the checksum validation. It would be a better UX for curl to indicate that the download failed.  To do this, the `-f` option should be included.

mthalman:
Fixes #2998

### DIFF
--- a/documentation/scenarios/installing-dotnet.md
+++ b/documentation/scenarios/installing-dotnet.md
@@ -58,7 +58,7 @@ RUN apt-get update \
 # Install .NET
 ENV DOTNET_VERSION=5.0.0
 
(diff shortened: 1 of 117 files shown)


jander-msft:
Could you also update the dotnet-monitor Dockerfile templates and Dockerfiles?

mthalman:
> Could you also update the dotnet-monitor Dockerfile templates and Dockerfiles?

The Monitor Dockerfiles use wget which will fail by default. Curl is the one which needs the option.

MichaelSimons (approved):

Issue #2938, opened 2021-07-01
Windows Server Core self-contained sample fails to run


mthalman:
When attempting to run the container built from dotnetapp's sample [Dockerfile.windowsservercore-x64-slim](https://github.com/dotnet/dotnet-docker/blob/main/samples/dotnetapp/Dockerfile.windowsservercore-x64-slim) file, it results in the following error:

```
Failed to map file. CreateFileMappingW(C:\app\dotnetapp.exe) failed with error 5
Failure processing application bundle.
Failed to map bundle file [C:\app\dotnetapp.runtimeconfig.json]
```

The same error occurs for aspnetapp's sample [Dockerfile.windowsservercore-x64-slim](https://github.com/dotnet/dotnet-docker/blob/main/samples/aspnetapp/Dockerfile.windowsservercore-x64-slim) file.

This file was supposed to have been tested by the builds but due to an [issue in how the test data was defined](https://github.com/dotnet/dotnet-docker/pull/2908/commits/8390955a4bfda3b027940963d729c77a852f1093), it was never being  tested.

mthalman:
When fixing this, the testing of these Dockerfiles should be re-enabled.  See https://github.com/dotnet/dotnet-docker/pull/2908/commits/e19d9b9c9f555f554db10295ae3eedb2712e2a44

mthalman:
This appears to be an issue in the runtime.  I've logged https://github.com/dotnet/runtime/issues/55405.  We'll keep this issue open as a tracking issue and use it to re-enable the test when the underlying issue is fixed.

PR #3318, opened 2021-12-02
Enable sample testing of SCD on WSC

mthalman:
Now that https://github.com/dotnet/runtime/issues/55405 is fixed, the testing of the sample Dockerfile for self-contained deployment on Windows Server Core can be enabled.

Fixes https://github.com/dotnet/dotnet-docker/issues/2938


mthalman:
I didn't actually see that this scenario is already being tested by the next entry after the commented-out entry. So all that needs to be done here is remove the commented-out entry.

MichaelSimons (approved):

Issue #2152, opened 2020-08-06
Support Docker image builds from private drops of .NET Core


mthalman:
On occasion, there's a need to produce Docker images for internal use based on private drops of .NET Core.  This currently requires a fair amount of changes to the infrastructure and Dockerfiles.  We should investigate the changes that are necessary for this and determine how feasible it is to include those changes in the public sources.

mthalman:
Here are the set of changes that needed to be made to get this to work internally:

* The blob URL used in Dockerfiles to download archive files needs to be modified:
  * The host name used in the URL must target the internal blob storage account instead of the public one.
  * A SAS query string value needs to be appended to the blob URL in order to have access to the resource.  This requires having a Dockerfile `ARG` that can pass this value in dynamically as part of the build, retrieved from a Key Vault.  This is complicated by the characters used in the SAS query string, containing characters like `&` that mess up command invocations unless appropriately encoded.  This is tricky to get right in a way that satisfies both Linux and Windows. I hacked around this by storing the base-64 encoding of the value in Key Vault, passing that encoded value down into the Dockerfile where it was finally decoded within the Dockerfile logic.  This is ugly and not a great solution to use as a permanent solution.
  * In order to flow the SAS query string from a build variable all the way to the Dockerfile as an `ARG`, it must be defined as a `manifestVariable` similar to what is done for the [official Image Builder pipeline](https://github.com/dotnet/docker-tools/blob/a9b75c5c340d2aaf7b6a334b99020c3aac55448b/eng/pipelines/templates/variables/image-builder.yml#L7). And then that variable must be referenced as a `buildArg` in the manifest in order to pass that value as an `ARG` to the Dockerfile.
* The update-dependencies tool needs its SHA updating logic changed to be able to retrieve checksum values from the internal storage account instead of the public one.  And this requires a SAS query string that is passed as an argument in order to be able to access that storage account.
* Test app Dockerfiles ([Linux](https://github.com/dotnet/dotnet-docker/blob/master/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/Dockerfile.linux), [Windows](https://github.com/dotnet/dotnet-docker/blob/master/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/Dockerfile.windows)) need the ability to restore from the internal NuGet feed. This consists of a variety of sub-tasks:
[...]

PR #3364, opened 2021-12-20
Get dotnet monitor from blob storage.

jander-msft:
The dotnet-monitor repo can now publish its nupkg file to blob storage in addition to the dotnet-tools feed. This change switches the Dockerfiles to pull from blob storage. This will be especially useful when shipping the final versions of 7.0+ such that the package can be provided using the same mechanism as the rest of the dotnet packages and the checksum file value will match the package (as compared to the 6.0 releases, where the package had to be published to nuget.org, which changes the signing of the package and thus the checksum file doesn't match).


mthalman (approved):
Very nice. This will also allow us to more easily support [internal builds](https://github.com/dotnet/dotnet-docker/issues/2152) should you ever have a need for that.

PR #3369, opened 2021-12-28
Add health checks

TruebloodOF:
Adding health checks would be useful for my teams current use case of using this aspnet core image as a dummy placeholder for new ecs deployments. Adding health checks will allow the load balancers to perceive these deployments as 'healthy'


mthalman:
In addition to having the health check in the service, it seems to be appropriate to also add a [`HEALTHCHECK` instruction in the Dockerfile](https://docs.microsoft.com/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-6.0#docker-healthcheck) since this sample is specific to Docker scenarios.

mthalman:
Closing due to lack of activity. Feel free to reopen if you wish to make further updates.

PR #3370, opened 2021-12-28
Mariner reference to prebuilt-ca-certificates pkg

mthalman:
The changes from https://github.com/dotnet/dotnet-docker/pull/3271 never got applied to the 7.0 runtime-deps Dockerfile.

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/7.0/Dockerfile.mariner
+++ b/eng/dockerfile-templates/runtime-deps/7.0/Dockerfile.mariner
@@ -1,7 +1,7 @@
 FROM cblmariner.azurecr.io/base/core:{{OS_VERSION_NUMBER}}
 
 RUN tdnf install -y \
-        ca-certificates-microsoft \
+        prebuilt-ca-certificates \
         \
         # .NET dependencies
         glibc \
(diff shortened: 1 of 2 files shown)


eng/dockerfile-templates/runtime-deps/7.0/Dockerfile.mariner
```diff
@@ -1,7 +1,7 @@
 FROM cblmariner.azurecr.io/base/core:{{OS_VERSION_NUMBER}}
 
 RUN tdnf install -y \
-        ca-certificates-microsoft \
+        prebuilt-ca-certificates \
```
MichaelSimons, line 4: Is the bash dependency the only difference in the packages?  Why do we want to change the package in the non-distro images.  If 6.0 and 7.0 reference `prebuilt-ca-certificates` should 5.0 and 3.1 be updated as well?

mthalman: > Is the bash dependency the only difference in the packages?

Yes



> Why do we want to change the package in the non-distro images

I prefer it for consistency reasons.



> If 6.0 and 7.0 reference `prebuilt-ca-certificates` should 5.0 and 3.1 be updated as well?

3.1 and 5.0 still use `ca-certificates` and were never updated as part of https://github.com/dotnet/dotnet-docker/pull/3247 because it would have technically been a breaking change.

MichaelSimons (approved):

PR #3381, opened 2022-01-06
Update .NET 7 Dockerfiles to new blob account

mthalman:
Builds for .NET 7 have migrated to a new blob storage account named dotnetbuilds. This account also has different container names compared to dotnetcli. So the `dotnet` container name is replaced with `public`.

I've made the changes to update the templates and also regenerated to target a recent .NET 7 build. I've also updated the update-dependencies tool to account for the URL differences between .NET 7 and older versions.


mthalman:
More changes for this are coming.

mthalman:
I've made some updates based on new learnings. The dotnetcli storage account is not considered legacy. It is still relevant to 7.0. It will be the location where release bits will be stored. For daily builds, the dotnetbuilds storage account will be used. This means update-dependencies needs to account for either location. I've updated its checksum logic to try dotnetbuilds first and fall back to dotnetcli if necessary.

This means that once a preview version of 7.0 is released, then main and nightly will differ in which storage account is being referenced in the 7.0 Dockerfiles.

eng/dockerfile-templates/aspnet/7.0/Dockerfile.alpine
```diff
@@ -11,7 +11,7 @@ ENV \
     Logging__Console__FormatterName=Json
 
 # Install ASP.NET Core
-RUN wget -O aspnetcore.tar.gz https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$ASPNET_VERSION/aspnetcore-runtime-$ASPNET_VERSION-linux-musl-{{ARCH_SHORT}}.tar.gz \
+RUN wget -O aspnetcore.tar.gz https://dotnetbuilds.azureedge.net/public/aspnetcore/Runtime/$ASPNET_VERSION/aspnetcore-runtime-$ASPNET_VERSION-linux-musl-{{ARCH_SHORT}}.tar.gz \
```
MichaelSimons, line 14: Have you considered moving this to the manifest?  I thought you were considering this with your template/servicing changes - perhaps you are deferring to make the change until later?  Given the difference between main and nightly, having a single location, the manifest, is advantageous.  A thought is to store separate nightly and non-nightly values in the manifest therefore eliminating differences in the manifest between branches.

mthalman: I was thinking of doing that later. But it might make sense to do that now anyway.

> A thought is to store separate nightly and non-nightly values in the manifest therefore eliminating differences in the manifest between branches.

I'd prefer the manifest to differ rather than the Dockerfiles. _Something_ has to differ between the branches. With what you're proposing, what would that be?

MichaelSimons: What I was referring to was to construct the Dockerfile templates and manifest in a way that they don't have to be changed during the merge process.  The Dockerfiles would need to be regenerated but no manual changes would be necessary.

Current"
```
RUN wget -O aspnetcore.tar.gz https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$ASPNET_VERSION/aspnetcore-runtime-$ASPNET_VERSION-linux-musl-{{ARCH_SHORT}}.tar.gz
```

Proposed:
```
RUN wget -O aspnetcore.tar.gz {{{{VARIABLES[cat("7.0|download-url",VARIABLES["branch"])]}}}}/aspnetcore/Runtime/$ASPNET_VERSION/aspnetcore-runtime-$ASPNET_VERSION-linux-musl-{{ARCH_SHORT}}.tar.gz
```

MichaelSimons: Manifest:

"7.0|download-url|nightly": "https://dotnetbuilds.azureedge.net/public"
"7.0|download-url|main": "https://dotnetcli.azureedge.net/dotnet"

mthalman: In order to set `VARIABLES["branch"]` correctly, it would require the ability to dynamically determine whether the source branch is main or nightly. This is non-trivial. For now, I'd prefer to just have a single line diff between main and nightly in the manifest.versions.json file.

Main:
```json
"base-url|7.0": "https://dotnetcli.azureedge.net/dotnet"
```

Nightly:
```json
"base-url|7.0": "https://dotnetbuilds.azureedge.net/public"
```

MichaelSimons: > In order to set VARIABLES["branch"] correctly, it would require the ability to dynamically determine whether the source branch is main or nightly

FYI we are currently using this variable in the readme template.  It's handled by the `Get-generatedReadmes.ps1` script.

mthalman: Thanks for the tip!

manifest.versions.json
```diff
@@ -35,16 +35,19 @@
...
+    "aspnet|7.0|linux-rpm|x64|sha": "80b4efafc6c61bb49c6d51c1e854096260a3c6e85844c3ce916f7b6dec9234dfc05a416960e9a42b8269a7d673015e46977fffdbebf6ceed8152c203517f1752",
+    "aspnet|7.0|linux|arm|sha": "b90538e9fc8f012f9e78372f75de080b3994f69b32df1db27e3feb535237d7c14491964c78b0c698c5fa4fa8f41fe4d3478b82df79f75e4b8337b369ea2635d2",
+    "aspnet|7.0|linux|x64|sha": "5d70059032ce0e867d511ac86bd9e5b18e47551c7e134b12b7d5ef8116aa3ec49a742256412149095479394bee050e13970d30e5648e8aaf7e0f99be20844386",
+    "aspnet|7.0|linux|arm64|sha": "6e8cd1398775117f780334474ede4acfc308c2a7e97863b3a31017747eaf0d9a51fd95eaaa209fdb0bbe93d74d850156c1c576f253100887d7e349a47ad9012b",
+    "aspnet|7.0|win|x64|sha": "150d7f76673faadb5bcd8ce0eb6501c4c49359c047edb98cbc2abb25bf7cb4018bf7adf2acc770c5991d50e324ea27fef4e28a7a75e54d2dd943b916e212f11f",
+    "aspnet-runtime-targeting-pack|7.0|linux-rpm|x64|sha": "b58adf389af98237ad066990ff32326943d0b92117112213236c223cb4c73e8c016ee2234ac9d5f7dc9d6a2cb31ab2ff382d0214ff6aca9dba5ed4f531e00fc3",
+
+    "base-url|7.0|main": "https://dotnetcli.azureedge.net/dotnet",
```
MichaelSimons, line 49: How do you see this evolving going forward?  Would we end up having separate variables defined for each version or do you see this as being a version agnostic variable?  I am wondering if we can get away with a single value going forward and then a legacy value used for 3.1/5.0.  At least until they start getting published to the new storage location if ever.

mthalman: In order to support the two storage accounts as well as [internal builds](https://github.com/dotnet/dotnet-docker/issues/2152), I see this eventually looking something like this:

```json
"base-url|default|main": "https://dotnetcli.azureedge.net/dotnet",
"base-url|internal|main": "https://dotnetbuilds.azureedge.net/internal",
"base-url|default|nightly": "https://dotnetbuilds.azureedge.net/public",
"base-url|internal|nightly": "https://dotnetbuilds.azureedge.net/internal",
"base-url|3.1|main": "$(base-url|default|main)",
"base-url|3.1|nightly": "$(base-url|default|nightly)",
"base-url|5.0|main": "$(base-url|default|main)",
"base-url|5.0|nightly": "$(base-url|default|nightly)",
"base-url|6.0|main": "$(base-url|default|main)",
"base-url|6.0|nightly": "$(base-url|default|nightly)",
"base-url|7.0|main": "$(base-url|default|main)",
"base-url|7.0|nightly": "$(base-url|default|nightly)"
```

Because of internal builds, there will be a need to control the base URL of a specific .NET version. So we'd have an entries for each of the versions like this.

MichaelSimons (approved):

manifest.versions.json
mthalman: The internal base URLs could probably be collapsed into one.

From this:

```json
"base-url|internal|main": "https://dotnetbuilds.azureedge.net/internal",
"base-url|internal|nightly": "https://dotnetbuilds.azureedge.net/internal",
```

To this:
```json
"base-url|internal": "https://dotnetbuilds.azureedge.net/internal"
```

PR #3382, opened 2022-01-06
Define PowerShell install Dockerfile sub-template

mthalman:
This takes advantage of the changes from https://github.com/dotnet/docker-tools/pull/930 and https://github.com/dotnet/docker-tools/pull/932 to define a template file that encapsulates the logic for installing PowerShell in the sdk Dockerfiles. The sdk Dockerfile templates just insert that template at the appropriate position and pass in arguments to control the output that is generated.

This does make the PowerShell template a bit more verbose and less easy to read because it's referencing variables instead of hardcoding values like version numbers. Where reasonable, I've used Cottles feature of defining template variables with the `set` keyword to encapsulate some of the frequently used verbose expressions.

This change is just a stepping stone in reaching a more comprehensive solution where sub-templates are used more extensively. This change is intended to just establish the pattern that will be used for future work.


mthalman:
I'm going to make some more changes to this to get early feedback on.

mthalman:
I've made some changes to factor out the downloading of files into a separate template. This is to establish the pattern for other future templates to make use of this template as well. It encapsulates the variation of commands between Alpine and other Linux distros. I don't necessarily think we need smaller templates like this for all the "primitive" commands that we run (e.g. downloading a file from Windows container, extracting a tarball). But I think it's useful in this case because of the need to vary the generation between Alpine and other Linux distros.

This is dependent on https://github.com/dotnet/docker-tools/pull/935 being fixed.

eng/dockerfile-templates/sdk/3.1/Dockerfile.linux
```diff
@@ -36,16 +36,9 @@ RUN dotnet_sdk_version={{VARIABLES["sdk|3.1|build-version"]}} \
...
-    && rm PowerShell.Linux.{{ARCH_NUPKG}}.$powershell_version.nupkg \
-    && ln -s /usr/share/powershell/pwsh /usr/bin/pwsh \
-    && chmod 755 /usr/share/powershell/pwsh \
-    # To reduce image size, remove the copy nupkg that nuget keeps.
-    && find /usr/share/powershell -print | grep -i '.*[.]nupkg$' | xargs rm
+{{InsertTemplate("../Dockerfile.linux.install-powershell",
+    [
+        "dotnet-version": "3.1",
+        "nupkg-platform": cat("Linux.", ARCH_NUPKG),
+        "sha-platform": cat("Linux|", ARCH_NUPKG)
```
mthalman, line 43: Is it worthwhile to factor out a wrapper template that defaults these settings? This would avoid duplication of these settings across all Dockerfile.linux and Dockerfile.mariner files for the sdk.

eng/dockerfile-templates/sdk/Dockerfile.linux.install-powershell
```diff
@@ -0,0 +1,24 @@
+{{
+    _ ARGS:
+        nupkg-platform: platform specifier for the PowerShell NuGet pkg filename (e.g. "Linux.Alpine")
+        sha-platform: platform specifier for the PowerShell NuGet pkg sha variable name (e.g. "Linux|x64")
+        dotnet-version: .NET Dockerfile version (e.g. 6.0)
+}}{{
+    set isAlpine to ge(find(OS_VERSION, "alpine"), 0) }}{{
+    set nupkgName to cat("PowerShell.", ARGS["nupkg-platform"], ".$powershell_version.nupkg")}}{{
+    set url to cat("https://pwshtool.blob.core.windows.net/tool/$powershell_version/", nupkgName)
```
MichaelSimons, line 9: I see you defined a local template variable for the url which is only referenced in one place.  What was the reason for this?  Curious if this pattern would help the readability if used for other values such of the powershell_version and powershell_sha as well?

eng/dockerfile-templates/sdk/Dockerfile.linux.install-powershell
```diff
@@ -0,0 +1,24 @@
+{{
+    _ ARGS:
+        nupkg-platform: platform specifier for the PowerShell NuGet pkg filename (e.g. "Linux.Alpine")
+        sha-platform: platform specifier for the PowerShell NuGet pkg sha variable name (e.g. "Linux|x64")
+        dotnet-version: .NET Dockerfile version (e.g. 6.0)
+}}{{
+    set isAlpine to ge(find(OS_VERSION, "alpine"), 0) }}{{
```
MichaelSimons, line 7: Cottle's limited expression support is rather unfortunate.  I have to think about how the braces are being used to achieve multiple lines with aligned expressions.

eng/dockerfile-templates/sdk/Dockerfile.windows.install-powershell
```diff
@@ -0,0 +1,15 @@
+{{
+    _ ARGS:
+        dotnet-version: .NET Dockerfile version (e.g. 6.0)
+}}{{if ge(find(OS_VERSION, "nanoserver"), 0):{{set rootDir to "\" }}^else:{{set rootDir to "$Env:ProgramFiles\"}}}}# Install PowerShell global tool
```
MichaelSimons, line 4: Separating the local template variable init logic like what was done in the linux template would help the readability IMO.

eng/dockerfile-templates/sdk/3.1/Dockerfile.linux
MichaelSimons: Rather than factoring this out into another template, can't these simply get computed in the PS template.  

```
if isAlpine 
  set nupkg-platform = "Linux.Alpine"
  set sha-platform = "Linux.Alpine"
else
  set nupkg-platform = cat("Linux.", ARCH_NUPKG)
  set sha-platform = cat("Linux|", ARCH_NUPKG)
```

This would be the same pattern used in the windows template to set `rootDir`.

eng/dockerfile-templates/sdk/Dockerfile.linux.install-powershell
mthalman: It was purely for readability to keep lines shorter and less complex. I don't see it benefiting `powershell_version` and `powershell_sha512` too much since those lines are already pretty simple (`<var>=<val>`). But I'd be open to declaring them as template variables simply to co-locate those similar concepts. Let me know.

eng/dockerfile-templates/sdk/Dockerfile.linux.install-powershell
mthalman: Yeah, not being able to have multiple commands in a block is limiting. I've logged https://github.com/r3c/cottle/issues/135 to request this feature.

mthalman:
I've been able to refactor things to completely remove the need for any args to the PS templates. This takes advantage of https://github.com/dotnet/docker-tools/pull/936 so that it can compute the .NET version. This would need to be duplicated in each template that needs the version. But I do like that it makes the Dockerfile templates that much simpler.

mthalman:
The latest commit requires the new version of Cottle from https://github.com/dotnet/docker-tools/pull/938.

MichaelSimons (commented):
New Cottle support for multiple statements makes these changes nice and readable.  I also like the elimination of the ARGS.

MichaelSimons (approved):
:shipit:

PR #3386, opened 2022-01-07
Fix various inconsistencies in aspnet and runtime Dockerfiles

mthalman:


eng/dockerfile-templates/runtime/7.0/Dockerfile.nanoserver
```diff
@@ -5,7 +5,7 @@ FROM mcr.microsoft.com/windows/servercore:{{OS_VERSION_NUMBER}}-amd64 AS install
 
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
-# Retrieve .NET Runtime
+# Retrieve .NET
```
MichaelSimons, line 8: Arguably `Retrieve .NET Runtime` is a better choice as this is specifically installing the runtime.  Installing .NET is rather ambiguous.

MichaelSimons (approved):

MichaelSimons (approved):

PR #3387, opened 2022-01-07
Add runtime environment variable to dotnet-monitor image

wiktork:
Add an environment variable to dotnet-monitor to simplify detection of sidecar scenarios.

- I am not aware of a way to share ARG's between different bases.
- I can push the inline ENV to top level if it's an issue for subsequent dependency updates.

### DIFF
--- a/eng/dockerfile-templates/monitor/6.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/monitor/6.0/Dockerfile.alpine
@@ -31,6 +31,8 @@ ENV \
     # Default Filter
     DefaultProcess__Filters__0__Key=ProcessId \
     DefaultProcess__Filters__0__Value=1 \
+    # Use the environment variable to determine if we are in a sidecar scenario.
+    DOTNET_MONITOR_VERSION={{VARIABLES["monitor|6.0|build-version"]}} \
     # Logging: JSON format so that analytic platforms can get discrete entry information
     Logging__Console__FormatterName=json \
     # Logging: Use round-trip date/time format without timezone information (always logged in UTC)
--- a/eng/dockerfile-templates/monitor/7.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/monitor/7.0/Dockerfile.alpine
@@ -31,6 +31,8 @@ ENV \
     # Default Filter
     DefaultProcess__Filters__0__Key=ProcessId \
     DefaultProcess__Filters__0__Value=1 \
+    # Use the environment variable to determine if we are in a sidecar scenario.
+    DOTNET_MONITOR_VERSION={{VARIABLES["monitor|7.0|build-version"]}} \
     # Logging: JSON format so that analytic platforms can get discrete entry information
     Logging__Console__FormatterName=json \
     # Logging: Use round-trip date/time format without timezone information (always logged in UTC)
--- a/src/monitor/6.0/alpine/amd64/Dockerfile
+++ b/src/monitor/6.0/alpine/amd64/Dockerfile
@@ -31,6 +31,8 @@ ENV \
     # Default Filter
     DefaultProcess__Filters__0__Key=ProcessId \
     DefaultProcess__Filters__0__Value=1 \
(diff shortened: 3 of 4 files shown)


jander-msft (approved):

eng/dockerfile-templates/monitor/6.0/Dockerfile.alpine
```diff
@@ -31,6 +31,8 @@ ENV \
     # Default Filter
     DefaultProcess__Filters__0__Key=ProcessId \
     DefaultProcess__Filters__0__Value=1 \
+    # Use the environment variable to determine if we are in a sidecar scenario.
```
mthalman, line 34: Is the intention that the implementation of .NET Monitor would read this variable to determine the action it should take? That there's specific logic which needs to execute when .NET Monitor is running as a container? If so, it seems that the comment here is very implementation-specific and doesn't actually describe what the variable is.

In addition, if the whole reason for this is to determine whether .NET Monitor is running from the container, should the name perhaps be more specific to that fact? Something like `DOTNET_MONITOR_RUNNING_IN_CONTAINER`? In fact, is there an opportunity here to use the existing `DOTNET_RUNNING_IN_CONTAINER` variable?

https://github.com/dotnet/dotnet-docker/blob/21319881546109b45399ba89265b4e295341c5df/src/runtime-deps/3.1/alpine3.14/amd64/Dockerfile#L17-L18

Would that variable be sufficient to satisfy the scenarios? If someone were to run .NET Monitor in an sdk container, for example, would you want this logic to be in effect?

jander-msft: > Is the intention that the implementation of .NET Monitor would read this variable to determine the action it should take? That there's specific logic which needs to execute when .NET Monitor is running as a container? If so, it seems that the comment here is very implementation-specific and doesn't actually describe what the variable is.
> 
> 
> 
> In addition, if the whole reason for this is to determine whether .NET Monitor is running from the container, should the name perhaps be more specific to that fact? Something like `DOTNET_MONITOR_RUNNING_IN_CONTAINER`? In fact, is there an opportunity here to use the existing `DOTNET_RUNNING_IN_CONTAINER` variable?
> 
> 
> 
> https://github.com/dotnet/dotnet-docker/blob/21319881546109b45399ba89265b4e295341c5df/src/runtime-deps/3.1/alpine3.14/amd64/Dockerfile#L17-L18
> 
> 
> 
> Would that variable be sufficient to satisfy the scenarios? If someone were to run .NET Monitor in an sdk container, for example, would you want this logic to be in effect?

We'll want to detect when the tool is running in its sidecar container. If it's running in the runtime or SDK container, we don't necessarily want to know that (like you said, we can use `DOTNET_RUNNING_IN_CONTAINER` to check that it is running in a container vs not). We need some way to determine when it's running in a process and file system namespace other than that of the target processes that are being diagnosed.

mthalman: > We'll want to detect when the tool is running in its sidecar container. If it's running in the runtime or SDK container, we don't necessarily want to know that (like you said, we can use `DOTNET_RUNNING_IN_CONTAINER` to check that it is running in a container vs not). We need some way to determine when it's running in a process and file system namespace other than that of the target processes that are being diagnosed.

In that case, would it be better to be explicit with the meaning of the variable here? Something like `IS_DOTNET_MONITOR_SIDECAR`?

wiktork: We can make it more specific, but I'm going to wait a bit on these changes. The scenario that we needed this for is likely not going to be needed after all.

mthalman: Any update on this? Can this be closed?

wiktork: I will close this for now, and reopen when it is more applicable (most likely for one of our 7.x releases)

PR #3404, opened 2022-01-13
Break aspnet Dockerfile templates into finer-grained sub-templates

mthalman:
These changes define a set of fine-grained aspnet Dockerfile sub-templates that enables more sharing of content than the previous templates. This has allowed all the templates to be distilled down primarily to a small set of templates pivoting on two characteristics: OS type and single-stage/multi-stage.

The primary Dockerfile templates are `eng/dockerfile-templates/aspnet/Dockerfile.linux` and `eng/dockerfile-templates/aspnet/Dockerfile.windows`. From there, it breaks off into a sub-template for either single-stage or multi-stage implementation. There are even further sub-templates for handling common content between them all such as installing ASP.NET and defining the environment variables.


mthalman:
I realized I used operator functions instead of operator symbols so I'm going to fix that up to make things cleaner to read.

eng/dockerfile-templates/runtime/Dockerfile.linux.install-aspnet
```diff
@@ -0,0 +1,21 @@
+{{
```
MichaelSimons, line 1: Is this file referenced?  Its existence doesn't seem right.  It is an aspnet template in the runtime directory.  Same goes for the next file (windows version).

eng/dockerfile-templates/aspnet/Dockerfile.linux.multi-stage
```diff
@@ -0,0 +1,23 @@
+{{
```
MichaelSimons, line 1: Did you consider combining the multi-stage and single-stage into a single template?  Obviously there are tradeoffs.  The value is seeing both patterns together to help ensure consistency and that the two final stages do not diverge.  To be clear, I don't see value in trying to integrate the two inline rather combining the two templates into an if else but sharing the common variables.

eng/dockerfile-templates/runtime/Dockerfile.linux.install-aspnet
mthalman: Deleted them.

MichaelSimons (approved):
It's really nice to see so many duplicated templates removed/reduced.

PR #3426, opened 2022-01-21
Add major-only version tags for .NET Monitor

jander-msft:
.NET Monitor now has multiple versions for 6.x images. It would be useful to add major-only version tags to allows customers to pull the latest 6.x image. All 6.x images are backward compatible with lower minor versions.

I've applied the same for the 7.x image line so that customers can use the 7 major tags from the beginning of the 7.x development.

cc @dotnet/dotnet-monitor

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -6007,8 +6007,10 @@
           "sharedTags": {
             "$(monitor|6.1|product-version)-alpine": {},
             "6.1-alpine": {},
+            "6-alpine": {},
             "$(monitor|6.1|product-version)": {},
-            "6.1": {}
+            "6.1": {},
+            "6": {}
           },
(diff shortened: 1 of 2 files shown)


manifest.json
```diff
@@ -6032,8 +6035,10 @@
           "sharedTags": {
             "$(monitor|7.0|product-version)-alpine": {},
             "7.0-alpine": {},
+            "7-alpine": {},
```
jander-msft, line 6038: nit: remove extra whitespace

jander-msft: And now it seems to be gone. Might be rendering artifact in GitHub UI. I'll check in more detail before merging.

mthalman (approved):

PR #3430, opened 2022-01-24
Fix Dockerfile inconsistencies

mthalman:
Addresses the following inconsistencies:
* Refer to ".NET Core Runtime" instead of just ".NET Core".
* Use consistent ordering of `tar` command options.
* Deletion of tarball should occur immediately after extracting it, before any other commands.
* Set build version to a local shell variable if it doesn't exist as an environment variable instead of inlining the version in a URL.

### DIFF
--- a/eng/dockerfile-templates/runtime/3.1/Dockerfile.nanoserver
+++ b/eng/dockerfile-templates/runtime/3.1/Dockerfile.nanoserver
@@ -17,7 +17,7 @@ RUN powershell -Command `
         }; `
         `
         mkdir dotnet; `
-        tar -C dotnet -oxzf dotnet.zip; `
+        tar -oxzf -C dotnet dotnet.zip; `
         Remove-Item -Force dotnet.zip
 
 
--- a/eng/dockerfile-templates/runtime/3.1/Dockerfile.nanoserver
+++ b/eng/dockerfile-templates/runtime/3.1/Dockerfile.nanoserver
@@ -17,7 +17,7 @@ RUN powershell -Command `
         }; `
         `
         mkdir dotnet; `
-        tar -oxzf -C dotnet dotnet.zip; `
+        tar -oxzf dotnet.zip -C dotnet; `
         Remove-Item -Force dotnet.zip
 
 
--- a/src/aspnet/6.0/cbl-mariner1.0/amd64/Dockerfile
+++ b/src/aspnet/6.0/cbl-mariner1.0/amd64/Dockerfile
@@ -8,8 +8,7 @@ ENV \
     Logging__Console__FormatterName=Json
 
 # Install ASP.NET Core
-RUN aspnetcore_version=6.0.1 \
-    && curl -fSL --output aspnetcore.rpm https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-x64.rpm \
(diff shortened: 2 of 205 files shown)


src/aspnet/6.0/cbl-mariner1.0/amd64/Dockerfile
```diff
@@ -8,7 +8,8 @@ ENV \
     Logging__Console__FormatterName=Json
 
 # Install ASP.NET Core
-RUN curl -fSL --output aspnetcore.rpm https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$ASPNET_VERSION/aspnetcore-runtime-$ASPNET_VERSION-x64.rpm \
+RUN aspnetcore_version=6.0.1 \
```
MichaelSimons, line 11: Can `ASPNET_VERSION` be used in place of this?  It doesn't seem right that a local var is defined when an `ENV` exists.

eng/dockerfile-templates/runtime/3.1/Dockerfile.nanoserver
```diff
@@ -17,7 +17,7 @@ RUN powershell -Command `
         }; `
         `
         mkdir dotnet; `
-        tar -C dotnet -oxzf dotnet.zip; `
+        tar -oxzf -C dotnet dotnet.zip; `
```
MichaelSimons, line 20: Ordering of the parameters is inconsistent here.  

```suggestion
        tar -oxzf dotnet.zip -C dotnet; `
```

src/runtime/5.0/bullseye-slim/amd64/Dockerfile
```diff
@@ -15,7 +16,9 @@ RUN curl -fSL --output dotnet.tar.gz https://dotnetcli.azureedge.net/dotnet/Runt
 # .NET runtime image
 FROM $REPO:5.0.13-bullseye-slim-amd64
 
-ENV DOTNET_VERSION=5.0.13
+ENV \
```
MichaelSimons, line 19: Why use a multiline ENV versus placing the comment before a single line ENV?  Does it help keep the templates simpler?  I think it detracts from the readability of the Dockerfile.

mthalman: Yes, it does keep the templates a bit simpler. Consider the case of Linux and Windows:

Linux:

```Dockerfile
ENV DOTNET_VERSION=6.0.1
```

Windows:

```Dockerfile
ENV `
    # Configure web servers to bind to port 80 when present
    ASPNETCORE_URLS=http://+:80 `
    DOTNET_RUNNING_IN_CONTAINER=true `
    # .NET Runtime version
    DOTNET_VERSION=6.0.1
```

For consistency, it would make sense for Linux to be commented in the same way as Windows:

```Dockerfile
# .NET Runtime version
ENV DOTNET_VERSION=6.0.1
```

But now to have a common template between Linux and Windows, it gets messy because the common is followed by `ENV` in the Linux case but in the Windows case is just followed by the variable name.

After looking at this again, I agree that this doesn't look all that great. It's probably ok to sacrifice a little bit of reusability within the templates in order to not uglify the Dockerfiles.

eng/dockerfile-templates/runtime/3.1/Dockerfile.nanoserver
mthalman: Fixed

src/aspnet/6.0/cbl-mariner1.0/amd64/Dockerfile
mthalman: Yep, that was unintended

MichaelSimons (approved):

PR #3432, opened 2022-01-24
Refactor update-dependencies pipeline for multiple .NET Monitor versions

jander-msft:
I've refactored the update-dependencies pipeline to more easily add .NET Monitor versions that need to participate in updates. Each major version of .NET Monitor is updated by separate jobs and the pipeline will have three variables per major version:
- Minor version number
- Channel quality
- Use stable branding

The generated PR should have titles that look like `[nightly] Update dependencies from dotnet/dotnet-monitor/6.1` which branches that look like `nightly-UpdateDependencies-nightly-From-dotnet-dotnet-monitor-6.1`

cc @dotnet/dotnet-monitor


mthalman (commented):
A feel like a more maintainable and cleaner solution would be to use a [pipeline matrix strategy](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/jobs-job-strategy?view=azure-pipelines). That will automatically generate the jobs instead of needing to explicitly define all of them in YAML.

jander-msft:
> A feel like a more maintainable and cleaner solution would be to use a [pipeline matrix strategy](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/jobs-job-strategy?view=azure-pipelines). That will automatically generate the jobs instead of needing to explicitly define all of them in YAML.

Are you suggesting that I keep the refactored template or to inline the template again (or you have no opinion on that), and use matrix strategy in either case?

mthalman:
> Are you suggesting that I keep the refactored template or to inline the template again (or you have no opinion on that), and use matrix strategy in either case?

Inlining it would probably be fine but I'm open to either.

jander-msft:
> > Are you suggesting that I keep the refactored template or to inline the template again (or you have no opinion on that), and use matrix strategy in either case?
> 
> Inlining it would probably be fine but I'm open to either.

I've inlined it as it was originally and now use matrix strategy.

mthalman (approved):

PR #3446, opened 2022-01-28
Use build versions in .NET Monitor URLs and stable branding.

jander-msft:
This change updates the .NET Monitor templates to use the product version for the package version and the build version as part of the blob container name for release and servicing builds. The builds of .NET Monitor publish their nupkg files under a blob container that has the build version as one of its folders to allow stable version builds to not collide. Prerelease builds produce package versions that are unique for every build, hence why the 7.0 template still uses the build version for its package version.

The 6.0 and 6.1 lines have been updated to the latest builds using stable branding.

cc @dotnet/dotnet-monitor

### DIFF
--- a/eng/dockerfile-templates/monitor/7.0/Dockerfile.alpine
+++ b/eng/dockerfile-templates/monitor/7.0/Dockerfile.alpine
@@ -6,7 +6,7 @@ FROM $SDK_REPO:{{VARIABLES["sdk|6.0|product-version"]}}-{{OS_VERSION}}{{ARCH_TAG
 
 # Install .NET Monitor
 ENV DOTNET_MONITOR_VERSION={{VARIABLES["monitor|7.0|build-version"]}}
-RUN wget -O dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg https://dotnetcli.azureedge.net/dotnet/diagnostics/monitor/$DOTNET_MONITOR_VERSION/dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg \
+RUN wget -O dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg https://dotnetcli.azureedge.net/dotnet/diagnostics/monitor/{{VARIABLES["monitor|7.0|build-version"]}}/dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg \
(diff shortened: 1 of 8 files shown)


eng/dockerfile-templates/monitor/7.0/Dockerfile.alpine
```diff
@@ -6,7 +6,7 @@ FROM $SDK_REPO:{{VARIABLES["sdk|6.0|product-version"]}}-{{OS_VERSION}}{{ARCH_TAG
 
 # Install .NET Monitor
 ENV DOTNET_MONITOR_VERSION={{VARIABLES["monitor|7.0|build-version"]}}
```
mthalman, line 8: I noticed that this didn't get changed to `product-version` like the others. Is that intentional?

jander-msft: Yes, builds without a final version have package versions that are the same as the build version.  When it comes closer to when we want to GA the 7.0 product, it will switch to product version.

So it boils down to the fact that 7.0 does not have a final version for its package whereas 6.0 and 6.1 do have final versions.

mthalman: Can't this be reflected in the way on which the build versions are specified in the manifest.versions.json file? As an example, for the other image types, the Dockerfiles reference the build version. For 7.0 preview versions, which is the only version we're actively updating in the nightly branch, the build version will keep updating for each new build we integrate. But for release and merging to main, the build version gets set to the stable version, matching the product version. Can this same approach be taken?

jander-msft: > But for release and merging to main, the build version gets set to the stable version, matching the product version. Can this same approach be taken?

I agree that we would like it to be this way for .NET Monitor, but at this time it cannot be. The reason why runtime/aspnet/sdk can do this is because someone on the release team effectively copies the GA build from the build version folder to the stable version folder in the blob storage account on release day (or shortly before that), hence why those products can use a stable version for their build version. .NET Monitor cannot do that because we don't have the ability to do so and I've been told there is no way for .NET Monitor to participate in the existing mechanism. I will try to work with them to give us that ability, but the answer might ultimately be a flat-out 'no' at the end of the day.

mthalman (approved):

PR #3457, opened 2022-02-02
Use consistent variable names between .NET versions

mthalman:
All Dockerfiles 5.0 and above use `sdk_version` instead of `dotnet_sdk_version` for the variable in the Dockerfile so I'm standardizing on that. Plus it's just a shorter name which is preferable.


MichaelSimons (approved):
`dotnet_sdk_version` would align with  `ENV DOTNET_SDK_VERSION`.  That might Not sure if that would help the readability or not.  The shorter name is nice though.

PR #3473, opened 2022-02-07
Servicing release for .NET 5.0/6.0

mthalman:
Related to https://github.com/dotnet/dotnet-docker/issues/3471


MichaelSimons (commented):
LGTM albeit a lot of changes.  These two commits confused me at first

1. https://github.com/dotnet/dotnet-docker/pull/3473/commits/ee4e7e651873d884eb90106997ea2f88431b94a5
1. https://github.com/dotnet/dotnet-docker/pull/3473/commits/af44940d5b1ecc8d3778be203bca0468a176b9cc

One possible practice going forward would be to try limit PRs to a single fix.  Formatting changes or infra should be checked in separately.  This may cause more work up-front but it makes integrations like this easier and makes the commit history easier to follow.

mthalman:
> One possible practice going forward would be to try limit PRs to a single fix. Formatting changes or infra should be checked in separately. This may cause more work up-front but it makes integrations like this easier and makes the commit history easier to follow.

Good suggestion. I did mark those particular PRs as "partial" in https://github.com/dotnet/dotnet-docker/issues/3471. But I could have called that out here in the PR comment as well.

PR #3524, opened 2022-02-25
Breakup dotnet-monitor's entrypoint

kelltrick:
Break's up `dotnet-monitor`'s `ENTRYPOINT` to more easily allow access to other `dotnet-monitor` commands and configurations.
Closes dotnet/dotnet-monitor#1279

### DIFF
--- a/eng/dockerfile-templates/monitor/Dockerfile
+++ b/eng/dockerfile-templates/monitor/Dockerfile
@@ -72,4 +72,5 @@ ENV \
 
 RUN chmod 755 dotnet-monitor
 
-ENTRYPOINT [ "dotnet-monitor", "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
+ENTRYPOINT [ "dotnet-monitor" ]
+CMD [ "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
--- a/eng/dockerfile-templates/monitor/Dockerfile
+++ b/eng/dockerfile-templates/monitor/Dockerfile
@@ -2,6 +2,7 @@
     _ .NET major version matches the major version of dotnet-monitor ^
     set dotnetMajor to split(PRODUCT_VERSION, ".")[0] ^
     set dotnetMajorMinor to cat(dotnetMajor, ".0") ^
+    set monitorMajor to split(PRODUCT_VERSION, ".")[0] ^
     set targetFrameworkMoniker to cat("net", dotnetMajorMinor) ^
     set targetFrameworkMonikerRegex to cat("net", dotnetMajor, "[.]0") ^
     set installerBaseTag to cat(VARIABLES[cat("sdk|", dotnetMajorMinor, "|product-version")], "-", OS_VERSION, ARCH_TAG_SUFFIX) ^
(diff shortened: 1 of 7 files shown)


kelltrick:
@jander-msft, here is the PR for dotnet/dotnet-monitor#1279

eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -72,4 +72,5 @@ ENV \
 
 RUN chmod 755 dotnet-monitor
 
-ENTRYPOINT [ "dotnet-monitor", "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
+ENTRYPOINT [ "dotnet-monitor" ]
```
jander-msft, line 75: This change needs to be conditioned so that it only applies to 7.0 and higher in order to not introduce a breaking change.

eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -72,4 +72,5 @@ ENV \
...
+CMD [ "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
```
jander-msft, line 76: Tests are failing due to this change. Please update the tests to understand that the `collect` command has been added to the `CMD` part of the image.

jander-msft (changes requested):

eng/dockerfile-templates/monitor/Dockerfile
jander-msft: I would suggest moving the ENTRYPOINT and CMD into two separate templates (both containing their corresponding values for each directive) and then include either template based on the `dotnet monitor` version.

jander-msft: My suggestion was that you put

```
ENTRYPOINT [ "dotnet-monitor", "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
````

in a separate template file named something like `Dockerfile.entrypoint.6` and then put

```
ENTRYPOINT [ "dotnet-monitor" ]
CMD [ "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
```

in a separate template file named something like `Dockerfile.entrypoint.7`.

Then you can include it with one line, something like:

```
{{InsertTemplate(cat("Dockerfile.entrypoint.", monitorMajor))}}
```

This will make it easier to just say "this major version of dotnet monitor has this entrypoint" rather than describing it in terms of how the entrypoint changes from version to version. If 8.0.0 comes along with another differentiation, then it's easy to drop in a new file rather than refactoring the existing logic.

File: tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs [hunk 3]
```diff
@@ -440,9 +447,19 @@ private static SampleImageData GetSampleImageData(MonitorImageData imageData)
                 .First(d => d.IsPublished = true && d.Arch == imageData.Arch);
         }
 
-        private static string GetMonitorAdditionalArgs(bool noAuthentication)
+        private static string GetMonitorAdditionalArgs(MonitorImageData imageData, bool noAuthentication)
         {
-            return noAuthentication ? Switch_NoAuthentication : null;
+            string cmdsToAdd = noAuthentication ? Switch_NoAuthentication : null;
```
jander-msft, line 452: Use a StringBuilder and start with the logic for adding Switch_DefaultImageCmd

eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -14,6 +14,10 @@
     _ All .NET Monitor versions publish to the same storage account, thus use the latest storage account scheme.
       However, the latest 6.0 release still uses dotnetcli storage account. ^
     set urlVersion to when(monitorMajorMinor != "6.0", "7.0", "6.0") ^
+    _ Using Entrypoint or Entrypoint + Cmd is tied to the dotnet-monitor version, however the existance of minor versions of 6.x make it easier to compare to dotnetMajorMinor. ^
+    _ This breaking change is intended to apply to all versions including 7.0 and after. ^
+    set fullEntrypoint to dotnetMajor = "6" ^
```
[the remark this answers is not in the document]
kelltrick: removed this and added `monitorMajor` up at the top.

eng/dockerfile-templates/monitor/Dockerfile
kelltrick: This is done with templates now,

eng/dockerfile-templates/monitor/Dockerfile
kelltrick: I think the test changes will fix everything... We'll see.

jander-msft (approved):

mthalman (approved):

Issue #3520, opened 2022-02-23
Remove workaround for Mariner ARM container images not setting architecture variant metadata


mthalman:
Tracking issue for https://microsoft.visualstudio.com/OS/_workitems/edit/38266794

mthalman:
Apparently the variant metadata is now set: https://github.com/microsoft/CBL-Mariner/issues/4856

This should allow us to remove the workaround from https://github.com/dotnet/dotnet-docker/blob/f68950a9244e5c24b68ab3826a0378d3678a92f2/eng/pipelines/steps/set-imagebuilder-build-args-var.yml

lbussell:
Fixed by https://github.com/dotnet/dotnet-docker/pull/5726

Issue #3455, opened 2022-02-02
New Linux Distro Version - CBL-Mariner 2.0


mthalman:
# New Linux Distro Version

Distro: CBL-Mariner 2.0

## ImageBuilder Tasks

- [x] Ensure that the ImageBuilder supports the new distro version in the [code](https://github.com/dotnet/docker-tools/blob/main/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs) to generate the correct README display name from the version specified in the manifest

## Nightly Branch Tasks

1. - [x] Copy the Dockerfiles of the most recent version of that distro (or author new ones for an entirely new distro) and place them in a version-specific folder under their respective variants (runtime-deps, runtime, aspnet, sdk). Note: not all variants have a corresponding runtime-deps image.
1. - [x] Modify the Dockerfiles as appropriate for any specific changes related to the new distro version
1. - [x] Update [manifest.json](https://github.com/dotnet/dotnet-docker/blob/nightly/manifest.json) to reference the new set of Dockerfiles with the appropriate tags
      - [ ] Move any distro-specific floating tags to the newer version (e.g. `3.1-alpine`)
1. - [x] Update the [test data](https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/TestData.cs) to include the new distro version
1. - [x] Update the [tags metadata templates](https://github.com/dotnet/dotnet-docker/tree/main/eng/mcr-tags-metadata-templates) to include the new distro version
1. - [x] Run the command to update the READMEs: `.\eng\readme-templates\Get-GeneratedReadmes.ps1`
1. - [x] Run the command to update the image size baseline file: `.\tests\performance\Validate-ImageSize.ps1 -UpdateBaselines`
1. - [x] Inspect generated changes for correctness
1. - [x] Consider whether sample Dockerfiles should be authored if this is a new distro and them to the [samples](https://github.com/dotnet/dotnet-docker/tree/main/samples)
1. - [x] Run the command to build and test your changes: `.build-and-test.ps1 -OS <os>`
1. - [x] Commit generated changes
1. - [x] Create PR
1. - [x] Get PR signoff
1. - [x] Merge PR to nightly branch
1. - [x] Wait for automatically queued CI build to finish on [dotnet-docker-nightly pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=359) (internal MSFT link)
[...]

PR #3532, opened 2022-03-02
Add Dockerfiles for CBL-Mariner 2.0

mthalman:
This adds Dockerfiles for both AMD64 and Arm64. For Arm64, tarballs are being used to install .NET instead of RPMs. Once we've got published Arm64 RPMs, we'll transition to them.

A build init step was added in order to workaround https://github.com/dotnet/dotnet-docker/issues/3520. This disables the platform validation check during build.

Dockerfiles of CBL-Mariner 1.0 for .NET 7 were removed because that version of CBL-Mariner will not be in support by the time .NET 7 is released.

Related to #3455


mthalman:
Looks like I still need to add image size baselines.

eng/package-scripts/cbl-mariner-distroless/get-upgradable-packages.2.0.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env sh
```
MichaelSimons, line 1: This shouldn't block this work but have explored alternative solutions that would avoid these version specific scripts?  Eliminating these would be nice as they reduce the manual steps when onboarding and deprecating new Mariner versions.

MichaelSimons (approved):

eng/package-scripts/cbl-mariner-distroless/get-upgradable-packages.2.0.sh
mthalman: Yes, it'd be a matter of adding support to define arguments to the scripts via the manifest file. But I don't see the investment to do that is worth it versus just defining a version-specific script.

PR #3565, opened 2022-03-15
Dockerfile template support for internal .NET builds

mthalman:
These changes update Dockerfile templates so that they are capable of generating Dockerfiles which can target internal .NET builds (builds that require a blob container SAS token in order to access).

The Dockerfile pattern used to do this is the following:
* Include `ARG SAS_QUERY_STRING` so that URL query string containing the SAS token can be passed as part of `docker build`.
* Append the `SAS_QUERY_STRING` variable to the end of the .NET blob container URL.
* Ensure that a multi-stage Dockerfile is used in order to prevent the value of `SAS_QUERY_STRING` from being included in the published container image. This accounts for most of the changes here because it requires restructuring or adding conditional checks throughout the Dockerfile in order to support this.

So how is it known whether the Dockerfile template should produce Dockerfiles that target internal builds? This information is embedded in the manifest.versions.json file. There are now a set of variables indicating the base URLs to use for the various branches:

```json
"base-url|public|main": "https://dotnetcli.azureedge.net/dotnet",
"base-url|public|nightly": "https://dotnetbuilds.azureedge.net/public",
"base-url|internal|main": "https://dotnetclimsrc.blob.core.windows.net/dotnet",
"base-url|internal|nightly": "https://dotnetbuilds.blob.core.windows.net/internal",

"base-url|public|legacy|main": "$(base-url|public|main)",
"base-url|public|legacy|nightly": "https://dotnetcli.azureedge.net/dotnet",
"base-url|internal|legacy|main": "$(base-url|internal|main)",
"base-url|internal|legacy|nightly": "https://dotnetclimsrc.blob.core.windows.net/dotnet",
```

The term 'legacy' used in the variable name indicates these are the variables which apply to legacy .NET versions (3.1, 5.0, 6.0).

As with before, there is a base-url variable for each .NET-version/branch combo. But rather than being hardcoded to a URL, it references the corresponding predefined base URL variable:

```json
"base-url|3.1|main": "$(base-url|public|legacy|main)",
```

In this way, it can be determined if the Dockerfile generation is targeting an internal build or not. For example, if the `$(base-url|3.1|main)` variable has a value that matches the value of `$(base-url|internal|legacy|main)`, then we know it's targeting an internal build. If it matches the value of `$(base-url|public|legacy|main)` then we know it's targeting a public build.

Related to #2152


mthalman:
Here's a diff of what the 6.0 Dockerfiles would like when generated to target an internal .NET build: https://github.com/mthalman/dotnet-docker/compare/issue2152-dockerfiles...issue2152-dockerfiles-6.0.

MichaelSimons (commented):
A couple high level questions.

1.  Have you considered making the Alpine Dockerfile always be multi-stage?  The concern is around maintenance and ensuring the same behavior between internal and external builds.  The drawback is obviously the public builds do not have to be multi-stage so they have a degree of complexity added.  The question is if this is egregious.
2. I see the mariner images no longer install the rpms.  What is the reason for this?  It seems like this is actually a functional change that could cause consumption issues - e.g. an first party wants to install an rpm app package which depends on the runtime package.

mthalman:
> 1. Have you considered making the Alpine Dockerfile always be multi-stage?  The concern is around maintenance and ensuring the same behavior between internal and external builds.  The drawback is obviously the public builds do not have to be multi-stage so they have a degree of complexity added.  The question is if this is egregious.

I default to the principal of keeping the public Dockerfiles concise and readable. Introducing multiple stages when it's unnecessary violates this principal. The community would then rightly question why it's multi-stage and if there's some hidden gotcha in the Dockerfile that requires it to be that way. To avoid introducing confusion and uncertainty, I would argue that we not change the structure of the public Dockerfiles. I understand your point about ensuring the same behavior between internal and external. But making a change like this only solves it for Alpine and nothing else. So I don't see it buying us much. There are other ways we can investigate to help ensure the internal Dockerfiles remain functional.

> 2. I see the mariner images no longer install the rpms.  What is the reason for this?  It seems like this is actually a functional change that could cause consumption issues - e.g. an first party wants to install an rpm app package which depends on the runtime package.

Fair point. This was done as a means to simplify the implementation. I should be able to accomplish this by downloading the RPMs in the builder stage and then in the final stage copying the RPMs from the builder stage and installing them.

mthalman:
I was able to update the templates to allow for RPM installation to be used by downloading the RPMs in the installer stage and then in the final stage it copies the RPM files and installs them. You can see the difference in the generated Dockerfiles with this commit: https://github.com/mthalman/dotnet-docker/commit/3f40f1ce26805c2a551475d3e5f1673d2639fa28.

PR #3575, opened 2022-03-22
Add authentication validation for .NET Monitor images

### LINKED ISSUE #3209: Add authentication test for dotnet-monitor image
Add tests for the dotnet-monitor image to ensure that authentication is working as expected using the default arguments as specified by the container.

jander-msft:
Only applies to .NET Monitor 6.1 and later, which have the `machinejson` output option for API key generation.

cc @dotnet/dotnet-monitor

closes #3209

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
@@ -150,13 +150,13 @@ namespace Microsoft.DotNet.Docker.Tests
                     if (!Config.IsHttpVerificationDisabled)
                     {
                         // Verify processes returns 401 (Unauthorized) since authentication was not configured.
-                        (await ImageScenarioVerifier.GetHttpResponseFromContainerAsync(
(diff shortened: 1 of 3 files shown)


tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -141,15 +149,15 @@ public Task VerifyMonitorNoHttps(MonitorImageData imageData)
                 {
                     if (!Config.IsHttpVerificationDisabled)
                     {
-                        // Verify processes returns 401 (Unauthorized) since authentication was not provided.
+                        // Verify processes returns 401 (Unauthorized) since authentication was not configured.
                         using HttpResponseMessage processesMessage =
```
mthalman, line 153: It doesn't look like this variable is used. So I think you could be using `VerifyHttpResponseFromContainerAsync` instead. Same comment applies to the `VerifyMonitorNoHttpsWithAuth` test.

mthalman (approved):

mthalman (approved):

PR #3598, opened 2022-03-28
Aspnet targeting pack RPM arch changes & update to .NET 7 Preview 3

mthalman:
New builds for 6.0 and 7.0 now include the architecture label in the filename for the ASP.NET Core targeting pack RPM. These changes update the update-dependencies tool and generated Dockerfiles to account for this change.

Because the nightly branch is still targeting the last publicly released version of 6.0, the 6.0 Dockerfiles need to continue referencing the RPM file without the arch in the name. So the Dockerfile template is only applying this change to 7.0 Dockerfiles currently. This will need to be updated to include 6.0 as part of the merge to main for the next 6.0 release.


MichaelSimons (approved):
Title of this PR is a bit confusing with the inclusion of updating to 7.0 preview 3.

mthalman:
> Title of this PR is a bit confusing with the inclusion of updating to 7.0 preview 3.

Updated the title. Both the tooling change and the Dockerfile updates are needed together in order to have a valid set of repo changes.

PR #3611, opened 2022-04-01
Add CBL Mariner 2.0 images in Arm64v8 for Monitor 6.2 and 7.0 images

kelltrick:
Adds dotnet-monitor images:
- 6.2 dotnet-monitor, CBL Mariner 2.0, Arm64v8
- 7.0 dotnet-monitor, CBL Mariner 2.0, Arm64v8

### DIFF
--- a/eng/mcr-tags-metadata-templates/monitor-tags.yml
+++ b/eng/mcr-tags-metadata-templates/monitor-tags.yml
@@ -6,10 +6,6 @@ $(McrTagsYmlTagGroup:7.0-alpine-arm64v8)
 $(McrTagsYmlTagGroup:6.2-alpine-amd64)
     customSubTableTitle: .NET Monitor Preview Tags
 $(McrTagsYmlTagGroup:6.2-alpine-arm64v8)
-    customSubTableTitle: .NET Monitor Preview Tags
-$(McrTagsYmlTagGroup:7.0-cbl-mariner-arm64v8)
-    customSubTableTitle: .NET Monitor Preview Tags
-$(McrTagsYmlTagGroup:6.2-cbl-mariner-arm64v8)
-    customSubTableTitle: .NET Monitor Preview Tags
+    customSubTableTitle: .NET Monitor Preview Tags10
 $(McrTagsYmlTagGroup:6.1-alpine-amd64)
 $(McrTagsYmlTagGroup:6.0-alpine-amd64)
--- a/eng/mcr-tags-metadata-templates/monitor-tags.yml
+++ b/eng/mcr-tags-metadata-templates/monitor-tags.yml
@@ -7,5 +7,9 @@ $(McrTagsYmlTagGroup:6.2-alpine-amd64)
     customSubTableTitle: .NET Monitor Preview Tags
 $(McrTagsYmlTagGroup:6.2-alpine-arm64v8)
     customSubTableTitle: .NET Monitor Preview Tags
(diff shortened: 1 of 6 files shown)


eng/mcr-tags-metadata-templates/monitor-tags.yml
```diff
@@ -7,5 +7,9 @@ $(McrTagsYmlTagGroup:6.2-alpine-amd64)
     customSubTableTitle: .NET Monitor Preview Tags
 $(McrTagsYmlTagGroup:6.2-alpine-arm64v8)
     customSubTableTitle: .NET Monitor Preview Tags
+$(McrTagsYmlTagGroup:7.0-cbl-mariner-arm64v8)
```
jander-msft, line 10: The CBL-Mariner images are undocumented, so these should not be added here

jander-msft (changes requested):
One correction that should also unblock the builds

tests/performance/ImageSize.nightly.linux.json
```diff
@@ -235,8 +235,10 @@
     "src/monitor/6.2/alpine/amd64": 107698176,
     "src/monitor/6.2/alpine/arm64v8": 118521282,
     "src/monitor/6.2/cbl-mariner/amd64": 350489871,
+    "src/monitor/6.2/cbl-mariner/arm64v8": 360000000,
```
kkeirstead, line 238: Since we allow a range around the actual value this probably doesn't matter, but is there a reason why we're dropping precision for these?

eng/mcr-tags-metadata-templates/monitor-tags.yml
```diff
@@ -6,6 +6,6 @@ $(McrTagsYmlTagGroup:7.0-alpine-arm64v8)
 $(McrTagsYmlTagGroup:6.2-alpine-amd64)
     customSubTableTitle: .NET Monitor Preview Tags
 $(McrTagsYmlTagGroup:6.2-alpine-arm64v8)
-    customSubTableTitle: .NET Monitor Preview Tags
+    customSubTableTitle: .NET Monitor Preview Tags10
```
kkeirstead, line 9: I'm assuming this was accidental?

mthalman:
Make sure to run `eng/dockerfile-templates/Get-GeneratedDockerfiles.ps1`. Looks like the Dockerfiles you added are not up-to-date.

tests/performance/ImageSize.nightly.linux.json
jander-msft: Not sure how to calculate these values locally since you cannot build arm64 images without arm64 hardware. Maybe we can check these in, wait for the images to be built, run the [size validation pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=379&_a=summary), and then update the baseline values from that run.

Maybe @mthalman has a better suggestion

mthalman: I usually run a build targeting a private branch to produce the images. But that seems overkill for this change. I think it's fine to just specify a good estimate of the size.

jander-msft:
Consider splitting this PR into two: one for the 6.2 addition and one for the 7.0 addition.
This will make it easier to port changes from nightly to main in the event that these two versions are shipped at different times.

eng/mcr-tags-metadata-templates/monitor-tags.yml
kelltrick: yep, that was a typo while I was deleting the doc line

Issue #3642, opened 2022-04-13
Breaking Change: Revert the Console Formatter Default in .NET 6


mthalman:
In the .NET 6 `aspnet` container images we made the decision to default the console logger in `CreateDefaultBuilder()` to structured JSON logs in order to support container environments where folks use tools like fluentd or Azure Monitor to collect those logs. However, this resulted in non-human readable output that does not work well when viewing this output to analyze logs. [Please see this issue for more details](https://github.com/dotnet/dotnet-docker/issues/2725).  We have made the decision to revert the default behavior back to multiline human-readable output.

### Servicing Update
In the May 2022 servicing update for .NET 6, we will change the `Logging__Console__FormatterName` environment variable to be unset for the ASP.NET Core container images. This will cause it to revert back to the original simple output.

Output of original .NET 6 behavior with `Logging__Console__FormatterName` set to `Json`:

```{"EventId":37,"LogLevel":"Warning","Category":"Microsoft.AspNetCore.Server.HttpSys.MessagePump","Message":"Overriding address(es) \u0027\u0027. Binding to endpoints added to UrlPrefixes instead.","State":{"Message":"Overriding address(es) \u0027\u0027. Binding to endpoints added to UrlPrefixes instead.","{OriginalFormat}":"Overriding address(es) \u0027\u0027. Binding to endpoints added to UrlPrefixes instead."}}
{"EventId":0,"LogLevel":"Information","Category":"Microsoft.Hosting.Lifetime","Message":"Now listening on: http://localhost:7000/","State":{"Message":"Now listening on: http://localhost:7000/","address":"http://localhost:7000/","{OriginalFormat}":"Now listening on: {address}"}}
{"EventId":0,"LogLevel":"Information","Category":"Microsoft.Hosting.Lifetime","Message":"Now listening on: http://localhost:7001/","State":{"Message":"Now listening on: http://localhost:7001/","address":"http://localhost:7001/","{OriginalFormat}":"Now listening on: {address}"}}
{"EventId":0,"LogLevel":"Information","Category":"Microsoft.Hosting.Lifetime","Message":"Now listening on: http://localhost:7002/","State":{"Message":"Now listening on: http://localhost:7002/","address":"http://localhost:7002/","{OriginalFormat}":"Now listening on: {address}"}}
[...]

PR #3703, opened 2022-04-13
Remove Json as default console logging formatter in aspnet images for 6.0

mthalman:
The change to default this to Json in 6.0 has been problematic for some customers as evidenced by the thread in https://github.com/dotnet/dotnet-docker/issues/2725. We're making a breaking change to revert this back to the original behavior as announced in https://github.com/dotnet/dotnet-docker/issues/3642. This leaves the `Logging__Console__FormatterName` environment variable unset.


MichaelSimons (changes requested):
The VerifyEnvironmentVariables tests are going to need updating for the aspnet and sdk images

e.g. - https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs#L64

MichaelSimons (approved):

PR #3709, opened 2022-04-13
Update Baseline for dotnet-monitor 7.0 cbl-mariner arm64v8

kelltrick:
This updates the baseline for `src/monitor/7.0/cbl-mariner/arm64v8` to the real value generated by the validate build.


kelltrick:
@mthalman Here is the update for the real baseline value.

jander-msft (approved):

mthalman (approved):

PR #3731, opened 2022-04-25
Remove ICU package from distroless version of Mariner 2.0

### LINKED ISSUE #3394: Exclude icu from distroless CBL-Mariner
The `icu` package is currently included in the distroless version of CBL-Mariner:

https://github.com/dotnet/dotnet-docker/blob/c80eae7c725c4c51a2ea18d0f0bb888f47d49e30/src/runtime-deps/6.0/cbl-mariner1.0-distroless/amd64/Dockerfile#L11

In order to reduce the image size/content, I propose that the `icu` package be excluded and have invariant mode enabled, just as is done for Alpine:

https://github.com/dotnet/dotnet-docker/blob/c80eae7c725c4c51a2ea18d0f0bb888f47d49e30/src/runtime-deps/3.1/alpine3.14/amd64/Dockerfile#L19-L20

This would be a breaking change and thus should be done starting with .NET 7.

mthalman:
This removes the ICU package from distroless Mariner 2.0 for both .NET 6 and 7.

Alpine Dockerfiles ended up having some comments updated in order to have a consistent comment that isn't specific to the name of the ICU package.

Fixes #3394

### DIFF
--- a/eng/dockerfile-templates/Dockerfile.common-dotnet-envs
+++ b/eng/dockerfile-templates/Dockerfile.common-dotnet-envs
@@ -1,10 +1,12 @@
 {{
+    set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set isAlpine to find(OS_VERSION, "alpine") >= 0 ^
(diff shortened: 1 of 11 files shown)


MichaelSimons (commented):
LGTM - friendly reminder to post announcement when this releases in the appropriate channels.

PR #3739, opened 2022-04-27
Adds Dockerfiles for CBL-Mariner 2.0 release

mthalman:
This is primarily adding new Dockerfiles for .NET 6 to be available for CBL-Mariner 2.0 since .NET 7 Preview is already been made available for that version of Mariner. Interestingly, runtime-deps Dockerfiles for .NET 6 on Mariner 2.0 had mistakenly been added to the main branch a while ago. So these changes update those.

Related to https://github.com/dotnet/dotnet-docker/issues/3455


src/runtime-deps/6.0/cbl-mariner2.0/amd64/Dockerfile
```diff
@@ -14,9 +14,9 @@ RUN tdnf install -y \
     && tdnf clean all
 
 # Install dotnet-runtime-deps package
-RUN dotnet_version=6.0.2 \
+RUN dotnet_version=6.0.4 \
```
MichaelSimons, line 17: What was the reason this was not updated previously?

MichaelSimons (approved):

src/runtime-deps/6.0/cbl-mariner2.0/amd64/Dockerfile
mthalman: Some runtime-deps Dockerfiles were mistakenly added to main from a previous merge. But they were never referenced by the manifest so were never updated.

MichaelSimons (approved):

mthalman:
Ok, this is finally ready to take a look at now. Please see the last two commits for the new changes. There were some more package updates needed in order to have gzip and tar as well.

MichaelSimons (approved):

Issue #3769, opened 2022-05-13
BaseUrlUpdater in update-dependencies tool incorrectly identifies dotnet-monitor base urls


jander-msft:
The update-dependencies tool fails when it attempts to update dotnet-monitor versions that are not the same as the overall .NET versions (e.g. 6.1, 6.2). See example failing usage: https://dev.azure.com/dnceng/internal/_build/results?buildId=1769734&view=logs&j=00fa4588-b6b6-574f-ae82-807fb3798490&t=14e894eb-e4bd-5b2c-c189-a6aebe347da5

The issue is that the BaseUrlUpdater constructors the base urls by concatenating `baseUrl`, the docker filer version (e.g. 6.1), the branch. However, the dotnet-monitor urls are not of this format; they look like `baseUrl|6.1-monitor|nightly`. The updater is unable to find the calculated key in the manifest variables dictionary and throws an NRE when it calls ToString on the result.

The same cause is also a separate issue for dotnet-monitor 7.0, where it calculates `baseUrl|7.0|nightly`, but that is not the expected variable for dotnet-monitor, and may cause URL calculation problems if dotnet-monitor's baseUrls every diverge for the versions that happen to align with .NET.

mthalman:
Fixed

PR #3773, opened 2022-05-16
Update dependencies fixes

mthalman:
There are two issues with the update-dependencies tool caused by the changes in https://github.com/dotnet/dotnet-docker/pull/3745:

* An NRE occurs when attempting to update the version for .NET Monitor. This happens because `BaseUrlUpdater` is not constructing the URL variable correctly. This is described in https://github.com/dotnet/dotnet-docker/issues/3769. I've fixed this by updating the logic to construct a variable name specific to .NET Monitor when appropriate.
* The will throw this error for the base .NET Dockerfiles: `System.Exception: 'git status' does not match DependencyInfo information. Git has modified files: False. DependencyInfo is updated: True.` This happens because of `NuGetConfigUpdater` causing it to seem like the file is updated without actually updating the file. The behavior of the `Microsoft.DotNet.VersionTools` infra requires that the `GetUpdateTasks` method only return tasks for things that are actually going to be updated. In thise case, `NuGetConfigUpdater` is doing it all the time regardless of whether the file is updated. I've fixed this to not return anything if the file will have no changes.

Fixes https://github.com/dotnet/dotnet-docker/issues/3769

### DIFF
--- a/eng/update-dependencies/NuGetConfigUpdater.cs
+++ b/eng/update-dependencies/NuGetConfigUpdater.cs
@@ -21,39 +21,54 @@ internal class NuGetConfigUpdater : IDependencyUpdater
     private const string PkgSrcSuffix = "_internal";
     private readonly string _repoRoot;
     private readonly Options _options;
+    private readonly string _configPath;
 
     public NuGetConfigUpdater(string repoRoot, Options options)
     {
         _repoRoot = repoRoot;
         _options = options;
+
+        string configSuffix = (_options.Branch == "nightly" ? ".nightly" : string.Empty);
+        _configPath = Path.Combine(_repoRoot, $"tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config{configSuffix}");
     }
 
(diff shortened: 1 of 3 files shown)


eng/update-dependencies/NuGetConfigUpdater.cs
```diff
@@ -21,39 +21,54 @@ internal class NuGetConfigUpdater : IDependencyUpdater
...
     private readonly string _repoRoot;
     private readonly Options _options;
+    private readonly string _configPath;
 
     public NuGetConfigUpdater(string repoRoot, Options options)
     {
         _repoRoot = repoRoot;
         _options = options;
+
+        string configSuffix = (_options.Branch == "nightly" ? ".nightly" : string.Empty);
```
MichaelSimons, line 31: Curious why this is wrapped in parens.

MichaelSimons (approved):

jander-msft (approved):

eng/update-dependencies/NuGetConfigUpdater.cs
mthalman: Just a personal preference when using the conditional ternary operator.

PR #3774, opened 2022-05-16
Generate readmes for MCR portal

mthalman:
* Updates the manifest to conform to the new schema defined by https://github.com/dotnet/docker-tools/pull/1001.
* Defines new readme files to be used for the MCR portal.
* Refactors the readme templates to enable more common content between the GH and MCR readmes.

### DIFF
--- a/.mcr/portal/README.aspnet.portal.md
+++ b/.mcr/portal/README.aspnet.portal.md
@@ -1,8 +1,9 @@
 ## About
 
-The images from the dotnet/nightly repositories include last-known-good (LKG) builds for the next release of [.NET](https://github.com/dotnet/core).
+**IMPORTANT**
+**The images from the dotnet/nightly repositories include last-known-good (LKG) builds for the next release of [.NET](https://github.com/dotnet/core).**
 
-See [dotnet](https://hub.docker.com/_/microsoft-dotnet/) for images with official releases of [.NET](https://github.com/dotnet/core).
+**See [dotnet](https://hub.docker.com/_/microsoft-dotnet/) for images with official releases of [.NET](https://github.com/dotnet/core).**
 
 This image contains the ASP.NET Core and .NET runtimes and libraries and is optimized for running ASP.NET Core apps in production.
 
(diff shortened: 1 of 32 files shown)


README.aspnet.md
```diff
@@ -1,21 +1,21 @@
-The images from the dotnet/nightly repositories include last-known-good (LKG) builds for the next release of [.NET](https://github.com/dotnet/core).
```
MichaelSimons, line 1: Why move this from the top.  This is very important to highlight upfront for the nightly repos.  We don't want new folks to miss this callout.

manifest.json
```diff
@@ -1,6 +1,8 @@
 {
-  "readme": "README.md",
-  "readmeTemplate": "eng/readme-templates/README.md",
+  "readme": {
```
MichaelSimons, line 2: Did MCR explicitly leave out product family repos?  It feels like there is an opportunity here to provide a better experience for grouping related repos.

mthalman: Yes, they're not supported yet.

MichaelSimons: So it is something on their backlog?

mthalman: Yes

README.aspnet.md
MichaelSimons: My concern with Docker Hub is that the Featured Tags section is first.  My preference is to have this before the Featured Tags on DH.  On MCR the About section is first so it isn't an issue.  The bold Important does stand out and help though.

MichaelSimons (approved):

.mcr/portal/README.aspnet.portal.md
```diff
@@ -0,0 +1,77 @@
...
+.NET Framework:
+
+* [dotnet/framework](https://hub.docker.com/_/microsoft-dotnet-framework/): .NET Framework, ASP.NET and WCF
+* [dotnet/framework/samples](https://hub.docker.com/_/microsoft-dotnet-framework-samples/): .NET Framework, ASP.NET and WCF Samples
+
+## Usage
+
+The [.NET Docker samples](https://github.com/dotnet/dotnet-docker/blob/main/samples/README.md) show various ways to use .NET and Docker together. See [Building Docker Images for .NET Applications](https://docs.microsoft.com/dotnet/core/docker/building-net-docker-images) to learn more.
+
+## Container sample: Run a web application
```
MichaelSimons, line 39: Should this be ###?

mthalman: Good find. Fixed

PR #3775, opened 2022-05-16
[Monitor] Update 6.1 cbl-mariner images from 1.0 to 2.0

kkeirstead:


jander-msft (approved):

manifest.json
```diff
@@ -6975,7 +6975,7 @@
               "dockerfile": "src/monitor/6.1/cbl-mariner/amd64",
               "dockerfileTemplate": "eng/dockerfile-templates/monitor/Dockerfile",
               "os": "linux",
-              "osVersion": "cbl-mariner1.0",
+              "osVersion": "cbl-mariner2.0",
```
jander-msft, line 6978: I believe that the following needs to be added:

```
"packageQueryOverrides": {
  "getInstalledPackagesPath": "eng/package-scripts/cbl-mariner/get-installed-packages.sh"
}
```

similar to the other CBL-Mariner 2.0 images.

jander-msft (changes requested):

manifest.json
kkeirstead: I followed the link to that script, and saw this comment was linked: https://github.com/dotnet/dotnet-docker/issues/1455#issuecomment-1116135242 - should I still be adding it?

mthalman: Yes, it's still necessary. Eventually we can rid of them but we're not there yet.

jander-msft (approved):

PR #3790, opened 2022-05-23
Update image size baselines

mthalman:


tests/performance/ImageSize.nightly.linux.json
```diff
@@ -23,14 +23,14 @@
     "src/runtime-deps/6.0/cbl-mariner1.0/amd64": 229251169,
     "src/runtime-deps/6.0/cbl-mariner1.0-distroless/amd64": 63825551,
     "src/runtime-deps/6.0/cbl-mariner2.0/amd64": 117182974,
-    "src/runtime-deps/6.0/cbl-mariner2.0/arm64v8": 235008327,
+    "src/runtime-deps/6.0/cbl-mariner2.0/arm64v8": 101425521,
     "src/runtime-deps/6.0/cbl-mariner2.0-distroless/amd64": 34587577,
-    "src/runtime-deps/6.0/cbl-mariner2.0-distroless/arm64v8": 23735961,
+    "src/runtime-deps/6.0/cbl-mariner2.0-distroless/arm64v8": 30541634,
```
MichaelSimons, line 28: Curious where the increase is coming from?

mthalman: This change was actually missed from last week so it's not specific to Arm. But looking back at the change made for amd64 and I see what the issue is now. I've logged https://github.com/dotnet/dotnet-docker/issues/3795.

PR #3801, opened 2022-05-26
Reference MAR URLs in readmes for MAR portal

mthalman:
The two sets of readmes that we have between DH and MAR should have links which are relative to those sites. In other words, readmes hosted in DH should contain links to DH pages while readmes hosted in MAR should contain links to MAR pages.

I've updated the readme templates to use a sub-template which will generate the appropriate URL depending on the context of the readme host site.

Since MAR doesn't have product family pages currently, I resorted to using a search query to approximate that.

### DIFF
--- a/.mar/portal/README.aspnet.portal.md
+++ b/.mar/portal/README.aspnet.portal.md
@@ -13,7 +13,6 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
 
 .NET:
 
-* [dotnet](https://mcr.microsoft.com/catalog/search=dotnet): .NET
 * [dotnet/sdk](https://mcr.microsoft.com/product/dotnet/sdk/about): .NET SDK
 * [dotnet/runtime](https://mcr.microsoft.com/product/dotnet/runtime/about): .NET Runtime
 * [dotnet/runtime-deps](https://mcr.microsoft.com/product/dotnet/runtime-deps/about): .NET Runtime Dependencies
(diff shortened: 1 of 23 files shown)


.mar/portal/README.aspnet.portal.md
```diff
@@ -13,26 +13,26 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
...
 .NET:
 
-* [dotnet](https://hub.docker.com/_/microsoft-dotnet/): .NET
-* [dotnet/sdk](https://hub.docker.com/_/microsoft-dotnet-sdk/): .NET SDK
-* [dotnet/runtime](https://hub.docker.com/_/microsoft-dotnet-runtime/): .NET Runtime
-* [dotnet/runtime-deps](https://hub.docker.com/_/microsoft-dotnet-runtime-deps/): .NET Runtime Dependencies
-* [dotnet/monitor](https://hub.docker.com/_/microsoft-dotnet-monitor/): .NET Monitor Tool
-* [dotnet/samples](https://hub.docker.com/_/microsoft-dotnet-samples/): .NET Samples
-* [dotnet/nightly/aspnet](https://hub.docker.com/_/microsoft-dotnet-nightly-aspnet/): ASP.NET Core Runtime (Preview)
+* [dotnet](https://mcr.microsoft.com/catalog/search=dotnet): .NET
```
MichaelSimons, line 16: This url results in - "We are sorry, the page you requested cannot be found."

I lean towards removing this link.  It is just adding one more click for users vs using one of the links below.  It is also showing Azure dotnet images which appear higher in the search results.

MichaelSimons (approved):

PR #3813, opened 2022-06-01
Pipeline support for internal .NET builds

mthalman:
Updates the pipeline infrastructure in order to enable support for internal .NET builds (related to https://github.com/dotnet/dotnet-docker/issues/2152).

These changes are dependent on the common infrastructure changes in https://github.com/dotnet/docker-tools/pull/1011.

The primary work here is to flow in the SAS query string needed to access internal files as well as the NuGet feed password for tests.


eng/pipelines/stages/build-test-publish-repo.yml
```diff
@@ -23,6 +23,48 @@ stages:
     customBuildInitSteps:
     - template: ../../../pipelines/steps/set-imagebuilder-build-args-var.yml
     - template: ../../../pipelines/steps/set-public-source-branch-var.yml
+    - powershell: |
+        $imageBuilderBuildArgs = "$(imageBuilderBuildArgs)"
+        if ("$(publishRepoPrefix)".Contains("/internal/")) {
+          if ("$env:ENABLEINTERNAL3_1" -eq "true") {
```
MichaelSimons, line 29: Are you planning to document the process on how to produce internal images?  This variable usage implies to me that you can only produce internal images for one release at a time.

MichaelSimons (approved):

eng/pipelines/stages/build-test-publish-repo.yml
mthalman: Yes, I can have a doc for this. And yes, the pipeline would only support one version at a time, which is preferable in my mind so that it isolates changes. This is necessary specifically for .NET Core 3.1 because it is stored in a separate blob container and requires different access tokens.

PR #3835, opened 2022-06-10
Set default port for distroless Mariner to a non-privileged port

### LINKED ISSUE #3796: Distroless image binds to privileged port
https://github.com/dotnet/dotnet-docker/blob/4adcfb2ba73d019049c4767b603d72ae299b5c8a/src/runtime-deps/6.0/cbl-mariner2.0-distroless/amd64/Dockerfile#L50

The container uses the non-root user `USER app` but that user doesn't have permission to bind to port 80. 
See https://github.com/dotnet/dotnet-docker/issues/3448

A container using a non-root user should default to using non-privileged ports
ex. http:5000 and https:5001
OR
http:8080 and https:8443

mthalman:
Fixes #3796


mthalman:
> Should a new section be added for Mariner?
> 
> https://github.com/dotnet/dotnet-docker/tree/main/samples/aspnetapp#build-an-image-for-alpine-debian-or-ubuntu

@richlander - Not here, no. We're not documenting Mariner scenarios on the public repo. I can update things for the internal doc.

richlander:
Fair enough. That sounds like something I would have said. Nice.

PR #3840, opened 2022-06-13
Fix Set-DotnetVersions script to allow positional parameter

mthalman:
The changes from https://github.com/dotnet/dotnet-docker/pull/3829 prevent the use of `ProductVersion` as a positional parameter:

```powershell
./Set-DotnetVersions 7.0

A positional parameter cannot be found that accepts argument '7.0'.
```

The fix is to explicitly set its position.

### DIFF
--- a/eng/Set-DotnetVersions.ps1
+++ b/eng/Set-DotnetVersions.ps1
@@ -7,7 +7,7 @@ Updates dependencies for the specified .NET version.
 [cmdletbinding()]
 param(
     # The major/minor version of the product (e.g. 6.0).
-    [Parameter(Mandatory = $true)]
+    [Parameter(Mandatory = $true, Position = 0)]
     [string]
     $ProductVersion,
 


jander-msft:
> ```powershell
> ./Set-DotnetVersions 7.0
> 
> A positional parameter cannot be found that accepts argument '7.0'.
> ```

Is running the script with only ProductVersion as a positional parameter intended to be supported? I've always seen it run (and have run it myself) with `-ProductVersion 7.0`.

mthalman:
> > ```powershell
> > ./Set-DotnetVersions 7.0
> > 
> > A positional parameter cannot be found that accepts argument '7.0'.
> > ```
> 
> Is running the script with only ProductVersion as a positional parameter intended to be supported? I've always seen it run (and have run it myself) with `-ProductVersion 7.0`.

It's invoked in that way in the update-dependencies pipeline: https://github.com/dotnet/dotnet-docker/blob/5ab738db61e9ced5bef232a2f9305ce8d1e27f1b/eng/pipelines/update-dependencies.yml#L23

jander-msft (approved):

MichaelSimons (approved):

PR #3863, opened 2022-06-22
Generalize distroless testing pattern

mthalman:
In order to more easily support distroless containers from a variety of Linux distros, I'm refactoring the testing pattern that we use for distroless.

In order to run the test logic against distroless containers, we need to be able to execute shell commands. This can't be done directly within the distroless container we produce because it contains no shell. The way this was solved with CBL-Mariner was to install the additional packages that we require to run the test on top of the distroless container image. This implementation is specific to CBL-Mariner. I wanted a more general solution that could be applied universally.

There are two categories of data that we need from the distroless container in order to execute a given test:

1. Metadata from the image
1. Specific file information from the container

The first category is easiest because we can use `docker inspect` to get the metadata. This is what was done for tests that need environment variable values and the default user.

The second category is relevant to the `VerifyInsecureFiles` and `VerifyPackageInstallation` tests. Both those tests need a running container in order to query file information. To do this, a Dockerfile was added which loads the entire contents of the distroless container file system into an individual directory within the base container. This base container does include a shell. So commands can then be run from that container targeting the commands against the path in which the distroless container file system was copied to.


tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -26,22 +26,12 @@ public void VerifyEnvironmentVariables(ProductImageData imageData)
 
         [LinuxImageTheory]
         [MemberData(nameof(GetImageData))]
-        public void VerifyDistrolessRunsAsNonRootUser(ProductImageData imageData)
+        public void VerifyDefaultUser(ProductImageData imageData)
```
MichaelSimons, line 29: This test feels like it has some value in running it against every image.

MichaelSimons (approved):
Nice abstraction that will be nice as we onboard additional distroless images.

richlander:
This looks solid. This is similar to copying rootfs from one installation to another.

MichaelSimons (approved):

PR #3868, opened 2022-06-23
.NET Monitor CBL-Mariner Distroless Images

jander-msft:
These changes add CBL-Mariner distroless images for .NET Monitor for the 6.2 and 7.0 versions.

Image baseline sizes were calculated using the relative difference between the ASP.NET and .NET Monitor images for the CBL-Mariner images (non-distroless).

Other adjustments were made so that file permissions were changed while on the SDK container instead of the ASP.NET container since `chmod` is not available in distroless images. Tested the permissions on a few of the images locally.

cc @dotnet/dotnet-monitor


mthalman (approved):
Very nice

PR #3873, opened 2022-06-24
Filter tests in build stage by runtime version

### LINKED ISSUE #3871: Tests for .NET Monitor 6.1/6.2 images are not being run in pipelines
No tests for version 6.1 and 6.2 of .NET Monitor are being executed at all in the pipelines. This is because the matrix that gets generated for those images gets associated with version "6.0". That's the .NET version those images have dependencies on. So the run-tests.ps1 script gets executed with 6.0 as the Version parameter: https://github.com/dotnet/dotnet-docker/blob/0961bc01f03010d4cfce01eefba0a88cabd62b20/tests/run-tests.ps1#L9

This excludes the 6.1 and 6.2 versions of .NET Monitor and never tests those images.

jander-msft:
### Background

The PR builds test images grouped by runtime version whereas CI builds test images grouped by product version. This is not an issue for runtime-deps/runtime/aspnet/sdk because their runtime version is the same as the product version. For .NET Monitor images, this is an issue because the runtime version (6.0 or 7.0) is different than the product version (6.1, 6.2, and 7.0).

### Fix

Add additional environment variable to the tests that allows filtering by runtime version. Update the build files to filter by runtime version in PR builds and to filter by product version in CI builds. Made minor changes to add a VersionedImageData class to the tests so that common filtering mechanisms can be used across all of the .NET images.

fixes #3871

### DIFF
--- a/eng/pipelines/stages/build-test-publish-repo.yml
+++ b/eng/pipelines/stages/build-test-publish-repo.yml
@@ -60,6 +60,11 @@ stages:
               # test step.
               $testInit='$Env:SAS_QUERY_STRING=' + "'$sasQueryString'" + '; $Env:NUGET_FEED_PASSWORD=''$(dn-bot-dnceng-artifact-feeds-rw)'''
           }
+
+          # Forward build reason for consumption by test script
+          if ($env:BUILD_REASON) {
+            $testRunnerOptions="$testRunnerOptions -e BUILD_REASON='$env:BUILD_REASON'"
+          }
         }
         
         echo "##vso[task.setvariable variable=testRunner.options]$testRunnerOptions"
--- a/eng/pipelines/stages/build-test-publish-repo.yml
+++ b/eng/pipelines/stages/build-test-publish-repo.yml
@@ -37,20 +37,16 @@ stages:
         }
         echo "##vso[task.setvariable variable=imageBuilderBuildArgs]$imageBuilderBuildArgs"
       displayName: Set Custom Build Variables
-    - ${{ if eq(variables['System.TeamProject'], parameters.publicProjectName) }}:
-      - powershell: |
-          $testRunnerOptions=""
+    - powershell: |
+        # Forward team project name for consumption by test script
+        $testRunnerOptions="-e SYSTEM_TEAMPROJECT='$env:SYSTEM_TEAMPROJECT'"
 
-          # Forward team project name for consumption by test script
-          if ($env:SYSTEM_TEAMPROJECT) {
-            $testRunnerOptions="$testRunnerOptions -e SYSTEM_TEAMPROJECT='$env:SYSTEM_TEAMPROJECT'"
-          }
-
-          echo "##vso[task.setvariable variable=testRunner.options]$testRunnerOptions"
-        displayName: Set Custom Test Variables
+        echo "##vso[task.setvariable variable=testRunner.options]$testRunnerOptions"
+      displayName: Set Custom Test Variables
     customTestInitSteps:
     - powershell: |
-        $testRunnerOptions=""
+        # Forward team project name for consumption by test script
+        $testRunnerOptions="-e SYSTEM_TEAMPROJECT='$env:SYSTEM_TEAMPROJECT'"
         $testInit=""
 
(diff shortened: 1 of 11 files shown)


mthalman (commented):
While I agree that these changes solve the issue, I think a different approach is needed. I don't know yet what that approach should be though. Still need to do some thinking. The issue here is a desire to keep the common infrastructure generalized to support other products using it (besides .NET). Introducing this `RuntimeVersion` parameter is customizing this solution to .NET. And that's what I'd like to be avoided.

mthalman (commented):
Here's a solution that doesn't require any changes to the common infrastructure:

* Revert all the changes in eng/common/templates
* Keep all the test project changes
* Change run-tests.ps1 to not add a `RuntimeVersion` parameter. Instead, have it detect whether it's in a test stage or not. That can be derived by checking the project name. Something like this:

```powershell
if ($env:SYSTEM_TEAMPROJECT -eq "internal") {
    $env:IMAGE_VERSION = $Version
}
else
{
    $env:RUNTIME_VERSION = $Version
}
```

This is going to need to some code comments too because this is certainly a non-obvious scenario.

jander-msft:
> > Here's a solution that doesn't require any changes to the common infrastructure:
> > 
> > * Revert all the changes in eng/common/templates
> > * Keep all the test project changes
> > * Change run-tests.ps1 to not add a `RuntimeVersion` parameter. Instead, have it detect whether it's in a test stage or not. That can be derived by checking the project name. Something like this:
> > 
> > ```powershell
> > if ($env:SYSTEM_TEAMPROJECT -eq "internal") {
> >     $env:IMAGE_VERSION = $Version
> > }
> > else
> > {
> >     $env:RUNTIME_VERSION = $Version
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is going to need to some code comments too because this is certainly a non-obvious scenario.
> 
> I think this solution would be okay for the current usage, but would preclude the ability to correctly test images for PRs created in the internal project.

However, I could use your solution but use BUILD_REASON instead of SYSTEM_TEAMPROJECT.

mthalman:
> I think this solution would be okay for the current usage, but would preclude the ability to correctly test images for PRs created in the internal project.

We don't currently have a scenario for that but we will eventually as part of https://github.com/dotnet/dotnet-docker/issues/2152.

eng/pipelines/stages/build-test-publish-repo.yml
```diff
@@ -37,6 +37,17 @@ stages:
         }
         echo "##vso[task.setvariable variable=imageBuilderBuildArgs]$imageBuilderBuildArgs"
       displayName: Set Custom Build Variables
+    - ${{ if eq(variables['System.TeamProject'], parameters.publicProjectName) }}:
+      - powershell: |
+          $testRunnerOptions=""
+
+          # Forward team project name for consumption by test script
+          if ($env:SYSTEM_TEAMPROJECT) {
```
mthalman, line 45: Is this condition necessary? In what scenario would it ever not be set? Same applies to the change below for `customTestInitSteps`.

eng/pipelines/stages/build-test-publish-repo.yml
```diff
@@ -37,6 +37,17 @@ stages:
         }
         echo "##vso[task.setvariable variable=imageBuilderBuildArgs]$imageBuilderBuildArgs"
       displayName: Set Custom Build Variables
+    - ${{ if eq(variables['System.TeamProject'], parameters.publicProjectName) }}:
```
mthalman, line 40: There's no harm in always setting the variable, even if it's in the internal project, right? To keep things simple, I would suggest removing this condition and always setting the variable.

eng/pipelines/stages/build-test-publish-repo.yml
jander-msft: I've removed the conditions

eng/pipelines/stages/build-test-publish-repo.yml
jander-msft: I've removed the condition

mthalman (approved):

Issue #3870, opened 2022-06-23
Distroless images for Mariner are not being tested in PR builds


mthalman:
The tests for https://github.com/dotnet/dotnet-docker/pull/3868 all succeeded but after merging that PR, the internal build failed: https://github.com/dotnet/dotnet-docker/pull/3869.

The reason for the discrepancy is that the PR build doesn't actually run any tests for the distroless versions of the Mariner images. The reason for this is in the run-tests.ps1 script: https://github.com/dotnet/dotnet-docker/blob/0961bc01f03010d4cfce01eefba0a88cabd62b20/tests/run-tests.ps1#L77-L80

That logic is correct when applied to internal builds because we test distroless images separately from the full images for internal builds. But for public PR builds, they're integrated into the same job and tested together. So the wildcard doesn't get added. The OS variable in this case is strictly `cbl-mariner2.0` which excludes the distroless images from being tested.

PR #3886, opened 2022-06-29
Set default entrypoint for distroless Mariner

### LINKED ISSUE #3866: Set default ENTRYPOINT for distroless Dockerfiles on Mariner
In our normal (non-distroless) Dockerfiles, we don't use an `ENTRYPOINT` instruction in the runtime/aspnet/sdk Dockerfiles to allow downstream consumers to easily decide which command/executable they want. But for distroless Dockerfiles, there's literally only one executable in the whole container: `dotnet`. So we may as well set that as the default:

```dockerfile
ENTRYPOINT ["/usr/bin/dotnet"]
CMD ["--info"]
```

In that case, the proposal would be to set the default command with the `--info` option.

mthalman:
This is applied for .NET 7 only.

Fixes #3866


mthalman:
Converted to a draft for now. This will require changes to the test code once a fix for https://github.com/dotnet/dotnet-docker/issues/3870 has been merged.

MichaelSimons (approved):

PR #3910, opened 2022-07-08
update-dependencies pipeline for internal builds

mthalman:
Defines a pipeline to run the update-dependencies tool to target internal builds. Also includes changes to update-dependencies for create a pull request to AzDO.

This follows the branch strategy described in https://github.com/dotnet/dotnet-docker/issues/2152#issuecomment-1152492044. The pipeline will be run from the `nightly` branch. As part of the process for creating the PR, it will create a new user-scoped source branch with a name of `users/dotnet-docker-bot/...`. The PR will be configured to target the `internal/release/nightly` branch. 

I also made some changes to the update-dependencies Dockerfile in order to improve its efficiency. It was previously configured to publish the app as self-contained and trimmed. But this offers no real benefits since the built image is only intended to be used locally. So it's been changed to use framework-dependent deployment instead. For the same reason, the base was changed to sdk instead of runtime-deps because this provides PowerShell for free.

Related to #2152


eng/pipelines/steps/update-dotnet-installer-dependencies.yml
```diff
@@ -0,0 +1,33 @@
+parameters:
```
MichaelSimons, line 1: To me the use of `installer` in the name is confusing.

eng/update-dependencies/Dockerfile
```diff
@@ -14,26 +14,17 @@ RUN dotnet restore -r $RID
...
 
 
 # runtime image
-FROM mcr.microsoft.com/dotnet/runtime-deps:6.0-alpine
+FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine
 
 # install git and Docker
 RUN apk add --no-cache \
         docker \
         git
```
MichaelSimons, line 26: git comes with the sdk base image.

MichaelSimons (approved):

MichaelSimons (approved):

PR #3939, opened 2022-07-22
Sample Dockerfile: explicitly restore test project

mthalman:
The complexapp sample Dockerfile should explicitly restore the test project. Otherwise, the test project won't get restored until the container is actually executed, implicitly by the `dotnet test` ENTRYPOINT. That's wasteful because every execution of the container would cause it to get restored again. It only needs to get restored once and embedded within the image.

### DIFF
--- a/samples/complexapp/Dockerfile
+++ b/samples/complexapp/Dockerfile
@@ -24,6 +24,8 @@ COPY tests/*.csproj .
 RUN dotnet restore tests.csproj
 
 COPY tests/ .
+RUN dotnet build --no-restore
+
 ENTRYPOINT ["dotnet", "test", "--logger:trx", "--no-restore"]
 
 FROM build AS publish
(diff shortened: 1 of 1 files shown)


MichaelSimons (commented):
Couldn't the same optimization could be done for building as well?

samples/complexapp/Dockerfile
```diff
@@ -19,8 +19,14 @@ RUN dotnet build -c release --no-restore
...
 WORKDIR /source/tests
+
+COPY tests/*.csproj .
+RUN dotnet restore tests.csproj
+
 COPY tests/ .
-ENTRYPOINT ["dotnet", "test", "--logger:trx"]
+RUN dotnet build --no-restore
+
+ENTRYPOINT ["dotnet", "test", "--logger:trx", "--no-restore"]
```
MichaelSimons, line 29: This should also specify `--no-build` for optimization purposes.

mthalman: Nice. I didn't even know that option existed.

MichaelSimons (approved):

richlander:
This is super related: https://github.com/dotnet/sdk/issues/26249

PR #3946, opened 2022-07-27
.NET Monitor: Remove stale UDS on startup

jander-msft:
Changes add a new environment variable to the 6.2+ images that will remove any existing Unix Domain Socket found at the listening server path (when that is configured) on startup. This allows for removing orphaned UDS from prior instances of dotnet-monitor and will avoid the Socket Error 48/98 address already is use issues.

cc @dotnet/dotnet-monitor

### DIFF
--- a/eng/dockerfile-templates/monitor/Dockerfile
+++ b/eng/dockerfile-templates/monitor/Dockerfile
@@ -17,7 +17,8 @@
       Otherwise, use the account that is associated with the current branch. ^
     set urlBranch to when(find(monitorVersion, "-") >= 0 || buildVersion != monitorVersion, VARIABLES["branch"], "main") ^
     _ Use the $DOTNET_MONITOR_VERSION variable for the version folder if the build and product versions are the same. ^
-    set versionFolder to when(buildVersion != monitorVersion, buildVersion, '$DOTNET_MONITOR_VERSION')
+    set versionFolder to when(buildVersion != monitorVersion, buildVersion, '$DOTNET_MONITOR_VERSION') ^
+    set supportsEndpointDeletion to monitorMajorMinor != "6.1"
 }}ARG ASPNET_REPO=mcr.microsoft.com/dotnet/aspnet
 ARG SDK_REPO=mcr.microsoft.com/dotnet/sdk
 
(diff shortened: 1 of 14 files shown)


tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -118,6 +118,11 @@ public void VerifyEnvironmentVariables(MonitorImageData imageData)
             // DefaultProcess filter should select a process with a process ID of 1
             variables.Add(new EnvironmentVariableInfo("DefaultProcess__Filters__0__Key", "ProcessId"));
             variables.Add(new EnvironmentVariableInfo("DefaultProcess__Filters__0__Value", "1"));
+            // Existing (orphaned) diagnostic port should be delete before starting server
+            if (imageData.Version >= ImageVersion.V6_2)
```
mthalman, line 122: I'd prefer:

```suggestion
            if (imageData.Version > ImageVersion.V6_1)
```

That way, when we eventually delete `V6_1`, the compiler will let us know about this and we can determine that there's no need for the conditional.

mthalman (approved):

eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -17,7 +17,8 @@
       Otherwise, use the account that is associated with the current branch. ^
     set urlBranch to when(find(monitorVersion, "-") >= 0 || buildVersion != monitorVersion, VARIABLES["branch"], "main") ^
     _ Use the $DOTNET_MONITOR_VERSION variable for the version folder if the build and product versions are the same. ^
-    set versionFolder to when(buildVersion != monitorVersion, buildVersion, '$DOTNET_MONITOR_VERSION')
+    set versionFolder to when(buildVersion != monitorVersion, buildVersion, '$DOTNET_MONITOR_VERSION') ^
+    set supportsEndpointDeletion to monitorMajorMinor != "6.1"
```
wiktork, line 21: There is no patch version right? I.e. 6.2.1 is not pullable once 6.2.2 is published?

wiktork (approved):

eng/dockerfile-templates/monitor/Dockerfile
jander-msft: We don't care about the patch version because we are no longer building 6.2.1 (or it won't be built shortly, after tomorrow's insertion is merged). Tomorrow's automated PR will insert the 6.2.2 candidate build which makes use of this environment variable.

jander-msft:
@mthalman, please merge this PR. It seems I no longer have the ability to merge PRs.

PR #3974, opened 2022-08-08
Enable signed NuGet package validation

### LINKED ISSUE #3932: Enable signed NuGet package validation for 6.0.4xx
The NuGet team is [re-enabling signed NuGet package validation](https://github.com/NuGet/NuGet.Client/pull/4706). It is enabled by default (for Linux) for .NET 7. It is opt-in for .NET 6. I'm proposing we opt-in to this capability in containers.

I previously mentioned that there was a problem with workloads. There is not. I was wrong. There is no expected breaking change due to enabling this feature. Both .NET 7 and .NET 6.0.4xx have the same behavior.

The only expected "break" is if NuGet finds a package with a bad cert. That's a good break, and the exact same thing would happen if someone used the Windows build of the .NET SDK.

`DOTNET_NUGET_SIGNATURE_VERIFICATION` is exposed to opt-in. If you set it to true, it enables signature verification by default during restore operations. The environment variable name is case-sensitive but the value is not: both `true` and `TRUE` work.

Note that this only affects signature verification during [restore operations](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore#description) (e.g.: `dotnet new`, `dotnet add package`, `dotnet restore`, etc.). Other NuGet commands like `dotnet nuget sign` and `dotnet nuget verify` always have verification enabled.

@dtivel

mthalman:
Fixes #3932 by setting the `DOTNET_NUGET_SIGNATURE_VERIFICATION` environment variable in Linux-based 6.0 Dockerfiles. This enables signature validation for NuGet packages once these Dockerfiles are updated to 6.0.400.

### DIFF
--- a/eng/dockerfile-templates/sdk/Dockerfile.envs
+++ b/eng/dockerfile-templates/sdk/Dockerfile.envs
@@ -13,7 +13,9 @@
     # Do not generate certificate
     DOTNET_GENERATE_ASPNET_CERTIFICATE=false {{lineContinuation}}{{if dotnetVersion != "3.1":
     # Do not show first run text
-    DOTNET_NOLOGO=true {{lineContinuation}}}}{{if dotnetVersion != "3.1":
+    DOTNET_NOLOGO=true {{lineContinuation}}}}{{if !isWindows && dotnetVersion = "6.0":
+    # Enable validation for signed NuGet packages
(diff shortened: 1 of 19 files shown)


MichaelSimons (commented):
Should the Sdk's VerifyEnvironmentVariables test should be updated for this change?

MichaelSimons (approved):

PR #3979, opened 2022-08-09
Delete Windows Server 20H2 Dockerfiles

mthalman:
Deleting the Dockerfiles based on Windows Server 20H2 because it is EOL as of today.


MichaelSimons (commented):
Curious why not make this in nightly first and then merge to main.  It feels low risk but there is still risk to the build.

mthalman:
> Curious why not make this in nightly first and then merge to main. It feels low risk but there is still risk to the build.

I thought the risk was low and I prefer to do a merge from main to nightly than a cherrypick from nightly to main. But I'm fine with rebasing on nightly.

MichaelSimons (approved):

PR #3989, opened 2022-08-11
Update samples to .NET 7

richlander:
- Update samples to .NET 7 (`.csproj` and `Dockerfile`)
- Fix bad patterns I find (I found some)
- Update docs to reference .NET 7
- Add non-root sample


samples/complexapp/complexapp/complexapp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
mthalman, line 5: The complexapp test project also needs to be updated.

samples/run-in-sdk-container.md
```diff
@@ -37,12 +37,12 @@ a8"    `Y42 a8"     "8a  42    42P'   `"8a a8P_____42   42
 "8a,   ,d42 "8a,   ,a8"  42,   42       42 "8b,   ,aa   42,
  `"8bbdP"Y8  `"YbbdP"'   "Y428 42       42  `"Ybbd8"'   "Y428
 
-.NET 6.0.0
+.NET 7.0.0-preview.7.22375.6
```
MichaelSimons, line 40: Would it be appropriate to edit out the version suffix - `-preview.7.22375.6`

samples/aspnetapp/Dockerfile.nanoserver-x64-slim
```diff
@@ -1,24 +1,22 @@
...
 # copy everything else and build app
-COPY aspnetapp/. ./aspnetapp/
-WORKDIR /source/aspnetapp
-RUN dotnet publish -c release -o /app -r win-x64 --self-contained true --no-restore /p:PublishTrimmed=true /p:PublishReadyToRun=true /p:PublishSingleFile=true
+COPY aspnetapp/. .
+RUN dotnet publish -c Release -o /app -r win-x64 --self-contained true --no-restore /p:PublishTrimmed=true /p:PublishReadyToRun=true /p:PublishSingleFile=true
 
 # final stage/image
-# Uses the ltsc2022 release; 20H2 and 1809 are other choices
+# Uses the ltsc2022 release
```
MichaelSimons, line 16: This comment no longer seems useful as the code is self documenting.  I see other edits like this removed the entire line.

MichaelSimons (approved):

samples/aspnetapp/aspnetapp/aspnetapp.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
ldillonel, line 4: Produces error if user follows Prerequisites that states .NET 6.

mthalman: Which prerequisites are you referring to?

ldillonel: https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/docker/building-net-docker-images?view=aspnetcore-6.0

mthalman: I had no idea these instructions existed.

Given that the doc page has a drop-down that allows you to select which .NET version you want, it already had out-of-sync content even prior to this change to .NET 7. For example, if you select the [3.1 version](https://docs.microsoft.com/aspnet/core/host-and-deploy/docker/building-net-docker-images?view=aspnetcore-3.1), you still have instructions to clone the same repo which previously was using .NET 6 samples, so still not matching the pre-reqs for that version. As an aside, that version of the doc also specifies the out-of-support 3.0 version in the Dockerfile snippets instead of using 3.1:

```Dockerfiles
FROM mcr.microsoft.com/dotnet/core/sdk:3.0 AS build
```

I see two options:

1. Don't rely on the dotnet-docker repo at all for the project code. Maintain a separate git repo specifically for the docs. I've seen other products do this, like Azure.
2. Use git tags in the dotnet-docker repo to allow customers to `git checkout` at a specific tag to be described in the documentation. The git tags would refer to a specific commit in the repo appropriate for that version. So we could have a `3.1-samples` tag that, when checked out, would give you the state of the repo when our samples were configured for 3.1. Similarly for other .NET versions.

@ldillonel, as a first step, please log an issue on the docs and link this comment thread. We can discuss further on that issue what the resolution should be.

ldillonel: @mthalman I logged the issue on the docs page first.  Here's the link https://github.com/dotnet/AspNetCore.Docs/issues/26770

PR #3996, opened 2022-08-15
Add Chiseled Ubuntu Dockerfiles

mthalman:
Adds Chiseled Ubuntu Dockerfiles for 6.0/7.0 on amd64 and arm64.


src/runtime-deps/6.0/jammy-chiseled/arm64v8/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+        base-files_base \
+        ca-certificates_data \
+        libc6_libs \
+        libgcc-s1_libs \
+        libssl3_libs \
+        libstdc++6_libs \
+        zlib1g_libs
+
+
+FROM scratch
```
MichaelSimons, line 39: The spacing format used in the final layer feels inconsistent as compared to the other Dockerfiles.  e.g the mariner-distroless images:

```
# .NET runtime-deps image
FROM mcr.microsoft.com/cbl-mariner/distroless/minimal:1.0

COPY --from=installer /staging/ /

ENV \
    # Configure web servers to bind to port 8080 when present
    ASPNETCORE_URLS=http://+:8080 \
    DOTNET_RUNNING_IN_CONTAINER=true

USER app
```

mthalman: Fixed

MichaelSimons (approved):
:shipit:

PR #4017, opened 2022-08-19
Refactor runtime-deps Dockerfile template

mthalman:
The runtime-deps Dockerfile template is pretty complex and hard for a human to parse. This is mainly due to the fact that it has logic used to generate distroless Mariner Dockerfiles in addition to the usual distroful Dockerfiles. Because distroless Mariner Dockerfiles are so different from the others, it causes a lot of extra stuff in the template to wade through.

As part of the implementation for [Chiseled Ubuntu](https://github.com/dotnet/dotnet-docker/pull/3996), those runtime-deps Dockerfiles had the same issue, being markedly different than all the other runtime-deps Dockerfiles. Because of that, a separate runtime-deps template was defined specifically for Chiseled Ubuntu. This kept all that content specific to Chiseled Ubuntu isolated from everything else.

These changes follow that same pattern that was used for Chiseled Ubuntu and applies it to distroless Mariner. A separate Dockerfile template for distroless Mariner is added. All the content specific to distroless Mariner is pulled out of the default runtime-deps template and placed in this new template file. This cleans things up a lot, making it much easier to read all the runtime-deps templates.


MichaelSimons (approved):
Nice refactoring.

PR #4022, opened 2022-08-22
Enable .NET Monitor distroless image scenario testing

jander-msft:
Currently, the distroless .NET Monitor image does not participate in the scenario tests because the filtering criteria requires that the .NET Monitor image and the sample image are the same type in terms of distroless vs non-distroless. This is not strictly necessary because non-distroless .NET Monitor can connect to either image type and distroless .NET Monitor can listen for either image type. These changes broaden the test combinations as described to allow distroless images to be mixed with non-distroless images in the scenario tests.


mthalman (approved):

jander-msft:
@mthalman, feel free to merge as I am not authorized to do so.

PR #4033, opened 2022-08-23
.NET Monitor Jammy Chiseled images

jander-msft:
These changes add Ubuntu Jammy Chiseled images for .NET Monitor 6.2 and 7.0. The sizes of the `amd64` images were inspected from a local build of the entire image stack whereas the `arm64` images were calculated based on relative size differences of the `amd64` images.

cc @dotnet/dotnet-monitor


eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -5,8 +5,9 @@
     set monitorMajor to split(PRODUCT_VERSION, ".")[0] ^
     set targetFrameworkMoniker to cat("net", dotnetMajorMinor) ^
     set targetFrameworkMonikerRegex to cat("net", dotnetMajor, "[.]0") ^
-    set isDistrolessMariner to defined(match(OS_VERSION, "^cbl-mariner\d+\.\d+-distroless$")) ^
-    set sdkDistro to when(isDistrolessMariner, cat("cbl-mariner", OS_VERSION_NUMBER), OS_VERSION) ^
+    set isMariner to find(OS_VERSION, "cbl-mariner") >= 0 ^
+    set isDistroless to find(OS_VERSION, "distroless") >= 0 || find(OS_VERSION, "chiseled") >= 0 ^
+    set sdkDistro to when(isDistroless, when(isMariner, cat("cbl-mariner", OS_VERSION_NUMBER), "jammy"), OS_VERSION) ^
```
mthalman, line 10: Instead of hardcoding `"jammy"`, can this be changed to dynamically extract the OS version.

I think this might work:

```suggestion
    set sdkDistro to when(isDistroless, when(isMariner, cat("cbl-mariner", OS_VERSION_NUMBER), split(OS_VERSION, "-")[0]), OS_VERSION) ^
```

mthalman:
I'm investigating the build failure.

mthalman:
> I'm investigating the build failure.

I think it's because my rebuild of one of the jobs took in new changes from https://github.com/dotnet/dotnet-docker/pull/4036 which caused it to be out of sync with the outputs of the other jobs. I'll queue a new build.

mthalman (approved):

PR #4043, opened 2022-08-29
Convert get-drop-versions scripts to PowerShell

mthalman:
Ports the get-drop-versions.sh and get-drop-versions-monitor.sh Bash scripts to PowerShell.

This will help to better enable some additional functionality in the future. It also provides a cross-platform script that can easily be executed in a local Windows environment.


eng/Get-DropVersions.ps1
```diff
@@ -0,0 +1,133 @@
...
+$tempDir = "$([System.IO.Path]::GetTempPath())/dotnet-docker-get-dropversions"
+New-Item -Path $tempDir -ItemType Directory -Force | Out-Null
+
+try {
+    $sdkInfo = GetSdkInfo
+    $versionDetails = GetVersionDetails $sdkInfo.CommitSha
+
+    $runtimeVersion = GetDependencyVersion "VS.Redist.Common.NetCore.SharedFramework.x64" $versionDetails
+    if (-not $runtimeVersion) {
+        $runtimeVersion = GetDependencyVersion "Microsoft.NETCore.App.Internal" $versionDetails
```
MichaelSimons, line 112: Why the name change for internal runtime but not aspnetcore?

mthalman: This fallback is actually outdated and not needed anymore. Removed.

MichaelSimons (approved):

PR #4057, opened 2022-09-06
Support multiple SDK versions in update-dependencies

mthalman:
This updates the infrastructure related to update-dependencies so that it can support the targeting of multiple SDK versions. This is necessary for https://github.com/dotnet/dotnet-docker/issues/2152. By having the update-dependencies pipeline generate a PR that targets multiple internal builds of .NET. This approach is mentioned in https://github.com/dotnet/dotnet-docker/issues/2152#issuecomment-1150402553. This allows the `internal/release/nightly` branch to remain up-to-date with the latest updates from `nightly` but also have the latest internal .NET builds applied to it for each PR.

When targeting internal builds, we'll have an `sdkVersions` pipeline variable that will be used to explicitly define which builds should be targeted. But for the public builds, those will continue to be dynamically retrieved by the use of an aka.ms URL with a channel specified. This means the `Get-DropVersions.ps1` script needs to handle either scenario. It's been updated to either use a Channel or an array of SDK versions. Because multiple versions can now be targeted, its output has been changed to in JSON format that represents an array of version sets (each one representing a Dockerfile version). This JSON representation is then consumed by the pipeline and fed into the Set-DotNetVersion.ps1 script for each of those Dockerfile versions. Lastly, the update-dependencies container is then invoked multiple times for each of the Dockerfile versions, each having its own specific set of version information specified. This is applied iteratively in the repo, allowing the repo to contain the full set of changes for all Dockerfile versions. Once they're all applied, the PR gets generated.

### DIFF
--- a/eng/Get-IsStableBranding.ps1
+++ b/eng/Get-IsStableBranding.ps1
@@ -1,0 +1,21 @@
+#!/usr/bin/env pwsh
+
+<#
+.SYNOPSIS
+Returns a value indicating whether the specified version is associated with stable branding.
+#>
+[cmdletbinding()]
+param(
+    # Build version of the product
+    [string]
+    $Version
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2.0
+
+if ($Version.Contains("-servicing") -or $Version.Contains("-rtm")) {
+    return $true
+}
+
+return $false
--- a/eng/Get-MonitorDropVersions.ps1
+++ b/eng/Get-MonitorDropVersions.ps1
@@ -15,4 +15,7 @@ param(
 $url = "https://aka.ms/dotnet/diagnostics/monitor$Channel/dotnet-monitor.nupkg.buildversion"
 $monitorVersion = $(Invoke-RestMethod $url).Trim()
 
+$stableBranding = & $PSScriptRoot/Get-IsStableBranding.ps1 -Version $monitorVersion
+
 Write-Output "##vso[task.setvariable variable=monitorVer]$monitorVersion"
+Write-Output "##vso[task.setvariable variable=stableBranding]$stableBranding"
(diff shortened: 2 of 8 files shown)


eng/Get-MonitorDropVersions.ps1
```diff
@@ -15,4 +15,7 @@ param(
 $url = "https://aka.ms/dotnet/diagnostics/monitor$Channel/dotnet-monitor.nupkg.buildversion"
 $monitorVersion = $(Invoke-RestMethod $url).Trim()
 
+$stableBranding = & $PSScriptRoot/Get-IsStableBranding.ps1 -Version $monitorVersion
```
jander-msft, line 18: I think this file is missing from the PR and I don't see it in the nightly branch

mthalman: Fixed

eng/Get-IsStableBranding.ps1
```diff
@@ -0,0 +1,21 @@
...
+    # Build version of the product
+    [string]
+    $Version
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2.0
+
+if ($Version.Contains("-servicing") -or $Version.Contains("-rtm")) {
+    return $true
```
jander-msft, line 18: If I understand this change, this would force us to use stable branding the moment we produce a package with `-rtm` or `-servicing` in its version, is that correct? The "use stable branding" variables on the pipeline would become defunct?

mthalman: Right, is that not going to work for .NET Monitor? I can change it back to use the queue-time variable for .NET Monitor. The intent here was to simplify the definition of the variables in the pipeline. Since we can now pass in a dynamic number of SDK versions, I wanted a simpler way to manage the stable-brandingness of those. By having the stable branding information derived from the version itself, it eliminates the need to somehow express that information dynamically from the pipeline variables.

I updated .NET Monitor to simply reuse this logic but it's not necessary to do so.

jander-msft: I think this is fine for .NET Monitor. Wanted to make sure what the expectation is when we do rtm and servicing releases.

jander-msft: FYI @dotnet/dotnet-monitor

PR #4078, opened 2022-09-14
Rename Ubuntu chiseled folder and tags for .NET Monitor

jander-msft:
Similar to how the folder and tag names for the .NET Monitor Alpine based images do not contain version information (because .NET Monitor will not produce images for more than one version of Alpine at a time), the Ubuntu Chiseled folders and tags should be named similarly, without version information.


mthalman:
I question whether this change should be made.

* Alpine is a different situation because of its update frequency and (relatively) minimal breaking changes. OTOH, Ubuntu has very long support and has more impactful breaking changes between versions. This is why we provide a floating `alpine` tag for the other .NET container images but not an `ubuntu` tag.
* Considering that the tagging is remaining as `jammy-chiseled`, the directory name should reflect that.
* There's not going to be a scenario where we change the Ubuntu version for a given .NET Monitor version. So keeping the directory name as the Ubuntu version seems non-problematic.

MichaelSimons:
@mthalman raises some good points.  One thing I noticed however is inconsistency with the mariner images.  Their folders and tags are distro specific but not version specific just like alpine.  With that precedence, this PR is the right direction but there should also be a tag change to use the distro name instead of the release name.  

To summarize, mariner and ubuntu should be consistent.  If you decide to go the versionless approach, it would be good to update the tagging documentation [here](https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-tags.md) to call out the exception for the monitor images.

jander-msft:
> @mthalman raises some good points. One thing I noticed however is inconsistency with the mariner images. Their folders and tags are distro specific but not version specific just like alpine. With that precedence, this PR is the right direction but there should also be a tag change to use the distro name instead of the release name.
> 
> To summarize, mariner and ubuntu should be consistent. If you decide to go the versionless approach, it would be good to update the tagging documentation [here](https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-tags.md) to call out the exception for the monitor images.

I mostly agree with the points made in general, except we need to consider that the .NET Monitor images are more appliance-like rather than base or builder images. This is especially true for distroless/chiseled images since nothing else will be running on them. Because of their appliance-like nature, the tagging scheme should be the same when upgrading from one version to another; the consumer shouldn't have to think about what distro version of the base image is being used because it's irrelevant to the usage of the .NET Monitor image. So I think tagging them without version information is the appropriate approach.

mthalman (approved):

MichaelSimons (approved):

PR #4082, opened 2022-09-17
.NET Monitor: Remove symbol files and non-matching arch assemblies

jander-msft:
The 7.0 package version in this change contains several new files that support new features that are in development. To mostly mitigate the size increase, I've amended the Dockerfiles to remove files that do not support the matching architecture for the Docker image as well as symbol files. This should be a net size reduction for the 6.x images and a minor net size increase for the 7.0 images.

cc @dotnet/dotnet-monitor

### DIFF
--- a/eng/dockerfile-templates/monitor/Dockerfile
+++ b/eng/dockerfile-templates/monitor/Dockerfile
@@ -44,7 +44,13 @@ RUN {{InsertTemplate("../Dockerfile.linux.download-file",
     # 4. Delete everything from step 3
     && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -v -i '.*/tools/{{targetFrameworkMonikerRegex}}' | xargs rm -rf \
     # To reduce image size further, remove the non-linux assemblies
-    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shims)/(win|osx)' | xargs rm -rf \
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(win|osx)' | xargs rm -rf \
+    # To reduce image size further, remove linux assemblies that do not match the {{ARCH_SHORT}} architecture{{if ARCH_SHORT != "x64":
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(linux-x64|linux-musl-x64)' | xargs rm -rf \}}{{if ARCH_SHORT != "arm64":
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(linux-arm64|linux-musl-arm64)' | xargs rm -rf \}}
+    # To reduce image size further, remove symbol files
+    && find /app -type f -name *.pdb | xargs rm -rf \
+    && find /app -type f -name *.dbg | xargs rm -rf \
     # Remove all the empty directories left by the previous step
     && find /app -type d -empty -delete \
     # Allow other users to run the tool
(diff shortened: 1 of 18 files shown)


eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -44,7 +44,13 @@ RUN {{InsertTemplate("../Dockerfile.linux.download-file",
     # 4. Delete everything from step 3
     && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -v -i '.*/tools/{{targetFrameworkMonikerRegex}}' | xargs rm -rf \
     # To reduce image size further, remove the non-linux assemblies
-    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shims)/(win|osx)' | xargs rm -rf \
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(win|osx)' | xargs rm -rf \
+    # To reduce image size further, remove linux assemblies that do not match the {{ARCH_SHORT}} architecture{{if ARCH_SHORT != "x64":
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(linux-x64|linux-musl-x64)' | xargs rm -rf \}}{{if ARCH_SHORT != "arm64":
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(linux-arm64|linux-musl-arm64)' | xargs rm -rf \}}
```
mthalman, line 50: What about doing this instead?

```suggestion
    # To reduce image size further, remove linux assemblies that do not match the {{ARCH_SHORT}} architecture
    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(linux-{{archShortOpposite}}|linux-musl-{{archShortOpposite}})' | xargs rm -rf \
```

And then defining the `archShortOpposite` variable at the top as:

```
set archShortOpposite to when(ARCH_SHORT = "x64", "arm64", "x64")
```

eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -44,7 +44,13 @@ RUN {{InsertTemplate("../Dockerfile.linux.download-file",
...
+    # To reduce image size further, remove symbol files
+    && find /app -type f -name *.pdb | xargs rm -rf \
+    && find /app -type f -name *.dbg | xargs rm -rf \
```
mthalman, line 53: This should also work:

```suggestion
    && find /app -type f \(-name *.dbg -o -name *.pdb \) | xargs rm -rf \
```

eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -44,7 +44,13 @@ RUN {{InsertTemplate("../Dockerfile.linux.download-file",
     # 4. Delete everything from step 3
     && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -v -i '.*/tools/{{targetFrameworkMonikerRegex}}' | xargs rm -rf \
     # To reduce image size further, remove the non-linux assemblies
-    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shims)/(win|osx)' | xargs rm -rf \
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(win|osx)' | xargs rm -rf \
+    # To reduce image size further, remove linux assemblies that do not match the {{ARCH_SHORT}} architecture{{if ARCH_SHORT != "x64":
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(linux-x64|linux-musl-x64)' | xargs rm -rf \}}{{if ARCH_SHORT != "arm64":
+    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(linux-arm64|linux-musl-arm64)' | xargs rm -rf \}}
+    # To reduce image size further, remove symbol files
+    && find /app -type f -name *.pdb | xargs rm -rf \
```
wiktork, line 52: Did you want to restrict this removal to native symbols only, which are considerably larger than the managed ones?

jander-msft: I'm not sure there is much benefit in leaving the managed ones in the image. It's rare that we are debugging the bits in the Docker image directly. The symbols package produced by the dotnet-monitor build should be uploading symbols to an accessible place.

PR #4086, opened 2022-09-22
Fix update-deps when targeting a single SDK version

mthalman:
This fixes an issue introduced from https://github.com/dotnet/dotnet-docker/pull/4057 where a single SDK version is provided. In that case, a PR will not be generated for the changes. The intent is that a PR will only be generated after the latest iteration of the set of SDK versions being targeted. But when providing only a single SDK version as input, the PowerShell logic ends up not executing as intended. 

When converting the customArgsArray parameter to JSON, PowerShell has a weird quirk where a JSON array with only a single element will be deserialized as just a string and not an array. This messes up the `for` loop where it checks if the value is the last item in the array. It's not recognized as being the last item, and so the credentials are never provided to update-dependencies and, thus, a PR never gets generated.

The fix is to simply wrap the result in an array. This works fine even in the case where the JSON has more than one item in the array. You don't get an array of arrays in that case, again because of the weirdness of PowerShell and single-item arrays.

### DIFF
--- a/eng/pipelines/steps/update-dependencies.yml
+++ b/eng/pipelines/steps/update-dependencies.yml
@@ -19,7 +19,9 @@ steps:
     $credArgs = "--user $(dotnetDockerBot.userName) --email $(dotnetDockerBot.email) --password $pat"
 
     # Execute update-deps for each of the items in the customArgsArray, representing different Dockerfile versions
-    $customArgsArray = '${{ parameters.customArgsArray }}' | ConvertFrom-Json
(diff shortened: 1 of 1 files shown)


mthalman:
> Powershell trying to he helpful but not ☹️

What's funny is that this behavior is known by GitHub Copilot. In my comments I typed "This deals with the quirk of" and Copilot finished it off with "how PowerShell treats a single item as a string instead of an array". Thanks Copilot!

PR #4088, opened 2022-09-22
Allow writable files for global mutexes

mthalman:
I've confirmed that globally writable files are expected for .NET global mutexes. This means that such files can show up in the /tmp/.dotnet directory. I've updated the unit test to allow for such files in this path only.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
@@ -77,7 +77,13 @@ namespace Microsoft.DotNet.Docker.Tests
                 optionalRunArgs: "--entrypoint /bin/sh"
             );
 
-            Assert.Empty(output);
+            IEnumerable<string> lines = output
+                .ReplaceLineEndings()
+                .Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries)
+                // Writable files in the /tmp/.dotnet directory are allowed for global mutexes
+                .Where(line => !line.StartsWith("/tmp/.dotnet"));
+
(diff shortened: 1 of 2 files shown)


eerhardt:
Do you know what code is using a Mutex now?

mthalman:
> Do you know what code is using a Mutex now?

No idea. It's related to running `dotnet tool install` in the SDK container as part of installing PowerShell.

eerhardt:
The best I could find is:

https://github.com/dotnet/runtime/blob/9b9b0c95c678065044d01049c56915e6b750b10a/src/coreclr/pal/src/include/pal/mutex.hpp#L66-L126

There's other info spread throughout:

https://github.com/dotnet/runtime/issues/49822
https://github.com/dotnet/coreclr/pull/5030

And the high-level explanation (doesn't go into details about the file structure, but talks about what "Global" prefixed Mutexes are) in the "Remarks" section of 

https://learn.microsoft.com/en-us/dotnet/api/system.threading.mutex.-ctor?view=net-6.0#system-threading-mutex-ctor(system-boolean-system-string)

cc @janvorli @kouvel

kouvel:
For global mutexes the shared memory and lock files/directories that go under `/tmp/.dotnet` are expected to be readable/writable by all users so that processes running as different users can synchronize on them.

eerhardt:
One thing to know/realize here is that the root folder `/tmp/.dotnet` needs to be readable/writable by all users, even when you only create a non-Global, named Mutex. That is because the same root folder is used between Global and non-Global Mutexes. The root directory needs to be created as globally RW so the Global Mutexes created under it are as well.

MichaelSimons (approved):

PR #4089, opened 2022-09-23
Add .NET Monitor 6.3 images; move Ubuntu Chiseled to 6.3

jander-msft:
This change adds .NET Monitor 6.3 images. They are same set of distros, architectures, and distroless as the 6.2 version. The Ubuntu Chiseled 6.2 images have been moved to 6.3 since they have not shipped yet.

cc @dotnet/dotnet-monitor


eng/mcr-tags-metadata-templates/monitor-tags.yml
```diff
@@ -7,9 +7,13 @@ $(McrTagsYmlTagGroup:7.0-ubuntu-chiseled-amd64)
     customSubTableTitle: .NET Monitor Preview Tags
 $(McrTagsYmlTagGroup:7.0-ubuntu-chiseled-arm64v8)
     customSubTableTitle: .NET Monitor Preview Tags
-$(McrTagsYmlTagGroup:6.2-alpine-amd64)
-$(McrTagsYmlTagGroup:6.2-alpine-arm64v8)
-$(McrTagsYmlTagGroup:6.2-ubuntu-chiseled-amd64)
+$(McrTagsYmlTagGroup:6.3-alpine-amd64)
+    customSubTableTitle: .NET Monitor Preview Tags
```
mthalman, line 11: When should 6.3 be merged to main? Will it be in preview at that time?

jander-msft: I can get it to produce non-final RTM builds in the next day or two if you'd rather see that initially in the nightly branch.

mthalman: In that case then I would suggest to not classify these tags as Preview in the nightly branch. The general philosophy is that nightly is the staging environment for what will go into main. If it's not going to be preview in main then we shouldn't label it that way in nightly.

mthalman: And I get that this was just temporary. But it's better to have it correct now so that we don't forget later on.

jander-msft: Done

mthalman (approved):

PR #4092, opened 2022-09-23
Include home directory for non-root user

### LINKED ISSUE #4083: Usability issue with distroless user's lack of home directory
In our distroless Dockerfiles, we create a non-root user in order to have things run without root privileges by default. When this was first defined, it was thought that a home directory for this user was unnecessary.

But there are some usability issues with not having a home directory. Some Linux tools, and perhaps the .NET application code itself, may expect that a home directory exists for the user.

For example, the docker CLI has functionality which will attempt to write to the `$HOME/.docker/config.json` path. When it attempts to create the `/home/app` directory in a .NET distroless container, it gets the following error: `Error saving credentials: mkdir /home/app: permission denied`.

As a result of not having this home directory by default, developers wishing to have one must resort to configuring it in their own Dockerfile which is made even more complicated by the fact that you can't create the directory directly in the distroless stage (due to no shell). By having a multi-stage Dockerfile with a staging directory, the workaround would look like this:

```Dockerfile
RUN homeDir="/staging/home/app" \
    && mkdir -p $homeDir \
    && chown -R app $homeDir
```

This issue proposes that the distroless Dockerfiles should be configured to create a home directory for the non-root user.

mthalman:
Ensures that a home directory exists for the non-root user that is created in distroless/chiseled images. This allows tools that are expecting this directory to exist to work without issue.
Fixes #4083

### DIFF
--- a/src/runtime-deps/6.0/jammy-chiseled/amd64/Dockerfile
+++ b/src/runtime-deps/6.0/jammy-chiseled/amd64/Dockerfile
@@ -18,9 +18,9 @@ RUN groupadd \
         --uid 101 \
         --gid 101 \
         --shell /bin/false \
-        --no-create-home \
         --system \
         app \
+    && mkdir -p "/rootfs/home/app" \
     && mkdir -p "/rootfs/etc" \
     && tail -1 < /etc/passwd > "/rootfs/etc/passwd" \
     && tail -1 < /etc/group > "/rootfs/etc/group"
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
@@ -4,6 +4,7 @@
         staging-dir: Path to the distroless staging directory
         create-dir (optional): Indicates whether the etc directory should be created in staging
         exclusive (optional): Indicates whether the app user is the only user and all other users are removed ^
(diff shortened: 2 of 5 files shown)


src/runtime-deps/6.0/jammy-chiseled/amd64/Dockerfile
```diff
@@ -18,9 +18,9 @@ RUN groupadd \
         --uid 101 \
         --gid 101 \
         --shell /bin/false \
-        --no-create-home \
         --system \
         app \
+    && mkdir -p "/rootfs/home/app" \
```
MichaelSimons, line 23: The chiseled images are still in preview so making this breaking change in 6.0 is acceptable.  It does however make the 6.0 images inconsistent in that the mariner images have a different behavior.  I am alright with that decision but did want to call it out.

MichaelSimons (approved):

src/runtime-deps/6.0/jammy-chiseled/amd64/Dockerfile
mthalman: Inconsistent in what way? Is this related to the home directory or to just the fact that the app user exists?

MichaelSimons: 6.0 chiseled images have a home directory defined for the non-root user while the 6.0 mariner distroless images don't.

PR #4108, opened 2022-09-29
Enable distroless Mariner images to be scanned correctly

mthalman:
The distroless Mariner images that we produce aren't configured properly to allow scanning tools to correctly identify the RPM packages that exist in them, specifically the RPM packages that are being directly installed by the .NET Dockerfiles. The scanning tools are only recognizing the RPM packages that exist in the base distroless Mariner image that the .NET image is based on.

This is because these scanning tools are looking for an [RPM manifest file](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/docs/how_it_works/5_misc.md#rpm-manifest-file) that's specific to Mariner. This file exists in the base distroless Mariner image but is not updated to reflect the RPM packages that are installed by the .NET Dockerfile.

There is a broader [work item](https://dev.azure.com/microsoft/OS/_queries/query/53e3fbc5-e40d-4e8a-824a-e924ef86b4f3/) (internal only link) already logged to provide a better overall experience in managing the scenario of installing additional packages in distroless Mariner images that should better address this. But we need a short-term solution now to address this issue.

These changes propose that we add commands to the Dockerfile to directly update the contents of the RPM manifest file using the [format as documented by Mariner](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/docs/how_it_works/5_misc.md#rpm-manifest-file). Care needs to be taken here to include the original contents of that file from the base distroless image and handling any duplicates between the original contents and the new output.

The effect of this is that scanning tools, such as [syft](https://github.com/anchore/syft), are able to correctly identify the packages that have been installed by .NET.

Before:

```console
> syft packages docker:mcr.microsoft.com/dotnet/runtime-deps:6.0-cbl-mariner2.0-distroless-amd64
NAME                           VERSION              TYPE
distroless-packages-minimal    0.1-2.cm2    rpm
filesystem                     1.1-10.cm2   rpm
mariner-release                2.0-21.cm2   rpm
prebuilt-ca-certificates-base  2.0.0-7.cm2  rpm
```

After:

```console
> syft packages docker:mcr.microsoft.com/dotnet/runtime-deps:6.0-cbl-mariner2.0-distroless-amd64
NAME                           VERSION              TYPE
distroless-packages-minimal    0.1-2.cm2            rpm
e2fsprogs-libs                 1.46.5-3.cm2         rpm
filesystem                     1.1-10.cm2           rpm
glibc                          2.35-2.cm2           rpm
gpg-pubkey                     be1229cf-5631588c    rpm
krb5                           1.19.3-1.cm2         rpm
libgcc                         11.2.0-2.cm2         rpm
libstdc++                      11.2.0-2.cm2         rpm
mariner-release                2.0-21.cm2           rpm
openssl                        1.1.1k-20.cm2        rpm
openssl-libs                   1.1.1k-20.cm2        rpm
prebuilt-ca-certificates       2549185:2.0.0-7.cm2  rpm
prebuilt-ca-certificates-base  2.0.0-7.cm2          rpm
zlib                           1.2.12-2.cm2         rpm
```

cc @mandeepsplaha

### DIFF
--- a/eng/dockerfile-templates/Dockerfile.linux.install-pkgs
+++ b/eng/dockerfile-templates/Dockerfile.linux.install-pkgs
@@ -24,8 +24,8 @@ else:apt-get update \
     for index, pkg in ARGS["pkgs"]:
         {{pkg}}{{if appendPkgSuffix(pkg, index):{{if pkg != "": }}\}}}}{{
 if isTdnf:
-    && tdnf clean all^
+    && tdnf clean all{{ARGS["pkg-mgr-opts"]}}^
 elif isDnf:
-    && dnf clean all^
+    && dnf clean all{{ARGS["pkg-mgr-opts"]}}^
 elif !isApk:
     && rm -rf /var/lib/apt/lists/*}}
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.distroless-mariner
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.distroless-mariner
@@ -1,24 +1,40 @@
 {{
-    set isDistrolessMariner to defined(match(OS_VERSION, "^cbl-mariner\d+\.\d+-distroless$")) ^
     set distrolessStagingDir to "/staging" ^
     set marinerRepo to "mcr.microsoft.com/cbl-mariner" ^
-    set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".")
+    set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
+    set baseImage to cat(marinerRepo, "/distroless/minimal:", OS_VERSION_NUMBER) ^
+    set finalStageBase to when(find(OS_VERSION, "1.0") < 0, "base", baseImage)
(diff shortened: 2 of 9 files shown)


MichaelSimons (commented):
Is there anything we should be doing to validate this within our repo tests?  Would it be valuable to baseline the file with the variable parts substituted?

mthalman:
> Is there anything we should be doing to validate this within our repo tests? Would it be valuable to baseline the file with the variable parts substituted?

I'd be willing to add a unit test that does this. Ideally, we could use the output of the syft tool as true end-to-end validation.

MichaelSimons (approved):

mandeepsplaha (approved):
This solution offers a great way to update the manifest file with additional packages.

mthalman:
After @mandeepsplaha had pointed out the existence of gpg-pubkey offline, I made an update to exclude that. Reason for that is in the commit description: https://github.com/dotnet/dotnet-docker/pull/4108/commits/70cefb610fb39e3d836a9234537a4c049e793c02

MichaelSimons (approved):

PR #4115, opened 2022-10-03
Set correct ownership of non-root user's home dir

mthalman:
The changes in https://github.com/dotnet/dotnet-docker/pull/4092 didn't correctly set the ownership of the home directory. So if an app running in the context of the non-root user `app` attempted to write to its `/home/app` directory, it would get an access denied error.

This happens because the directory is created while running as the root user and nothing is done further to set ownership.

This has now been updated to both create the directory and set ownership using the `install` command. That in itself isn't enough, however. Due to https://github.com/moby/moby/issues/38710, a workaround is necessary to ensure that the `/home/app` directory that gets copied into the final stage has the ownership that was initially set. This is done by including another `COPY` instruction that explicitly targets that directory and sets ownership with the `--chown` option.

I've also updated the unit test to verify the user can write to its home directory. I've generalized this to apply to all relevant environments.

### DIFF
--- a/src/runtime-deps/7.0/cbl-mariner2.0-distroless/amd64/Dockerfile
+++ b/src/runtime-deps/7.0/cbl-mariner2.0-distroless/amd64/Dockerfile
@@ -28,7 +28,7 @@ RUN tdnf install -y shadow-utils \
         --shell /bin/false \
         --system \
         app \
-    && mkdir -p "/staging/home/app" \
+    && install -d -m 0755 -o 101 -g 101 "/staging/home/app" \
     && tail -1 < /etc/passwd > "/staging/etc/passwd" \
     && tail -1 < /etc/group > "/staging/etc/group"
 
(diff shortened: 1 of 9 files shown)


src/runtime-deps/7.0/cbl-mariner2.0-distroless/amd64/Dockerfile
```diff
@@ -59,6 +59,9 @@ FROM mcr.microsoft.com/cbl-mariner/distroless/minimal:2.0
 
 COPY --from=installer /staging/ /
 
+# Workaround for https://github.com/moby/moby/issues/38710
+COPY --from=installer --chown=101:101 /staging/home/app /home/app
```
MichaelSimons, line 63: This is unfortunate but as I am sure you have pondered, I don't seem any better solutions.

MichaelSimons (approved):

PR #4122, opened 2022-10-05
Preserve root user in distroless-style containers

mthalman:
The intent with the distroless-style containers is to have a non-root user by default. As part of that implementation, the users that were actually registered in the container have been scoped down to just that non-root user `app`.

However, all of the files in the system are actually owned by the root user (technically owned by UID 0 since there is no user named root). So it's a bit odd that there are files in the container that are owned by a user that is not registered.

Not having a root user exist doesn't actually buy us any sort of protection. The container can still be run with root-level permissions by running the container with `--user 0` instead of `--user root`.

So to make things more correct and consistent, the root user will be preserved.

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
@@ -21,8 +21,9 @@ if dotnetVersion != "6.0" || !isMariner:
 && mkdir -p "{{ARGS["staging-dir"]}}/home/app" \}}{{
 if ARGS["exclusive"]:{{if ARGS["create-dir"]:
 && mkdir -p "{{ARGS["staging-dir"]}}/etc" \}}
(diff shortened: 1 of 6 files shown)


MichaelSimons (commented):
Is value in adding a test to validate which users are defined?

mthalman:
> Is value in adding a test to validate which users are defined?

I debated with myself on doing that. I had decided not to, thinking the time it would add to the test execution wouldn't be worth it. However, after thinking more about it, I think I have a low-cost way to do this.

mthalman:
Ok, added a simple verification for this. From the test log:

```
Executing: docker run --name 7.0-app-run-133096264721653707 --rm  7.0-fx_dependent_app-133096264688034514 
Hello, World!
Execution Elapsed Time: 00:00:03.6613909
Executing: docker run --name 7.0-app-run-133096264763844177 --rm -u root  7.0-fx_dependent_app-133096264688034514 
Hello, World!
Execution Elapsed Time: 00:00:02.2781849
```

MichaelSimons (approved):

PR #4128, opened 2022-10-06
Update readmes for how tags are listed and featured

mthalman:
This outlines our plans for classifying tags in the readmes for each .NET release:

* While a .NET version is in preview, its tags will be listed separately from the other supported versions in its own "Preview Tags" table.
* Once a .NET version has reached an RC release:
  * That separate table will now longer be used and the RC tags will be incorporated with the tags of the other supported versions.
  * The "Featured Tags" section will be updated to include the RC version.

### DIFF
--- a/.mar/portal/README.monitor.portal.md
+++ b/.mar/portal/README.monitor.portal.md
@@ -16,7 +16,7 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
 
 * `7` (RC)
   * `docker pull mcr.microsoft.com/dotnet/nightly/monitor:7`
-* `6` (Standard Support)
+* `6` (LTS)
   * `docker pull mcr.microsoft.com/dotnet/nightly/monitor:6`
 
 ## Related Repos
--- a/.mar/portal/README.monitor.portal.md
+++ b/.mar/portal/README.monitor.portal.md
@@ -14,9 +14,9 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
 
 ## Featured Tags
 
-* `7` (Preview)
+* `7` (RC)
   * `docker pull mcr.microsoft.com/dotnet/nightly/monitor:7`
-* `6` (Preview)
+* `6` (Standard Support)
   * `docker pull mcr.microsoft.com/dotnet/nightly/monitor:6`
 
 ## Related Repos
--- a/.mar/portal/README.aspnet.portal.md
+++ b/.mar/portal/README.aspnet.portal.md
@@ -12,9 +12,9 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
 
 ## Featured Tags
 
-* `7.0` (Preview)
+* `7.0` (RC)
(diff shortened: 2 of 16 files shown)


.mar/portal/README.monitor.portal.md
```diff
@@ -14,9 +14,9 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
 
 ## Featured Tags
 
-* `7` (Preview)
+* `7` (RC)
   * `docker pull mcr.microsoft.com/dotnet/nightly/monitor:7`
-* `6` (Preview)
+* `6` (Standard Support)
```
MichaelSimons, line 19: What does standard support mean?  Would is make sense to drop any non-preview callout?

MichaelSimons (approved):

.mar/portal/README.monitor.portal.md
mthalman: "Standard Support" is the new approved replacement for "Current" (non-LTS release). It's what .NET 7 will also be labeled as once it's GA. Actually, now that I think of it, this should be labeled as LTS. Is that correct, @jander-msft? The `monitor:6` tag will be supported for the lifetime of .NET 6?

mthalman: > Would is make sense to drop any non-preview callout?

@MichaelSimons - Are you asking whether it would make sense to only show Preview/RC tags in the Featured section in the nightly branch?

MichaelSimons: No, I am asking about the `(Standard Support)` verbiage specifically.  Looking at the dotnet monitor [release page](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/releases.md), I am not seeing any differentiated support types.  If this is really the case, I am asking if it would make sense to simple drop `(Standard Support)`?  It implies to me there are other support types.

mthalman: @jander-msft - Let us know your thoughts on how you think this should be represented.

MichaelSimons: I missed your earlier response @mthalman on the switch to `Standard Support`.  That makes sense, I think the dotnet monitor [release page](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/releases.md) should use this terminology though so that there is a mapping.  @jander-msft

jander-msft: In terms of how long each version of dotnet-monitor is supported, we haven't made a public stance on that yet. I don't think the standard [.NET Support Policy](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) necessarily covers it either. I will discuss this with the team this morning and get back to you today.

jander-msft: Because .NET Monitor will align its support lifecycle with the [.NET Support Policy](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), the `6*` images can be labeled as `LTS` and the `7*` images labeled as `Standard Support` at GA. We will update the dotnet/dotnet-monitor repository with the appropriate lifecycle information.

MichaelSimons (approved):

jander-msft (approved):

PR #4144, opened 2022-10-14
dotnet-monitor: Set default GC mode

schmittjoseph:
This PR explicitly sets the default GC mode for `dotnet monitor` images to Server. This PR **does not** change the default behavior of any of the images as this is their current default, and instead ensures that future images keep their current behavior going forward as we change `dotnet monitor`'s GC defaults for other environments.

cc @dotnet/dotnet-monitor

### DIFF
--- a/eng/dockerfile-templates/monitor/Dockerfile
+++ b/eng/dockerfile-templates/monitor/Dockerfile
@@ -79,6 +79,8 @@ ENV \
     Logging__Console__FormatterOptions__TimestampFormat=yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffffff'Z' \
     # Logging: Write timestamps using UTC offset (+0:00)
     Logging__Console__FormatterOptions__UseUtcTimestamp=true \
+    # Server GC mode
+    DOTNET_gcServer=1 \
     # Add dotnet-monitor path to front of PATH for easier, prioritized execution
     PATH="/app:${PATH}"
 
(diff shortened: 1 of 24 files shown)


src/monitor/6.3/cbl-mariner/arm64v8/Dockerfile
```diff
@@ -53,6 +53,8 @@ ENV \
     Logging__Console__FormatterOptions__TimestampFormat=yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffffff'Z' \
     # Logging: Write timestamps using UTC offset (+0:00)
     Logging__Console__FormatterOptions__UseUtcTimestamp=true \
+    # Server GC mode
+    DOTNET_gcServer=1 \
```
jander-msft, line 57: Please alphabetize with the other environment variables

eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -73,6 +73,8 @@ ENV \
     DefaultProcess__Filters__0__Value=1 \
     # Remove Unix Domain Socket before starting diagnostic port server
     DiagnosticPort__DeleteEndpointOnStartup=true \
+    # Server GC mode
+    DOTNET_gcServer=1 \
```
jander-msft, line 77: Add this to the [VerifyEnvironmentVariables](https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs#L122) test.

jander-msft (approved):

jander-msft (approved):

mthalman (approved):

PR #4207, opened 2022-11-14
Use Dockerfile paths to determine which test data is in scope

### LINKED ISSUE #4185: .NET Monitor 6.X Tests Sometimes Not Run
### Describe the Bug

Sometimes, the [dotnet-docker-nightly](https://dev.azure.com/dnceng/internal/_build?definitionId=359&_a=summary) pipeline does not run tests for .NET Monitor 6.x. 

### Steps to Reproduce

See the following pipeline run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2036707&view=results

The .NET Monitor 6.2 and 6.3 images for cbl-mariner2.0 were rebuilt and published, however the tests for .NET Monitor 6.2 and 6.3 were not run. See the [Linux_amd64 6.0-cbl-mariner2.0-distroless-aspnet](https://dev.azure.com/dnceng/internal/_build/results?buildId=2036707&view=logs&j=e38aee6b-5c96-541e-b5a1-cdfc93948036&t=a89fa43f-4c2a-5df0-6f87-e9a312f4c614) test job and step where the .NET Monitor tests were skipped. Typically, the .NET Monitor tests for 6.2 and 6.3 run in their own job because the test jobs in CI builds are grouped by image version (instead of runtime version, like in PR builds) among other groupings.

mthalman:
Test jobs are not executing the tests that they should be, specifically for .NET Monitor. This is due to the dependency on the `productVersion` variable from the pipeline matrix. This variable is basically unreliable because it's a single value that isn't necessarily representative of the products that are targeted by the job.

For example, a test job can get generated for testing this set of paths:
* src/aspnet/6.0/cbl-mariner2.0-distroless/arm64v8
* src/runtime/6.0/cbl-mariner2.0-distroless/arm64v8
* src/runtime-deps/6.0/cbl-mariner2.0-distroless/arm64v8
* src/monitor/6.2/cbl-mariner-distroless/arm64v8
* src/sdk/6.0/cbl-mariner2.0/arm64v8
* src/aspnet/6.0/cbl-mariner2.0/arm64v8
* src/runtime/6.0/cbl-mariner2.0/arm64v8
* src/runtime-deps/6.0/cbl-mariner2.0/arm64v8
* src/monitor/6.2/cbl-mariner/arm64v8
* src/monitor/6.3/cbl-mariner/arm64v8
* src/monitor/6.3/cbl-mariner-distroless/arm64v8

The `productVersion` matrix variable for this test job is set to `6.0`. Because the test code filters out test data based the value of the product version, only test data which matches 6.0 will actually get tested. This causes all the Monitor tests for 6.2 and 6.3 to be skipped.

It's also not correct to simply change the `productVersion` variable to an array of versions. There's no indication of what products those versions are specific to.

The correct way to handle this is to base the testing of the Dockerfile paths in the same way that Image Builder uses the paths to determine what to build. To do this, I've updated the contract between the common pipeline and the test script to use a `Paths` argument instead of `Version`. The test project then filters the test data according to those paths by synthesizing Dockerfile paths with which to compare against.

By updating the tests to rely on paths instead of product version, it allowed me to clean up special logic that was added for .NET Monitor tests that attempted to compensate for this disparity between the main .NET product version and .NET Monitor product version. That logic is no longer necessary.

Since these changes also affect files in the eng/common location, I'll wait to get approval of the changes here first. Then I'll get those changes submitted to https://github.com/dotnet/docker-tools.

Fixes #4185


tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs
```diff
@@ -20,8 +20,10 @@ public AspnetImageTests(ITestOutputHelper outputHelper)
 
         protected override DotNetImageType ImageType => DotNetImageType.Aspnet;
 
+        public static IEnumerable<object[]> GetAspnetImageData() => GetImageData(DotNetImageType.Aspnet);
```
MichaelSimons, line 23: This doesn't need to be named GetAspnetImageData does it?  GetImageData is shorter, the aspnet core part isn't necessary given the context and GetImageData makes the MemberData attributes more consistent across the various image tests - e.g. SdkImageTests

tests/Microsoft.DotNet.Docker.Tests/Config.cs
```diff
@@ -34,6 +34,9 @@ public static class Config
             Environment.GetEnvironmentVariable("SAS_QUERY_STRING") ?? string.Empty;
         public static string NuGetFeedPassword { get; } =
             Environment.GetEnvironmentVariable("NUGET_FEED_PASSWORD") ?? string.Empty;
+        public static string[] Paths { get; } =
+        Environment.GetEnvironmentVariable("DOCKERFILE_PATHS")?
```
MichaelSimons, line 38: Indentation seems off.

tests/Microsoft.DotNet.Docker.Tests/DotNetImageType.cs
```diff
@@ -10,5 +10,6 @@ public enum DotNetImageType
         Runtime,
         Runtime_Deps,
         Aspnet,
+        Monitor
```
MichaelSimons, line 13: I like how cleaner things are by treating all image types the same (e.g. their image data).

eng/common/templates/stages/build-test-publish-repo.yml
```diff
@@ -60,6 +60,7 @@ stages:
         # values aren't actually used for the pre-build tests.
         - powershell: |
             echo "##vso[task.setvariable variable=productVersion]"
```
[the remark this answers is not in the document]
mthalman: It's only used for the testRunTitle: https://github.com/dotnet/dotnet-docker/blob/f668146eddfc052e6895fa3a994dd8e1ea4c53ff/eng/common/templates/steps/test-images-linux-client.yml#L90

Even though it's not completely representative of the versions being tested, I think it's still useful to have it here to better organize the test results.

MichaelSimons (approved):

PR #4214, opened 2022-11-16
Split update-deps pipeline into a separate one for Monitor

### LINKED ISSUE #4168: Split .NET Monitor updates out of the mainline .NET updates pipeline
### Describe the Problem

The release process for .NET Monitor is rotated amongst several folks on the team. As part of the release process, the team needs the ability to update the [dotnet-docker-update-dependencies](https://dev.azure.com/dnceng/internal/_build?definitionId=470) pipeline to acquire the correct builds of .NET Monitor. However, I am the only team member that has the ability to edit the pipeline; in my absence, my other team members would like to be able to update the pipeline as well.

This pipeline is responsible for updating all .NET products in the dotnet-docker repository. If our team was given edit access to the pipeline wholesale, we could accidentally change settings that would impact the mainline .NET products. Similarly, if our credentials were compromised, the pipeline may be updated in a malicious way. Both of these are mitigated by the fact that the pipeline only creates PRs into the public repository, which require manual approval and completion by the dotnet-docker team.

### Describe the Solution

We would request that the .NET Monitor portion of the pipeline is bifurcated into its own pipeline, for which the .NET Monitor team can be given edit access. The original pipeline would only update the mainline .NET products whereas the new pipeline would only update the .NET Monitor product.

mthalman:
Fixes #4168


eng/pipelines/update-dependencies-monitor.yml
```diff
@@ -0,0 +1,39 @@
...
+  displayName: M-F daily build
+  branches:
+    include:
+    - nightly
+  always: true
+variables:
+- template: ../common/templates/variables/dotnet/common.yml
+stages:
+- stage: Monitor
+  condition: and(not(canceled()), eq(variables['update-monitor-enabled'], 'true'))
```
MichaelSimons, line 14: Why is this variable needed now that the pipeline is split?  Isn't it implied you want to update the monitor images if the pipeline is invoked?

MichaelSimons (approved):

jander-msft (approved):

PR #4231, opened 2022-12-01
Update apps to cover memory reservation scenario

richlander:


samples/dotnetapp/Program.cs
```diff
@@ -101,22 +102,19 @@ string GetInBestUnit(long size)
     }
 }
 
-long GetBestValue(string[] paths)
+bool GetBestValue(string[] paths, out long limit, [NotNullWhen(true)] out string? bestPath)
```
mthalman, line 105: Rather than two `out` parameters, I prefer the pattern of just returning a nullable tuple and have no `out` parameters. Just a suggestion but I'm fine either way.

axljoker03 (approved):

samples/dotnetapp/Program.cs
richlander: Interesting pattern. I've never come across that before, but it makes sense. I'll just keep the code as-is, but may make that switch for next time.

richlander:
I just re-tested the two apps (per `Dockerfile` and `Dockerfile.chiseled`. They are working correctly.

samples/dotnetapp/Program.cs
```diff
@@ -69,18 +71,17 @@
...
-    long currentMemory = GetBestValue(currentMemoryPaths);
+    GetBestValue(currentMemoryPaths, out long currentMemory, out string? memoryPath);
 
-    if (memoryLimit > 0)
-    {
-        WriteLine($"cgroup memory limit: {memoryLimit} ({GetInBestUnit(memoryLimit)})");
-        WriteLine($"cgroup memory usage: {currentMemory} ({GetInBestUnit(currentMemory)})");
-        WriteLine($"GC Hard limit %: {(double)totalMemoryBytes/memoryLimit * 100:N0}");
-    }
+    WriteLine($"cgroup memory constraint: {bestPath}");
```
mthalman, line 81:
```suggestion
    WriteLine($"cgroup memory constraint: {bestMemoryLimitPath}");
```

PR #4242, opened 2022-12-07
Add .NET Monitor 8 Dockerfiles

jander-msft:
For .NET Monitor 8, the initial offering will be to only provide distroless and chiseled images. These images will follow the same tagging policy introduced for .NET 8.

cc @dotnet/dotnet-monitor


jander-msft:
Fix for failing tests is in #4244

mthalman (approved):

PR #4244, opened 2022-12-07
Use runtime-deps image for verifying insecure files

jander-msft:
The new .NET Monitor 8 version will only be offered as distroless images. This requires updating tests to not assume there is a companion distroful image that can be used for validating the existence of insecure files. The change uses the corresponding distroful runtime-deps image rather than assuming a companion distroful image exists.


mthalman (approved):
Nice clean solution

jander-msft:
@mthalman, feel free to merge. I don't have permission to do so.

PR #4248, opened 2022-12-09
Update expected packages for distroless Mariner

mthalman:
The latest servicing version of Mariner has caused a change in the expected packages that exist. We verify these expected packages for the distroless version and this new version has caused `tzdata` to be added and `prebuilt-ca-certificates-base` to be removed. This doesn't seem right and I'll continue to investigate what happened and log issues if necessary. But these changes update the tests to reflect the current status and unblock the build.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
@@ -118,7 +118,7 @@ namespace Microsoft.DotNet.Docker.Tests
                 "openssl",
                 "openssl-libs",
                 "prebuilt-ca-certificates",
-                "prebuilt-ca-certificates-base",
+                "tzdata",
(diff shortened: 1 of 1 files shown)


mthalman:
Ok, looks like these changes were intentional to make distroless Mariner a drop-in replacement for Google distroless to better support AKS: https://github.com/microsoft/CBL-Mariner/pull/4248.

The addition of `tzdata` changes the minimal base image size from 3.3 MB to 4.1 MB.

The "removal" of `prebuilt-ca-certificates-base` is actually a good thing. That used to exist in the minimal base image but we install `prebuilt-ca-certificates` which should be a replacement of it but because of distroless being distroless, removal of replaced packages isn't done automatically by the way we install packages. So before we ended up with both packages when all we really wanted was `prebuilt-ca-certificates`. Now that is happening. So that part is good.

MichaelSimons (approved):

PR #4255, opened 2022-12-12
Updates for December servicing

mthalman:
Note on .NET Monitor:
Since there is no release for .NET Monitor 6.2 this month, I've left it configured to be based on Alpine 3.16. But for 6.3 and 7.0, I've updated them to Alpine 3.17.

Related to https://github.com/dotnet/release/issues/446


eng/dockerfile-templates/monitor/Dockerfile
```diff
@@ -72,7 +72,7 @@ ENV \
     DefaultProcess__Filters__0__Key=ProcessId \
     DefaultProcess__Filters__0__Value=1 \
     # Remove Unix Domain Socket before starting diagnostic port server
-    DiagnosticPort__DeleteEndpointOnStartup=true \{{ if dotnetMajor = "7":
+    DiagnosticPort__DeleteEndpointOnStartup=true \{{ if monitorMajorMinor != "6.2":
```
jander-msft, line 75: This change does not appear to be in `nightly` and `main` has the condition as `dotnetMajor = "7"`. Could you revert this to what is in `main` to keep the released images consistent and not impact 6.3 (it shouldn't have a behavioral change, but lets not chance it)? Once this is merged back to `nightly`, I can adjust it to be what it really should be (e.g. `dotnetMajor >= 7`).

mthalman: Ok, so nightly is configured incorrectly then? It should only be applied to 7.0?

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -163,7 +156,7 @@ public void VerifyEnvironmentVariables(ProductImageData imageData)
             variables.Add(new EnvironmentVariableInfo("DefaultProcess__Filters__0__Value", "1"));
             // Existing (orphaned) diagnostic port should be delete before starting server
             variables.Add(new EnvironmentVariableInfo("DiagnosticPort__DeleteEndpointOnStartup", "true"));
-            if (imageData.Version.Major == 7)
+            if (imageData.Version.ToString() != "6.2")
```
jander-msft, line 159: Same here in reference to https://github.com/dotnet/dotnet-docker/pull/4255#discussion_r1046326882

eng/dockerfile-templates/monitor/Dockerfile
jander-msft: Correct. When this addition was shipping, we were only updating the 7.0 images, so you added a change to scope it to only 7.0 (which is what we really should have done in `nightly`).

mthalman: Well, the reason I added the condition to main was to prevent the change to be applied in 6.3 at the time because 6.3 wasn't being serviced. Since 6.3 is now being serviced, I made the change to have it applied. Ok, so I'll revert this. Thanks for the clarification.

mthalman: Actually, I'll change it to explicitly disable it for 6.2/6.3 so it will get applied for 8.0.

jander-msft (approved):

MichaelSimons (approved):

PR #4265, opened 2022-12-14
Vulnerability doc improvements

mthalman:
The vulnerability doc intended to use the `click` functionality of mermaid to provide hyperlink behavior to the elements in the flowchart. This doesn't work as intended. When clicking such a link, it appears that the mermaid renderer attempts to open the link in the mermaid "box" rather than on the main browser. It then shows an error that says "This content is blocked. Contact the site owner to fix the issue".

So I've removed the use of those `click` definitions. Without that, the document lacks a good experience to walk through the flowchart. To improve things, I've updated each details section of the doc that represents a flowchart question so that it will now include hyperlinks to each of the possible responses. This allows the reader to guide themselves via the details section rather than the flowchart graphic. The graphic is kept as a visualization of the overall process.

### DIFF
--- a/documentation/vulnerability-reporting.md
+++ b/documentation/vulnerability-reporting.md
@@ -1,9 +1,7 @@
 # Container Vulnerability Workflow
 
-This documented is intended to help guide you to the appropriate course of action when encountering reported vulnerabilities in the .NET container images.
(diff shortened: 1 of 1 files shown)


MichaelSimons (approved):
It's too bad the links aren't supported, I liked them.  

PS when reviewing this I was reminded of the _choose your own adventure books I read growing up_

Issue #4314, opened 2023-01-10
Automated .NET Monitor Updates for Releases


jander-msft:
### Describe the Problem

The .NET Monitor team is making an effort to automate more of the release process for the .NET Monitor product to ensure a more consistent release process that requires less manual intervention steps. One part of this effort is the desire to perform automated updates of the .NET Monitor release candidates into the dotnet/dotnet-docker repository.

Currently, the way that release candidates are inserted is using a scheduled nightly trigger on the [update-dependencies-monitor](https://github.com/dotnet/dotnet-docker/blob/nightly/eng/pipelines/update-dependencies-monitor.yml) pipeline (corresponding [Azure DevOps pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=1207)). This pipeline has two primary pitfalls as it is:
1. The pipeline scans for each major version of .NET Monitor and updates them simulatneously; it doesn't afford for only updating a single major version which can unintentionally update major versions that are not ready to be released.
2. A person has to manually update the pipeline with the correct release variables before a release candidate is made available and then has to manually update the pipeline again after the release is published to revert it to appropriate values.

### Describe the Solution

1. Update the pipeline to only scan on a scheduled trigger for the in-development .NET Monitor version (currently the 8.0 version).
2. Add a [pipeline resource trigger](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/resources?view=azure-devops&tabs=schema#define-a-pipelines-resource) that references the .NET Monitor release pipeline. When triggered from the pipeline resource, version information is read from the release pipeline to indicate the product version and major/minor version that needs to be updated instead of reading the information from a blob URL. The release pipeline will be in the same organization and project as the update pipeline.

### Additional Context

- These changes do not affect the target branch (currently `nightly`).
- These changes preserve the nightly update of the in-development .NET Monitor version at the existing schedule (currently daily Monday through Friday). One way to think of this is that the in-development version is updated using a pull model (where the update pipeline polls for the updates as it does today) and releases are ingested using a push model (where the .NET Monitor release pipeline instructs the update pipeline to execute for a specific release).
- These changes do not affect the need for PRs generated from the update pipeline to be manually reviewed and approved.

PR #4315, opened 2023-01-10
Only update one .NET Monitor version in update dependencies pipeline

jander-msft:
This change simplifies the update dependencies pipeline for .NET Monitor to only scan for updates for the version configured by the `MajorMinorVersion` and `ChannelQuality` pipeline variables, which will be set to `8.0` and `daily` respectively.

Post-merge, the [Azure DevOps pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=1207&_a=summary) will be updated to reflect the variable changes.

Addresses the first issue in #4314

cc @dotnet/dotnet-monitor


mthalman (approved):

mthalman:
@jander-msft - You should have permission to merge now. I'll let you do that when you're ready so you can coordinate the pipeline configuration with it.

PR #4317, opened 2023-01-10
Update projects to .NET 7

lbussell:
Part of https://github.com/dotnet/docker-tools/issues/1078

This might need to wait on https://github.com/dotnet/docker-tools/pull/1104, not sure.

### DIFF
--- a/eng/common/Install-DotNetSdk.ps1
+++ b/eng/common/Install-DotNetSdk.ps1
@@ -40,7 +40,7 @@ if (!(Test-Path $DotnetInstallScriptPath)) {
     & "$PSScriptRoot/Invoke-WithRetry.ps1" "Invoke-WebRequest 'https://dot.net/v1/$DotnetInstallScript' -OutFile $DotnetInstallScriptPath"
 }
 
-$DotnetChannel = "6.0"
+$DotnetChannel = "7.0"
 
 $InstallFailed = $false
 if ($IsRunningOnUnix) {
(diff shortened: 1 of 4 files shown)


eng/common/Install-DotNetSdk.ps1
```diff
@@ -40,7 +40,7 @@ if (!(Test-Path $DotnetInstallScriptPath)) {
     & "$PSScriptRoot/Invoke-WithRetry.ps1" "Invoke-WebRequest 'https://dot.net/v1/$DotnetInstallScript' -OutFile $DotnetInstallScriptPath"
 }
 
-$DotnetChannel = "6.0"
+$DotnetChannel = "7.0"
```
MichaelSimons, line 43: We will have to be on the watch for any eng/common dependency flows otherwise this will get reverted until this change is made to docker-tools.

MichaelSimons (approved):

eng/common/Install-DotNetSdk.ps1
mthalman: Yes, this change is already made in https://github.com/dotnet/docker-tools/pull/1104 so that should get merged first. That will flow into this repo.

lbussell:
Superseded by https://github.com/dotnet/dotnet-docker/pull/4327.

richlander:
I'm wondering if we should take a more aggressive posture and help test the product in the process. We could starting using .NET 8 after Preview 1. In the past, we've hosted the marketing site starting with Preview 1.

Issue #4322, opened 2023-01-11
.NET 8 Debian images should be based on Bookworm


richlander:
We expect that Debian Bookworm will ship before .NET 8. We did this same race with .NET 6 and Debian Bullseye. Fortunately, .NET 6 lost that race and we expect that .NET 8 will similarly lose the race with Bookworm.

For .NET 6, we shipped Bullseye images starting with .NET 6 Preview 1. We should use the same model here.

- https://www.debian.org/releases/bookworm/
- https://wiki.debian.org/DebianBookworm
- https://devblogs.microsoft.com/dotnet/announcing-net-6/#debian-11-bullseye

richlander:
We need to decide on broad test coverage for Bookworm. Do we need a new test leg for this on dotnet/runtime (for example)?

@mmitche @ViktorHofer

ViktorHofer:
In dotnet/runtime we currently test both Debian 10 and 11. Based on https://wiki.debian.org/DebianReleases, Debian 10 already reached EOL. Would it make sense to remove 10 and add 12?

ViktorHofer:
cc @wfurt @ericstj @carlossanlop

mthalman:
[Triage]
We should target .NET 8 Preview 1 to get the Debian Dockerfiles updated to be on Bookworm.

lbussell:
Task tracking info for this work:

Distro: Debian 12 "Bookworm"

## ImageBuilder Tasks

- [ ] Ensure that the ImageBuilder supports the new distro version in the [code](https://github.com/dotnet/docker-tools/blob/main/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs) to generate the correct README display name from the version specified in the manifest

## Nightly Branch Tasks

1. - [ ] Copy the Dockerfiles of the most recent version of that distro (or author new ones for an entirely new distro) and place them in a version-specific folder under their respective variants (runtime-deps, runtime, aspnet, sdk). Note: not all variants have a corresponding runtime-deps image.
1. - [ ] Modify the Dockerfiles as appropriate for any specific changes related to the new distro version
1. - [ ] Update [manifest.json](https://github.com/dotnet/dotnet-docker/blob/nightly/manifest.json) to reference the new set of Dockerfiles with the appropriate tags
      - [ ] Move any distro-specific floating tags to the newer version (e.g. `6.0-alpine`)
[...]

PR #4330, opened 2023-01-13
Move .NET 8 preview images to Debian 12 Bookworm

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/4322


mthalman (approved):
Looks great! Glad to see the only updates needed were libicu and libssl versions.

MichaelSimons (commented):
There appear to be some aspnet/8.0 Dockerfiles that were copied incorrectly.  The paths contain both bookworm and bullseye.

See https://github.com/dotnet/dotnet-docker/blob/27e052fe1e49a2aca3f0bbfd48e39f2630dc68f3/src/aspnet/8.0/bookworm-slim/bullseye-slim/amd64/Dockerfile

MichaelSimons (approved):

mthalman:
Looks like pipeline triggers are having issues.

image

Issue #4300, opened 2023-01-05
Consider defining a helper "health check" utility for distroless scenarios


mthalman:
The use of the `HEALTHCHECK` instruction often involves the use of a command that executes an HTTP request to the application in order to verify its health. That command might be `curl` or `wget` for Linux containers. This becomes problematic for distroless/chiseled containers that would likely want to exclude such commands. How then would a call be made to verify the health of the app?

One solution is to include a basic .NET application that execute the HTTP request (see https://github.com/dotnet/dotnet-docker/discussions/4296). Depending on the ubiquity of such a solution, it might make sense for .NET to provide its own implementation of this application that can be easily injected into .NET distroless/chiseled solutions.

I'll illustrate this with an example. Let's say .NET publishes an `httpservice` image that contains a .NET app which simply sends an HTTP request to a URL provided as an argument. It returns 0 if the request succeeds and 1 if it fails. A distroless implementation could consume this app like the following:

```Dockerfile
# Build stage excluded for brevity
...

FROM mcr.microsoft.com/dotnet/nightly/runtime-deps:7.0-jammy-chiseled
WORKDIR /app
COPY --from=build /app .

# Inject the .NET httpservice into the image to use for the health check
COPY --from=mcr.microsoft.com/dotnet/httpservice:7.0 /httpservice.exe /
HEALTHCHECK CMD [ "/httpservice.exe", "http://localhost/healthz" ]

ENTRYPOINT ["./app"]
```

I think it'd be important to understand the landscape of how customers are using health checks to determine the suitability of a pre-defined tool like this. We obviously don't want to be reimplementing curl in .NET but allowing for a little bit of behavior customization in order to fulfill a large portion of customer scenarios seems reasonable.

mthalman:
[Triage]
As a first step, we'd like to provide documentation in the repo to help guide customer's on the pattern to implement this on their own in a way that doesn't depend on .NET providing a new tool. This is a low-cost solution and can also be used as a means to gather feedback from customers to determine the optimal solution that would meet their needs if we were to publish our own tool.

64J0:
Does this affect Kubernetes' probes? 

I'm not 100% sure, but I think the probes commands are handled outside the container, so it must work even in a distroless environment. Is it correct?

richlander:
> Does this affect Kubernetes' probes?

Not needed for Kubernetes. This is a docker compose need.

Kralizek:
[...]

PR #4352, opened 2023-01-24
Add basic HTTP health check

richlander:
I followed these instructions: https://learn.microsoft.com/aspnet/core/host-and-deploy/health-checks.

I tested with `curl` and Kubernetes. Both worked.

```bash
% curl http://localhost:8000/healthz
Healthy
```

### DIFF
--- a/samples/aspnetapp/Dockerfile
+++ b/samples/aspnetapp/Dockerfile
@@ -15,7 +15,7 @@ FROM mcr.microsoft.com/dotnet/aspnet:7.0
 RUN apt-get update \ 
     && apt-get install -y --no-install-recommends curl \
     && rm -rf /var/lib/apt/lists/*
-# This port needs to map the port being used
+# This port needs to match the port being used
 HEALTHCHECK CMD curl http://localhost:80/healthz
 WORKDIR /app
 COPY --from=build /app .
--- a/samples/aspnetapp/Dockerfile
+++ b/samples/aspnetapp/Dockerfile
@@ -12,11 +12,6 @@ RUN dotnet publish -c Release -o /app --use-current-runtime --self-contained fal
 
 # final stage/image
 FROM mcr.microsoft.com/dotnet/aspnet:7.0
-RUN apt-get update \ 
-    && apt-get install -y --no-install-recommends curl \
-    && rm -rf /var/lib/apt/lists/*
-# This port needs to match the port being used
-HEALTHCHECK CMD curl http://localhost:80/healthz
 WORKDIR /app
 COPY --from=build /app .
 ENTRYPOINT ["dotnet", "aspnetapp.dll"]
--- a/samples/aspnetapp/Dockerfile
+++ b/samples/aspnetapp/Dockerfile
@@ -15,7 +15,8 @@ FROM mcr.microsoft.com/dotnet/aspnet:7.0
 RUN apt-get update \ 
     && apt-get install -y --no-install-recommends curl \
     && rm -rf /var/lib/apt/lists/*
-HEALTHCHECK CMD curl http://localhost/healthz
+# This port needs to map the port being used
+HEALTHCHECK CMD curl http://localhost:80/healthz
 WORKDIR /app
 COPY --from=build /app .
 ENTRYPOINT ["dotnet", "aspnetapp.dll"]
--- a/samples/aspnetapp/Dockerfile
+++ b/samples/aspnetapp/Dockerfile
@@ -12,6 +12,10 @@ RUN dotnet publish -c Release -o /app --use-current-runtime --self-contained fal
 
 # final stage/image
 FROM mcr.microsoft.com/dotnet/aspnet:7.0
+RUN apt-get update \ 
+    && apt-get install -y --no-install-recommends curl \
+    && rm -rf /var/lib/apt/lists/*
+HEALTHCHECK CMD curl http://localhost/healthz
 WORKDIR /app
 COPY --from=build /app .
 ENTRYPOINT ["dotnet", "aspnetapp.dll"]
--- a/samples/aspnetapp/Dockerfile.alpine-slim
+++ b/samples/aspnetapp/Dockerfile.alpine-slim
@@ -8,14 +8,14 @@ RUN dotnet restore --use-current-runtime /p:PublishReadyToRun=true
 
 # copy everything else and build app
 COPY aspnetapp/. .
-RUN dotnet publish -c Release -o /app --use-current-runtime --self-contained true --no-restore /p:PublishTrimmed=true /p:PublishReadyToRun=true /p:PublishSingleFile=true
+RUN dotnet publish -c Release -o /app --use-current-runtime --no-restore /p:PublishTrimmed=true /p:PublishReadyToRun=true /p:PublishSingleFile=true
 
 # final stage/image
 FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine
-# This port needs to match the port being used
-HEALTHCHECK CMD wget http://localhost:80/healthz
 WORKDIR /app
 COPY --from=build /app .
+# This port needs to match the port being used
+HEALTHCHECK CMD wget http://localhost:80/healthz
 ENTRYPOINT ["./aspnetapp"]
 
 # See: https://github.com/dotnet/announcements/issues/20
--- a/samples/aspnetapp/Dockerfile.alpine-slim
+++ b/samples/aspnetapp/Dockerfile.alpine-slim
@@ -16,7 +16,7 @@ WORKDIR /app
 COPY --from=build /app .
 
 # This port needs to match the port being used
-HEALTHCHECK CMD wget http://localhost:80/healthz
+HEALTHCHECK CMD wget -qO- -t1 http://localhost:80/healthz
 ENTRYPOINT ["./aspnetapp"]
 
 # See: https://github.com/dotnet/announcements/issues/20
--- a/samples/aspnetapp/Dockerfile.alpine-slim
+++ b/samples/aspnetapp/Dockerfile.alpine-slim
@@ -16,7 +16,7 @@ WORKDIR /app
 COPY --from=build /app .
 
 # This port needs to match the port being used
-HEALTHCHECK CMD wget -qO- -t1 http://localhost:80/healthz
+HEALTHCHECK CMD wget -qO- -t1 http://localhost:80/healthz || exit 1
 ENTRYPOINT ["./aspnetapp"]
 
 # See: https://github.com/dotnet/announcements/issues/20
--- a/samples/aspnetapp/Dockerfile.alpine-slim
+++ b/samples/aspnetapp/Dockerfile.alpine-slim
@@ -14,6 +14,7 @@ RUN dotnet publish -c Release -o /app --use-current-runtime --no-restore /p:Publ
 FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine
 WORKDIR /app
 COPY --from=build /app .
+
 # This port needs to match the port being used
 HEALTHCHECK CMD wget http://localhost:80/healthz
 ENTRYPOINT ["./aspnetapp"]
--- a/samples/aspnetapp/README.md
+++ b/samples/aspnetapp/README.md
@@ -81,7 +81,7 @@ CONTAINER ID   IMAGE       COMMAND         CREATED         STATUS               
 b143cf4ac0d1   aspnetapp   "./aspnetapp"   8 seconds ago   Up 7 seconds (health: starting)   0.0.0.0:8000->80/tcp   fervent_lichterman
 ```
 
-After a while, you will see that the app is no longer "starting".
+After 30s, the status should transition to "healthy" from "health: starting".
 
 You can also look at health status with `docker inspect`. The following pattern uses `jq`, which makes it much easier to drill in on the interesting data.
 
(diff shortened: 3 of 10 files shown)


mthalman (commented):
In my mind, having the app capable of doing a health check and leaving it at that, it doesn't go far enough. We should also demonstrate how it's used in the Dockerfile. I think we should update the Dockerfiles to include a `HEALTHCHECK` instruction. This is still useful in plain old Docker, outside the context of Kubernetes. The health check status will show up in a `docker ps` output of the running sample container.

richlander:
I will do that Matt. That's a good idea.

richlander:
Before I go any further, I want to validate the change for just `Dockerfile`. I'm happy to make the same change to the others.

Before:

```bash
$ docker run --rm -d -p 8000:80 mcr.microsoft.com/dotnet/samples:aspnetapp
42f412fdb600bbdc515ec7839f2632bdb3658400f76e021b1da83b61544f8033
$ docker ps -l
CONTAINER ID   IMAGE                                        COMMAND         CREATED         STATUS         PORTS                  NAMES
42f412fdb600   mcr.microsoft.com/dotnet/samples:aspnetapp   "./aspnetapp"   4 seconds ago   Up 3 seconds   0.0.0.0:8000->80/tcp   elegant_buck
```

After:

```bash
$ docker run --rm -d -p 8000:80 aspnetapp
5a992022ff999d10e82992b7f729105ba12ef630655a3a5ccf0f76424d61f6bc
$ docker ps -l
CONTAINER ID   IMAGE       COMMAND                  CREATED          STATUS                             PORTS
       NAMES
5a992022ff99   aspnetapp   "dotnet aspnetapp.dll"   10 seconds ago   Up 10 seconds (health: starting)   0.0.0.0:8000->80/tcp   beautiful_kilby
docker ps -l
CONTAINER ID   IMAGE       COMMAND                  CREATED          STATUS                    PORTS                  NAMES
5a992022ff99   aspnetapp   "dotnet aspnetapp.dll"   34 seconds ago   Up 33 seconds (healthy)   0.0.0.0:8000->80/tcp   beautiful_kilby
```

Looks good?

samples/aspnetapp/Dockerfile
```diff
@@ -12,6 +12,10 @@ RUN dotnet publish -c Release -o /app --use-current-runtime --self-contained fal
 
 # final stage/image
 FROM mcr.microsoft.com/dotnet/aspnet:7.0
+RUN apt-get update \ 
+    && apt-get install -y --no-install-recommends curl \
+    && rm -rf /var/lib/apt/lists/*
+HEALTHCHECK CMD curl http://localhost/healthz
```
jander-msft, line 18: Does this need to somehow account for if the user specifies HTTPS_PORTS OR HTTPS_PORTS environment variables on the container?

richlander: That's a great point. Ports suck.

With .NET 8, that will be easier. This is still .NET 7. I don't what to do. I added a comment.

samples/aspnetapp/Dockerfile
```diff
@@ -12,6 +12,11 @@ RUN dotnet publish -c Release -o /app --use-current-runtime --self-contained fal
 
 # final stage/image
 FROM mcr.microsoft.com/dotnet/aspnet:7.0
+RUN apt-get update \ 
+    && apt-get install -y --no-install-recommends curl \
+    && rm -rf /var/lib/apt/lists/*
+# This port needs to map the port being used
```
richlander, line 18:
```suggestion
# This port needs to match the port being used
```

samples/aspnetapp/Dockerfile
```diff
@@ -12,6 +12,11 @@ RUN dotnet publish -c Release -o /app --use-current-runtime --self-contained fal
 
 # final stage/image
 FROM mcr.microsoft.com/dotnet/aspnet:7.0
+RUN apt-get update \
```
MichaelSimons, line 15: This is intended to be a multi-platform Dockerfile.  This breaks that contract.  This is another case for needing https://github.com/dotnet/dotnet-docker/issues/4300.

samples/aspnetapp/Dockerfile
```diff
@@ -12,6 +12,11 @@ RUN dotnet publish -c Release -o /app --use-current-runtime --self-contained fal
 
 # final stage/image
 FROM mcr.microsoft.com/dotnet/aspnet:7.0
+RUN apt-get update \ 
+    && apt-get install -y --no-install-recommends curl \
+    && rm -rf /var/lib/apt/lists/*
+# This port needs to match the port being used
+HEALTHCHECK CMD curl http://localhost:80/healthz
```
MichaelSimons, line 19: When I first saw this, I thought it was a bit weird to see the HEALTHCHECK instruction prior to the app being copied into the image.  After thinking about it, it makes a bit of sense in that it never changes and therefore can better utilize the docker layer cache in a build environment.  That aside I wonder if it will be confusing to the reader.

samples/aspnetapp/Dockerfile
richlander: I also think that this approach is ugly and we should provide a helper app. My intent was to demonstrate the basic case and see where the conversation led.

samples/aspnetapp/Dockerfile
richlander: I think this approach is the correct one. The caching doesn't add much but it is caching all the same. The `ASPNET_CORE_URLS` ENV is added before all of this. It's just layering of one form or another.

samples/aspnetapp/Dockerfile
mthalman: > I also think that this approach is ugly and we should provide a helper app.

Even with the helper app, I'm wondering whether this is going to overcomplicate the sample and it should be factored out into a separate Dockerfile that demonstrates the health check.

richlander:
We could do the healthcheck in a separate Dockerfile. However, we'd want to push that version to MCR.

richlander:
I switched to Alpine (because `wget` is there) and added a new pattern that could be used to collapse other Alpine Dockerfiles. I'm happy to do that in another PR.

In our initial conversations, we were going to call this one `Dockerfile.health`. It would be best to avoid that if we have the goal of enable health checks throughout our fleet of samples.

I will add this to the manifest.json file once everyone is happy with the approach. I tested it and it works as expected.

mthalman (commented):
I think these file names should indicate their difference from the result of the samples. How about `Dockerfile.healthcheck-alpine-slim` and `Dockerfile.healthcheck-nanoserver-slim`?

The https://github.com/dotnet/dotnet-docker/blob/main/manifest.samples.json file should be updated to reference these Dockerfiles so that they will be published.

Did we also want to update https://github.com/dotnet/dotnet-docker/blob/main/samples/aspnetapp/README.md to have a blurb that pointed people to these Dockerfiles and explained the health check?

samples/aspnetapp/Dockerfile.alpine-slim
```diff
@@ -0,0 +1,29 @@
...
+
+# copy everything else and build app
+COPY aspnetapp/. .
+RUN dotnet publish -c Release -o /app --use-current-runtime --no-restore /p:PublishTrimmed=true /p:PublishReadyToRun=true /p:PublishSingleFile=true
+
+# final stage/image
+FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine
+WORKDIR /app
+COPY --from=build /app .
+# This port needs to match the port being used
```
MichaelSimons, line 17: IMO this would benefit by having more whitespace.  Specifically an empty line before this comment would help make it standout more.

richlander:
```suggestion

# This port needs to match the port being used
```

richlander:
On the naming, it's a question of where we are headed with this. I chose the naming w/o healthcheck with the idea that:

- These new Dockerfiles would replace the existing slim ones (for Alpine and Nanoserver to start).
- It would be odd to not have the health check in the Alpine and Nanoserver ones given that it is just one more (useful) line.
- We'll design an approach for the rest of the dockerfiles in a bit.
- We can point out that we have healthcheck examples from other docs/READMEs.

richlander:
You can check the output of health checks via `docker inspect`. How you use `curl` and `wget` matters for that. I also discovered that the details of using these tools differs a fair bit.

On Alpine, the following appears to be the best approach: `HEALTHCHECK CMD wget -qO- -t1 http://localhost:80/healthz`.

It produces the following output for the success case:

$ docker inspect 21ec64322542 | jq .[-1].State.Health
{
  "Status": "healthy",
  "FailingStreak": 0,
  "Log": [
    {
      "Start": "2023-01-26T23:18:42.790331305Z",
      "End": "2023-01-26T23:18:42.953258349Z",
      "ExitCode": 0,
      "Output": "Healthy"
    },
    {
      "Start": "2023-01-26T23:19:12.962127063Z",
      "End": "2023-01-26T23:19:13.095388607Z",
      "ExitCode": 0,
      "Output": "Healthy"
    }
  ]
}

And for the failure case:

```bash
$ docker inspect 0fd5206a0f1b | jq .[-1].State.Health
{
  "Status": "starting",
  "FailingStreak": 2,
  "Log": [
    {
      "Start": "2023-01-26T23:25:14.920091363Z",
      "End": "2023-01-26T23:25:15.090543745Z",
      "ExitCode": 1,
      "Output": "wget: server returned error: HTTP/1.1 404 Not Found\n"
    },
    {
      "Start": "2023-01-26T23:25:45.098815567Z",
      "End": "2023-01-26T23:25:45.168346131Z",
      "ExitCode": 1,
      "Output": "wget: server returned error: HTTP/1.1 404 Not Found\n"
    }
  ]
}
```

I found that Alpine and Ubuntu had different behavior, in terms of the silent behavior while still printing error information. I ended up with the following.

- On Alpine, use `-q`.
-  On Ubuntu, use `-nv`.

Works?

samples/aspnetapp/README.md
```diff
@@ -45,35 +62,49 @@ We recommend that you do not use `--rm` in production. It cleans up container re
...
 
-```console
-docker build -t aspnetapp -f .\Dockerfile.windowsservercore-iis-x64 .
-docker run --rm -it -p:8080:80 aspnetapp
+After a while, you will see that the app is no longer "starting".
+
+You can also look at health status with `docker inspect`. The following pattern uses `jq`, which makes it much easier to drill in on the interesting data.
+
+```bash
+$ docker inspect b143cf4ac0d1 | jq .[-1].State.Health
```
mthalman, line 89: Consider providing a PowerShell version of this as well.

mthalman (commented):
You'll need to also run `eng/readme-templates/Get-GeneratedReadmes.ps1` to update the readmes due to the Dockerfile path change in the manifest.

samples/aspnetapp/README.md
richlander: Ah, you caught me being lazy. Gah!

Saibamen:
This feature will be enabled in base Docker images?
It could be enabled by setting ENV variable (to make sure that Developer enables Healthz on code-level).

richlander:
That occurred to me, but it doesn't sound like the best model. Setting an ENV and opting into `HEALTHCHECK` seems pretty similar. We want to ensure that all the capabilities are in place to use `HEALTHCHECK`. It's 100% possible today, just less pretty and convenient than we realized.

samples/aspnetapp/README.md
```diff
@@ -45,35 +62,64 @@ We recommend that you do not use `--rm` in production. It cleans up container re
...
-aspnetapp           latest              b2f0ecb7bdf9        About an hour ago   353MB
-aspnetapp           nanoserver          d4b7586827f2        About an hour ago   353MB
+```bash
+$ docker ps
+CONTAINER ID   IMAGE       COMMAND         CREATED         STATUS                            PORTS                  NAMES
+b143cf4ac0d1   aspnetapp   "./aspnetapp"   8 seconds ago   Up 7 seconds (health: starting)   0.0.0.0:8000->80/tcp   fervent_lichterman
 ```
 
-## Build an image for Windows Server Core
+After a while, you will see that the app is no longer "starting".
```
richlander, line 84:
```suggestion
After 30s, the status should transition to "healthy" from "health: starting".
```

richlander:
Can I do args for the nanoserver Dockerfile? Ideally, we could avoid duplicating the Dockerfile just to accommodate the difference between the two FROMs.

mthalman:
> Can I do args for the nanoserver Dockerfile? Ideally, we could avoid duplicating the Dockerfile just to accommodate the difference between the two FROMs.

Can you explain with an example? What are you looking to reuse that you'd use ARGs for?

richlander:
`Docker.nanoserver-slim`, for multiple versions of nanoserver. Right now, it is targeting just one and I'm not sure how to resolve that.

Saibamen:
> Is this a capability that you will use? Have you been using healthcheck to date? If we only implement it in these samples, will you apply that pattern to your Dockerfiles?

No, I was not using healthchecks before, but I think I will be using it for Docker images where we don't have implemented `"api/version" controller's action

mthalman:
> `Docker.nanoserver-slim`, for multiple versions of nanoserver. Right now, it is targeting just one and I'm not sure how to resolve that.

Ah, got it. I don't have a problem using an ARG for that, as long as it gets defaulted.

richlander:
I was super unclear as to what I was asking. How do I specify an arg in the manifests.json file? I didn't see that used?

mthalman:
> I was super unclear as to what I was asking. How do I specify an arg in the manifests.json file? I didn't see that used?

Here's an example: https://github.com/dotnet/docker-tools/blob/bf5e6ad6053b16d023b4021bef892403783709d2/src/Microsoft.DotNet.ImageBuilder/manifest.json#L28

mthalman (approved):

PR #4376, opened 2023-02-01
.NET Monitor: Use tarballs instead of tool package

### LINKED ISSUE #4337: .NET Monitor: Use archive files instead of .NET tool package for installation
### Describe the Problem

The .NET Monitor Dockerfiles install the dotnet-monitor tool using `dotnet tool install` which installs the platform neutral variant of dotnet-monitor into the image. Post installation steps are necessary to remove files in order to minimize the image, such as the alternate TFMs (the NuGet package includes two TFMs), native files for other platforms, and extra NuGet package files that `dotnet tool install` writes out. The removal process looks like:

```
    dotnet tool install dotnet-monitor --tool-path /app --add-source / --version $DOTNET_MONITOR_VERSION --framework net7.0 --no-cache \
    # To reduce image size, remove all non-net7.0 TFMs
    # To do this safely, we need to first find everything that dotnet tool installed named "dotnet-monitor".
    # The /dotnet-monitor/ folder exists under .store which is not a stable constant to rely on. The result is multi stage search:
    # 1. Find any files in /app
    # 2. Match anything that is under a *folder* called 'dotnet-monitor'
    # 3. Match anything from step 2 that isn't in a '/tools/net7.0' folder (/tools is the folder we use in a nuget file that is stable)
    # 4. Delete everything from step 3
    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -v -i '.*/tools/net7[.]0' | xargs rm -rf \
    # To reduce image size further, remove the non-linux assemblies
    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(win|osx)' | xargs rm -rf \
    # To reduce image size further, remove linux assemblies that do not match the x64 architecture
    && find /app -type f -print | grep -i '.*/dotnet-monitor/.*' | grep -i -E '/(runtimes|shared|shims)/(linux-arm64|linux-musl-arm64)' | xargs rm -rf \
    # To reduce image size further, remove symbol files
    && find /app -type f \( -name *.pdb -o -name *.dbg \) -print | xargs rm -rf \
    # Remove all the empty directories left by the previous step
    && find /app -type d -empty -delete \
    # Allow other users to run the tool
    && chmod 755 /app/dotnet-monitor \
    && rm dotnet-monitor.$DOTNET_MONITOR_VERSION.nupkg
```

As the tool evolves over time, this procedure requires manual adjustment to trim down the files to what is necessary for the image.

### Describe the Solution

Currently, the .NET Monitor team is working on new packages in the form of archives (zip for Windows, tar.gz for non-Windows) in addition to maintaining the current .NET Tool offering. The archives are a RID-specific, single TFM offering that only include the necessary files for the target platform and TFM. These archives will be available on the typical blob storage accounts (e.g. `dotnetbuilds` and `dotnetcli`) thus acquisition would be similar to runtime, aspnet, and SDK.

An example file name is `dotnet-runtime-7.0.2-linux-x64.tar.gz` which targets Linux x64 platforms with a .NET 7 shared framework installation.

The Dockerfiles should be updated to use the new archive offering when it is available.

jander-msft:
This change uses the newly produced tarball files in favor of the .NET tool package version in order to minimize the cleanup logic when installing the .NET tool package (this package has multiple TFMs in it, which bloats the image significantly without the cleanup logic).

The Dockerfile templates were also refactored in the same way that the runtime and aspnet templates are done but with .NET Monitor specific changes. This allows for better consistency across the product templates and removes the need for a multi-stage build for the Alpine images (the CBL-Mariner images require multi-stage due to .NET Monitor packages being tarballs rather than rpm packages like on the runtime/aspnet/etc images, so extra dependencies for downloading and unpacking the tarballs are needed in an installer image).

closes #4337 

cc @dotnet/dotnet-monitor


eng/dockerfile-templates/monitor/Dockerfile.linux
```diff
@@ -0,0 +1,55 @@
...
+    set isDistroless to find(OS_VERSION, "distroless") >= 0 || find(OS_VERSION, "chiseled") >= 0 ^
+    set isSingleStage to isAlpine ^
+    set aspnetBaseTag to
+        cat("$REPO:", VARIABLES[cat("dotnet|", dotnetMajorMinor, "|product-version")], "-", OS_VERSION, ARCH_TAG_SUFFIX) ^
+    set osVersionBase to when(isDistroless, match(OS_VERSION, ".+(?=.*-)")[0], OS_VERSION_BASE) ^
+    set installerImageTag to when(isMariner,
+        cat("mcr.microsoft.com/cbl-mariner/base/core:", OS_VERSION_NUMBER),
+        when(isAlpine,
+            aspnetBaseTag,
+            cat(ARCH_VERSIONED, "/buildpack-deps:", osVersionBase, "-curl")))
```
mthalman, line 17: There should be no reason you can't use the aspnet base image as the base for everything. So this can be simplified to:

```suggestion
    set installerImageTag to aspnetBaseTag
```

Or just replace all usages of `installerImageTag` with `aspnetBaseTag`.

eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -64,7 +64,7 @@ public class DockerfileShaUpdater : FileRegexUpdater
             {
                 { "powershell", new string[] { "https://pwshtool.blob.core.windows.net/tool/$VERSION_DIR/PowerShell.$OS.$ARCH.$VERSION_FILE.nupkg" } },
 
-                { "monitor", new string[] { $"$DOTNET_BASE_URL/diagnostics/monitor/$VERSION_DIR/dotnet-monitor.$VERSION_FILE.nupkg" } },
+                { "monitor", new string[] { $"$DOTNET_BASE_URL/diagnostics/monitor/$VERSION_DIR/dotnet-monitor-$VERSION_FILE$OPTIONAL_OS-{GetRuntimeSdkArchFormat()}.$ARCHIVE_EXT" } },
```
mthalman, line 67: You should be able to use `$OS` instead of `$OPTIONAL_OS`:

```suggestion
                { "monitor", new string[] { $"$DOTNET_BASE_URL/diagnostics/monitor/$VERSION_DIR/dotnet-monitor-$VERSION_FILE$OS-{GetRuntimeSdkArchFormat()}.$ARCHIVE_EXT" } },
```

`$OPTIONAL_OS` is only needed for file entries that may be for an RPM file.

eng/dockerfile-templates/monitor/Dockerfile.linux
jander-msft: > There should be no reason you can't use the aspnet base image as the base for everything. 

What does using the aspnet base image provide that these base images don't? Is it mostly a convenience because the dockerfile already depends on it?

mthalman: No functional difference. It just simplifies things, specifically the template.

jander-msft: Kindof... I think I would need to use the non-distroless ASP.NET image for the installer image when it's doing a multi-stage build of a distroless image.

jander-msft: I ended up removing '-distroless' and '-chiseled' from the OS_VERSION to determine the non-distroless OS and version and use that for the installer image. Let me know what you think.

mthalman: Oh, totally forgot that we need to account for distroless.

jander-msft: Sigh, `curl` doesn't exist on the Ubuntu Jammy ASP.NET images. I think I'm going to revert to my original solution (the same thing that runtime and aspnet are doing with the buildpack-deps installer image).

mthalman (approved):

mthalman (approved):

PR #4387, opened 2023-02-03
Add more flexible Dockerfile

richlander:
This Dockerfile has two basic modes:

For x64, by default:

```bash
docker build -t aspnetapp -f Dockerfile.debian .
```

For Arm64:

```bash
docker build -t aspnetapp -f Dockerfile.debian --build-arg ARCH=arm64v8 --build-arg RID=linux-arm64 .
```

This isn't super pretty, but it works. Also, this is showing the model. Clearly, you can pivot between x64 and Arm64 being the default.

One option is to make one symbol work for both ARCH and RID. That's part of what makes this super ugly. I think the easiest thing might be to push `x64` and `arm64` tags to align with our RIDs, as they are `amd64` and `arm64v8`, today.

### DIFF
--- a/samples/aspnetapp/Dockerfile.debian
+++ b/samples/aspnetapp/Dockerfile.debian
@@ -1,0 +1,20 @@
+# https://hub.docker.com/_/microsoft-dotnet
+ARG RID=linux-x64
+ARG ARCH=amd64
+FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
+ARG RID
+WORKDIR /source
+
+# copy csproj and restore as distinct layers
+COPY aspnetapp/*.csproj .
+RUN dotnet restore -r $RID
+
+# copy everything else and build app
+COPY aspnetapp/. .
+RUN dotnet publish -c Release -o /app -r $RID --self-contained false --no-restore
+
+# final stage/image
+FROM mcr.microsoft.com/dotnet/aspnet:7.0-bullseye-slim-$ARCH
+WORKDIR /app
+COPY --from=build /app .
+ENTRYPOINT ["./aspnetapp"]


mthalman:
I'd prefer a solution that made use of docker's `--platform` option rather than inventing new ARGs. I think it's better to try to find a solution that can fit with that option to better integrate with the ecosystem, especially buildx and building multi-arch images.

e.g. if somebody ran a build like this:
`docker build --platform linux/arm64 .`

But that still requires emulation if using a multi-arch tag for the sdk stage. Because if you had `FROM mcr.microsoft.com/dotnet/sdk:7.0`, you'd end up with an arm64 sdk image which may not match the arch of the machine it's being run on. This really demonstrates the need for emulation to work in the scenario of building a project.

An alternative is to explicitly set the architecture of the sdk tag:

```Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:7.0-jammy-amd64 AS build

ARG TARGETARCH

RUN arch=$TARGETARCH \
    && if [ "$arch" = "amd64" ]; then arch="x64"; fi \
    && echo "linux-$arch" > /tmp/rid

WORKDIR /source

COPY *.csproj .
RUN dotnet restore -r $(cat /tmp/rid)

COPY . .
RUN dotnet publish -c Release -f net7.0 -o /app -r $(cat /tmp/rid) --self-contained false
```

Of course, that limits the cross-platform nature of the Dockerfile wrt to the build machine: it can only be built on an x64 machine. But on the other hand, it can be used to produce images for multiple architectures.

richlander:
That brings us back to the base issue. We don't support the `--platform` tag, at least not in a way that results in using QEMU.

I still claim that my proposal, with the secondary proposal of moving to a single ARG` is the best option. The bash-foo is inventive, but pretty hostile to the UX. I think we either have to create aliases in our container images or the RID graph. I believe that the former is far easier.

mthalman:
> That brings us back to the base issue. We don't support the `--platform` tag, at least not in a way that results in using QEMU.

Yes, as long as you avoid executing `dotnet`, using `--platform` is fine. I used the pattern I proposed above at https://github.com/mthalman/dredge/blob/main/src/Valleysoft.Dredge/Dockerfile. That works fine for me because I don't use the Dockerfile for developer scenarios, meaning that I don't need to worry about supporting different architecture hosts; I only care about the architecture of the build machine. This lets me set the platforms of my target images idiomatically: https://github.com/mthalman/dredge/blob/b4e91e40384a186a843d09707f62ac4c91602a36/.github/workflows/release.yml#L150. Most importantly, this usage allows me to get multi-arch tags for free by using the `--platform` option. With your custom ARGs proposal, you're on your own for producing multi-arch tags.

richlander:
I think I'm conflating things. Is `--platform` only used for MA tags to select the right concrete tag? If so (and your example suggests that), then I misunderstood what was happening. I thought `--platform` was causing QEMU to kick in, but I can see that `--platform` is a step earlier in the causal chain.

I think you are saying:

- Use `--platform` + our MA tags to determine the final image being produced.
- Use an actual build-arg only if you want to pivot the SDK from one native tag to another.

I can change the sample after you validate my understanding.

That still leaves the question of how to make the Dockerfile prettier. My proposal still has better UX. Disagree? I'm not enthusiastic about getting people to adopt your (clever) pattern as general guidance.

richlander:
OK. I need to play with this more.

richlander:
I didn't realize these ENVs were set.

https://docs.docker.com/build/building/multi-platform/

```bash
$ cat Dockerfile
FROM debian

ARG TARGETPLATFORM
ARG TARGETARCH
ARG TARGETOS
ARG BUILDPLATFORM
RUN echo "I am running on $BUILDPLATFORM, building for $TARGETPLATFORM"
RUN echo "I am targeting ${TARGETOS} and ${TARGETARCH}"
$ docker build -t test .
[+] Building 0.1s (7/7) FINISHED
 => [internal] load build definition from Dockerfile                                                               0.0s
 => => transferring dockerfile: 38B                                                                                0.0s
 => [internal] load .dockerignore                                                                                  0.0s
 => => transferring context: 2B                                                                                    0.0s
 => [internal] load metadata for docker.io/library/debian:latest                                                   0.0s
 => [1/3] FROM docker.io/library/debian                                                                            0.0s
 => CACHED [2/3] RUN echo "I am running on linux/amd64, building for linux/amd64"                                  0.0s
 => CACHED [3/3] RUN echo "I am targeting linux and amd64"                                                         0.0s
 => exporting to image                                                                                             0.0s
 => => exporting layers                                                                                            0.0s
 => => writing image sha256:cb369cc34bb85653e611045554732592b28ebee906a2ee062d23f7b5e0a3285a                       0.0s
 => => naming to docker.io/library/test
```

richlander:
I am going to start this over again.

Issue #3968, opened 2022-08-05
Consider switching default port for ASP.NET Core to `8080` or `5000`


richlander:
Currently, the default port for our images is port `80`: https://github.com/dotnet/dotnet-docker/blob/main/src/runtime-deps/6.0/jammy/amd64/Dockerfile#L19

It isn't possible for non-root images to bind to port 80 or 443: https://github.com/containerd/containerd/issues/2516. We should consider using only non-root ports so that it is easy to use our assets for both root and non-root scenarios.

There are a few things to consider:

- Folks will need to update script and manifests that do port mapping. As a result, it's an upgrade-time breaking change.
- Folks can straightforwardly opt-out by [re-setting the ENV to what it is today](https://github.com/dotnet/dotnet-docker/blob/2aeeffadbf4f31c062ba20d46efb4782d62e9a43/src/runtime-deps/3.1/bullseye-slim/amd64/Dockerfile#L19) in their layer.
- This change would bifurcate our `runtime-deps` Dockerfiles.
- `8080` might not be the best port to choose. `5000` might be a much better idea since that's the default port for dev.

Is it worth it? I think so. The motivation is embracing non-root as a key scenario.

I'm thinking that this change is a bit late for .NET 7. It's probably best left until .NET 8 Preview 1.

Related: https://github.com/dotnet/designs/pull/271

richlander:
@bradygaster @davidfowl @LadyNaggaga @glennc

davidfowl:
How do other frameworks handle this? What do they recommended?

richlander:
I just played with node.js in docker. It doesn't have this problem.  I think it is because ASP.NET Core is being too fancy. AFAICT, it's security first posture is taking us down a bad path.

Today's behavior:

```bash
rich@kamloops:~$ docker run --rm -it -p 8000:80 mcr.microsoft.com/dotnet/samples:aspnetapp
      Now listening on: http://[::]:80
```

Don't set the [ASPNETCORE_URLS ENV](https://github.com/dotnet/dotnet-docker/blob/2aeeffadbf4f31c062ba20d46efb4782d62e9a43/src/runtime-deps/3.1/alpine3.15/amd64/Dockerfile#L16):

```bash
      Now listening on: http://localhost:5000
```

I cannot access that port from outside of the container because that's a private loopback port.

I can outsmart that. I'll set the ENV but to `5000`:

```bash
[...]

PR #4397, opened 2023-02-07
Add non-root user support

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/2249. Also see https://github.com/dotnet/designs/pull/271.

Adds non-root runtime-deps files so that any 8.0 image may be run with `-u app` or new images can be created with `USER app` at the end of the Dockerfile to run as non-root.

Other changes:
- switch the default port from 80 to 8080
- use the new environment variable ASPNETCORE_HTTP_PORTS instead of ASPNETCORE_URLS
- don't unset ASPNETCORE_URLS in sdk images. Aspnetcore sees the (empty) URLS variable and overrides the new HTTP_PORTS variable


richlander:
Other related issues:

- https://github.com/dotnet/aspnetcore/pull/44194
- https://github.com/dotnet/dotnet-docker/issues/3968

richlander:
> Configure web servers to bind to port 8080/8443 when present

It seems like it should only mention ports that are configured.

>  Aspnetcore sees the (empty) URLS variable and overrides the new HTTP_PORTS variable

That doesn't seems right. The diff suggests that the old variable is gone.

> Home directory

It looks like we set the home directory for regular images and not for distroless/chiseled. We should document our policy.

mthalman:
FYI: Looks like tests are failing for all versions other than 8.0.

src/runtime-deps/8.0/cbl-mariner2.0/amd64/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+        ca-certificates \
+        \
+        # .NET dependencies
+        glibc \
+        icu \
+        krb5 \
+        libgcc \
+        libstdc++ \
+        openssl-libs \
+        shadow-utils \
```
mthalman, line 13: We should avoid having `shadow-utils` be included in the final image. It's not a required package for the operation of a .NET app. Unfortunately, that adds a bit of complexity to things. I see three options:

1. A multi-stage Dockerfile where the first stage installs `shadow-utils` and creates the group/user. The final stage would then copy the passwd and group files from the first stage. This also requires manual creation of the HOME directory.
2. Directly write to the passwd and group files with `echo` commands to populate the necessary config for the user and group. This also requires manual creation of the HOME directory.
3. Keep the installation of `shadow-utils` (although I would prefer it to be placed next to `ca-certificates` to not give the impression that it's a dependency of .NET). But then uninstall the package after running the commands. It'd be good to manually verify the contents of the resulting image to see if there's any leftover files as a result of this that look suspect.

I prefer option 3. @MichaelSimons - what are your thoughts?

tests/Microsoft.DotNet.Docker.Tests/CommonRuntimeImageTests.cs
```diff
@@ -28,7 +28,15 @@ public static IEnumerable<object[]> GetImageData(DotNetImageType imageType)
...
-            variables.Add(new EnvironmentVariableInfo("ASPNETCORE_URLS", $"http://+:{imageData.DefaultPort}"));
+
+            if (imageData.VersionFamily.Major >= 8)
+            {
+                variables.Add(new EnvironmentVariableInfo("ASPNETCORE_HTTP_PORTS", imageData.DefaultPort.ToString()));
+            }
+            else
+            {
+                variables.Add(new EnvironmentVariableInfo("ASPNETCORE_URLS", $"http://+:{imageData.DefaultPort}"));
+            }
```
mthalman, line 39: This logic should be inverted. The condition should be `imageData.VersionFamily.Major <= 7` and the `else` should apply to `ASPNETCORE_HTTP_PORTS`. That will allow things to naturally get cleaned up as versions go EOL.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -66,34 +68,45 @@ public async Task Execute()
...
+                if (!_isWeb && DockerHelper.IsLinuxContainerModeEnabled && _imageData.IsDistroless &&
                     (!_imageData.OS.StartsWith(OS.Mariner) || _imageData.Version.Major > 6))
                 {
-                    await RunTestAppImage(fxDepTag, runAsAdmin: true);
+                    await RunTestAppImage(fxDepTag, user: _adminUser);
+
+                    if (_nonRootUserSupported)
+                    {
+                        await RunTestAppImage(fxDepTag, user: _nonRootUser);
+                    }
```
mthalman, line 92: I don't think this is needed. It's essentially a duplicate of `await RunTestAppImage(fxDepTag, user: fxDepUser);` from above. That runs with a user of `<null>` which will default to the non-root user for distroless. So there's not a need to test again that explicitly sets that as the user.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -66,34 +68,45 @@ public async Task Execute()
...
                     // Use `sdk` image to publish self contained app and run with `runtime-deps` image
                     string selfContainedTag = BuildTestAppImage("self_contained_app", solutionDir, customBuildArgs);
                     tags.Add(selfContainedTag);
-                    await RunTestAppImage(selfContainedTag, runAsAdmin: runAsAdmin);
+                    await RunTestAppImage(selfContainedTag, user: _adminUser);
+
+                    if (_nonRootUserSupported)
+                    {
+                        await RunTestAppImage(selfContainedTag, user: _nonRootUser);
+                    }
```
mthalman, line 105: I see the non-root user scenario is tested here for self-contained but not above for the framework-dependent scenario (it's only done for the distroless case, not the general case).

src/runtime-deps/8.0/cbl-mariner2.0/amd64/Dockerfile
MichaelSimons: #3 would require it all be done within a single stage though.  I still think that is preferrable over #1.

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
mthalman: This code block is for distroless which already does explicitly set `USER app`.

mthalman:
Because new runtime-deps Dockerfiles were created, the https://github.com/dotnet/dotnet-docker/blob/main/tests/performance/ImageSize.nightly.linux.json file needs to be updated to have new entries for these files. It should just be a matter of copying the entry from the other .NET version that it was originally sharing.

src/runtime-deps/8.0/cbl-mariner2.0/amd64/Dockerfile
mthalman: By stage, I think you mean layer. So what that means, @lbussell, is that the installation of the packages, creation of the user, and removal of the shadow-utils should all be done in one `RUN` instruction.

MichaelSimons: >By stage, I think you mean layer. 

Yes.

EklipZgit:
May want to take a look at 
https://github.com/dotnet/runtime/issues/81456
May want to consider having the non-root users home directory not default to the AppServices mapped /home because certificate chains cannot be written to the default home location in default linux appservices due to the issues outlined at the end of that issue.

src/runtime-deps/8.0/cbl-mariner2.0/amd64/Dockerfile
lbussell: I took a pass at this - there's still some whitespace weirdness to figure out. I didn't finish fixing the tests yet on Mariner either.

src/runtime-deps/8.0/alpine3.17/amd64/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+        ca-certificates \
+        \
+        # .NET dependencies
+        krb5-libs \
+        libgcc \
+        libintl \
+        libssl3 \
+        libstdc++ \
+        zlib
+RUN addgroup \
```
mthalman, line 13: Add a line break here

src/runtime-deps/8.0/cbl-mariner2.0/amd64/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM mcr.microsoft.com/cbl-mariner/base/core:2.0
+
+RUN tdnf install -y \
+        shadow-utils \
+        ca-certificates \
```
mthalman, line 5: Swap the order of these two so they're alphabetical. Cottle's templating does have a [sort function](https://cottle.readthedocs.io/en/stable/page/03-builtin.html#sort-map-callback).

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -56,7 +59,7 @@ public async Task Execute()
                     // Use `sdk` image to build and run test app
                     string buildTag = BuildTestAppImage("build", solutionDir, customBuildArgs);
                     tags.Add(buildTag);
-                    string dotnetRunArgs = _isWeb ? $" --urls http://0.0.0.0:{_imageData.DefaultPort}" : string.Empty;
+                    string dotnetRunArgs = _isWeb && _imageData.Version.Major < 8 ? $" --urls http://0.0.0.0:{_imageData.DefaultPort}" : string.Empty;
```
mthalman, line 62: I don't think this `--urls` arg is needed at all, for any version.

eng/dockerfile-templates/Dockerfile.linux.install-deps
```diff
@@ -1,7 +1,9 @@
 {{
     _ ARGS:
         isSdk (optional): Indicates whether the dependencies are installed for an sdk Dockerfile
-        distroless-staging-dir (optional): Location of the staging directory for distroless installation ^
+        distroless-staging-dir (optional): Location of the staging directory for distroless installation
+        continue-cmd (optional): Whether to append the line continuation character at the end of the commands
```
mthalman, line 5: Instead of passing this as an arg, can't the calling template just explicitly add the continuation character itself, when necessary, after inserting the template?

lbussell: Leaving the argument off breaks some older versions, for example .NET 6.0 on alpine3.16. This may be due to some of my testing infrastructure changes with the default port, I'm not sure, but I haven't touched those Dockerfiles.

mthalman: To clarify, not saying the argument should be removed completely without additional changes. The templates which reference this one should compensate and add the continuation character when necessary. This should be considered a refactor and produce identical Dockerfiles as before the change.

lbussell: Sorry, that reply was meant to be on a different comment. I have removed the continue-cmd argument and pulled the logic out into the outer Dockerfile(s).

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
lbussell: Leaving the argument off breaks some older versions, for example .NET 6.0 on alpine3.16. This may be due to some of my testing infrastructure changes with the default port, I'm not sure, but I haven't touched those Dockerfiles.

richlander:
FYI ... This is what Ubuntu 23.04 has.

```bash
$ docker run --rm ubuntu:23.04 cat /etc/passwd | grep ubuntu
ubuntu:x:1000:1000:Ubuntu:/home/ubuntu:/bin/bash
```

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -56,8 +59,8 @@ public async Task Execute()
                     // Use `sdk` image to build and run test app
                     string buildTag = BuildTestAppImage("build", solutionDir, customBuildArgs);
                     tags.Add(buildTag);
-                    string dotnetRunArgs = _isWeb ? $" --urls http://0.0.0.0:{_imageData.DefaultPort}" : string.Empty;
-                    await RunTestAppImage(buildTag, command: $"dotnet run{dotnetRunArgs}");
+                    string dotnetRunArgs = _isWeb && (_imageData.Version.Major == 6 || _imageData.Version.Major == 7) ? $" --urls http://0.0.0.0:{_imageData.DefaultPort}" : string.Empty;
```
mthalman, line 62:
```suggestion
                    string dotnetRunArgs = _isWeb && (_imageData.Version.Major <= 7) ? $" --urls http://0.0.0.0:{_imageData.DefaultPort}" : string.Empty;
```

eng/dockerfile-templates/monitor/Dockerfile.envs
```diff
@@ -2,8 +2,9 @@
     _ .NET major version matches the major version of dotnet-monitor ^
     set dotnetMajor to split(PRODUCT_VERSION, ".")[0]
 }}ENV \
-    # Unset ASPNETCORE_URLS from aspnet base image
-    ASPNETCORE_URLS= \
+    {{if dotnetMajor != "6" && dotnetMajor != "7":# Unset ASPNETCORE_HTTP_PORTS from aspnet base image
+    ASPNETCORE_HTTP_PORTS= \^else:# Unset ASPNETCORE_URLS from aspnet base image
```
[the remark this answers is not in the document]
jander-msft: > @jander-msft - I know you requested this. Can you explain the need for this?

.NET Monitor already runs its HTTP server at ports 52323 and 52325 by default. Either setting ASPNETCORE_HTTP_PORTS would override that behavior (which we don't want by default) or it is not observed (which would be bad to insinuate that it has some effect when it does not); I think the former would be the case if the environment variable is specified. I will very later today that this is the case.

> Also, should the monitor Dockerfile be configured to run as non-root by default?

That would be great if that could be added too. Although, if this change is only scoped to .NET 8+, then this work shouldn't be necessary because .NET Monitor is only offering distroless and chiseled images for .NET 8+, which should already be using the non-root user.

src/runtime-deps/8.0/alpine3.17/arm32v7/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+        krb5-libs \
+        libgcc \
+        libintl \
+        libssl3 \
+        libstdc++ \
+        zlib
+
+# Create a non-root user and group
+RUN addgroup \
+        --system \
```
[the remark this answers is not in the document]
MichaelSimons: A 3rd option would be to use the existing pattern and log a follow-up issue to correct the ordering across the board.

src/runtime-deps/8.0/cbl-mariner2.0-distroless/amd64/Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+    && tdnf clean all --releasever=2.0 --installroot /staging
+
+# Generate RPM manifest file by appending to the original manifest file from base distroless image
+COPY --from=mcr.microsoft.com/cbl-mariner/distroless/minimal:2.0 /var/lib/rpmmanifest/container-manifest-2 /tmp/rpmmanifest
+RUN tmpManifestPath="/tmp/rpmmanifest" \
+    && rpm --query --all --queryformat "%{NAME}\t%{VERSION}-%{RELEASE}\t%{INSTALLTIME}\t%{BUILDTIME}\t%{VENDOR}\t%{EPOCH}\t%{SIZE}\t%{ARCH}\t%{EPOCHNUM}\t%{SOURCERPM}\n" --root /staging | grep -v gpg-pubkey >> $tmpManifestPath \
+    && mkdir -p /staging/var/lib/rpmmanifest \
+    # Remove duplicates that match on the first field (package name)
+    && tac $tmpManifestPath | gawk '!x[$1]++' | sort > /staging/var/lib/rpmmanifest/container-manifest-2
+
```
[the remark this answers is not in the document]
lbussell: The extra line is present in old mariner 2.0 distroless dockerfiles as well, which I'm trying not to overwrite: https://github.com/dotnet/dotnet-docker/blob/b8063871d5eb200704aa84f88e53c1f22a813aeb/src/runtime-deps/6.0/cbl-mariner2.0-distroless/amd64/Dockerfile#L30-L31

Should we file an issue for this as well?

src/runtime-deps/8.0/cbl-mariner2.0/amd64/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+        # .NET dependencies
+        glibc \
+        icu \
+        krb5 \
+        libgcc \
+        libstdc++ \
+        openssl-libs \
+        zlib \
+    && tdnf clean all \
+    # Create a non-root user and group
```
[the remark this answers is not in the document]
lbussell: The shadow-utils issue is the reason it's all in one stage. Separating it into multiple stages anywhere in-between will increase the image sizeI fixed the duplicate cleaning logic, that was a bug. Do you think that we should install shadow-utils only for the new user creation logic, and uninstall immediately after? i.e.

```bash
RUN tdnf install -y \
      shadow-utils \
    && groupadd \
        --system \
        ...
    && tdnf remove -y \
        shadow-utils \
    && tdnf clean all \
```

src/runtime-deps/8.0/alpine3.17/arm32v7/Dockerfile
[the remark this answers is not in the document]
mthalman: I agree that this can be done as a follow-up issue to apply consistently.

src/runtime-deps/8.0/cbl-mariner2.0/amd64/Dockerfile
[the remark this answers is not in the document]
MichaelSimons: > Do you think that we should install shadow-utils only for the new user creation logic, and uninstall immediately after?

Yes, the instruction you included is the pattern I was expecting to see.

src/runtime-deps/8.0/cbl-mariner2.0-distroless/amd64/Dockerfile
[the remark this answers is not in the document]
MichaelSimons: > Should we file an issue for this as well?

Yes, please.

src/runtime-deps/8.0/cbl-mariner2.0/amd64/Dockerfile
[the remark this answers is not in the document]
lbussell: Got it, misunderstood the original discussion, I'll work on that change now.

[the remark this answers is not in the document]
mthalman: That seems rather inefficient to me. We'd then need to "update" and clean up packages twice. What benefit does that approach have?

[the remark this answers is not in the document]
MichaelSimons: Consistency of the patterns and layers across our Dockerfiles.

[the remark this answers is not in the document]
lbussell: @mthalman there is the `--cacheonly / -C` option if we want to avoid updating the cache from the previous layer.

MichaelSimons (approved):
LGTM

eng/dockerfile-templates/Dockerfile.linux.install-deps
```diff
@@ -1,7 +1,7 @@
 {{
     _ ARGS:
-        isSdk (optional): Indicates whether the dependencies are installed for an sdk Dockerfile
-        distroless-staging-dir (optional): Location of the staging directory for distroless installation ^
+        distroless-staging-dir (optional): Location of the staging directory for distroless installation
+        additionalPkgs (optional): Additional packages to install ^
```
mthalman, line 4: This arg is no longer used by callers to this template.

src/runtime-deps/8.0/alpine3.17/amd64/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+        \
+        # .NET dependencies
+        krb5-libs \
+        libgcc \
+        libintl \
+        libssl3 \
+        libstdc++ \
+        zlib
+
+# Create a non-root user and group
```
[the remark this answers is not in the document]
lbussell: Filed https://github.com/dotnet/dotnet-docker/issues/4419 to address this.

src/runtime-deps/8.0/cbl-mariner2.0-distroless/amd64/Dockerfile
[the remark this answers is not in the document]
lbussell: Filed https://github.com/dotnet/dotnet-docker/issues/4418.

jander-msft (approved):
.NET Monitor LGTM

eng/dockerfile-templates/Dockerfile.linux.install-deps
```diff
@@ -45,15 +45,24 @@
...
+            [
+                when(isDistrolessMariner, "prebuilt-ca-certificates", "ca-certificates"),
+                "",
+                dotnetDepsComment
+            ],
+            [
+                "ca-certificates",
+                "",
+                dotnetDepsComment
+            ]) ^
```
mthalman, line 58: This looks to be indented more than everything else.

lbussell: I reverted to the original version of this file that exists on nightly - but since it does look weird, I can address it here.

mthalman (approved):

mthalman:
Documented breaking changes at https://github.com/dotnet/docs/issues/35958, https://github.com/dotnet/docs/issues/35959

PR #4413, opened 2023-02-12
Add more instructions

richlander:
- For building platform-specific images (or not).
- Enabling (or disabling) globalization (and potentially reducing size)

Please review @tarekgh


samples/build-for-a-platform.md
```diff
@@ -0,0 +1,67 @@
...
+
+In addition, Docker [Buildkit exposes multiple environment variables](https://github.com/dotnet/dotnet-docker/pull/4387#issuecomment-1416565213) that can be used to further conditionalize behavior. These environment variables can be controlled with the pattern demonstrated in [Dockerfile](https://github.com/mthalman/dredge/blob/main/src/Valleysoft.Dredge/Dockerfile). As mentioned, .NET doesn't support being run in emulation. The pattern in that Dockerfile results in the SDK always being run natively while the final image is affected by the `--platform` switch. This model also has the best performance since the bulk of computation is run natively.
+
+Note: We are enabling the following model in a future release: https://github.com/dotnet/dotnet-docker/issues/4388#issuecomment-1421401384.
```
mthalman, line 59: Why not make these changes as part of this PR? I think it'd be more appropriate to reference the Dockerfile in this repo rather than the Dredge Dockerfile.

samples/build-for-a-platform.md
```diff
@@ -0,0 +1,67 @@
...
+
+## .NET and QEMU
+
+Docker desktop uses [QEMU](https://www.qemu.org/) for emulation, for example running x64 code on an Arm64 machine. .NET doesn't support being run in QEMU. That means that the SDK needs to always be run natively, to enable [multi-stage build](https://docs.docker.com/build/building/multi-stage/). Multi-stage build is used by all of our samples.
```
mthalman, line 63:
```suggestion
Docker Desktop uses [QEMU](https://www.qemu.org/) for emulation, for example running x64 code on an Arm64 machine. [.NET doesn't support being run in QEMU](https://github.com/dotnet/core/blob/main/release-notes/7.0/supported-os.md#qemu). That means that the SDK needs to always be run natively, to enable [multi-stage build](https://docs.docker.com/build/building/multi-stage/). Multi-stage build is used by all of our samples.
```

samples/enable-globalization.md
```diff
@@ -0,0 +1,88 @@
+# Enabling (or disabling) globalization functionality
+
+.NET includes various [globalization](https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-and-localization) capabilities, including support for processing natural language text, calendars, currency, and timezones. The .NET implementations for these capabilities comes primarily from operating system libraries, such as [International Components for Unicode (ICU)](https://icu.unicode.org/) and [tzdata](https://en.wikipedia.org/wiki/Tz_database). In some cases, this libraries and their accompanying databases are always available and in other cases it is desired that they be absent because they may be considered prohibitively large.
```
mthalman, line 3:
```suggestion
.NET includes various [globalization](https://learn.microsoft.com/dotnet/core/extensions/globalization-and-localization) capabilities, including support for processing natural language text, calendars, currency, and time zones. The .NET implementations for these capabilities comes primarily from operating system libraries, such as [International Components for Unicode (ICU)](https://icu.unicode.org/) and [tzdata](https://wikipedia.org/wiki/Tz_database). In some cases, these libraries and their accompanying databases are always available and in other cases it is desired that they be absent because they may be considered prohibitively large.
```

samples/enable-globalization.md
```diff
@@ -0,0 +1,88 @@
...
+- Alpine `aspnet`, `monitor`, `runtime`, `runtime-deps` images
+- Ubuntu chiseled images
+
+Images that do not include ICU enable [Globalization Invariant Mode](https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-invariant-mode.md), which provides more basic globalization behaviors in absence of using ICU.
+
+Some users want to add ICU to one of the image types that doesn't include it. It is counter-productive to remove ICU from an image that already includes it.
+
+### Alpine images
+
+ICU can be added to a .NET Alpine image by adding the following statements to the final stage within a `Dockerfile`, as demonstrated in [Dockerfile.alpine-icu](aspnetapp/Dockerfile.alpine-icu). This Dockerfile fragment adds and configures ICU and disables Globalization invariant mode.
```
mthalman, line 28:
```suggestion
ICU can be added to a .NET Alpine image by adding the following instructions to the final stage within a `Dockerfile`, as demonstrated in [Dockerfile.alpine-icu](aspnetapp/Dockerfile.alpine-icu). This Dockerfile fragment adds and configures ICU and disables Globalization invariant mode.
```

samples/enable-globalization.md
```diff
@@ -0,0 +1,88 @@
...
+    icu-libs
+```
+
+### Ubuntu chiseled images
+
+ICU can be added to an Ubuntu chiseled image, as demonstrated by https://github.com/ubuntu-rocks/dotnet/issues/21.
+
+## Tzdata
+
+Tzdata provides data about timezones. It is needed in applications that deal with time.
```
mthalman, line 46:
```suggestion
Tzdata provides data about time zones. It is needed in applications that deal with time.
```

samples/enable-globalization.md
```diff
@@ -0,0 +1,88 @@
...
+
+.NET container images that include tzdata:
+
+- Debian images
```
mthalman, line 50: Yes, but not as a matter of policy. If Debian decided to remove it, we wouldn't explicitly add it. Should we call that out?

samples/aspnetapp/README.md
```diff
@@ -190,9 +190,9 @@ docker build -t aspnetapp -f .\Dockerfile.windowsservercore-iis-x64 .
 docker run --rm -it -p:8080:80 aspnetapp
 ```
 
-## Build an image for ARM32 and ARM64
+## Build an image for Arm32 and Arm64
 
-By default, distro-specific .NET tags target x64, such as `6.0-alpine` or `6.0-focal`. You need to use an architecture-specific tag if you want to target ARM. Note that .NET is only supported on Alpine on ARM64 and x64, and not ARM32.
+By default, distro-specific .NET tags target x64, such as `6.0-alpine` or `6.0-jammy`. You need to use an architecture-specific tag if you want to target Arm. Note that .NET is only supported on Alpine on ARM64 and x64, and not Arm32.
```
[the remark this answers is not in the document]
richlander: That's shocking! Indeed.

samples/build-for-a-platform.md
richlander: I plan to switch that PR to use the new format and then block on merging it until it is ready.

samples/enable-globalization.md
richlander: Fixed.

samples/aspnetapp/Dockerfile.alpine-icu
```diff
@@ -0,0 +1,27 @@
+FROM mcr.microsoft.com/dotnet/sdk:7.0-alpine AS build
+
+ARG TARGETARCH
```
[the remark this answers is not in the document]
richlander: fixed.

samples/enable-globalization.md
```diff
@@ -0,0 +1,94 @@
...
+
+The best practice is to pass timezone information into a container via environment variable.
+
+```bash
+$ cat /etc/timezone
+America/Los_Angeles
+docker run --rm -it -e TZ=$(cat /etc/timezone) app
+```
+
+This approach enables a container image to be launched matching the host, as opposed to setting the information as part of `docker build`.
```
mthalman, line 94: > enables a container image to be launched matching the host

Is that a common practice? In the world of cloud computing, what time zone the host is in seems irrelevant. I did a [quick search](https://grep.app/search?current=6&q=docker%20run.%2ATZ%3D&regexp=true) and couldn't find any examples that use the host time zone. All are either hardcoding it or have it set to an environment variable that can be set dynamically. That approach seems better to me.

samples/aspnetapp/Dockerfile.alpine-icu
```diff
@@ -0,0 +1,27 @@
...
+
+# copy everything else and build app
+COPY aspnetapp/. .
+RUN dotnet publish --use-current-runtime -c Release --self-contained false --no-restore -o /app
+
+# To enable globalization support, see
+# https://github.com/dotnet/dotnet-docker/blob/main/samples/enable-globalization.md
+# final stage/image
+FROM mcr.microsoft.com/dotnet/aspnet:7.0-alpine
+ENV \
```
MichaelSimons, line 17:
```suggestion

ENV \
```

samples/aspnetapp/Dockerfile.alpine-icu
```diff
@@ -0,0 +1,27 @@
...
+
+# copy csproj and restore as distinct layers
+COPY aspnetapp/*.csproj .
+RUN dotnet restore --use-current-runtime
+
+# copy everything else and build app
+COPY aspnetapp/. .
+RUN dotnet publish --use-current-runtime -c Release --self-contained false --no-restore -o /app
+
+# To enable globalization support, see
```
MichaelSimons, line 13: The wording `To enable globalization` is a bit confusing because globalization is already enabled.  Maybe something like `To read more about globalization support, see...`

samples/build-for-a-platform.md
```diff
@@ -0,0 +1,67 @@
...
+
+Notes:
+
+- `amd64` is used for historical reasons and is synonymous with "x64", however, `x64` is not an accepted alias.
+- .NET tags are described in [.NET Container Tags -- Patterns and Policies](../documentation//supported-tags.md).
+- This document applies to building Linux containers, not Windows containers (which are x64 only). Some aspects apply to Windows containers, but they are not specifically addressed.
+
+## Dockerfiles that build everywhere
+
+The default scenario is using multi-arch tags, which will work in multiple environments.
```
MichaelSimons, line 17: Will the reader understand the `multi-arch` nomenclature here?  The [earlier link](https://docs.docker.com/build/building/multi-platform/) referred to these are multi-platform images.

samples/build-for-a-platform.md
```diff
@@ -0,0 +1,67 @@
...
+FROM mcr.microsoft.com/dotnet/aspnet:6.0-alpine
+```
+
+The [Dockerfile](aspnetapp/Dockerfile) example demonstrates this case, and can be built with the following pattern.
+
+```bash
+docker build -t app .
+```
+
+It can be built on any supported operating system and architecture. For example, if built on an Apple M1 machine, Docker will produce a `linux/arm64` image, while on a Windows x64 machine, it will produce a `linux/amd64` image.
```
MichaelSimons, line 32: `while on a Windows x64 machine, it will produce a `linux/amd64` image`

Not if your windows machine is configured for Windows containers :)

samples/build-for-a-platform.md
```diff
@@ -0,0 +1,67 @@
...
+
+Note: We are enabling the following model in a future release: https://github.com/dotnet/dotnet-docker/issues/4388#issuecomment-1421401384.
+
+## .NET and QEMU
+
+Docker Desktop uses [QEMU](https://www.qemu.org/) for emulation, for example running x64 code on an Arm64 machine. [.NET doesn't support being run in QEMU](https://github.com/dotnet/core/blob/main/release-notes/7.0/supported-os.md#qemu). That means that the SDK needs to always be run natively, to enable [multi-stage build](https://docs.docker.com/build/building/multi-stage/). Multi-stage build is used by all of our samples.
+
+As a result, we need a reliable pattern that enables produces multiple variants of images on one machine, but that doesn't use emulation.
```
MichaelSimons, line 65: Broken wording - `that enables produces multiple variants`

samples/aspnetapp/Dockerfile.alpine-icu
richlander: Good call.

samples/build-for-a-platform.md
richlander: Yup.

samples/build-for-a-platform.md
richlander: Yes. I was being lazy.

richlander:
I believe I have addressed all feedback.

samples/build-for-a-platform.md
```diff
@@ -0,0 +1,90 @@
...
+```bash
+docker build -t app .
+```
+
+It can be built on any supported operating system and architecture. For example, if built on an Apple M1 machine, Docker will produce a `linux/arm64` image.
+
+For Linux containers on an x64 machine:
+
+```bash
+$ docker inspect app | jq .[0].Architecture
```
MichaelSimons, line 37: The docker CLI will do this for you and the nice this is the command is platform agnostic.

``` docker inspect app -f "{{ .Architecture }}"```

richlander:
I believe I have now addressed the feedback.

samples/build-for-a-platform.md
richlander: Thanks!

I submit to being a `jq` addict.

MichaelSimons: How do you like translating that into PowerShell 🥲

MichaelSimons (approved):

PR #4414, opened 2023-02-12
Add globalization sample

richlander:
Globalization invariant mode is a major part of our size optimization strategy, yet, we don't have much guidance on how to use it correctly. However, we do receive feedback about these topics. Also, we don't have great policies of when to install `icu` and `tzdata`. This sample will hopefully help inform that.

Related to https://github.com/dotnet/dotnet-docker/pull/4413


tarekgh:
Here is good example for the collation:

```C#
    CompareInfo turkishCompare = new CultureInfo("tr-TR").CompareInfo;
    Console.WriteLine(turkishCompare.Compare("i", "\u0130", CompareOptions.IgnoreCase)); // print 0 as "i" is equal to Turkish I "\u0130".
```

richlander:
I like the Turkish I example. While it's not exactly obscure, it is also a bit of a novelty. I wonder if there are other cases we should cover. For example, if you were building an e-commerce app in Westerrn Europe or US, can you get away with globalization invariant mode? If not, what's an obvious example for why not?

Clockwork-Muse:
For timezone specifically, we've had multiple issues in the main repository with people [volume mounting `/etc/localtime`](https://github.com/dotnet/runtime/issues/62545), and I feel like the mention of using `/etc/timezone` is going to get them to host-mount that (as opposed to using the `TZ` environment variable, which is what has been recommended previously). 
That's for the **extremely rare** cases you actually want to run a server/container in a locale/zone in the first place, since in almost all cases the relevant locale/zone should be picked up from some referenced entity.

If we're dealing with just timezone data, there's all sorts of fun timezones with relatively recent "fun" adjustments, or quite unusual offsets that are lesser known in the western world.

richlander:
I didn't know about the pattern to mount the directory. I can include that.

Yes, I am familiar with local time being an anti-pattern. The current text makes it sound like we're encouraging it, which isn't the intent. I can update that. People appear to be doing this (from my research), which led me to want to include it.

When the timezone file is mounted, what does it typically content? Is it the UTC value?

Thanks for the feedback. Appreciated.

Clockwork-Muse:
Uh, no.
The recommendation, at least previously, has been to just add [the `TZ` environment variable](https://github.com/dotnet/runtime/issues/62545#issuecomment-1115349195), not mount any host files:

```csharp
docker run --rm -e TZ=$(cat /etc/timezone) mcr.microsoft.com/dotnet/aspnet:6.0-focal date
```
We especially don't want people mounting `/etc/localtime`, since that doesn't work.  From a "best practices" standpoint, we likely also want to not encourage baking configuration into containers.

tarekgh:
@richlander please don't use the mounting options. The issue https://github.com/dotnet/runtime/issues/60469 has detailed discussions about that. In short, Most Linux systems do file system symbolic/hard linking to the zone files. If you are mounting any TZ file to the container, it will interfere with that and will start producing inconsistent results for the time conversions and reporting zone names. As @Clockwork-Muse mentioned, please use `TZ` environment variable to set the default time zone on the container. This is much cleaner and shouldn't cause any problem.

tarekgh:
@richlander 

> I like the Turkish I example. While it's not exactly obscure, it is also a bit of a novelty. I wonder if there are other cases we should cover. For example, if you were building an e-commerce app in Westerrn Europe or US, can you get away with globalization invariant mode? If not, what's an obvious example for why not?

Here is a good example:

```C#
            // Å LATIN CAPITAL LETTER A WITH RING ABOVE is equal to 'A'  + '\u030A'  

            CompareInfo englishCompare = new CultureInfo("en-US").CompareInfo;
            Console.WriteLine(englishCompare.Compare("\u00C5", "A\u030A"));
```

richlander:
@Clockwork-Muse Thanks for that it. It makes sense. In practice, what value would you expect in `/etc/timezone` from the host?

Clockwork-Muse:
`TZ` should be set to one of the normal tzdb/IANA timezone identifiers, whatever you happen to pick (the example command previously was for somebody who wanted to use the host timezone, but for testing purposes you could/should pass anything).

In my case it's `America/Los_Angeles`.  [Wikipedia has a list](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), which you could also use to find some unusual zones.

richlander:
ChatGPT knows all.

image

Clockwork-Muse:
... using NTP to interface with the **host's** time seems a bit much (you'd have to set it at a server), and NTP doesn't handle timezones anyways.  ChatGPT is wong.

If it was better, it would try to warn you away from setting timezone for the container in the first place.

richlander:
Just saw your comment now, but I removed that extra info. Good now?

I very much appreciate the help!

samples/globalapp/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+// Hello world, indeed
+WriteLine("Hello, World!");
+
+DateTime nowUtc = DateTime.UtcNow;
+DateTime now = DateTime.Now;
+
+// Accessing UTC and Local time will always work
+PrintHeader("Print baseline timezones");
+WriteLine($"{nameof(TimeZoneInfo.Utc)}: {TimeZoneInfo.Utc}; {nowUtc}");
+// Local will match UTC unless `/etc/timezone` is set to another timezone
```
Clockwork-Muse, line 13: This should also be a reference to the `TZ` environment variable.

samples/globalapp/README.md
```diff
@@ -0,0 +1,89 @@
+# Globalization Test App
+
+This [app](Program.cs) is intended for testing .NET APIs that use globalization APIs and concepts. The behavior of these APIs varies depending on how a container (or other environment) is configured. The [Dockerfile](Dockerfile) is intended to aid testing and isn't itself an example of a production Dockerfile.
+
+The behavior of these APIs is affected by:
+
+- Presence of `icu` and `tzdata` packages.
+- The value of [globalization invariant mode](https://aka.ms/GlobalizationInvariantMode).
+- The value (or absence) of `/etc/timezone`
```
Clockwork-Muse, line 9: This also needs to be a reference to the `TZ` environment variable.

samples/globalapp/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+The recommended way to configure tzdata and timezones in containers is to set the container timezone to UTC and pass the host system's timezone information to the container at runtime using environment variables. This can be achieved by setting the environment variable TZ to [the timezone you want to use](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), for example, `TZ=America/Los_Angeles`. This will ensure that the container uses the correct timezone information without requiring additional setup or configuration.
```
Clockwork-Muse, line 11: In almost all cases the container timezone will be in UTC, because anything else is an extra layer to move it away from UTC.

richlander:
You might want to take a look at this other PR, too @Clockwork-Muse . It is dealing with some of the same concepts.

https://github.com/dotnet/dotnet-docker/pull/4413

samples/globalapp/README.md
```diff
@@ -0,0 +1,102 @@
...
+The first approach passes the `Etc/UTC` timezone directly, while the second approach passes the host timezone to the container.
+
+A machine configured to UTC will produce the following:
+
+```bash
+# cat /etc/timezone
+Etc/UTC
+```
+
+The app produces the following output, for "America/Los_Angeles" timezone:
```
richlander, line 28:
```suggestion
The app produces the following output, for the "America/Los_Angeles" timezone:
```

richlander:
I believe I have addressed all the feedback.

richlander:
Fixed merge conflict with other PR.

Issue #4388, opened 2023-02-06
Enable using docker build `--platform` switch (easily)


richlander:
There is no way to write a Dockerfile for .NET that is (A) succinct, (B) easy to build from the command line, (C) can equally produce Arm64 or x64 images, (D) that will run equally well on both Arm64 and x64 machines, and (E) avoids running the SDK in an emulator (since .NET doesn't support running in QEMU).

The following Dockerfile (which doesn't currently work) would satisfy all four of these requirements.

```Dockerfile
ARG BUILDARCH
FROM mcr.microsoft.com/dotnet/sdk:7.0-bullseye-slim-$BUILDARCH AS build
ARG TARGETARCH
WORKDIR /source

# copy csproj and restore as distinct layers
COPY aspnetapp/*.csproj .
RUN dotnet restore -r linux-$TARGETARCH

# copy everything else and build app
COPY aspnetapp/. .
RUN dotnet publish -c Release -r linux-$TARGETARCH --self-contained false --no-restore -o /app

# final stage/image
FROM mcr.microsoft.com/dotnet/aspnet:7.0-bullseye-slim
WORKDIR /app
COPY --from=build /app .
ENTRYPOINT ["./aspnetapp"]
```

Note: The two ENVs are set via Docker, not my invention. Context: https://github.com/dotnet/dotnet-docker/pull/4387#issuecomment-1416565213. The reason this `Dockerfile` doesn't work is because `-r` expects `x64` not `amd64` (which `$TARGETARCH` returns) and our container tags expect `arm64v8` not `arm64` (which `$BUILDARCH` returns). Funny enough, `$TARGETARCH` returns `arm64` for Arm64, which `-r` likes fine and `$BUILDARCH` returns `amd64` for x64, which our container tags like fine.

Note: Close readers may wonder why `$TARGETARCH` is not needed for the last `FROM` statement. That's because `--platform` is picking the correct image from the multi-arch tag. We could use $`TARGETARCH` if we wanted, but it is unnecessary since the underlying mechanics will do the right thing. How would one know that the tag is a multi-arch tag just by looking at it? It's because it doesn't include an architecture.

It would enable the following scenarios:

- `docker build` -- will produce a native-arch result on both Arm64 and x64 and use the native arch SDK.
- `docker build --platform linux/arm64` -- Will produce an Arm64 container image on both Arm64 and x64 machines but run the  SDK as native-arch, avoiding emulation (on x64 machines).
[...]

Issue #3392, opened 2022-01-10
Disable UseAppHost in sample FDD Dockerfiles


mthalman:
The sample FDD Dockerfiles generated all use the following line:

```Dockerfile
RUN dotnet publish -c release -o /app --no-restore
```

This ends up outputting both an app executable and a library file in the output directory. See the [publish documentation](https://docs.microsoft.com/dotnet/core/deploying/#publish-framework-dependent):

> Publishing an app as framework-dependent produces a cross-platform binary as a dll file, and a platform-specific executable that targets your current platform.

Either file can be used to execute the app. The executable can be executed directly while the DLL requires the use of the .NET CLI (`dotnet aspnetapp.dll`). From an image size perspective, it seems wasteful to include both files in the output. Only one of them is actually needed in order to execute the app. The Dockerfile is configured to only make use of the DLL:

```Dockerfile
ENTRYPOINT ["dotnet", "aspnetapp.dll"]
```

So it seems reasonable to simply not include the executable in the output.

By configuring the `UseAppHost` property, the publish logic can be configured to not include the executable in the output.

I propose that the sample FDD Dockerfiles disable this property by default:

```Dockerfile
RUN dotnet publish -c release -o /app --no-restore /p:UseAppHost=false
```

richlander:
This approach seems like overkill. I agree that the apphost is unnecessary for this scenarios, however, I think we should continue using it for a few reasons.

- The additional CLI complexity (as guidance) isn't worth the size win (~200k).
- Native platforms start with a single exe, not a launcher.
- This launcher pattern breaks when you switch to single file or native aot.

mthalman:
[Triage]
The samples have been updated to make use of the app host which reflects the points that were made in https://github.com/dotnet/dotnet-docker/issues/3392#issuecomment-1491278642. Based on those reasons, we've determined that the disabling of UseAppHost should not be done.

PR #4441, opened 2023-02-22
Update samples for platform targeting

richlander:
- Update samples to use [new pattern](https://github.com/dotnet/dotnet-docker/issues/4388#issuecomment-1421401384).
- Update guidance relating to platform targeting.
- Update other docs at the same time.
- Dependent on https://github.com/dotnet/sdk/pull/30762


mthalman:
> * Dependent on [fix issue 30369 enable docker build platform switch sdk#30762](https://github.com/dotnet/sdk/pull/30762)

Is this going to be backported to 7.0?

samples/container-best-practices.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Basic pattern
+
+The [dotnetapp](dotnetapp/README.md) samples has all assets in one directory. It uses the following pattern.
+
+```bash
+docker build --pull -t dotnetapp .
+```
+
+The context is the current directory.
```
mthalman, line 17:
```suggestion
The build context is the current directory.
```

samples/container-best-practices.md
```diff
@@ -0,0 +1,185 @@
...
+
+Multiple Dockerfile variations are provided. They can be specified with the `-f` argument, as follows.
+
+```bash
+docker build --pull -t dotnetapp -f Dockerfile.alpine .
+```
+
+### Multiple projects
+
+The [complexapp](complexapp/README.md) samples is more complex, being composed of multiple projects
```
mthalman, line 27:
```suggestion
The [complexapp](complexapp/README.md) samples is more complex, being composed of multiple projects.
```

samples/container-best-practices.md
```diff
@@ -0,0 +1,185 @@
...
+## Setting the Docker context
+
+The [build context](https://docs.docker.com/build/building/context/) is a key aspect to using `docker build`. In the most basic case, all assets are in a single directory and the context can match. In other cases, assets are spread out more. Getting the context right and the location of the Dockerfile in relation is perhaps the most important part of building container images.
+
+### Basic pattern
+
+The [dotnetapp](dotnetapp/README.md) samples has all assets in one directory. It uses the following pattern.
+
+```bash
+docker build --pull -t dotnetapp .
```
mthalman, line 14: Since the whole point of this is to explain how build context work, I think it's useful to indicate what the current directory is when executing the commands in this section. That's a key piece of info to help with understanding. I think it's important to show it represented with the command rather than only in prose-form.

samples/container-best-practices.md
```diff
@@ -0,0 +1,185 @@
...
+
+```bash
+docker build --pull -t app -f src/app/Dockerfile .
+```
+
+This pattern makes sense if the whole repo is required for the build.
+
+## Annotated Dockerfile -- Basic scenarios
+
+The following [Dockerfile](dotnetapp/Dockerfile) can be thought of as the "basic model" for building .NET apps. It satisfies a lot of needs and the other more advanced Dockerfiles offer relatively small variations on the same model. The annotations describe the beav
```
mthalman, line 74: The last sentence here is incomplete. Not sure where you were intending to go with that.

samples/container-best-practices.md
```diff
@@ -0,0 +1,185 @@
...
+```dockerfile
+# Establishes the `build` stage using an SDK image.
+# This multi-platform tag will provide a compatible image
+# on Linux or Windows, for Arm32, Arm64, and x64 architectures.
+FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
+WORKDIR /source
+
+# Copies msbuild assets into the image.
+# Other assets like `nuget.config` or `Directory.build.props` (if they exist) would need to be copied.
+# Copying a mininmal set of assets and the restoring packages is good for caching.
```
mthalman, line 85:
```suggestion
# Copying a mininmal set of assets and restoring packages is good for caching.
```

samples/container-best-practices.md
```diff
@@ -0,0 +1,185 @@
...
+# If a .cs file is changed, Docker will realize that and jump past the restore line.
```
mthalman, line 86:
```suggestion
# If only a .cs file is changed, Docker will realize that and jump past the restore line.
```

samples/container-best-practices.md
```diff
@@ -0,0 +1,185 @@
...
+# Sets the entrypoint for the app.
+# This launch pattern was chosen since it works the same in both Linux and Windows containers.
+ENTRYPOINT ["dotnet", "dotnetapp.dll"]
+```
+
+This Dockerfile above works well when the machines you build and host on are similar (both x64, for example). If your build and production machine are different (for example, Arm64 and x64), then you need to adopt a different pattern.
+
+## Annotated Dockerfile -- Advanced platform targeting
+
+The following [Dockerfile](aspnetapp/Dockerfile.alpine) offers more control for targeting, with the `--platform` switch. It is very similar to the prior sample, with a couple variations to change the behavior. In particular, we want the .NET SDK to always run natively (best performance and avoids running in QEMU, which .NET doesn't support).
```
mthalman, line 121: Provide a link to build-for-a-platform.md?

samples/container-best-practices.md
```diff
@@ -0,0 +1,185 @@
...
+# We are setting `-a` in the same way as above.
+RUN dotnet publish -a $TARGETARCH --self-contained false --no-restore -o /app
+
+FROM mcr.microsoft.com/dotnet/aspnet:7.0-alpine
+# This healthcheck directive queries the `healthz` endpoint exposed by ASP.NET Core middlware.
+HEALTHCHECK CMD wget -qO- -t1 http://localhost:80/healthz || exit 1
+WORKDIR /app
+COPY --from=build /app .
+# This launch pattern launches an executable (since this Dockerfile is Linux-specific).
+ENTRYPOINT ["./aspnetapp"]
```
mthalman, line 154: I think we should continue to use the DLL. See https://github.com/dotnet/dotnet-docker/issues/3392

Issue #1623, opened 2020-01-24
Consider changing samples to use an ephemeral port


richlander:
Per https://github.com/dotnet/dotnet-docker/pull/1622 and @glennc, consider using port 5050 instead of 8000 for samples. The 8xxx range is likely in more common use by production workloads. The samples are aimed at development scenarios, not production.

Anyone have a better suggestion than `5050`?

Example of where ports are defined:

* https://github.com/dotnet/dotnet-docker/blob/master/samples/aspnetapp/aspnetapp/Properties/launchSettings.json#L29
* https://github.com/dotnet/dotnet-docker/blob/master/samples/aspnetapp/README.md
* https://github.com/dotnet/dotnet-docker/blob/master/samples/run-in-sdk-container.md#aspnet-core-app

PR #4442, opened 2023-02-22
Add non-root Dockerfile

richlander:
This one is .NET 8.

Like the file name?

We need a Dockerfile like this for the blog post.

Fixes https://github.com/dotnet/dotnet-docker/issues/1623

### DIFF
--- a/samples/aspnetapp/Dockerfile.alpine-non-root
+++ b/samples/aspnetapp/Dockerfile.alpine-non-root
@@ -14,6 +14,7 @@ RUN dotnet publish --use-current-runtime -c Release --self-contained false --no-
 # https://github.com/dotnet/dotnet-docker/blob/main/samples/enable-globalization.md
 # final stage/image
 FROM mcr.microsoft.com/dotnet/aspnet:8.0-preview-alpine
+# These ENVs are only needed because the sample uses .NET 7
 ENV DOTNET_ROLL_FORWARD=Major
 ENV DOTNET_ROLL_FORWARD_PRE_RELEASE=1
 WORKDIR /app
--- a/samples/aspnetapp/Dockerfile.alpine-non-root
+++ b/samples/aspnetapp/Dockerfile.alpine-non-root
@@ -10,15 +10,19 @@ RUN dotnet restore --use-current-runtime
 COPY aspnetapp/. .
 RUN dotnet publish --use-current-runtime -c Release --self-contained false --no-restore -o /app
 
+
 # To enable globalization:
 # https://github.com/dotnet/dotnet-docker/blob/main/samples/enable-globalization.md
 # final stage/image
 # .NET 8 tag includes `-preview`
 FROM mcr.microsoft.com/dotnet/aspnet:8.0-preview-alpine
+
 # These ENVs are only needed because the sample uses .NET 7
 ENV DOTNET_ROLL_FORWARD=Major
 ENV DOTNET_ROLL_FORWARD_PRE_RELEASE=1
+
 WORKDIR /app
 COPY --from=build /app .
+
 USER app
 ENTRYPOINT ["./aspnetapp"]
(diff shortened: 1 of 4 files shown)


samples/aspnetapp/Dockerfile.nanoserver-containeruser
```diff
@@ -0,0 +1,20 @@
+# https://hub.docker.com/_/microsoft-dotnet
```
mthalman, line 1: This entire Dockerfile seems unnecessary since Nano Server runs as `ContainerUser` by default. So it's really just a no-op. It's Windows Server Core that runs as `ContainerAdministrator` by default.

MichaelSimons:
Are there going to be any updates to the sample markdown files to reference these?

samples/aspnetapp/Dockerfile.nanoserver-containeruser
richlander: K. I'll switch it to Windows Server Core. Good call.

samples/aspnetapp/Dockerfile.windowsservercore-containeruser
```diff
@@ -13,7 +13,11 @@ RUN dotnet publish -c Release -o /app -r win-x64 --self-contained false --no-res
 # final stage/image
 # Relies on 7.0 multi-arch tag to pick the same Windows version as the host. 
 # Alternatively, a release specific tag can be used, like: `7.0-nanoserver-ltsc2022`
-FROM mcr.microsoft.com/dotnet/aspnet:7.0
+# .NET 8 tag includes `-preview`
+FROM mcr.microsoft.com/dotnet/aspnet:8.0-preview-windowsservercore-ltsc2022
```
mthalman, line 15: I'm curious whether this made you think we should have a `8.0-preview-windowsservercore` tag.

samples/aspnetapp/Dockerfile.windowsservercore-containeruser
```diff
@@ -13,7 +13,11 @@ RUN dotnet publish -c Release -o /app -r win-x64 --self-contained false --no-res
 # final stage/image
 # Relies on 7.0 multi-arch tag to pick the same Windows version as the host. 
 # Alternatively, a release specific tag can be used, like: `7.0-nanoserver-ltsc2022`
-FROM mcr.microsoft.com/dotnet/aspnet:7.0
+# .NET 8 tag includes `-preview`
```
mthalman, line 14: So technically this doesn't need to be 8.0. Was this done just for consistency of versions when comparing with Linux?

samples/aspnetapp/Dockerfile.windowsservercore-containeruser
richlander: Good question, but it did not. I'm also recently not keen on multi-platform tags for Windows, as you've seen.

samples/aspnetapp/Dockerfile.windowsservercore-containeruser
richlander: That's an excellent point. I'll change that. I like it.

mthalman (approved):

samples/aspnetapp/Dockerfile.alpine-non-root
```diff
@@ -0,0 +1,24 @@
...
+
+# copy csproj and restore as distinct layers
+COPY aspnetapp/*.csproj .
+RUN dotnet restore --use-current-runtime
+
+# copy everything else and build app
+COPY aspnetapp/. .
+RUN dotnet publish --use-current-runtime -c Release --self-contained false --no-restore -o /app
+
+# To enable globalization:
```
MichaelSimons, line 13: IMO this final stage could use some whitespace to improve the readability.
```
# final stage/image
# .NET 8 tag includes `-preview`
# to enable globalization see https://github.com/dotnet/dotnet-docker/blob/main/samples/enable-globalization.md
FROM mcr.microsoft.com/dotnet/aspnet:8.0-preview-alpine

# These ENVs are only needed because the sample uses .NET 7
ENV DOTNET_ROLL_FORWARD=Major
ENV DOTNET_ROLL_FORWARD_PRE_RELEASE=1

WORKDIR /app
COPY --from=build /app .

USER app
ENTRYPOINT ["./aspnetapp"]
```

MichaelSimons: I also like to include two blank lines between stages.

richlander: Good call. Fixed.

Issue #4368, opened 2023-01-30
Build script and test script take different format for "Version" parameter


lbussell:
The issue here is that when you provide a Version argument to `build-and-test.ps1`, it will either build or test the images but not both.

`eng/common/build.ps1` passes the `Version` argument directly to ImageBuilder, which expects a format of "8.0.*": https://github.com/dotnet/dotnet-docker/blob/cf9aef9e150e1383b4bf474cd3c9a427a44704de/eng/common/build.ps1#L53

`tests/run-tests.ps1` uses the `Version` argument directly in a path that filters what images to test. The paths only include the major and minor version, e.g. "8.0":
https://github.com/dotnet/dotnet-docker/blob/cf9aef9e150e1383b4bf474cd3c9a427a44704de/tests/run-tests.ps1#L53-L57

We should probably document that we expect a version format of "X.Y.Z", and then strip the last version segment off of that string before we use it to calculate the path in `run-tests.ps1`.

lbussell:
This is effectively fixed by https://github.com/dotnet/dotnet-docker/pull/4445.

PR #4445, opened 2023-02-22
Check and fix version formatting in build-and-test script

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/4368

### DIFF
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -24,6 +24,16 @@ param(
     [string[]]$TestCategories = @("runtime", "runtime-deps", "aspnet", "sdk", "pre-build", "sample", "image-size", "monitor")
 )
 
+if ($Version -notmatch '^\d+\.\d+(\.[\d*])?$') {
+    Write-Error "Error: Input version is not in the expected format of X.Y or X.Y.*"
+    exit 1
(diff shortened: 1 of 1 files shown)


build-and-test.ps1
```diff
@@ -24,6 +24,16 @@ param(
     [string[]]$TestCategories = @("runtime", "runtime-deps", "aspnet", "sdk", "pre-build", "sample", "image-size", "monitor")
 )
 
+if ($Version -notmatch '^\d+\.\d+(\.[\d*])?$') {
+    Write-Error "Error: Input version is not in the expected format of X.Y or X.Y.*"
+    exit 1
+}
+
+# If we call the script with a format like "8.0", add the trailing ".*" that ImageBuilder expects
+if ($Version -notmatch '\.\*') {
```
mthalman, line 33: Wouldn't this cause `.*` to be appended when the version already specifies a three-part version like `7.0.3`? That would create a version value of `7.0.3.*`. Since we've already validated the correctness of the syntax above, for this check we could simply check for number of `.` characters. If there's only 1, we tack on the `.*`.

mthalman (approved):

Issue #4423, opened 2023-02-15
Image for SDK 7.0.200 missing


nesc58:
Hello,
the download for 7.0.200 is available on https://dotnet.microsoft.com/en-us/download/dotnet/7.0 but no docker image exists for 7.0.200. The docker image only exists for 7.0.103. The february release (https://github.com/dotnet/core/issues/8191) should add images for 7.0.200.

Could you please add it?

Kind regards

dave-yotta:
If this is added, everyone using `7.0` will get the breaking changes from `7.0.200`, which has caused a bit of a stir:
https://github.com/dotnet/sdk/pull/29065
https://github.com/dotnet/sdk/issues/30624

mthalman:
Please see our recent announcement about the 7.0.2xx feature band: https://github.com/dotnet/dotnet-docker/discussions/4434. Our plan is to publish this in the March release.

richlander:
You can close it. We are acting on this report.

mthalman:
The 7.0.201 SDK was released early and so new SDK images have been published for this version. See https://github.com/dotnet/dotnet-docker/discussions/4460.

PR #4452, opened 2023-02-23
Update to .NET SDK 7.0.201

lbussell:
https://github.com/dotnet/release/issues/479

Also see https://github.com/dotnet/dotnet-docker/issues/4423 and https://github.com/dotnet/dotnet-docker/discussions/4434.


mthalman:
This likely needs to be refreshed. There was a new build produced since these SHAs were last collected.

mthalman (approved):

Issue #4438, opened 2023-02-21
Please build the chiseled ASP.NET Core Runtime for Linux arm32


ali50m:
I would like to build docker images for both linux/amd64 and linux/arm/v7(Raspbian). But I cannot find the chiseled
tag for arm32 on https://hub.docker.com/_/microsoft-dotnet-nightly-aspnet/.

Please consider adding one for the Raspbian users.

#4436

mthalman:
[Triage]
At the time of implementation of the chiseled Dockerfiles, there was a technical limitation with the infrastructure that made it not straightforward to support Arm32 Dockerfiles. This is captured in https://github.com/dotnet/docker-tools/issues/1017. But I think that issue is now resolved as a result of updates to the build machines. That can be verified easily by simply producing the Dockerfiles and attempting to build them through the build pipeline. If it works, then the issue can be resolved.

But we support having Arm32 Dockerfiles as it would fill in the gaps that currently exist in the architecture matrix. We want to be consistent in the images that are provided.

mthalman:
The Docker repositories under the `dotnet/nightly` name is meant for testing/evaluation purposes and not supported for production use. Currently, chiseled images for .NET 6 and 7 are only available in the `dotnet/nightly` repositories. However, we did just publish chiseled images as part of the [.NET 8 Preview 1 release](https://github.com/dotnet/dotnet-docker/discussions/4440). Since that is a preview release, that is also only intended for testing purposes. The new Arm32 images would fall into this same bucket of non-production support. You can follow the status of the availability of chiseled images for production use at https://github.com/dotnet/dotnet-docker/discussions/4137.

lbussell:
This is completed with https://github.com/dotnet/dotnet-docker/pull/4453. The images are still only in dotnet/nightly as mentioned above.

PR #4453, opened 2023-02-23
Add arm32 version of Ubuntu chiseled images

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/4438


MichaelSimons (commented):
I was expecting to see changes to the [TestData](https://github.com/dotnet/dotnet-docker/blob/main/tests/Microsoft.DotNet.Docker.Tests/TestData.cs#L66) so that the new images get tested.

MichaelSimons (approved):

mthalman (approved):

mthalman (approved):
Readmes need to be regenerated

Issue #4451, opened 2023-02-23
New `app` user for .NET 8 images isn't documented as a potentially breaking change


rwkarg:
### Describe the Bug

The new `app` user for .NET 8 images (https://github.com/dotnet/dotnet-docker/pull/4397) is an undocumented breaking change for users that are currently creating a user named `app` on top of .NET 7 and earlier images.

[Documented breaking changes](https://learn.microsoft.com/en-us/dotnet/core/compatibility/8.0)

### Steps to Reproduce

Have a Dockerfile that builds off of one of the Microsoft images (ex. `dotnet/aspnet`) in use with .NET 7 or earlier tags

```Dockerfile
ARG TAG_VERSION
FROM mcr.microsoft.com/dotnet/aspnet:${TAG_VERSION}

# Minimal steps to produce error
RUN groupadd -r app && \
    useradd -m -u 10005 -r -g app -d /home/app -s /sbin/nologin -c "Docker image user" app
USER app
```

Now Set TAG_VERSION to a .NET 8 tag (ex. `8.0.0-preview.1`) and `docker build`

--> Container image build error since `app` user already exists.

### Other Information

Preferred Outcome: Use a higher number user id (>10,000) and document how this addition may break existing workflows when upgrading to .NET 8 where an `app` user is currently being created on top of .NET 7 or earlier tags.

We currently build images off the base Microsoft images to add a non-root user (as well as other unrelated specializations not relevant to this issue) and just happen to currently use `app` as the user name.

We could address this by omitting the RUN step to add a user and use the provided `app` user, but that still runs in to an issue with our deployments to Kubernetes. The `securityContext` for the pods/containers specify a `runAsUser` that requires the user id.

The .NET 8 images create the user with id 101 which will not work for our existing deployments. This would require coordinating our .NET 8 upgrades, per service, with similar updates to our deployments to align the ids.

Additionally, recommendations are to use a high numbered user id (>10,000 is a recommendation I've seen in several places, including the security scanning tool [Trivy](https://github.com/aquasecurity/trivy-operator/blob/main/docs/docs/configuration-auditing/built-in-policies.md) regarding this [specific vulnerability](https://avd.aquasec.com/misconfig/kubernetes/general/avd-ksv-0020/))

Blog post regarding container user colliding with a host user having the same id: [Understanding how uid and gid work in Docker containers](https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf)

[...]

PR #4476, opened 2023-03-06
Use higher number UID

lbussell:
See #4451 and specifically https://github.com/dotnet/dotnet-docker/issues/4451#issuecomment-1445187919


eng/dockerfile-templates/runtime-deps/Dockerfile.distroless-mariner
```diff
@@ -4,6 +4,11 @@
     set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set baseImage to cat(marinerRepo, "/distroless/minimal:", OS_VERSION_NUMBER) ^
     set username to "app" ^
+    set uid to when(dotnetVersion = "6.0" || dotnetVersion = "7.0",
+        when(find(OS_VERSION, "1.0") >= 0,
+            1000,
+            101),
+        64198) ^
     set uid to when(find(OS_VERSION, "1.0") >= 0, 1000, 101) ^
```
mthalman, line 12: This can be removed because it's incorporated in the previous statement, right?

MichaelSimons:
Since we are explicitly making this change to ensure a user ID >10,000, consider adding a unit test to validate this.   There are two valuable outcomes of this.  One, we ensure our ID is > 10,000.  Two, we ensure a consistent UID across all of our images (at least the subset we are able to make this change in).

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -73,6 +73,17 @@ public void VerifyDefaultUser(ProductImageData imageData)
             VerifyCommonDefaultUser(imageData);
         }
 
+        [LinuxImageTheory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifyNonRootUID(ProductImageData imageData)
```
MichaelSimons, line 78: Would it make sense to incorporate this into https://github.com/dotnet/dotnet-docker/blob/main/tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs#L98?  The primary benefit is we test this across all image types in the same place we test the default user.

lbussell: I moved it in with the CommonDefaultUser test. At first it felt redundant, but I think we would want to fail if we ever messed with the UID/users in any images deriving from runtime-deps.

tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
```diff
@@ -117,6 +117,37 @@ protected void VerifyCommonDefaultUser(ProductImageData imageData)
             }
 
             Assert.Equal(expectedUser, actualUser);
+
+            if (((imageData.Version.Major == 6 || imageData.Version.Major == 7) && (!imageData.IsDistroless || imageData.OS.StartsWith(OS.Mariner)))
+                || imageData.IsWindows)
+            {
+                OutputHelper.WriteLine("UID check is only relevant to .NET 8.0+ and distroless images besides CBL Mariner.");
```
MichaelSimons, line 124: Comment is confusing because it doesn't match the check - e.g. Windows images are excluded as well.

tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
```diff
@@ -117,6 +117,37 @@ protected void VerifyCommonDefaultUser(ProductImageData imageData)
             }
 
             Assert.Equal(expectedUser, actualUser);
+
```
MichaelSimons, line 120: IMO the readability would be improved by keeping this in the previous method you had - VerifyDefaultUserID.

MichaelSimons (approved):
I left a couple comments - overall LGTM.

mthalman (approved):

Issue #4467, opened 2023-02-28
The shell for `app` should be the same as `root`


richlander:
The two shells are different. I don't believe that's intentional. We're only wanting to differentiate the users on permissions.

```bash
# docker run --rm mcr.microsoft.com/dotnet/aspnet:8.0-preview cat /etc/passwd | grep /bin/
root:x:0:0:root:/root:/bin/bash
sync:x:4:65534:sync:/bin:/bin/sync
app:x:101:101::/home/app:/bin/false
```

mthalman:
[Triage]
This configuration occurred because the implementation of creating the user is shared with the implementation for our distroless Dockerfiles. In the distroless scenario, it does make sense to not have the shell set since there is none. It was just an oversight to not update the shell definition when definining the non-root user for the distroful scenario.

lbussell:
https://github.com/dotnet/dotnet-docker/pull/4477 makes it so that non-distroless images leave the shell unchanged when creating the non-root user.

PR #4477, opened 2023-03-06
Use default shell for non-root user in non-distroless images

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/4467

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
@@ -16,7 +16,8 @@
     "name": ARGS["name"],
     "uid": ARGS["uid"],
     "gid": ARGS["gid"],
-    "no-create-home": ARGS["no-create-home"]
+    "no-create-home": ARGS["no-create-home"],
+    "no-shell": "true"
 ])}} \{{if !ARGS["no-create-home"]:
     && install -d -m 0755 -o {{ARGS["uid"]}} -g {{ARGS["gid"]}} "{{ARGS["staging-dir"]}}/home/{{ARGS["name"]}}" \}}{{
     if ARGS["exclusive"]:{{if ARGS["create-dir"]:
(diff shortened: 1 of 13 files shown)


MichaelSimons (commented):
Consider added test coverage for this in VerifyCommonDefaultUser.  The goal would be to ensure consistency across our images and to prevent unintentional regressions in the future.

lbussell:
OK, I tried to write a test for this and came away with some learning.

1. When we remove the `--shell /bin/false` line, the default shell for the `app` user becomes `/sbin/nologin`, which is like `/bin/false` but instead of immediately exiting, prints "This account is not available". They're functionally equivalent.
2. The reason you still get a shell when running `docker run -it -u app <container>` is because the `-u app` argument only tells docker which user to run your process as, not which shell to use in interactive mode. For that, it looks at the entrypoint/CMD, which for alpine (where I've been testing), is `/bin/sh`: https://github.com/alpinelinux/docker-alpine/blob/d8ed1701dac37e1b6db026bec0a26be683288074/x86_64/Dockerfile#L3

Therefore, I'm not sure if there's a point to ever specify `/bin/false`, unless another distro has different behavior without specifying a shell. I'll do some research on that. I guess we could write a test to check `/etc/passwd` and make sure that non-root users, if they exist, have either `/bin/false` or `/sbin/nologin` set as their login shell.

mthalman:
> Consider added test coverage for this in VerifyCommonDefaultUser. The goal would be to ensure consistency across our images and to prevent unintentional regressions in the future.

Given that this has essentially the same functional behavior between the old definition and the new definition, the change here is basically cosmetic. I don't see much value having a test for it. Consistency across the Dockerfiles should be provided by the templates. So the test's conditional logic would essentially mirror the conditions in the templates. I'm all for tests but I want to them to have high value. Each test we add has an impact on pipeline performance and I want to avoid death by a thousand paper cuts.

mthalman (approved):

mthalman:
I think all distroless Dockerfiles should use `/bin/false` and everything else should be left unset (keep the default).

MichaelSimons (approved):

Issue #4462, opened 2023-02-28
Latest SDK docker image fails to execute `dotnet build`


mbreaton:
### Describe the Bug

Latest SDK docker image fails to execute `dotnet build`

### Steps to Reproduce
```
FROM mcr.microsoft.com/dotnet/aspnet:7.0 AS base
WORKDIR /app
EXPOSE 80
EXPOSE 443

FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
WORKDIR /src
COPY . .
RUN dotnet build "Test.sln" -c Release -o /app/build
```

### Other Information

Falling back to SDK 7.0.103 resolves the issues
```
FROM mcr.microsoft.com/dotnet/sdk:7.0.103 AS build
```

PR #4485, opened 2023-03-12
Change Publish Output Option to PublishDir

jeremyruffell:
Since the release of the **.NET 7.0.200 SDK** the `--output` Option is no longer supported for Build Related Commands.
I have created this PR to replace the `--output {OUTPUT_DIR}` with `/p:PublishDir={OUTPUT_DIR}` in the Dockerfile's and Documentation per the Microsoft Documentation mentioned above.

I believe this to be a cause of issue #4462 as reverting to a pre **.NET 7.0.100 SDK** solved there issue.

### DIFF
--- a/documentation/scenarios/nuget-credentials.md
+++ b/documentation/scenarios/nuget-credentials.md
@@ -46,7 +46,7 @@ RUN dotnet restore
 
 # Copy and publish app and libraries
 COPY . .
-RUN dotnet publish -c Release -o out  --no-restore
+RUN dotnet publish -c Release /p:PublishDir=out  --no-restore
 
 
 FROM mcr.microsoft.com/dotnet/runtime:6.0 AS runtime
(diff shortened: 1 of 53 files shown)


jeremyruffell:
@dotnet-policy-service agree

jeremyruffell:
> Background for the `output` option breaking change - https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid
> 
> I believe this was made in 7.0.200 not 7.0.100.

Wops! Typo on my end!

lbussell:
Per the documentation, the breaking change is only for solutions. So, this shouldn't change the behavior of the samples since we only use project files. Did you encounter this issue when building one of our samples or when building your own solution?

I think the warning is valuable. If we change the samples to include the `/p:PublishDir={OUTPUT_DIR}` workaround, a customer could potentially experience unexpected behavior in their application with no warnings.

Would it benefit you if we had a more complex sample that included a `.sln` file that follows the guidelines in the [breaking change documentation](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid)?

mthalman:
@jeremyruffell - Any response on this https://github.com/dotnet/dotnet-docker/pull/4485#issuecomment-1468991601?

mthalman:
Closing based on the response at https://github.com/dotnet/dotnet-docker/pull/4485#issuecomment-1468991601.

PR #4502, opened 2023-03-19
Add Kubernetes samples

richlander:
The goal is to add samples for a variety of scenarios. The described scenarios are just an initial set to get us started.

@karolz-ms


kubernetes/non-root-user/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+If the UID form was used in the Dockerfile, then `runAsUser` would not be needed, in the manifest fragment above.
+
+Given common patterns and in the interests of security, we recommend the following practice:
+
+- Set `USER` with the name in a Dockerfiles.
+- Set `runAsUser` with UID in Kubernetes manifests.
+
+This approach has the advantage of making Dockerfiles easy to inspect and secure by default, while establishing Kubernetes manifests as the final arbiter of deployment permissions.
```
karolz-ms, line 87: I do not quite understand the rationale for this recommendation; it seems error-prone. One needs to understand what the mapping between user name --> UID is (which is not part of the app Dockerfile--presumably it is established in the .NET base image?), and also explicitly put correct UID into the K8s manifest.

Why not rely on `runAsNonRoot` in K8s manifest instead? That makes the K8s manifest simpler without sacrificing security. Or am i missing something?

richlander: From my testing, there are three models that work (and what I was trying to describe):

1. Docker: `USER 64198`; Kubernetes: `runAsNonRoot: true`
2. Docker `USER app`; Kubernetes: `runAsNonRoot: true` + `runAsUser: 64198`
3. Docker: default root user; Kubernetes: `runAsNonRoot: true` + `runAsUser: 64198`

Which one are you advocating for? Or is there a model I'm missing?

karolz-ms: The first model

richlander: BTW: I found this space unintuitive. Also, I was looking at Dockerfile best practices and didn't notice anything about using uids. I suspect most people are not very familiar with that.

I think option 3 is bad. Option 1 is technically the best, which is what I suspect you are advocating for. That's also the one I was most attracted to, but I thought it would be harder for some folks to adopt. I'm happy to change the wording. I'm curious what @mthalman and @MichaelSimons think.

karolz-ms: +1 on getting feedback from M & M.

But a couple things to consider:
- Docker's own docs name explicit UID as the most reliable for setting up non-root user: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
- Explicit UID is almost always required when mounting volumes from host filesystem into non-root container, because if you do not use a UID that is meaningful on the host OS, you will get an access denied error. And volume mounting is how our tools set up debugging for .NET containers. Which makes me think at least some developers are familiar with using numeric UIDs in a Dockerfile.

richlander: Yup. I saw that point on explicit UID. That guidance is unrelated to this topic. It's saying that creating users w/o a UID is an anti-pattern. It isn't guidance (on how I read it) on what the value of `USER` should be.

All that said, I think the UID model is better. Let's turn this around. If we were to recommend that, what would be do in our samples? Remove the `runAsUser` line and then mention it in docs? I actually started with it in the sample but commented out.

karolz-ms: Yes, that is what I would expect, namely, that the image author should configure the image with appropriate user setting. 

The only scenario I can think of that would warrant explicit `runAsUser` setting in the Kubernetes manifest would be that the image has multiple users defined, and the image has been constructed in such a way that the consumer of the image should choose the user, depending on their use case for the image. I have never encountered a need like this in practice.

mthalman: > * Docker's own docs name explicit UID as the most reliable for setting up non-root user: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user

Note that this guidance is about _creating_ the user because of the non-determinism in which UID would be assigned if not explicitly set, not about configuring the `USER` instruction. So when it comes to what type of value to set for the `USER` instruction, Docker's best practices seem to have no opinion.

kubernetes/non-root-user/non-root-user.yaml
```diff
@@ -0,0 +1,33 @@
...
+    matchLabels:
+      app: aspnetapp
+  template:
+    metadata:
+      labels:
+        app: aspnetapp
+    spec:
+      containers:
+      - name: aspnetapp
+        image: dotnetnonroot.azurecr.io/aspnetapp
```
mthalman, line 18: This will need to change.

kubernetes/non-root-user/non-root-user.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: hello-dotnet
```
mthalman, line 4: It looks like all the deployment YAML files use `hello-dotnet` for the name. This can be confusing when going through all the examples here without cleaning up because they're all named the same.

kubernetes/non-root-user/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+```yml
+    spec:
+      containers:
+      - name: aspnetapp
+        image: dotnetnonroot.azurecr.io/aspnetapp
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsNonRoot: true
+          runAsUser: 64198
```
mthalman, line 55: This can be removed once the image is set with the new pattern.

samples/README.md
```diff
@@ -2,6 +2,8 @@
 
 The following samples and guidance demonstrate how to use .NET and Docker for development, testing and production. You can use the samples as the basis of your own Docker images or just to learn.
 
+Kubernetes samples are provided in the [kubernetes](../kubernetes/README.md) directory.
```
mthalman, line 5: I'd prefer having the kubernetes directory a sub-directory of samples. It can be "advertised" as its own thing in the root readme if necessary. The only argument to be made for having it at the repo root is to cater to people that would accidentally discover it. I don't think that's a compelling argument over keeping the root level of the repo clean. Let's link to it instead.

kubernetes/non-root-user/non-root-user.yaml
richlander: Right, but to what?

samples/README.md
richlander: The thing that most motivated me on this is the length of the URLs for `kubectl apply -f URL`.

I don't have an issue with moving this content underneath `samples`.

Here's what it looks like right now.

```bash
kubectl apply -f https://raw.githubusercontent.com/dotnet/dotnet-docker/main/kubernetes/non-root-user/non-root-user.yaml
```

kubernetes/non-root-user/non-root-user.yaml
mthalman: An issue should be logged to get a non-root sample image published.

samples/README.md
mthalman: The length of the URL doesn't really concern me. People will just be copy-and-pasting it anyway, right?

mthalman (approved):

samples/README.md
MichaelSimons: My preference is to place it under samples.

richlander: I'll move it.

richlander:
I believe I've made all the changes requested.

samples/aspnetapp/Dockerfile.alpine-non-root
```diff
@@ -20,9 +20,10 @@ FROM mcr.microsoft.com/dotnet/aspnet:8.0-preview-alpine
 # These ENVs are only needed because the sample uses .NET 7
 ENV DOTNET_ROLL_FORWARD=Major
 ENV DOTNET_ROLL_FORWARD_PRE_RELEASE=1
+ENV APP_UID=64198
```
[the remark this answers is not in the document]
mthalman: @richlander - Ping on this change.

samples/kubernetes/graceful-shutdown/README.md
```diff
@@ -0,0 +1,142 @@
+# Graceful shutdown
+
+Enable graceful shutdown of an app, particularly with long-running code. This pattern id described in more depth in [Graceful shutdown of ASP.NET-based applications](graceful-shutdown.md).
```
mthalman, line 3:
```suggestion
Enable graceful shutdown of an app, particularly with long-running code. This pattern is described in more depth in [Graceful shutdown of ASP.NET-based applications](graceful-shutdown.md).
```

samples/kubernetes/graceful-shutdown/README.md
```diff
@@ -0,0 +1,142 @@
...
+    return new DelayValue(value);
+});
+```
+
+Test the pattern with a single replica deployment. This test requires multiple terminal windows (with careful timing). The pod names and other data won't look the same.
+
+First terminal:
+
+```bash
+$ kubectl apply -f https://raw.githubusercontent.com/richlander/dotnet-k8s/main/hello-dotnet/hello-dotnet.yaml
```
mthalman, line 36:
```suggestion
$ kubectl apply -f https://raw.githubusercontent.com/dotnet/dotnet-docker/main/samples/kubernetes/hello-dotnet/hello-dotnet.yaml
```

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
```diff
@@ -0,0 +1,107 @@
...
+
+The default ASP.NET host [handles SIGTERM signal](https://learn.microsoft.com/aspnet/core/fundamentals/host/generic-host#ihostlifetime) and will raise the `ApplicationStopping` event when that happens. You should leverage that event to implement graceful application shutdown. How you handle the event, varies depending on what type of code is performing a long-running operation. There are two broad categories of such code: long-running requests and background processing.
+
+## Long-running network requests
+
+In general, it is better to avoid situations where a network requests takes a long time to process (several seconds or more). These kinds of requests tend to limit application scalability and are prone to failing due to network connection errors and client timeouts. It is better, if possible, to replace a long-running requests with a pair of related requests: the first request starts the operation, and the second request allows the client to ask for the result. If this is not feasible, you can make long-running requests more robust by judicious use of `CancellationTokens`.
```
mthalman, line 13:
```suggestion
In general, it is better to avoid situations where a network request takes a long time to process (several seconds or more). These kinds of requests tend to limit application scalability and are prone to failing due to network connection errors and client timeouts. It is better, if possible, to replace a long-running requests with a pair of related requests: the first request starts the operation, and the second request allows the client to ask for the result. If this is not feasible, you can make long-running requests more robust by judicious use of `CancellationTokens`.
```

samples/kubernetes/replicas-and-health/README.md
```diff
@@ -0,0 +1,79 @@
...
+Events:
+  Type    Reason     Age   From               Message
+  ----    ------     ----  ----               -------
+  Normal  Scheduled  31s   default-scheduler  Successfully assigned default/dotnet-replica-health-64d49554c9-f8mlp to minikube
+  Normal  Pulled     31s   kubelet            Container image "mcr.microsoft.com/dotnet/samples:aspnetapp" already present on machine
+  Normal  Created    31s   kubelet            Created container aspnetapp
+  Normal  Started    30s   kubelet            Started container aspnetapp
+```
+
+Much of the output has removed in this example. The key part is the events. Kubernetes will show a failing container, as is demonstrated in [Configure Liveness, Readiness and Startup Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/), as follows.
```
mthalman, line 44:
```suggestion
Much of the output has been removed in this example. The key part is the events. Kubernetes will show a failing container, as is demonstrated in [Configure Liveness, Readiness and Startup Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/), as follows.
```

mthalman (approved):

richlander:
Thanks for the feedback. I thought I'd removed the ENV for APP_UID. Ughh. Gone now.

Good now @MichaelSimons?

richlander:
I added a `LoadBalancer` example since that works much better with a cloud service. The `NodePort` examples work better for local experimentation.

Covered, in part, by:

- https://minikube.sigs.k8s.io/docs/handbook/accessing/
- https://learn.microsoft.com/en-us/azure/aks/tutorial-kubernetes-deploy-application?tabs=azure-cli#test-the-application

mthalman (approved):

samples/kubernetes/README.md
```diff
@@ -0,0 +1,24 @@
...
+Most of the samples use a `NodePort` service since that works better locally. Use the [Cloud ready app](hello-cloud-apps/README.md) -- which uses a `LoadBalancer` service -- if you want to test the app in a cloud Kubernetes service.
+
+You can host a .NET sample with a [few quick commands](hello-dotnet/README.md).
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/dotnet/dotnet-docker/main/samples/kubernetes/hello-dotnet/hello-dotnet.yaml
+kubectl port-forward service/hello-dotnet 8080:80
+```
+
+If you are new to Kubernetes, you'll need to establish a [local environment](https://kubernetes.io/docs/tasks/tools/), such as [Docker Desktop](https://www.docker.com/products/kubernetes/), [K3s](https://k3s.io/) and [Rancher](https://rancherdesktop.io/).
```
MichaelSimons, line 16:
```suggestion
If you are new to Kubernetes, you'll need to establish a [local environment](https://kubernetes.io/docs/tasks/tools/), such as [Docker Desktop](https://www.docker.com/products/kubernetes/), [K3s](https://k3s.io/) or [Rancher](https://rancherdesktop.io/).
```

samples/kubernetes/hello-cloud-dotnet/README.md
```diff
@@ -0,0 +1,35 @@
...
+kubectl apply -f hello-cloud-dotnet.yaml
+```
+
+To run on AKS or another cloud service, configure `kubectl` to access that kubernetes cluster control plane. Otherwise, you can configure your local environment to [create a `LoadBalancer` tunnel](https://minikube.sigs.k8s.io/docs/handbook/accessing/#example-of-loadbalancer).
+
+You can [discover the external IP](https://learn.microsoft.com/en-us/azure/aks/tutorial-kubernetes-deploy-application?tabs=azure-cli#test-the-application) for the service using `kubectl`.
+
+```bash
+% kubectl get service -w
+NAME                 TYPE           CLUSTER-IP    EXTERNAL-IP   PORT(S)          AGE
```
MichaelSimons, line 23: Curious if this was hand edited?  I am surprised the last line is not column aligned with the header.

samples/kubernetes/graceful-shutdown/README.md
```diff
@@ -0,0 +1,142 @@
+# Graceful shutdown
+
+Enable graceful shutdown of an app, particularly with long-running code. This pattern is described in more depth in [Graceful shutdown of ASP.NET-based applications](graceful-shutdown.md).
+
+The [following pattern](https://github.com/richlander/dotnet-docker/blob/fe9ec76b68acca1c2319a08482d9d0d594751422/samples/aspnetapp/aspnetapp/Program.cs#L30-L55) can be used:
```
MichaelSimons, line 5: Smells to be referencing your fork here - shouldn't this reference the dotnet source?

MichaelSimons (commented):
Few minor comments, rest looks great.

samples/kubernetes/graceful-shutdown/README.md
richlander: Ooops. Fixed.

samples/kubernetes/hello-cloud-dotnet/README.md
richlander: I don't think it was.

samples/kubernetes/README.md
```diff
@@ -0,0 +1,37 @@
+# Using .NET with Kubernetes
+
+[Kubernetes](https://kubernetes.io/) provides an orchestration system for containers. You can host .NET in Kubernetes in the same way as other dev platforms. We've provided samples that demonstrate best practices for configuring .NET apps with Kubernetes. More general container samples are provided in the [samples](../README.md) directory.
+
+These instruction use [kubectl](https://kubernetes.io/docs/reference/kubectl/) and should work with any [Kubernetes](https://kubernetes.io/) environment.
+
+If you are new to Kubernetes, you'll need to establish a [local environment](https://kubernetes.io/docs/tasks/tools/), such as [Docker Desktop](https://www.docker.com/products/kubernetes/), [K3s](https://k3s.io/) and [Rancher](https://rancherdesktop.io/).
```
tmds, line 7: Maybe you include [OpenShift Local](https://developers.redhat.com/products/openshift-local/overview)?

samples/kubernetes/README.md
```diff
@@ -0,0 +1,37 @@
...
+- [Replicas and health checks](health-and-replicas/README.md)
+- [Non-root user](non-root-user/README.md)
+- [Graceful shutdown](graceful-shutdown/README.md)
+- [Manual deployment](manual-deployment/README.md)
+
+The [`hello-dotnet`](hello-dotnet/README.md) example is demonstrates most of these aspects together, in a real-world configuration.  `NodePort` and `LoadBalancer` variants are provided.
+
+## Run .NET app on your local cluster
+
+You can host a `NodePort` sample on your local cluster:
```
tmds, line 23: I've never used the `NodePort` type.
I always use `ClusterIP` independent of whether I'm deploying local or on a cluster.
I'm not sure why this example uses `NodePort`.

samples/kubernetes/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```bash
+$ kubectl apply -f https://raw.githubusercontent.com/dotnet/dotnet-docker/main/samples/kubernetes/hello-dotnet/hello-dotnet-nodeport.yaml
+$ kubectl port-forward service/hello-dotnet 8080:80
```
tmds, line 27: Can you use the `port-forward` command with a service?
I thought it only works with pods.

samples/kubernetes/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## Run .NET app on your cloud cluster
+
+You can host a `LoadBalancer` sample on your cloud cluster:
```
tmds, line 32: Maybe include something about `LoadBalancer` vs `ClusterIP`?

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
```diff
@@ -0,0 +1,107 @@
...
+    var originalCt = httpContext.RequestAborted;
+    var combinedCt = CancellationTokenSource.CreateLinkedTokenSource(originalCt, hostLifetime.ApplicationStopping).Token;
+    httpContext.RequestAborted = combinedCt;
+    return next(httpContext);
+});
+```
+
+## Adding a shutdown delay
+
+Services that handle requests coming from [Kubernetes ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) may run into an issue where they continue to receive new requests *after* Kubernetes requests them to shut down via `SIGTERM` signal. This usually lasts for only a few seconds, but if the service shuts down immediately and these requests fail, it puts a burden on the client to retry them and the overall service responsiveness goes down.
```
tmds, line 54: > may run into an issue where they continue to receive new requests *after* Kubernetes requests them to shut down via `SIGTERM` signal. This usually lasts for only a few seconds, but if the service shuts down immediately and these requests fail, it puts a burden on the client to retry them and the overall service responsiveness goes down.

I wonder: when ASP.NET receives SIGTERM, will it continue to accept new requests while it is stopping?

cc @halter73

samples/kubernetes/non-root-user/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Delete the resources (remote URL or local manifest).
+
+```bash
+kubectl delete -f https://raw.githubusercontent.com/dotnet/dotnet-docker/main/samples/kubernetes/non-root-user/non-root-user.yaml
+```
+
+## SecurityContext
+
+[`SecurityContext`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#securitycontext-v1-core) holds security configuration that will be applied to a container, in a Pod or [Deployment manifest](non-root-user.yaml). It includes a few fields that are required to configure a container to correctly run as a non-root user. This sample follows [Kubernetes "Restricted" hardening best practices](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted).
```
tmds, line 51: > It includes a few fields that are required to configure a container to correctly run as a non-root user.

These fields aren't required to run a rootless image.
They prevent running a non-rootless image.

If your image is rootless, you don't need to set these fields.
You can still set `runAsNonRoot` to ensure your only running rootless images.

samples/kubernetes/replicas-and-health/README.md
```diff
@@ -0,0 +1,79 @@
+# Replicas backed by health-checks
+
+[Kubernetes](https://kubernetes.io/) enables creating multiple [replicas](https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/) and registering [health checks](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/).
```
tmds, line 3: It would be good to have some explanation about the different probe types.

These examples show only liveness probes.

Quoting from the doc:

> Caution: Liveness probes can be a powerful way to recover from application failures, but they should be used with caution. Liveness probes must be configured carefully to ensure that they truly indicate unrecoverable application failure, for example a deadlock.

samples/kubernetes/README.md
richlander: I picked that up from reading Kubernetes and Minikube docs. They seem to use `NodePort` in a lot of their examples. Certainly, there is also `LoadBalancer` and `ClusterIP`.

This doc (which I found just now) does a good job of describing the different service types: https://learn.microsoft.com/en-us/azure/aks/concepts-network#services

Perhaps I should be using `ClusterIP` and `LoadBalancer`. The former seems to work fine for local usage. `LoadBalancer` works really well for getting a public IP.

samples/kubernetes/README.md
richlander: https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/#forward-a-local-port-to-a-port-on-the-pod

samples/kubernetes/README.md
richlander: Feel free to suggest the change.

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
richlander: @karolz-ms

samples/kubernetes/non-root-user/README.md
richlander:
```suggestion
[`SecurityContext`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#securitycontext-v1-core) holds security configuration that will be applied to a container, in a Pod or [Deployment manifest](non-root-user.yaml). These fields can be used to configure and/or validate that the container image is running as non-root. This sample follows [Kubernetes "Restricted" hardening best practices](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted).
```

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
karolz-ms: @tmds great point that I have missed--it seems that once the ApplicationStopping CancellationToken is activated, ASP.NET no longer accepts new requests. Let me do some research to figure out what is the best way to delay it.

karolz-ms: The behavior I really want is this: after receiving SIGTERM, accept new requests and process in-flight requests for `n` seconds more, THEN stop accepting new requests, cancel requests in-flight, and shut down. 

I believe setting `HostOptions.ShutdownTimeout` only helps with the second part, and only in imperfect way: it is an abort. I would like to cancel the in-flight requests instead and return `503 Service Unavailable`. This way the clients immediately know the request failed and can retry it, as opposed to encountering silence on the wire and timing out eventually.

tmds: > This way the clients immediately know the request failed and can retry it, as opposed to encountering silence on the wire and timing out eventually.

Have you observed this, or is there an issue that describes this?

karolz-ms: No, I have not observed this. 

I think you are hinting at the fact that, normally, when the service reaches the shutdown timeout and just quits, the .NET runtime, or OS, will just close remaining socket automatically and notify any clients?

karolz-ms: I got the behavior that I wanted by replacing the default `IHostLifetime` implementation. Will send an update shortly

tmds: > just close remaining socket automatically and notify any clients?

Yes, I expect the close of the TCP connection to be observable by the client without a time out. The client can then make another request.

It would be good to describe specifically what problem gets solved with the suggested code.

I expect the client to have some responsibility in a robust, distributed application. The server can't solve the problem on its own.

karolz-ms: > > just close remaining socket automatically and notify any clients?
> 
> Yes, I expect the close of the TCP connection to be observable by the client without a time out. The client can then make another request.

Makes sense, but I think an argument still can be made for graceful request cancellation, if not for client's sake, then to ensure that the resources used by the service instance are released/cleaned up.

> It would be good to describe specifically what problem gets solved with the suggested code.

I took care of explaining the rationale in the doc. Very briefly: preventing long-running requests from delaying the shutdown, and having a bit of a shutdown delay to give ingress time to stop sending requests to the instance.

> I expect the client to have some responsibility in a robust, distributed application. The server can't solve the problem on its own.

Agreed.

tmds: > preventing long-running requests from delaying the shutdown

This is what `HostOptions.ShutdownTimeout` provides, right?

> and having a bit of a shutdown delay to give ingress time to stop sending requests to the instance.

It's not clear to me what this solves.

What happens when ingress sends requests to an instance that it stopping?

Doesn't the connection fail, causing it to be made against another pod of the deployment?

karolz-ms: > > preventing long-running requests from delaying the shutdown
> 
> This is what `HostOptions.ShutdownTimeout` provides, right?

Not quite, this does not give request handlers opportunity to cancel work cleanly.

> > and having a bit of a shutdown delay to give ingress time to stop sending requests to the instance.
> 
> It's not clear to me what this solves.
> 
> What happens when ingress sends requests to an instance that it stopping?
> 
> Doesn't the connection fail, causing it to be made against another pod of the deployment?

Quoting from the doc:

> Services that handle requests coming from [Kubernetes ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) may run into an issue where they continue to receive new requests *after* Kubernetes requests them to shut down via `SIGTERM` signal. This usually lasts for only a few seconds, but if the service shuts down immediately and these requests fail, it puts a burden on the client to retry them and the overall service responsiveness goes down.
>
> The reason for this behavior is that ingress and Kubernetes control plane are separate entities, operating independently. When Kubernetes decides to shut down some pod(s), there is inevitable delay between control plane taking a pod out of Service backend, and ingress "noticing" that this has happened and actually stopping routing requests to that pod. So it is useful to introduce some delay between the time the pod receivers `SIGTERM` signal, and the time when it stops responding to requests.

samples/kubernetes/README.md
tmds, line 7:
```suggestion
If you are new to Kubernetes, you'll need to establish a [local environment](https://kubernetes.io/docs/tasks/tools/), such as [Docker Desktop](https://www.docker.com/products/kubernetes/), [K3s](https://k3s.io/), [Rancher](https://rancherdesktop.io/) and [OpenShift Local](https://developers.redhat.com/products/openshift-local).
```

samples/kubernetes/README.md
tmds: Also looked through the docs.

A `LoadBalancer` allows to make the service available through the cloud platform (by leveraging the cloud platform load balancer).

While a `ClusterIP` makes the service available in the Kubernetes cluster.

> Perhaps I should be using ClusterIP and LoadBalancer. The former seems to work fine for local usage. LoadBalancer works really well for getting a public IP.

Based on these definitions I think that makes sense.

samples/kubernetes/README.md
tmds: Good to know! I've always looked up the pod names.

richlander: Someone was joking with me yesterday about another Kubernetes topic:

"There are ten thousand things to know about Kubernetes. How could you not know this?"

samples/kubernetes/README.md
richlander: I re-ordered these so that they would be in alpha-order (no favorites).

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
karolz-ms: Yes, that is what I am proposing. The current behavior is:

- when `SIGTERM` arrives, stop accepting new requests immediately
- run "in flight" requests to completion, subject to `HostOptions.ShutdownTimeout`
- additional coding is needed if request handlers need to be `ApplicationStopping`-aware (e.g. cancel long work)

I found a good, existing issue about this problem space in ASP.NET repo and added a comment there: https://github.com/dotnet/aspnetcore/issues/30387#issuecomment-1503891178 

I have also added a reference to `HostOptions.ShutdownTimeout` to the doc: https://github.com/dotnet/dotnet-docker/pull/4502/commits/d00581a801a16efcd8b6a26223b39e72fe6425b9  -- thanks for the suggestion (and for good discussion!)

richlander:
I added and updated everything I want for now.

I still need to do more work on the `hello-dotnet` sample. I want to add `dotnet-monitor` to it and add prometheus annotations so that cloud services, like Azure Monitor, will be able to collect data automagically. I'll do that in a follow-up PR.

You may notice that I added load test and Prometheus manifests. The Prometheus one is pretty straightforward, but I need to move the bombardier container image to another location. I can do that later.

samples/kubernetes/graceful-shutdown/README.md
```diff
@@ -0,0 +1,142 @@
+# Graceful shutdown
+
+Enable graceful shutdown of an app, particularly with long-running code. This pattern is described in more depth in [Graceful shutdown of ASP.NET-based applications](graceful-shutdown.md).
+
+The [following pattern](https://github.com/dotnet/dotnet-docker/blob/main/samples/aspnetapp/aspnetapp/Program.cs#L30-L55) can be used:
+
+```bash
```
ilmax, line 7:
```suggestion
```csharp
```

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Long-running network requests
+
+In general, it is better to avoid situations where a network request takes a long time to process (several seconds or more). These kinds of requests tend to limit application scalability and are prone to failing due to network connection errors and client timeouts. It is better, if possible, to replace a long-running requests with a pair of related requests: the first request starts the operation, and the second request allows the client to ask for the result. If this is not feasible, you can make long-running requests more robust by judicious use of `CancellationTokens`.
```
tmds, line 19: > you can make long-running requests more robust

More explicit:

The use of `CancellationToken` in this section is to make a request that takes longer than `HostOptions.ShutdownTimeout` abort early.

By aborting early, the client will retry the request sooner.

(note that this request may have been started before the `SIGTERM`, so it could have finish in the `ShutdownTimeout`)

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
```diff
@@ -0,0 +1,129 @@
...
+        return TypedResults.StatusCode(StatusCodes.Status503ServiceUnavailable);
+    }
+    
+    return TypedResults.Ok($"Worked {value} seconds, looks good");
+});
+```
+
+The same technique can be applied to traditional (controller-based) ASP.NET code. You just need to add `IHostApplicationLifetime` parameter to controller method(s) that might take a long time to execute, construct a linked `CancellationToken`, and then use it inside the request method.
+
+You can also change ASP.NET behavior so that the default `CancellationToken` passed to your request handlers *does get activated* when `ApplicationStopping` event occurs. This code snippet (a custom middleware) will do it:
```
tmds, line 45: Is this different from `opts.ShutdownTimeout = TimeSpan.FromSeconds(0)`?

karolz-ms: Yes, in the sense that the request handler now is notified about the need to shut down and has a chance to react.

tmds: Ah, yes. Got it.

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
```diff
@@ -0,0 +1,129 @@
...
+
+## How Kubernetes requests application shutdown
+
+Kubernetes will send `SIGTERM` signal to the main container process as means of requesting shutdown. By default the container has 30 seconds to exit gracefully before it gets killed, but this can be changed via [container lifecycle spec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#lifecycle).
+
+The default ASP.NET host [handles SIGTERM signal](https://learn.microsoft.com/aspnet/core/fundamentals/host/generic-host#ihostlifetime) and will raise the `ApplicationStopping` event when that happens. The host will stop accepting new requests, but the requests already "in flight" will run to completion, subject to a timeout. The default shutdown timeout in ASP.NET is also 30 seconds, and can be changed via `HostOptions.ShutdownTimeout`, for example:
```
tmds, line 9: > The default shutdown timeout in ASP.NET is also 30 seconds

Where did you find the default?

karolz-ms: https://github.com/dotnet/dotnet/blob/main/src/runtime/src/libraries/Microsoft.Extensions.Hosting/src/HostOptions.cs

karolz-ms: That's probably wrong value though https://github.com/dotnet/dotnet/blob/main/src/aspnetcore/src/Hosting/Hosting/src/Internal/WebHostOptions.cs good catch, will fix

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
karolz-ms: I am proposing

> ...  you can make long-running requests cancel work promptly (and avoid being aborted by the ASP.NET host) by leveraging the `CancellationToken` that is associated with `ApplicationStopping` event.

samples/kubernetes/graceful-shutdown/graceful-shutdown.md
tmds: The `30s` is correct. I was wondering where you found it.

```cs
var builder = WebApplication.CreateBuilder(args);
Console.WriteLine(builder.Configuration.Get<HostOptions>().ShutdownTimeout);
```
prints
```
00:00:30
```

ArminShoeibi:
Hi Rich @richlander 

We are using this method 

```c#
public class ApplicationLifetimeHostedService(IHostApplicationLifetime hostApplicationLifetime, ILogger<ApplicationLifetimeHostedService> logger) : BackgroundService
{
    protected override Task ExecuteAsync(CancellationToken stoppingToken)
    {
        hostApplicationLifetime.ApplicationStopping.Register(() =>
        {
            logger.LogError("SIGTERM received, waiting for 90 seconds");
            Thread.Sleep(TimeSpan.FromSeconds(90));
            logger.LogError("Termination delay complete, continuing stopping process");
        });
        return Task.CompletedTask;
    }
}
```

Should we override the default IHostlifetime AKA ConsoleLifetime and delete this background service?

PR #4508, opened 2023-03-20
Add ENV variable for app user ID

### LINKED ISSUE #4506: Add `ENV` for `app` UID
We had a discussion about how folks should use the new `app` user and whether they should use its name or UID for the `USER` instruction. We decided that the UID is best since it works better with the `runAsNonUser` Kubernetes feature.

Plan:

- Add `ENV APP_UID=64198` to `runtime-deps` layer
- Encourage users to use the following pattern: `USER $APP_UID`

It's possible that having an ENV for the UID may be useful for other scenarios.

@karolz-ms

lbussell:
Fixes #4506

@mthalman, are the changes to the 6.0/7.0 CBL Mariner distroless images appropriate? They can be reversed if we don't want to add anything new and potentially breaking there.

### DIFF
--- a/eng/dockerfile-templates/Dockerfile.common-dotnet-envs
+++ b/eng/dockerfile-templates/Dockerfile.common-dotnet-envs
@@ -8,7 +8,7 @@
     set isDistroless to find(OS_VERSION, "distroless") >= 0 || find(OS_VERSION, "chiseled") >= 0 ^
     set lineContinuation to when(isWindows, "`", "\") ^
     set port to when(isDistroless || (dotnetVersion != "6.0" && dotnetVersion != "7.0"), "8080", "80")
-}}ENV {{lineContinuation}}{{if (!isWindows && dotnetVersion != "6.0" && dotnetVersion != "7.0") || isDistroless:
+}}ENV {{lineContinuation}}{{if (!isWindows && dotnetVersion != "6.0" && dotnetVersion != "7.0") || (isDistroless && !isMariner):
     # UID of the non-root user 'app'
     APP_UID={{ARGS["uid"]}} {{lineContinuation}}}}
     # Configure web servers to bind to port {{port}} when present
--- a/eng/dockerfile-templates/Dockerfile.common-dotnet-envs
+++ b/eng/dockerfile-templates/Dockerfile.common-dotnet-envs
@@ -9,6 +9,7 @@
     set lineContinuation to when(isWindows, "`", "\") ^
     set port to when(isDistroless || (dotnetVersion != "6.0" && dotnetVersion != "7.0"), "8080", "80")
 }}ENV {{lineContinuation}}{{if (!isWindows && dotnetVersion != "6.0" && dotnetVersion != "7.0") || isDistroless:
+    # UID of the non-root user 'app'
     APP_UID={{ARGS["uid"]}} {{lineContinuation}}}}
     # Configure web servers to bind to port {{port}} when present
     {{if dotnetVersion = "6.0" || dotnetVersion = "7.0":ASPNETCORE_URLS=http://+:{{port}}^else:ASPNETCORE_HTTP_PORTS={{port}}}} {{lineContinuation}}
--- a/eng/dockerfile-templates/Dockerfile.common-dotnet-envs
+++ b/eng/dockerfile-templates/Dockerfile.common-dotnet-envs
@@ -1,4 +1,6 @@
 {{
+    _ ARGS:
+        uid (optional): ID of the non-root user ^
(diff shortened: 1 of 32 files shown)


eng/dockerfile-templates/Dockerfile.common-dotnet-envs
```diff
@@ -1,12 +1,15 @@
...
     set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set isAlpine to find(OS_VERSION, "alpine") >= 0 ^
     set isWindows to find(OS_VERSION, "nanoserver") >= 0 || find(OS_VERSION, "windowsservercore") >= 0 ^
     set isMariner to find(OS_VERSION, "cbl-mariner") >= 0 ^
     set isDistroless to find(OS_VERSION, "distroless") >= 0 || find(OS_VERSION, "chiseled") >= 0 ^
     set lineContinuation to when(isWindows, "`", "\") ^
     set port to when(isDistroless || (dotnetVersion != "6.0" && dotnetVersion != "7.0"), "8080", "80")
-}}ENV {{lineContinuation}}
+}}ENV {{lineContinuation}}{{if (!isWindows && dotnetVersion != "6.0" && dotnetVersion != "7.0") || isDistroless:
+    APP_UID={{ARGS["uid"]}} {{lineContinuation}}}}
```
mthalman, line 12: Include a comment for this variable. Something like: `UID of the non-root 'app' user`

mthalman (commented):
Also need to update the unit tests to validate the new environment variable.

src/runtime-deps/8.0/jammy/arm64v8/Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN groupadd \
         app
 
 ENV \
+    APP_UID=64198 \
```
MichaelSimons, line 29: Ideally the Dockerfiles would make use of this ENV during the creation of the user.  Thoughts on refactoring to support this?  In the runtime, aspnet, and sdk Dockerfiles we define the ENVs at the beginning.  Doing the same in the runtime-deps would make it consistent and support utilizing the ENVs anywhere.

lbussell: It makes sense to me, since the ENV variables will change less often than the runtime dependencies.

mthalman:
> @mthalman, are the changes to the 6.0/7.0 CBL Mariner distroless images appropriate?

I originally didn't think this was going to be a breaking change. But after seeing that the `VerifyDefaultUser` test is failing, I realize now that it is technically a breaking change because it is changing the value of the User in the image metadata. So I think we should exclude Mariner distroless from these changes.

richlander:
I was also wondering that, about Mariner 6/7. I agree. Let's got with .NET 8.

tests/Microsoft.DotNet.Docker.Tests/ProductImageData.cs
```diff
@@ -55,6 +55,27 @@ public ImageVersion VersionFamily
 
         public override int DefaultPort => (IsDistroless | Version.Major >= 8) ? 8080 : 80;
 
+        public override int? NonRootUID {
+            get {
+                if (IsWindows)
```
MichaelSimons, line 60: Having a windows condition here is a type of duplication.  If the value ever needed to be changed, it would require changing two places in the code.  This introduces a bit of maintenance burden.  Consider refactoring the checks so that base.NonRootUID is used for the Windows case.  It feels like the Windows check could be dropped completely.

```
if (OS == Tests.OS.Mariner10Distroless)
{
    return 1000;
}
else if (OS == Tests.OS.Mariner20Distroless && (Version.Major == 6 || Version.Major == 7))
{
    return 101;
}
else
{
    return base.NonRootUID;
}
```

MichaelSimons (commented):
One minor suggestion, rest looks good.

tests/Microsoft.DotNet.Docker.Tests/ProductImageData.cs
MichaelSimons: > I was trying to capture that Windows doesn't really have a numeric UID, the whole property itself doesn't make sense for Windows, which is why it's nullable. 

Right I get that, but you have this logic duplicated.  Once here and again in the base property https://github.com/dotnet/dotnet-docker/pull/4508/files#diff-821437012d57821526073deb389b2d8379631e855337319e5e3dae32a7f8ee92R26.  If you remove the windows check here, you would share the windows logic via the else case.

lbussell: Oh yep, forgot about that in the base value. Makes sense, will remove the windows check here.

mthalman (approved):

Issue #4343, opened 2023-01-20
Add Dockerfiles for Linux .NET Composite Images


ivdiazsa:
### Describe the Problem

We have been working on .NET composites where, by bundling together the most needed assemblies, we are able to deliver a lower startup time, and overall better performance when running .NET apps. Our next step is to have the official Docker images recurrently built, alongside the existing .NET Linux images, so that customers can consume them.

### Describe the Solution

We currently build them by hand but that can only be used for internal testing. We would like to have them officially published in the website: https://hub.docker.com/_/microsoft-dotnet

### Additional Context

In order to achieve this purpose, we would like to add the corresponding Dockerfiles in this repo, which will pull the composites from the ASP.NET repo artifacts, and use them in a similar fashion as the currently existing Dockerfiles.

ivdiazsa:
@mangod9 @richlander Here's the issue we can use to track progress and discussions to get the .NET composite images, and Dockerfiles, published to customers.

ivdiazsa:
Adding @trylek as well as he has been a key contributor to this project.

tmds:
> We have been working on .NET composites where, by bundling together the most needed assemblies

Is a .NET composite a single file that has bundles parts of different assemblies?

> we are able to deliver a lower startup time, and overall better performance

What causes it to start faster? and achieve a better performance?

trylek:
The .NET composite image bundles together the ready-to-run native code from all component assemblies. The managed assemblies themselves remain on disk but they are basically pure MSIL. (Everything can then be bundled into one file using the Single EXE .NET feature though.) The better startup and to a certain extent performance (so far we've mostly been focusing on startup) stems from the fact that a composite image can be treated as a "versioning bubble". The limitation that it always naturally needs to be shipped / serviced as a whole lets us do the following:

* JIT can drop ready-to-run type / field layout compatibility checks at assembly boundaries;
* We can inline more code across assemblies within the versioning bubble;
* We can directly encode field offsets for types from a different assembly within the versioning bubble;
* We can compile a larger set of generic instantiations over types from all assemblies within the versioning bubble.

[...]

PR #4532, opened 2023-03-30
Add Dockerfiles for Linux .NET Composite Images

lbussell:
See discussion at https://github.com/dotnet/dotnet-docker/issues/4343

A few things:
- The musl changes have not flown in yet
- There are no updates to tests yet
- SDK images are based on aspnet-composite but will fail to build if sdk/aspnet's runtime is incoherent
- The Dockerfile.linux-composite template is copied mostly from the runtime template, so there's probably room to clean up some unnecessary conditions

I believe at this point the preview 3 build might be coherent, not sure how that will affect the SDK build.

### DIFF
--- a/eng/dockerfile-templates/aspnet/Dockerfile.linux-composite
+++ b/eng/dockerfile-templates/aspnet/Dockerfile.linux-composite
@@ -7,52 +7,47 @@
     set isDistrolessMariner to isMariner && isDistroless ^
     set baseUrl to VARIABLES[cat("base-url|", dotnetVersion, "|", VARIABLES["branch"])] ^
     set isInternal to find(baseUrl, "msrc") >= 0 || find(baseUrl, "internal") >= 0 ^
-    set isRpmInstall to isFullMariner && dotnetVersion = "6.0" ^
+    set isRpmInstall to false ^
     set isSingleStage to (isAlpine || isRpmInstall) && !isInternal ^
-    set runtimeBaseTag to
-        cat("$REPO:", VARIABLES[cat("dotnet|", dotnetVersion, "|product-version")], "-", OS_VERSION, ARCH_TAG_SUFFIX) ^
-    set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
+    set runtimeDepsBaseTag to cat(
+        "$REPO:", VARIABLES[cat("dotnet|", dotnetVersion, "|product-version")], "-", OS_VERSION, ARCH_TAG_SUFFIX) ^
     set osVersionBase to when(isDistroless, match(OS_VERSION, ".+(?=.*-)")[0], OS_VERSION_BASE) ^
     set installerImageTag to when(isDistrolessMariner,
         cat("mcr.microsoft.com/cbl-mariner/base/core:", OS_VERSION_NUMBER),
         when(isAlpine || isFullMariner,
-            runtimeBaseTag,
-            cat(ARCH_VERSIONED, "/buildpack-deps:", osVersionBase, "-curl"))) ^
-    set copyFromSrcPath to when(isDistroless, "/dotnet", "/shared/Microsoft.AspNetCore.App") ^
-    set copyFromDstPath to when(isDistroless, "/usr/share/dotnet", "/usr/share/dotnet/shared/Microsoft.AspNetCore.App")
-}}ARG REPO=mcr.microsoft.com/dotnet/{{
-    if dotnetVersion = "6.0" || dotnetVersion = "7.0":{{
-        echo "runtime"
-    }}^
-    else:{{
-        echo "runtime-deps"
-    }}}}{{if isSingleStage:
+            runtimeDepsBaseTag,
+            cat(ARCH_VERSIONED, "/buildpack-deps:", osVersionBase, "-curl")))
+}}ARG REPO=mcr.microsoft.com/dotnet/runtime-deps{{if isSingleStage:
 {{
 
-    _ SINGLE STAGE
+_ SINGLE STAGE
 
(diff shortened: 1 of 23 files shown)


lbussell:
@trylek FYI, the error when running the SDK here is different than when the runtime version is incoherent: https://dev.azure.com/dnceng-public/public/_build/results?buildId=223533&view=logs&j=122e9d27-e507-5abe-8693-a1b4223451d1&t=9a62c243-bd71-5ad9-80c3-4732047202bc&l=1518

```
Process terminated. MVID mismatch between loaded assembly 'Microsoft.Extensions.Logging.Abstractions' (MVID = {bac9dcef-a261-4318-95e1-e275aa1f3054}) and an assembly with the same simple name embedded in the native image 'full-composite.r2r.dll' (MVID = {ee18884a-ce2c-4fc2-9032-f3e6dd125475})
   at Microsoft.DotNet.Cli.NewCommandParser.GetCommand()
   at Microsoft.DotNet.Cli.Parser..cctor()
   at Microsoft.DotNet.Cli.Program.ProcessArgs(System.String[], System.TimeSpan, Microsoft.DotNet.Cli.Telemetry.ITelemetry)
   at Microsoft.DotNet.Cli.Program.Main(System.String[])
```

trylek:
Thanks Logan, I think this basically means that the ASP.NET assembly <code>Microsoft.Extensions.Logging.Abstractions</code> coming from the ASP.NET composite image was somehow overwritten or overridden by its non-composite version. I'm experimenting with the script now based on Michael's feedback, I hope to be able to propose a better change than I did yesterday.

ivdiazsa:
> Thanks Logan, I think this basically means that the ASP.NET assembly `Microsoft.Extensions.Logging.Abstractions` coming from the ASP.NET composite image was somehow overwritten or overridden by its non-composite version. I'm experimenting with the script now based on Michael's feedback, I hope to be able to propose a better change than I did yesterday.

Yup, this is exactly what that error message means.

ivdiazsa:
We've got a problem on our hands. So, @lbussell is facing the MVID mismatch error again here: https://github.com/dotnet/dotnet-docker/pull/4538#issuecomment-1506080743

The problem here is... our culprit lies elsewhere. It is the `Microsoft.Extensions.Logging.Abstractions` located under the `sdk` path. For some reason, it also wants the composite. At some point I dug into that potential rabbit hole, and if I copied the `Microsoft.Extensions.Logging.Abstractions` from `shared/Microsoft.AspNetCore.App` to `sdk`, the error went away and it worked well.

We could hack our way into getting it copied when generating the composite, but I fear we will get a lot of push back from the SDK people, since our project's scope is just ASP.NET.

lbussell:
@ivdiazsa @trylek please take a look at the latest build failures in Azure Pipelines. There are MVID mismatches when trying to install powershell on the composite SDK images. I was able to reproduce this from https://github.com/dotnet/dotnet-docker/pull/4538 by running `dotnet tool install --global powershell` on the built SDK image.

Maybe @adityapatwardhan can take a look as well?

```bash
root@e0c7592da6e2:/# dotnet tool install --global powershell
Process terminated. MVID mismatch between loaded assembly 'System.Security.Cryptography.Pkcs' (MVID = {2ee770f7-a962-42d8-b3f4-b4f4319dbe57}) and an assembly with the same simple name embedded in the native image 'full-composite.r2r.dll' (MVID = {45d68915-11ab-4e4d-bafc-abcfd64de8ec})

   at System.Runtime.Loader.AssemblyLoadContext.<LoadFromPath>g____PInvoke|5_0(IntPtr, UInt16*, UInt16*, System.Runtime.CompilerServices.ObjectHandleOnStack)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromPath(IntPtr, System.String, System.String, System.Runtime.CompilerServices.ObjectHandleOnStack)
   at System.Runtime.Loader.AssemblyLoadContext.InternalLoadFromPath(System.String, System.String)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(System.String)
   at System.Runtime.Loader.AssemblyLoadContext.ResolveUsingLoad(System.Reflection.AssemblyName)
   at System.Runtime.Loader.AssemblyLoadContext.Resolve(IntPtr, System.Reflection.AssemblyName)
   at NuGet.Packaging.Signing.PrimarySignature.Decode(Byte[])
   at NuGet.Packaging.Signing.PrimarySignature.Load(Byte[])
   at NuGet.Packaging.Signing.PrimarySignature.Load(System.IO.Stream)
   at NuGet.Packaging.PackageArchiveReader+<GetPrimarySignatureAsync>d__30.MoveNext()
...
```

ivdiazsa:
Sigh... At this point, we might need to either wait until the SDK coupling is fixed, or ship both images for the time being: Normal ASP.NET and Composite ASP.NET. This way, you could still use the normal ASP.NET to generate the SDK image because well... Who knows how much deeper this Pandora's Box will go?

trylek:
Hmm, do you think the problem is caused by the dotnet app? I was under the impression that the error is displayed when running the published powershell i.o.w. that the problem is in some non-standard assembly versions it might be using. I'm also trying to understand what exactly "powershell 7.4.0" refers to - the very design of composite images requires the runtime and ASP.NET to be updated and used in lockstep so if PowerShell uses custom versions of some framework assemblies, there's not much we can do.

mangod9:
yup feels like using asp.net composite in sdk is going to lead to trouble. SDK should probably continue to use non-composite if possible. Not sure what this change entails though. 

This particular exception looks to be Nuget related, wonder if it has its own version of that specific assembly it ships with.

lbussell:
Superseded by https://github.com/dotnet/dotnet-docker/pull/4590

Issue #4520, opened 2023-03-29
Publish non-root sample in MCR


richlander:
We're well into our productization of the non-root scenario. It would make it easier for users to test the new experience if we published a .NET 8 app in our samples repo as non-root. For Kubernetes in particular, that would be very useful.

We typically wait for RC1 to push new images. In a lot of ways, that's a nice to have. This time around, it seems warranted to do something earlier since we're changing the way container images work, not just the product they contain.

It isn't entirely clear what to do that makes sense long-term.

Here's a proposal:

- Push and document a new `non-root` tag (with the aspnetapp sample), starting now.
- Leave the existing tags as .NET 7-based.
- With .NET 8 RC1, move the existing samples to .NET 8 and make them non-root.
- Continuing pushing the `non-root` tag and make it an alias of `aspnetapp`.
- Undocument the `non-root` tag at that time and push it through RC2 and then no longer push it after that.
- At GA, we're left with just `dotnetapp` and `aspnetapp` tags.

Use this Dockerfile: https://github.com/dotnet/dotnet-docker/blob/main/samples/aspnetapp/Dockerfile.alpine-non-root

mthalman:
[Triage]

Let's introduce a new `non-root` tag to use for publishing this .NET 8-based image. At RC1 we will update the rest of the Dockerfiles to be based on .NET 8. And then at GA, we will delete the `non-root` tag.

richlander:
We are waiting until .NET 8 RC1.

PR #4542, opened 2023-04-04
Publish aspnetapp-non-root sample images

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/4520


mthalman:
Build error: 

` ---> System.InvalidOperationException: A value was not found for the ARG '$BUILDPLATFORM' in 'samples/aspnetapp/Dockerfile.alpine-non-root'`

Oh no! Our lovely Dockerfile isn't incompatible with our own infrastructure.

@richlander - Do you mind if we remove the usage of `$BUILDPLATFORM` and `$TARGETARCH` from Dockerfile.alpine-non-root? This pattern isn't applied consistently to the Dockerfiles anyway. We'd need to make changes to Image Builder in order to support the use of variable references in the `FROM` instruction.

richlander:
That's unfortunate. I referenced that Dockerfile in this post: https://devblogs.microsoft.com/dotnet/improving-multiplatform-container-support/. Should I make a new one that is multi-arch specific and then update the link?

mthalman:
The build error may be caused by the lack of `customBuildLegGroups` for the new entries in the manifest. I didn't think it was needed, but it might be the reason that the matrix generation is getting screwed up. I would try adding `customBuildLegGroups` to those entries, copying from the other ones, to see if that fixes things.

mthalman:
The change to the matrix type should not be necessary, nor it is desired. Did my [suggestion](https://github.com/dotnet/dotnet-docker/pull/4542#issuecomment-1498183722) above not work in your testing?

mthalman:
Build error says `failed to parse platform : "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument`

I think this is because BuildKit is not enabled by default on the build agents. I _think_ we can workaround this by adding the following to the top of the sample Dockerfile:

```Dockerfile
# syntax=docker/dockerfile:1
```

richlander:
That would be unfortunate to have to add. These images are intended to be best practice and that directive doesn't seem otherwise needed. Certainly, if we need that as a temporary measure, that's fine.

lbussell:
> That would be unfortunate to have to add. These images are intended to be best practice and that directive doesn't seem otherwise needed. Certainly, if we need that as a temporary measure, that's fine.

That's alright, it still doesn't work even with the directive.

mthalman:
I think this will require a change to Image Builder then to explicitly ensure BuildKit is enabled so that it executes commands like this: `DOCKER_BUILDKIT=1 docker build ...`

mthalman:
Still getting the [same error](https://github.com/dotnet/dotnet-docker/pull/4542#issuecomment-1499428268). I don't think the buildkit changes have flowed in yet. https://github.com/dotnet/docker-tools/pull/1129 is still open.

lbussell:
New ImageBuilder should come in with https://github.com/dotnet/dotnet-docker/pull/4598.

mthalman:
This is blocked after the need to revert the build kit change in https://github.com/dotnet/docker-tools/pull/1134. That's required in order to finish this work. Since that work is planned for any time soon, we'll just close this for now.

PR #4555, opened 2023-04-12
Split .NET Monitor 8 image into base and extensions

jander-msft:
This change splits the .NET Monitor 8 image into two separate images:
- A base image that contains the base installation of the .NET Monitor tool, environment variables, and entrypoint specification. The repository for this image is `dotnet/monitor/base`.
- A deriving image that takes the base image and installs the .NET Monitor extensions into the tool. Currently, the only extensions included are the Azure Blob storage and S3 storage extensions. The repository for this image is `dotnet/monitor`. It provides equivalent functionality compared to the prior 8.0 Preview images.

This split allows others to more easily produce a custom .NET Monitor image with a custom list of extensions installed on top of it without having to reimplement the entire image.

The image size baseline for the amd64 images were determined by building the images and inspecting their sizes. The arm64v8 image size baselines were calculated by using the relative difference between the amd64 images and applying those same offsets.

Other changes include replacing most of the template code that compares PARENT_REPO to `dotnet` and `nightly` with code that looks for these values at specific offsets. The problem with the prior implementation is that the new base image repository is `dotnet/nightly/monitor/base`, so its PARENT_REPO is `monitor` and its SHORT_REPO is `base`.

cc @dotnet/dotnet-monitor

### DIFF
--- a/eng/readme-templates/FeaturedTags.md
+++ b/eng/readme-templates/FeaturedTags.md
@@ -12,6 +12,8 @@ elif match(SHORT_REPO, "monitor"):* `7` (Standard Support)
   * `docker pull {{FULL_REPO}}:7`
 * `6` (Long-Term Support)
   * `docker pull {{FULL_REPO}}:6`^
+elif match(REPO, "monitor/base"):* `8` (Long-Term Support)
+  * `docker pull {{FULL_REPO}}:8`^
 else:* `7.0` (Standard Support)
   * `docker pull {{FULL_REPO}}:7.0`
 * `6.0` (Long-Term Support)
--- a/eng/readme-templates/FeaturedTags.md
+++ b/eng/readme-templates/FeaturedTags.md
@@ -1,6 +1,7 @@
 {{
     _ ARGS:
-      top-header: The string to use as the top-level header.
+      top-header: The string to use as the top-level header. ^
+    set isNightlyRepo to match(split(REPO, "/")[1], "nightly")
 }}{{ARGS["top-header"]}} Featured Tags
 
 {{if match(SHORT_REPO, "samples")
@@ -12,8 +13,8 @@ elif match(SHORT_REPO, "monitor"):* `7` (Standard Support)
   * `docker pull {{FULL_REPO}}:7`
 * `6` (Long-Term Support)
   * `docker pull {{FULL_REPO}}:6`^
-elif match(REPO, "monitor/base"):* `8` (Long-Term Support)
-  * `docker pull {{FULL_REPO}}:8`^
+elif match(REPO, "monitor/base") && isNightlyRepo:* `8` (Preview)
+  * `docker pull {{FULL_REPO}}:8-preview`^
 else:* `7.0` (Standard Support)
   * `docker pull {{FULL_REPO}}:7.0`
 * `6.0` (Long-Term Support)
--- a/eng/readme-templates/FeaturedTags.md
+++ b/eng/readme-templates/FeaturedTags.md
@@ -1,7 +1,6 @@
 {{
     _ ARGS:
-      top-header: The string to use as the top-level header. ^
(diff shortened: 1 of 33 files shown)


eng/readme-templates/FeaturedTags.md
```diff
@@ -12,6 +12,8 @@ elif match(SHORT_REPO, "monitor"):* `7` (Standard Support)
   * `docker pull {{FULL_REPO}}:7`
 * `6` (Long-Term Support)
   * `docker pull {{FULL_REPO}}:6`^
+elif match(REPO, "monitor/base"):* `8` (Long-Term Support)
+  * `docker pull {{FULL_REPO}}:8`^
```
mthalman, line 16: * The `8` tag doesn't exist yet. It's still `8-preview`.
* It should be labeled as `Preview` instead of `Long-Term Support`.
* It should be conditioned to output this only in the nightly branch. We don't want preview tags shown in the Featured section in main, not until we hit RC.
  * I've also logged https://github.com/dotnet/dotnet-docker/issues/4563

eng/readme-templates/Use.monitor-base.md
```diff
@@ -0,0 +1,12 @@
+{{
+    _ ARGS:
+      top-header: The string to use as the top-level header.
+      readme-host: Moniker of the site that will host the readme
+}}{{ARGS["top-header"]}}# Building a Custom .NET Monitor Image
+
+The following Dockerfiles demonstrate how you can use this base image to build a .NET Monitor image with a custom set of extensions.
+
+* [CBL Mariner Distroless - amd64](https://github.com/dotnet/dotnet-docker/blob/main/src/monitor/8.0/cbl-mariner-distroless/amd64/Dockerfile)
+* [CBL Mariner Distroless - arm64v8](https://github.com/dotnet/dotnet-docker/blob/main/src/monitor/8.0/cbl-mariner-distroless/arm64v8/Dockerfile)
```
mthalman, line 10: I would say we wouldn't include this in our readmes since the Mariner images aren't supported for public use.

src/monitor/8.0/cbl-mariner-distroless/amd64/Dockerfile
```diff
@@ -10,41 +10,23 @@ RUN tdnf install -y \
     && tdnf clean all
 
 # Retrieve .NET Monitor
```
mthalman, line 12: Use this instead?
```suggestion
# Retrieve .NET Monitor extensions
```

eng/readme-templates/FeaturedTags.md
jander-msft: I've fixed this (using preview and only including it for the nightly branch) only for the monitor/base image rather than removing it for now and getting it fixed in #4563 since it would otherwise not have any featured tags.

mthalman: Oh, I suppose when this gets merged to main for Preview 4, we'll still want to include `8-preview` in Featured Tags since that will be the only one available?

jander-msft: Up to you, I don't have a strong opinion on this. It might be odd that it doesn't have featured tags when merged over to `main` since this is the version first of that image, but it's also not strictly necessary. I think because the header `Featured Tags` would still exist even if there are no list tags and I don't want to over complicate that template for just this one image, I'll remove the `nightly` condition and let it be generated unconditionally. Sound good?

mthalman (approved):
Just as a sanity check, can you cherry pick these changes locally to the main branch and regen the readmes to make sure they look good in the context of the main branch?

jander-msft:
> Just as a sanity check, can you cherry pick these changes locally to the main branch and regen the readmes to make sure they look good in the context of the main branch?

Greate suggestion! It did find some differences, which I've fixed up in the latest changes.

mthalman (approved):

PR #4564, opened 2023-04-15
Use FileScoped namespaces and refactor tests.

jaliyaudagedara:
Use File Scoped namespaces
Proper use of `var`
Refactor Tests


samples/complexapp/complexapp/Program.cs
```diff
@@ -1,19 +1,17 @@
-﻿using System;
-using libbar;
+﻿using libbar;
 using libfoo;
 using static System.Console;
 
-namespace complexapp
+namespace complexapp;
```
mthalman, line 5: Even better than using file-scoped namespace would be to use top-level statements for Program.cs.

samples/complexapp/tests/StringUtilsTests.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace tests;
+
+public class StringUtilsTests
+{
+    [Fact]
+    public void ReverseString_WhenInputIsNull_ThrowsCorrectException()
+    {
+        string inputString = null;
+        Assert.Throws<ArgumentNullException>(() => StringUtils.ReverseString(inputString));
+    }
```
mthalman, line 14: The addition of this extra test is causing a test failure in the expected number of tests. These values need to be updated:

https://github.com/dotnet/dotnet-docker/blob/609446a3686940137628bbb67404f427294813ae/tests/Microsoft.DotNet.Docker.Tests/SampleImageTests.cs#L129-L130

samples/complexapp/complexapp/Program.cs
jaliyaudagedara: Of course, updated

samples/complexapp/tests/StringUtilsTests.cs
jaliyaudagedara: Updated

mthalman (approved):

PR #4571, opened 2023-04-19
Fix broken link

si618:


si618:
@dotnet-policy-service agree

si618:
@dotnet-policy-service agree

PR #4579, opened 2023-04-24
Fix url

chris-lerolle:
Hi, I was casually browsing the doc and found a small typo in that url. The ".md" extension is repeated.


MichaelSimons (approved):
Thanks for fixing the broken link!

PR #4590, opened 2023-05-03
[Continuation] - Add Dockerfiles for Linux .NET Composite Images

ivdiazsa:
This PR is a continuation of #4532. For the time being, we will be publishing two ASP.NET images: The normal one and the composite one. The reason we need the normal one as well is because it's part of the layering for the SDK image.


ivdiazsa:
cc @trylek @mangod9

jander-msft:
As the PR is, it looks like nothing is testing these images, so you may need to update the tests to add some ability to test them. Also, probably need to retarget to the `nightly` branch instead of the `main` branch.

If you add CBL Mariner distroless and/or Ubuntu Chiseled images, consider updating [monitor-base](https://github.com/dotnet/dotnet-docker/tree/nightly/src/monitor-base/8.0) to be layered on top of this. This will help demonstrate the readiness of the composite image with a real application sitting on top of it (that has a handful of tests in this repository) as well as provide a smoke test for updates to the composite layer.

ivdiazsa:
Thanks for the feedback @jander-msft. I will look into adding tests for these images. As for the `monitor-base` layer, I don't think that's necessary yet because we are currently only targeting Alpine.

src/aspnet-composite/8.0/alpine3.17/amd64/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+FROM $REPO:8.0.0-preview.3-alpine3.17-amd64
+
+# .NET globalization APIs will use invariant mode by default because DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=true is set
+# by the base runtime-deps image. See https://aka.ms/dotnet-globalization-alpine-containers for more information.
+
+# ASP.NET Core version
+ENV ASPNET_VERSION=8.0.0-preview.3.23177.8
+
+# Install ASP.NET Composite Runtime
+RUN wget -O dotnet.tar.gz https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$ASPNET_VERSION/aspnetcore-runtime-composite-$ASPNET_VERSION-linux-musl-x64.tar.gz \
```
lbussell, line 11: As @jander-msft said, we need to target the `nightly` branch, my apologies as I could have pointed that out earlier. Since this PR targets main, the URL is incorrect which is why we're getting a 404. The nightly URLs have a `dotnetbuilds` URL instead.

A simple rebase should be fine. If you get conflicts just regenerate the Dockerfiles instead of trying to resolve conflicts in the Dockerfiles. After it gets merged to `nightly` then it will get merged from `nightly` to `main` as part of our release process.

lbussell (changes requested):

src/aspnet-composite/8.0/alpine3.17/amd64/Dockerfile
lbussell: It also looks like in the rebase you accidentally picked up some extra commits from main.

ivdiazsa: Oh no, now we're at that mixed-commits-problem :/ Might as well close this PR and create a new one with another branch taken directly from `nightly` instead of `main`.

ivdiazsa:
Closing due to a lot of git garbage getting mixed in because of the rebase. Will open a cleaned-up version of the PR soon.

PR #4591, opened 2023-05-04
Update README.md

dascgit:
Removed "]" that was misplaced


dascgit:
@dotnet-policy-service agree

mthalman (approved):
Thanks for the fix!

PR #4593, opened 2023-05-04
Disable log init during non-root user creation

### LINKED ISSUE #4588: Adding non-root user in jammy issues results in image size increase
See https://github.com/dotnet/dotnet-docker/pull/4587. I investigated the difference between the two images:

- `mcr.microsoft.com/dotnet/nightly/runtime-deps@sha256:24cc8445fa87fb3771d40628f6745b3988f425dcd3d6fb80a2384305ae7c536e` (new)
- `mcr.microsoft.com/dotnet/nightly/runtime-deps@sha256:1dd42bff18ccd210bea44f481ee3b49db9cea509e7b7ff2a2e2b3e5aff3a1ac3` (old)

The culprit here is the `/var/log/lastlog` file. It has increased in size from 30kb to 19mb. `/var/log/faillog` has also increased in size from 3.3kb to 2.1mb. This requires some more investigation.

lbussell:
Fixes #4588

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.non-root-user
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.non-root-user
@@ -9,6 +9,7 @@
     set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set isAlpine to find(OS_VERSION, "alpine") >= 0 ^
     set isDebian to find(OS_ARCH_HYPHENATED, "Debian") >= 0 ^
+    set isUbuntu to find(OS_ARCH_HYPHENATED, "Ubuntu") >= 0 ^
     set isMariner to find(OS_VERSION, "cbl-mariner") >= 0 ^
     set isDistrolessMariner to defined(match(OS_VERSION, "^cbl-mariner\d+\.\d+-distroless$")) ^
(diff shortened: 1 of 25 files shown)


lbussell:
Couple notes:
Ubuntu has both `useradd` and `adduser`. So I switched it to `useradd` because it's more configurable.
Mariner appears to have both commands but actually just aliases `adduser` to `useradd`. I switched that to useradd so it's more clear what's going on.

mthalman (approved):
Looks good. I'll leave it to you to determine when to merge this. If it's merged before the 8.0 Preview 4 release, it will require additional changes when merging to main because we'd need to avoid making any changes to the 6.0/7.0 Dockerfiles. Since this change doesn't provide any functional benefits, I think it's ok to wait until after the Preview 4 release so that it can be applied together across all versions in June.

lbussell:
I have also done an internal test run to evaluate the image size changes: https://dev.azure.com/dnceng/internal/_build/results?buildId=2174607&view=results

lbussell:
I confirmed that this reduced the `dotnet/runtime-deps:8.0-preview-jammy` image size from 137MB -> 107MB.

lbussell:
Merging this now since 8.0 preview 4 is now out. We can this in for the June releases.

PR #4594, opened 2023-05-05
[Fixed] - Add Dockerfiles for Linux .NET Composite Images

ivdiazsa:
This PR follows up #4590, which got seriously messed up when rebasing.


ivdiazsa:
cc @trylek @mangod9 @jander-msft

README.aspnet.md
```diff
@@ -71,6 +71,7 @@ Tags | Dockerfile | OS Version
 -----------| -------------| -------------
 8.0.0-preview.4-bookworm-slim-amd64, 8.0-preview-bookworm-slim-amd64, 8.0.0-preview.4-bookworm-slim, 8.0-preview-bookworm-slim, 8.0.0-preview.4, 8.0-preview, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/8.0/bookworm-slim/amd64/Dockerfile) | Debian 12
 8.0.0-preview.4-alpine3.17-amd64, 8.0-preview-alpine3.17-amd64, 8.0-preview-alpine-amd64, 8.0.0-preview.4-alpine3.17, 8.0-preview-alpine3.17, 8.0-preview-alpine | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet/8.0/alpine3.17/amd64/Dockerfile) | Alpine 3.17
+8.0.0-preview.4-alpine3.17-amd64-composite, 8.0-preview-alpine3.17-amd64-composite, 8.0-preview-alpine-amd64-composite, 8.0.0-preview.4-alpine3.17, 8.0-preview-alpine3.17, 8.0-preview-alpine | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspnet-composite/8.0/alpine3.17/amd64/Dockerfile) | Alpine 3.17
```
MichaelSimons, line 74: I think `composite` should appear immediately after the distro - `8.0.0-preview.4-alpine3.17-composite-amd64`

This is consistent with the other image variants like `mariner2.0-distroless`, `jammy-chiseled`, and `debian-slim`

src/aspnet-composite/8.0/alpine3.17/amd64/Dockerfile
MichaelSimons, line 1: The folder structure should match the tagging pattern.  To follow my earlier comment regarding the tagging, this should be in a src/aspnet/8.0/alpine3.17-composite/amd64/ folder structure.

src/aspnet-composite/8.0/alpine3.17/amd64/Dockerfile
```diff
@@ -0,0 +1,18 @@
+ARG REPO=mcr.microsoft.com/dotnet/runtime-deps
+FROM $REPO:8.0.0-preview.4-alpine3.17-amd64
+
+# .NET globalization APIs will use invariant mode by default because DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=true is set
+# by the base runtime-deps image. See https://aka.ms/dotnet-globalization-alpine-containers for more information.
+
+# ASP.NET Core version
+ENV ASPNET_VERSION=8.0.0-preview.4.23251.12
```
MichaelSimons, line 8: In order to provide parity with the non-composite images, I wonder if this should also define DOTNET_VERSION.  This is necessary because of being based on runtime-deps versus runtime.

samples/aspnetapp/aspnetapp/NuGet.config
MichaelSimons, line 1: What precipitates the need for this?

MichaelSimons (commented):
Please work with @lbussell to define the appropriate tests.  Also the size baseline will need updating to include these new image variants.

samples/aspnetapp/aspnetapp/NuGet.config
lbussell: I think this is leftover from my PRs where I was using this sample to test the ASP.NET composite images. If we want to test an 8.0 sample on the composite runtime then we'd need to have this somewhere. I think there would be ways to test without that either though.

lbussell:
New imagebuilder version should come in with https://github.com/dotnet/dotnet-docker/pull/4598

eng/dockerfile-templates/runtime/Dockerfile.linux.install-runtime
```diff
@@ -19,15 +20,30 @@
...
-        "/Runtime/", runtimeVersionDir, "/") ^
+    set aspnetVersionFile to when(isInternalStableBranding,
+        VARIABLES[cat("dotnet|", dotnetVersion, "|product-version")],
+        aspnetVersionDir) ^
+    set fileArch to when(ARGS["is-rpm-install"] && ARCH_SHORT = "arm64", "aarch64", ARCH_SHORT) ^
+    set fileExt to when(ARGS["is-rpm-install"], "rpm", "tar.gz") ^
+    set filePlatform to when(ARGS["is-rpm-install"], "", when(isAlpine, "-linux-musl", "-linux")) ^
+    set baseUrl to VARIABLES[cat("base-url|", dotnetVersion, "|", VARIABLES["branch"])] ^
+    set runtimeBaseUrl to cat(baseUrl, "/Runtime/", runtimeVersionDir, "/") ^
+    set aspnetUrl to cat(baseUrl, "/aspnetcore/Runtime/", aspnetVersionDir, "/aspnetcore-runtime-composite-", aspnetVersionFile,
```
mthalman, line 42: Name this `aspnetCompositeUrl` instead of `aspnetUrl` to avoid confusion.

eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -86,6 +86,7 @@ public class DockerfileShaUpdater : FileRegexUpdater
                 { "runtime-deps-cm.2", new string[] { $"$DOTNET_BASE_URL/Runtime/$VERSION_DIR/dotnet-runtime-deps-$VERSION_FILE-cm.2-{GetRpmArchFormat()}.$ARCHIVE_EXT" } },
 
                 { "aspnet", new string[] { $"$DOTNET_BASE_URL/aspnetcore/Runtime/$VERSION_DIR/aspnetcore-runtime-$VERSION_FILE$OPTIONAL_OS-{GetRuntimeSdkArchFormat()}.$ARCHIVE_EXT" } },
+                { "aspnet", new string[] { $"$DOTNET_BASE_URL/aspnetcore/Runtime/$VERSION_DIR/aspnetcore-runtime-$COMPOSITE_SUFFIX$VERSION_FILE$OPTIONAL_OS-{GetRuntimeSdkArchFormat()}.$ARCHIVE_EXT" } },
```
mthalman, line 89: This isn't going to run. It's introducing a duplicate key, `aspnet`, into the dictionary. I would suggest naming it `aspnet-composite` instead. This variable name change then needs to be reflected in the spot where it's referenced at https://github.com/dotnet/dotnet-docker/pull/4594/files#diff-9ae9ccb196a27239e9c7f5c7cb97edec392356c2ac3451cb6e6582cee2455b94R73.

There also needs to be corresponding entries in the manifest.versions.json file at the root of the repo to define these SHA variables and their values. That's the file which defines the SHAs and are then injected into the Dockerfile.

This is the reason the build is failing with checksum mismatches. The Dockerfile template is referencing the `aspnet` prefix variable from manifest.versions.json file which is giving you the SHA of the non-composite aspnet tarball.

You should be able to run eng\Set-DotnetVersions.ps1 which will regenerate all Dockerfiles to reflect the checksums that are defined in manifest.versions.json.

samples/aspnetapp/aspnetapp/aspnetapp.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
mthalman, line 4: We don't want samples updated to 8.0 yet.

ivdiazsa:
@mthalman Thanks for your feedback Matt. I will get to work on it now.

ivdiazsa:
@mthalman Thanks for your feedback Matt. I will get to work on it now.

tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Docker.Tests;
+
+[Trait("Category", "aspnet-composite")]
+public class AspnetCompositeImageTests : CommonRuntimeImageTests
```
mthalman, line 14: It looks like, in general, this class has a very similar implementation to the existing `AspnetImageTests` class. I would suggest refactoring such that there is a common base class between them that implements the majority of these tests. It looks like much of the differences here are just conditional checks for whether the test should be skipped. That can be implemented via an abstract method in the base which is overridden appropriately in the derived classes.

README.aspnet.md
mthalman, line 1: We should really have some sort of documentation that describes the purpose of these composite images (i.e. what are they? how do they differ from the regular aspnet images? how do I determine which one to use?). If necessary, this can link to another doc page that describes things in more detail.

Note that the Docker repository readmes are generated from templates. For example, the `About` section of README.aspnet.md comes from https://github.com/dotnet/dotnet-docker/blob/nightly/eng/readme-templates/About.aspnet.md. Readmes can be regenerated from templates by running `eng\readme-templates\Get-GeneratedReadmes.ps1`

trylek: @mthalman - we have a technical description of the composite format here:

https://github.com/dotnet/runtime/blob/main/docs/design/features/readytorun-composite-format-design.md

In terms of our current shipping story the composite container is a tool a customer can use to reduce container size on disk while keeping performance of default R2R publishing with the caveat that the composite container has tighter version coupling - in particular, the final app is not allowed to use handpicked custom versions of framework / ASP.NET assemblies that are baked into the composite image. @ivdiazsa - please let me know if you need further help with the actual wording.

ivdiazsa:
In my latest commit, I added a base class for the common functionality between normal ASP.NET and composite ASP.NET tests, as suggested by @mthalman. However, I can't get it to work. I keep getting an error saying that those methods are not supported. Does anyone know why that's happening?

I pushed the changes to "save my progress".

mthalman:
> In my latest commit, I added a base class for the common functionality between normal ASP.NET and composite ASP.NET tests, as suggested by @mthalman. However, I can't get it to work. I keep getting an error saying that those methods are not supported. Does anyone know why that's happening?
> 
> I pushed the changes to "save my progress" and show easily what I added and modified.

We have no visibility into these errors because it's not even getting to that point in the pipeline. It's failing on the prebuild validation which ensures that both the Dockerfiles and readmes are up-to-date with respect to the templates. It says the Dockerfiles are not in sync. So you'll need to regenerate them with `.\eng\dockerfile-templates\Get-GeneratedDockerfiles.ps1`

ivdiazsa:
> We have no visibility into these errors because it's not even getting to that point in the pipeline. It's failing on the prebuild validation which ensures that both the Dockerfiles and readmes are up-to-date with respect to the templates. It says the Dockerfiles are not in sync. So you'll need to regenerate them with `.\eng\dockerfile-templates\Get-GeneratedDockerfiles.ps1`

I don't think that's the problem. I just ran `Get-GeneratedDockerfiles.ps1` and `Get-GeneratedReadmes.ps1` again, and there are no changes to commit. Tomas and I did some looking into and it's some problem with Xunit but we haven't devised a definitive fix yet.

MichaelSimons:
> I don't know why the CI is failing with that. Running those scripts doesn't change anything in my machine.

It is failing because CI always merges in the latest changes from the target branch before running.  The tip of nightly is on a newer .NET version therefore the new Dockerfiles you are adding are out of date.  Sync your local branch with the tip of nightly and regen the Dockerfiles.

mthalman:
The issue with the abstract class and xUnit can be solved without reverting the refactoring. The problem was that the test methods in `CommonAspnetImageTests` were referencing a method named `GetImageData`. Before the refactoring, that method was defined here:

https://github.com/dotnet/dotnet-docker/blob/df920782fdedd2d4b3bc74c3c7a3fe9a2ba76a07/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs#L23

But as a result of moving the test methods into a base class, they're now referencing an _inherited_ static method with the same name but takes a parameter:

https://github.com/dotnet/dotnet-docker/blob/df920782fdedd2d4b3bc74c3c7a3fe9a2ba76a07/tests/Microsoft.DotNet.Docker.Tests/CommonRuntimeImageTests.cs#L20

xUnit can't bind to that properly because no argument is provided from the `MemberData` attribute and it throws a very unhelpful exception.

This can be fixed by changing those implemented methods in `CommonAspnetImageTests` to be abstract and implementing them with the adorned attributes in the two derived classes. This will then bind to the locally defined `GetImageData` method.

In addition, my expectation was that the `CommonAspnetImageTests` class would have a common implementation of the test methods wherever possible. For example: [AspnetImageTests.cs/VerifyAppScenario](https://github.com/ivdiazsa/dotnet-docker/blob/8b6d6f11d9b96ed2a9b57907034b0faee4fc48c7/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs#L27-L38) and [AspnetCompositeImageTests.cs/VerifyAppScenario](https://github.com/ivdiazsa/dotnet-docker/blob/8b6d6f11d9b96ed2a9b57907034b0faee4fc48c7/tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs#L27-L42) only differ in the `if` block. The common implementation of that method should be moved to the base class and the `if` check factored out into an abstract method. The same applies to the other test methods between `AspnetImageTests` and `AspnetCompositeImageTests`.

trylek:
@mthalman - I have originally advised Ivan to get rid of the class hierarchy exactly because xUnit can bind neither <code>GetImageData</code>, nor <code>ImageType</code>, to a non-static method and all the tests are tagged as <code>Theory</code> with <code>MemberData</code>. Are you suggested that all these tests be duplicated in the final classes (<code>AspnetImageTests</code> vs. <code>AspnetCompositeImageTests</code>) as just wrappers calling into <code>CommonAspnetImageTests</code> in the vast majority of cases? My original impression was that it would be easier to just add the composite image as one more test case produced by <code>GetImageData</code> - after all, most of the test stack should be oblivious of the distinction.

mthalman:
> @mthalman - I have originally advised Ivan to get rid of the class hierarchy exactly because xUnit can bind neither `GetImageData`, nor `ImageType`, to a non-static method and all the tests are tagged as `Theory` with `MemberData`. Are you suggesting that all these tests be duplicated in the final classes (`AspnetImageTests` vs. `AspnetCompositeImageTests`) as just wrappers calling into `CommonAspnetImageTests` in the vast majority of cases? My original impression was that it would be easier to just add the composite image as one more test case produced by `GetImageData` - after all, most of the test stack should be oblivious of the distinction.

I just want there to be an elimination of duplication between `AspnetImageTests` and `AspnetCompositeImageTests`. For example, these two test methods have very similar implementations:

* https://github.com/ivdiazsa/dotnet-docker/blob/a981e7409c10c042958652508c94910c9b2cdf3a/tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs#L46-L63
* https://github.com/ivdiazsa/dotnet-docker/blob/a981e7409c10c042958652508c94910c9b2cdf3a/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs#L42-L56

My suggestion for accomplishing that was to have a base class with a common implementation. This pattern is already used for common test logic between the aspnet and runtime image types. For example, the derived test class implements a wrapper method necessary to satisfy xUnit:

https://github.com/dotnet/dotnet-docker/blob/df920782fdedd2d4b3bc74c3c7a3fe9a2ba76a07/tests/Microsoft.DotNet.Docker.Tests/AspnetImageTests.cs#L87C1-L92

While the base class provides the implementation:

https://github.com/dotnet/dotnet-docker/blob/df920782fdedd2d4b3bc74c3c7a3fe9a2ba76a07/tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs#L29-L35

Making use of inheritance to provide this common implementation is especially useful if there is any slight variation needed between derived classes as this can be handled via an abstract method implementation.

trylek:
Thanks Matt for clarifying, that sounds reasonable to me. @ivdiazsa, please let me know if you need help in addressing Matt's feedback and apologies for originally having advised you to remove the class hierarchy, I was not aware of the rationale Matt just explained.

mthalman:
Looks like there are checksum mismatches that need to be resolved.

tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public static EnvironmentVariableInfo GetAspnetCompositeVersionVariableInfo(
+                ProductImageData imageData,
+                DockerHelper dockerHelper)
+        {
+            string version = imageData.GetProductVersion(DotNetImageType.Aspnet_Composite,
+                                                         dockerHelper);
+
+            return new EnvironmentVariableInfo("ASPNET_VERSION", version)
```
MichaelSimons, line 93: This should validate the DOTNET_VERSION env as well.

tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            EnvironmentVariableInfo compositeVersionVariableInfo = GetAspnetCompositeVersionVariableInfo(
+                    imageData,
+                    DockerHelper);
+
+            base.VerifyAspnetEnvironmentVariables(imageData, compositeVersionVariableInfo);
+        }
+
+        [DotNetTheory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifyPackageInstallation(ProductImageData imageData)
```
MichaelSimons, line 52: This test doesn't seem necessary/correct.  RPM packages are only used in .NET 6.0 mariner images.  Because of this is the base method should be removed and restore the AspnetImageTests back to its original form for this single test.

tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -18,8 +18,8 @@ public class DockerHelper
     {
         public static string DockerOS => GetDockerOS();
         public static string DockerArchitecture => GetDockerArch();
-        public static string ContainerWorkDir => IsLinuxContainerModeEnabled ? "/sandbox" : "c:\\sandbox";
         public static bool IsLinuxContainerModeEnabled => string.Equals(DockerOS, "linux", StringComparison.OrdinalIgnoreCase);
+        public static string ContainerWorkDir => IsLinuxContainerModeEnabled ? "/sandbox" : "c:\\sandbox";
```
MichaelSimons, line 22: Curious why this change was made/is necessary.  Consider reverting.

tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        [DotNetTheory]
+        [MemberData(nameof(GetImageData))]
+        public override void VerifyDefaultUser(ProductImageData imageData)
+        {
+            base.VerifyCommonDefaultUser(imageData);
+        }
+
+        public static EnvironmentVariableInfo GetAspnetCompositeVersionVariableInfo(
```
MichaelSimons, line 86: Consider moving this to the common aspnet base class to share the common variables.

samples/aspnetapp/aspnetapp/NuGet.config
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
MichaelSimons, line 1: Why is this change necessary?

tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        protected override DotNetImageType ImageType => DotNetImageType.Aspnet_Composite;
+
+        public static IEnumerable<object[]> GetImageData() => GetImageData(DotNetImageType.Aspnet_Composite);
+
+        [DotNetTheory]
+        [MemberData(nameof(GetImageData))]
+        public async Task VerifyAppScenario(ProductImageData imageData)
+        {
+            if (imageData.OSTag != OS.Alpine317Composite)
```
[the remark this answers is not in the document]
ivdiazsa: Because for composites we only support Alpine. So there's no point of trying to run further if it's not Alpine.

samples/aspnetapp/aspnetapp/NuGet.config
ivdiazsa: This was added by Logan in his original PR. I wasn't sure whether we need this file so I just let it be. If we don't need it, I can remove it.

tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs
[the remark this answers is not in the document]
MichaelSimons: 1. I would expect the image data would work in such a way we are only invoking these test methods for images that exist.
2. It smells that we are only checking this condition in this test case and not the others.

tests/Microsoft.DotNet.Docker.Tests/AspnetCompositeImageTests.cs
ivdiazsa: Isn't that already validated? The `VerifyAspnetEnvironmentVariables()` method in the common class has this code:

```csharp
        List<EnvironmentVariableInfo> variables = new List<EnvironmentVariableInfo>
        {
            RuntimeImageTests.GetRuntimeVersionVariableInfo(imageData, DockerHelper)
        };

        if (aspnetVersionVariableInfo != null)
        {
            variables.Add(aspnetVersionVariableInfo);
        }

        base.VerifyCommonEnvironmentVariables(imageData, variables);
```

We're getting `DOTNET_VERSION` from `RuntimeImageTests.GetRuntimeVersionVariableInfo(imageData, DockerHelper)`, which is then added to the list that is passed to `base.VerifyCommonEnvironmentVariables(imageData, variables);`.

For clarity, `GetRuntimeVersionVariableInfo()` has this code, which is basically what we're doing in `VerifyAspnetEnvironmentVariables()`:

```csharp
            string version = imageData.GetProductVersion(DotNetImageType.Runtime, dockerHelper);
            return new EnvironmentVariableInfo("DOTNET_VERSION", version)
            {
                IsProductVersion = true
            };
```

MichaelSimons: Yes, you are right.

samples/aspnetapp/aspnetapp/NuGet.config
lbussell: This should no longer be necessary. It was for the test script that I had set up. The AspnetCompositeImageTests should make that no longer necessary.

ivdiazsa:
I need help with these checksums. I sync'd my branch with nightly today and ran the scripts, but there are no changes to be committed. The scripts I tried are:
- Get-GeneratedDockerfiles
- Get-GeneratedReadmes
- Set-DotnetVersions

ivdiazsa:
This last failure was a Windows test that timed out, which I assume is unrelated to my changes. Could someone re-run that pipeline? I don't have the appropriate permissions.

ivdiazsa:
It's now ready to be merged! Unless anyone has further issues with the PR.

ivdiazsa:
> > It's now ready to be merged! Unless anyone has further issues with the PR.
> 
> I think this comment should be resolved before merging - https://github.com/dotnet/dotnet-docker/pull/4594/files#r1189899424
> 
> We need to help the users of our aspnet images understand the differences of the standard and composite images and when to use one versus the other. I would expect the in depth details about composites would be provided by a link to someplace else.
> 
> This documentation can certainly be evolved over time with other check-ins but we should have some documentation to start with for the preview releases.

Where should I add such documentation? `dotnet-docker/documentation`?

MichaelSimons:
To the [aspnet readme](https://github.com/dotnet/dotnet-docker/blob/main/README.aspnet.md).  This is the readme that gets syndicated to Docker Hub and MAR.

MichaelSimons:
> Where should I add such documentation? `dotnet-docker/documentation`?

MichaelSimons:
Sorry for accidentally closing.

ivdiazsa:
Composite doc is up

trylek (approved):
Thanks Ivan for driving this to the finish line!

eng/readme-templates/Use.aspnet.md
```diff
@@ -15,3 +15,9 @@ docker run -it --rm -p 8000:80 --name aspnetcore_sample mcr.microsoft.com/dotnet
 After the application starts, navigate to `http://localhost:8000` in your web browser.
 
 See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotnet/dotnet-docker/blob/main/samples/host-aspnetcore-https.md) to use HTTPS with this image.
+
+{{ARGS["top-header"]}}# Composite images containers
```
mthalman, line 19:
```suggestion
{{ARGS["top-header"]}}# Composite container images
```

eng/readme-templates/Use.aspnet.md
```diff
@@ -15,3 +15,9 @@ docker run -it --rm -p 8000:80 --name aspnetcore_sample mcr.microsoft.com/dotnet
...
+
+Starting from .NET 8, a composite version of the ASP.NET images is being offered alongside the regular image. The main characteristics of these images are their smaller size on disk while keeping the performance of the default R2R setting. The caveat is that the composite images have tighter version coupling. This means the final app run on them cannot use handpicked custom versions of the framework and/or ASP.NET assemblies that are in-built into the composite binary.
```
mthalman, line 21:
```suggestion
Starting from .NET 8, a composite version of the ASP.NET images is being offered alongside the regular image. The main characteristics of these images are their smaller size on disk while keeping the performance of the default [ReadyToRun (R2R) setting](https://learn.microsoft.com/dotnet/core/deploying/ready-to-run). The caveat is that the composite images have tighter version coupling. This means the final app run on them cannot use handpicked custom versions of the framework and/or ASP.NET assemblies that are built into the composite binary.
```

src/aspnet/8.0/alpine3.17-composite/amd64/Dockerfile
```diff
@@ -0,0 +1,20 @@
+ARG REPO=mcr.microsoft.com/dotnet/runtime-deps
+FROM $REPO:8.0.0-preview.5-alpine3.17-amd64
+
+# .NET globalization APIs will use invariant mode by default because DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=true is set
+# by the base runtime-deps image. See https://aka.ms/dotnet-globalization-alpine-containers for more information.
+
+# .NET Runtime version
+ENV DOTNET_VERSION=8.0.0-preview.5.23260.3
+# ASP.NET Core version
+ENV ASPNET_VERSION=8.0.0-preview.5.23262.1
```
mthalman, line 10: I'm fine with getting these changes in now. But we should have a follow-up issue to clean this up. We should have just one `ENV` instruction:

```Dockerfile

ENV \
  # .NET Runtime version
  DOTNET_VERSION=8.0.0-preview.5.23260.3 \
  # ASP.NET Core version
  ASPNET_VERSION=8.0.0-preview.5.23262.1
```

That's going to require some refactoring of the templates.

README.aspnet.md
```diff
@@ -37,6 +37,12 @@ After the application starts, navigate to `http://localhost:8000` in your web br
 
 See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotnet/dotnet-docker/blob/main/samples/host-aspnetcore-https.md) to use HTTPS with this image.
 
+## Composite images containers
```
MichaelSimons, line 40: To me, this section does not fit well under `Usage`.  The Usage section is intended to contain instructions on how to properly execute and run the artifact.  IMO, this new section better fits in the about section to get the appropriate attention it deserves.

README.aspnet.md
```diff
@@ -37,6 +37,12 @@ After the application starts, navigate to `http://localhost:8000` in your web br
...
+
+Starting from .NET 8, a composite version of the ASP.NET images is being offered alongside the regular image. The main characteristics of these images are their smaller size on disk while keeping the performance of the default R2R setting. The caveat is that the composite images have tighter version coupling. This means the final app run on them cannot use handpicked custom versions of the framework and/or ASP.NET assemblies that are in-built into the composite binary.
```
MichaelSimons, line 42:
```suggestion
Starting from .NET 8, a composite version of the ASP.NET images, denoted with the `-composite` tag part,  is being offered alongside the regular image. The main characteristics of these images are their smaller size on disk while keeping the performance of the default R2R setting. The caveat is that the composite images have tighter version coupling. This means the final app run on them cannot use handpicked custom versions of the framework and/or ASP.NET assemblies that are in-built into the composite binary.
```

src/aspnet/8.0/alpine3.17-composite/amd64/Dockerfile
ivdiazsa: > 

#4632

README.aspnet.md
```diff
@@ -17,6 +17,12 @@
 
 This image contains the ASP.NET Core and .NET runtimes and libraries and is optimized for running ASP.NET Core apps in production.
 
+# Composite container images
```
MichaelSimons, line 20:
```suggestion
## Composite container images
```

.mar/portal/README.aspnet.portal.md
```diff
@@ -8,6 +8,12 @@
 
 This image contains the ASP.NET Core and .NET runtimes and libraries and is optimized for running ASP.NET Core apps in production.
 
+# Composite container images
```
MichaelSimons, line 11: This new section should be a subsection of About.

```suggestion
### Composite container images
```

ivdiazsa: Ok I'll fix it right now but please comment down all the feedback you have. We've been delaying a lot for non-blocking changes and this back-and-forth for nits is not helping anyone.

eng/readme-templates/About.aspnet.md
```diff
@@ -1,7 +1,3 @@
 This image contains the ASP.NET Core and .NET runtimes and libraries and is optimized for running ASP.NET Core apps in production.
 
-{{ARGS["top-header"]}}# Composite container images
-
-Starting from .NET 8, a composite version of the ASP.NET images, `denoted with the -composite` tag part, is being offered alongside the regular image. The main characteristics of these images are their smaller size on disk while keeping the performance of the default [ReadyToRun (R2R) setting](https://learn.microsoft.com/dotnet/core/deploying/ready-to-run). The caveat is that the composite images have tighter version coupling. This means the final app run on them cannot use handpicked custom versions of the framework and/or ASP.NET assemblies that are built into the composite binary.
-
-For a full technical description on how the composites work, we have a [feature doc here](https://github.com/dotnet/runtime/blob/main/docs/design/features/readytorun-composite-format-design.md).
+{{InsertTemplate("About.composite-aspnet.md")}}
```
mthalman, line 3: This needs to flow the `top-header` arg into the sub-template.

```suggestion
{{InsertTemplate("About.composite-aspnet.md", [ "top-header": ARGS["top-header"] ])}}
```

eng/readme-templates/About.aspnet.md
```diff
@@ -1,3 +1,3 @@
 This image contains the ASP.NET Core and .NET runtimes and libraries and is optimized for running ASP.NET Core apps in production.
 
-{{InsertTemplate("About.composite-aspnet.md")}}
+{{InsertTemplate("About.composite-aspnet.md", [ "top-header": ARGS["top-header"] ])}}
```
mthalman, line 3: Ah, no wonder why this didn't produce the expected generated readme. We're missing another piece.

Please update this:
https://github.com/dotnet/dotnet-docker/blob/be061376d55ee2d3c312f2a9a5c8642d0fc3307b/eng/readme-templates/About.md?plain=1#L14

to this:

```
{{InsertTemplate(join(["About", templateQualifier, "md"], "."), [ "top-header": ARGS["top-header"] ])}}
```

That should produce the correct headers after regenerating the readmes. If it doesn't, let me know.

ivdiazsa: Done

mthalman (approved):
Looks good. Thanks for all the work you put into this.

eng/dockerfile-templates/aspnet/Dockerfile.linux-composite
```diff
@@ -0,0 +1,98 @@
...
+    set osVersionBase to when(isDistroless, match(OS_VERSION, ".+(?=.*-)")[0], OS_VERSION_BASE) ^
+    set installerImageTag to when(isDistrolessMariner,
+        cat("mcr.microsoft.com/cbl-mariner/base/core:", OS_VERSION_NUMBER),
+        when(isAlpine || isFullMariner,
+            runtimeDepsBaseTag,
+            cat(ARCH_VERSIONED, "/buildpack-deps:", osVersionBase, "-curl")))
+}}ARG REPO=mcr.microsoft.com/dotnet/runtime-deps{{ if isSingleStage:
+{{
+
+    _ SINGLE STAGE
```
lbussell, line 23: I would just like to note that there are unused options in this file but we should leave them for if we decide to ship non-musl variants of composite images in the future.

lbussell (approved):

mangod9:
Thanks everyone for getting this change merged! Assume this will now push the new images, or is there more work required for these composite images to showup on dockerhub?

mangod9:
Ah nice! Thanks for the info.

lbussell:
The images are now available. I filed https://github.com/dotnet/dotnet-docker/issues/4634 for a minor issue with the manifest.json/readme.

mangod9:
Yeah I see them. Do the tags look ok? Seems like we need to add composite to the last two tags? 

<html>
<body>


8.0.0-preview.5-alpine3.17-arm64v8, 8.0-preview-alpine3.17-arm64v8, 8.0-preview-alpine-arm64v8, 8.0.0-preview.5-alpine3.17, 8.0-preview-alpine3.17, 8.0-preview-alpine | Dockerfile | Alpine 3.17 | 05/18/2023
-- | -- | -- | --
8.0.0-preview.5-alpine3.17-composite-arm64v8, 8.0-preview-alpine3.17-composite-arm64v8, 8.0-preview-alpine-composite-arm64v8, 8.0.0-preview.5-alpine3.17, 8.0-preview-alpine3.17, 8.0-preview-alpine | Dockerfile | Alpine 3.17 | 05/23/2023


</body>
</html>

lbussell:
Opened #4635

Issue #4544, opened 2023-04-05
Not all sample images are built and tested in CI/PR validation


lbussell:
https://github.com/dotnet/dotnet-docker/pull/4534 restructured some of the sample dockerfiles, but left some of them, like [Dockerfile.debian](https://github.com/richlander/dotnet-docker/blob/654dbd52d157901292a22f20bf34006ff0a27bb6/samples/aspnetapp/Dockerfile.debian) unable to build and/or run.

The [dotnet-docker-samples](https://dev.azure.com/dnceng/internal/_build?definitionId=376) pipeline only ends up building images that we publish (defined in manifest.samples.json). There should be some additional PR validation that runs `pwsh .\build-and-test.ps1 -Path *samples* -TestCategories sample`.

mthalman:
This was tested in CI at one point. I believe it inadvertently stopped as a result of the changes from https://github.com/dotnet/dotnet-docker/pull/4207.

PR #4615, opened 2023-05-12
Fix exec: "./aspnetapp.dll": permission denied

natict:
With only "./aspnetapp.dll" as the entrypoint the resulting container image fails to run on arm64 Linux (most likely also on amd64) with:
```
exec: "./aspnetapp.dll": permission denied: unknown.
```


natict:
@dotnet-policy-service agree

natict:
Changed the fix to match other Dockerfiles in samples directory

mthalman:
Note that this wasn't caught by CI because of https://github.com/dotnet/dotnet-docker/issues/4544.

natict:
@mthalman resolved the additional Dockerfiles and rebased the patch

mthalman (approved):

PR #4616, opened 2023-05-15
Optimize layers in runtime-deps images

### LINKED ISSUE #4419: Optimize layers in runtime-deps images
In runtime-deps images, we set the environment variables after we install dependencies and other instructions. Since the environment variables basically never change, we should be able to put the `ENV` instruction before the dependency installation to take advantage of layer caching.

Also, in some runtime-deps Dockerfiles with a non-root user defined, it is defined after the deps install. We should be able to make the same optimization with these instructions too.

Example Dockerfile with potential for both optimizations: https://github.com/dotnet/dotnet-docker/blob/nightly/src/runtime-deps/7.0/cbl-mariner2.0-distroless/amd64/Dockerfile

lbussell:
Fixes #4419

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
@@ -10,7 +10,9 @@
         no-create-home (optional): Indicates whether a home directory should be created for the user ^
     set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set isMariner to find(OS_VERSION, "cbl-mariner") >= 0 ^
-    set isAlpine to find(OS_VERSION, "alpine") >= 0
+    set isAlpine to find(OS_VERSION, "alpine") >= 0 ^
+    set uid to when(len(ARGS["uid"]) > 0, ARGS["uid"], "$APP_UID") ^
+    set gid to when(len(ARGS["gid"]) > 0, ARGS["gid"], "$APP_UID")
 }}{{InsertTemplate("Dockerfile.linux.non-root-user",
 [
     "name": ARGS["name"],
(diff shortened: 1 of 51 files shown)


lbussell:
@mthalman I haven't looked very closely, but [the error here](https://dev.azure.com/dnceng-public/public/_build/results?buildId=274547&view=logs&j=c977949b-2bac-537a-607c-ccfd6201d486&t=15285f4f-8cbb-599d-bbac-1b70b321dc8e&l=638) sure looks a lot like https://github.com/dotnet/docker-tools/pull/1132#issuecomment-1538760029...

eng/dockerfile-templates/runtime-deps/Dockerfile.linux.distroless-user
```diff
@@ -10,7 +10,9 @@
         no-create-home (optional): Indicates whether a home directory should be created for the user ^
     set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set isMariner to find(OS_VERSION, "cbl-mariner") >= 0 ^
-    set isAlpine to find(OS_VERSION, "alpine") >= 0
+    set isAlpine to find(OS_VERSION, "alpine") >= 0 ^
+    set uid to when(len(ARGS["uid"]) > 0, ARGS["uid"], "$APP_UID") ^
+    set gid to when(len(ARGS["gid"]) > 0, ARGS["gid"], "$APP_UID")
```
mthalman, line 15: It looks like `uid` and `gid` are optional now. The comments for the those two args should indicate that they're optional like you did in Dockerfile.linux.non-root-user.

mthalman (approved):
Looks great!

lbussell:
Same issue still even after rebase. I don't think the dotnet-docker repo ever got the broken ImageBuilder.

lbussell:
Was rebasing this and took the opportunity to throw a fix for https://github.com/dotnet/dotnet-docker/issues/4632 in since all these changes are about cleaning up layers.

src/runtime-deps/8.0/bookworm-slim/amd64/Dockerfile
```diff
@@ -13,23 +21,14 @@ RUN apt-get update \
...
 RUN groupadd \
         --system \
-        --gid=64198 \
+        --gid=$APP_UID \
         app \
     && useradd -l \
-        --uid 64198 \
-        --gid 64198 \
+        --uid=$APP_UID \
+        --gid= \
```
mthalman, line 31: Missing value here.

MichaelSimons:
> My guess is that the placement of ENV to be above COPY has somehow affected how Docker configures the resulting image

The Docker documentation used to state the first instruction of a scratch Dockerfile after the FROM must be ADD/COPY.  The [documentation is now a little more subtle](https://docs.docker.com/build/building/base-images/#create-a-simple-parent-image-using-scratch) - _Using the scratch “image” signals to the build process that you want the next command in the Dockerfile to be the first filesystem layer in your image._

mthalman (approved):

PR #4620, opened 2023-05-16
Add .NET Monitor section to CODEOWNERS

lbussell:


CODEOWNERS
```diff
@@ -2,3 +2,8 @@
 # See https://help.github.com/articles/about-code-owners/
 
 * @MichaelSimons @mthalman @lbussell
+
+monitor/ @jander-msft
```
jander-msft, line 6: Let's use `@dotnet/dotnet-monitor` instead of me directly, if that works. I'll likely be reviewing these, but in case the rest of the .NET Monitor team wants to contribute, they'll get notified.

CODEOWNERS
```diff
@@ -2,3 +2,8 @@
...
+monitor-base/ @jander-msft
```
mthalman, line 7: Don't these need to be prefixed with `src/`?

jander-msft: > Don't these need to be prefixed with `src/`?

If that happens, there would need to be carve-outs for the /eng/* paths, the readmes in the root, etc. I think the way it is, any file that has a segment of `monitor` or `montor-base` would include me/the team, which is probably correct.

CODEOWNERS
jander-msft: The `@dotnet/dotnet-monitor` team probably doesn't have write access to this repo, so that may not work: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax

jander-msft, line 6: [This](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file) suggests that latter rules supersede the prior rules, so you may have to duplicate your default set (e.g. `@MichaelSimons @mthalman @lbussell`) with these new rules.

jander-msft (approved):

CODEOWNERS
```diff
@@ -1,4 +1,9 @@
 # Users referenced in this file will automatically be requested as reviewers for PRs that modify the given paths.
 # See https://help.github.com/articles/about-code-owners/
 
-* @MichaelSimons @mthalman @lbussell
+*               @MichaelSimons @mthalman @lbussell
```
mthalman, line 4: I created a new team to manage this easier: dotnet-docker-reviewers

mthalman:
The PR is spitting out validation errors for the file due to dotnet-docker-reviewers team. I don't know why. It has access.

MichaelSimons:
> The PR is spitting out validation errors for the file due to dotnet-docker-reviewers team. I don't know why. It has access.

I think the reference is incorrect.  It needs to be prefixed with `@dotnet/`

CODEOWNERS
```diff
@@ -1,4 +1,9 @@
 # Users referenced in this file will automatically be requested as reviewers for PRs that modify the given paths.
 # See https://help.github.com/articles/about-code-owners/
 
-* @MichaelSimons @mthalman @lbussell
+*               @dotnet-docker-reviewers
+
+monitor/        @dotnet-docker-reviewers @jander-msft
```
MichaelSimons, line 6: I suspect this reference is wrong - see https://github.com/dotnet/installer/blob/main/CODEOWNERS

```suggestion
monitor/        @dotnet/dotnet-docker-reviewers @jander-msft
```

MichaelSimons (approved):

PR #4635, opened 2023-05-23
Separate regular and composite images in manifest

### LINKED ISSUE #4634: Aspnet Readmes incorrectly lists non-composite tags for composite images
Looking at the Aspnet readmes after composite images were added, the composite row in the table lists `8.0-preview-alpine` as a valid tag for these images:

https://github.com/dotnet/dotnet-docker/blob/4819bd175a20d5eed0cbcb5048ecf6e452f05ce3/README.aspnet.md?plain=1#L77-L84

However, pulling this tag (on my amd64 machine) always gives me the non-composite image. To fix this we should separate the composite from the non-composite images in the `manifest.json`, where they are currently grouped under the same shared tags:
https://github.com/dotnet/dotnet-docker/blob/4819bd175a20d5eed0cbcb5048ecf6e452f05ce3/manifest.json#L3863-L3964

lbussell:
Fixes #4634


manifest.json
```diff
@@ -3913,7 +3913,17 @@
                 "8.0-preview-alpine-arm64v8": {}
               },
               "variant": "v8"
-            },
```
MichaelSimons, line 3916: It seems like image-builder should have detected this manifest an being invalid.  Can you log an issue if one doesn't already exist?  TIA

mthalman (approved):

manifest.json
lbussell: File https://github.com/dotnet/docker-tools/issues/1139

PR #4676, opened 2023-06-16
Remove --system flag from non-root user and group creation

### LINKED ISSUE #4624: Non-root UID is greater than SYS_UID_MAX
### Describe the Bug

This is how we currently create the non-root user in Debian Bookworm:
https://github.com/dotnet/dotnet-docker/blob/e709fa5d158575e78187ed2b8866dc7f6113c535/src/runtime-deps/8.0/bookworm-slim/amd64/Dockerfile#L17-L27

We're creating a "system" user, but using a UID that exceeds the `SYS_UID_MAX` env variable, resulting in this [warning in CI](https://dev.azure.com/dnceng-public/public/_build/results?buildId=276807&view=logs&j=dd8712c6-8a1e-50f1-fa86-6250d6555872&t=a075e40c-529c-5ca3-4315-cb4ee61596f6&l=259): 
```
useradd warning: app's uid 64198 is greater than SYS_UID_MAX 999
```

As far as I can tell this doesn't have any consequences but it doesn't look great. We should look into if we need the user to be a "system" one and what the consequences of not being a system user are.

### Steps to Reproduce

Build a .NET 8.0 debian-bookworm runtime-deps Dockerfile from the nightly branch.

lbussell:
Fixes #4624. I have left 6.0 Mariner distroless alone to keep that image stable. 7.0 Mariner distroless isn't affected by this issue since it uses a UID of 101.


eng/dockerfile-templates/runtime-deps/Dockerfile.linux.non-root-user
```diff
@@ -14,17 +14,18 @@
     set isDistrolessMariner to defined(match(OS_VERSION, "^cbl-mariner\d+\.\d+-distroless$")) ^
     set isChiseledUbuntu to find(OS_VERSION, "chiseled") >= 0 ^
     set addUserCommand to when(isDebian || isUbuntu || isMariner, "useradd -l", "adduser") ^
+    set isSystemUser to ARGS["uid"] < 1000 || (dotnetVersion = "6.0" && isDistrolessMariner) ^
```
mthalman, line 17: I want to make sure that we clean up the logic which sets system user at the appropriate time. Can you determine what the trigger event is based on which .NET version/OS version goes EOL that we can then clean up this logic? Then please log an issue describing what needs to get cleaned up and when to do it.

mthalman (approved):

lbussell:
So, when you remove the `--system` flag from Alpine, it attempts to interactively set the new user's password. Additionally, Alpine doesn't complain about the UID ranges. In fact, the file with the UID ranges isn't even defined -

```powershell
PS> docker run --rm amd64/ubuntu:jammy cat /etc/login.defs | select-string "uid"
...
# Min/max values for automatic uid selection in useradd
UID_MIN                  1000
UID_MAX                 60000
#SYS_UID_MIN              100
#SYS_UID_MAX              999
```

```powershell
PS> docker run --rm amd64/alpine:3.18 cat /etc/login.defs 
cat: can't open '/etc/login.defs': No such file or directory
```

So I have added the `--system` flag back to Alpine so that it can create the new user without interaction.

mthalman:
> So, when you remove the `--system` flag from Alpine, it attempts to interactively set the new user's password.

This can be avoided by using the `-D` option. I think we'd want to use that and get rid of `--system` to make things consistent across the Dockerfiles.

mthalman (approved):

PR #4669, opened 2023-06-14
Move dotnet-monitor and Alpine floating tags to 3.18

lbussell:
Fixes #4663

Per https://github.com/dotnet/dotnet-docker/pull/4629#issuecomment-1557647204 I have also moved .NET monitor to Alpine 3.18.

PR #4681, opened 2023-06-19
Fix mislabeled Alpine ASP.NET composite images

lbussell:
These were missed from https://github.com/dotnet/dotnet-docker/pull/4669 and I believe they may be causing the CI failures in https://github.com/dotnet/dotnet-docker/pull/4677 and https://github.com/dotnet/dotnet-docker/pull/4676


mthalman (commented):
Readmes should be regenerated.

mthalman (approved):

PR #4688, opened 2023-06-21
Enable image scanning for Ubuntu Chiseled

### LINKED ISSUE #4667: Add scannability support for Ubuntu chiseled images
The Ubuntu chiseled images are essentially a black box from the perspective of container image scanning tools like [Syft](https://github.com/anchore/syft) and [Trivy](https://github.com/aquasecurity/trivy). This is because those tools typically look for a package database that describes which packages are installed in the image. Due to the way these images are constructed using [chisel](https://github.com/canonical/chisel), no such database exists. For that reason, scanning tools would describe the images as having no packages at all. This is not what we want. It's important that these tools provide a more accurate representation of these images for the purposes of vulnerability reporting and SBOM generation.

Canonical is in the process of developing a [wrapper script for chisel](https://github.com/canonical/rocks-toolbox/blob/main/chisel-wrapper) that will generate a dpkg status file. This file will describe what packages are to be installed by chisel. Later, when a scanning tool is run on the image, it will detect this dpkg status file, just as it would for normal full OS container images, and provide an accurate representation of what packages are in the image. We should update our chiseled Dockerfiles to incorporate this [script ](https://github.com/canonical/rocks-toolbox/blob/main/chisel-wrapper) in its execution.

lbussell:
Fixes #4667 

- Enable image scanning
- Enable image scanning test
- Tested against `syft` and `Trivy`:

```pwsh
PS> docker run --rm -v //var/run/docker.sock:/var/run/docker.sock anchore/syft chiseled
NAME             VERSION                   TYPE 
base-files       12ubuntu4.3               deb
ca-certificates  20230311ubuntu0.22.04.1   deb
libc6            2.35-0ubuntu3.1           deb
libgcc-s1        12.1.0-2ubuntu1~22.04     deb
libssl3          3.0.2-0ubuntu1.10         deb
libstdc++6       12.1.0-2ubuntu1~22.04     deb
zlib1g           1:1.2.11.dfsg-2ubuntu9.2  deb
```

It seems to have way fewer packages than CBL Mariner distroless.

Since we seem to be okay pulling directly from `main` in the `canonical/chisel` repo, I followed the same pattern for `rocks-toolbox`. If we want to take more precautions with this repo I would consider adding a `rocks-toolbox` sha/version/branch entry to `manifest.versions.json`.

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
@@ -6,16 +6,21 @@
     set gid to uid
 }}FROM {{ARCH_VERSIONED}}/golang:1.18 as chisel
 
-RUN git clone --depth 1 -b main https://github.com/canonical/chisel /opt/chisel
+RUN git clone --depth 1 -b main https://github.com/canonical/chisel /opt/chisel \
+    && git clone --depth 1 -b main https://github.com/canonical/rocks-toolbox /opt/rocks-toolbox
+
 WORKDIR /opt/chisel
+
 RUN go generate internal/deb/version.go \
     && go build ./cmd/chisel
 
 
 FROM {{ARCH_VERSIONED}}/ubuntu:{{osVersionBase}} as builder
 
-RUN apt-get update && \
-    apt-get install -y ca-certificates
+RUN apt-get update \
+    && apt-get install -y \
+        ca-certificates \
+        file
 
 RUN {{InsertTemplate("Dockerfile.linux.non-root-user-staged", [
         "staging-dir": "/rootfs",
@@ -27,27 +32,30 @@ RUN {{InsertTemplate("Dockerfile.linux.non-root-user-staged", [
         "create-home": "true"
     ])}}
 
-COPY --from=chisel /opt/chisel/chisel /usr/bin/
-RUN chisel cut --release "ubuntu-{{osVersionNumber}}" --root /rootfs \
-        base-files_base \
-        base-files_release-info \
-        ca-certificates_data \
-        libc6_libs \
-        libgcc-s1_libs \
-        libssl3_libs \
-        libstdc++6_libs \
-        zlib1g_libs
+COPY --from=chisel /opt/chisel/chisel /opt/rocks-toolbox/chisel-wrapper /usr/bin/
+
+RUN mkdir -p /rootfs/var/lib/dpkg/ \
+    && chisel-wrapper --generate-dpkg-status /rootfs/var/lib/dpkg/status -- \
+        --release "ubuntu-{{osVersionNumber}}" --root /rootfs \
+            base-files_base \
+            base-files_release-info \
+            ca-certificates_data \
+            libc6_libs \
+            libgcc-s1_libs \
+            libssl3_libs \
+            libstdc++6_libs \
+            zlib1g_libs
 
 
 FROM scratch
 
 COPY --from=builder /rootfs /
 
-# Workaround for https://github.com/moby/moby/issues/38710
-COPY --from=builder --chown={{uid}}:{{gid}} /rootfs/home/{{username}} /home/{{username}}
-
 {{InsertTemplate("../Dockerfile.common-dotnet-envs", [
     "uid": uid
 ])}}
+
+# Workaround for https://github.com/moby/moby/issues/38710
+COPY --from=builder --chown=$APP_UID:$APP_UID /rootfs/home/{{username}} /home/{{username}}
 
 USER $APP_UID
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
@@ -4,7 +4,12 @@
     set username to "app" ^
     set uid to 64198 ^
     set gid to uid
-}}FROM {{ARCH_VERSIONED}}/golang:1.18 as chisel
+}}FROM {{ARCH_VERSIONED}}/golang:1.20 as chisel
+
+RUN apt-get update \
+    && apt-get install -y \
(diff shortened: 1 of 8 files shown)


src/runtime-deps/8.0/jammy-chiseled/amd64/Dockerfile
```diff
@@ -1,15 +1,20 @@
...
 
 
 FROM amd64/ubuntu:jammy as builder
 
-RUN apt-get update && \
-    apt-get install -y ca-certificates
+RUN apt-get update \
+    && apt-get install -y \
+        ca-certificates \
+        file
```
lbussell, line 17: The `chisel-wrapper` script depends on the `file` command - it's about a 10MB install. I also checked `amd64/buildpack-deps:jammy-curl` but it doesn't contain the package by default either.

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
```diff
@@ -27,27 +32,30 @@ RUN {{InsertTemplate("Dockerfile.linux.non-root-user-staged", [
...
+    && chisel-wrapper --generate-dpkg-status /rootfs/var/lib/dpkg/status -- \
+        --release "ubuntu-{{osVersionNumber}}" --root /rootfs \
+            base-files_base \
+            base-files_release-info \
+            ca-certificates_data \
+            libc6_libs \
+            libgcc-s1_libs \
+            libssl3_libs \
+            libstdc++6_libs \
+            zlib1g_libs
```
mthalman, line 47: Looking at our other Dockerfiles, I see that we indent 4 spaces for package names.
```suggestion
        base-files_base \
        base-files_release-info \
        ca-certificates_data \
        libc6_libs \
        libgcc-s1_libs \
        libssl3_libs \
        libstdc++6_libs \
        zlib1g_libs
```

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
```diff
@@ -6,16 +6,21 @@
     set gid to uid
 }}FROM {{ARCH_VERSIONED}}/golang:1.18 as chisel
```
mthalman, line 7: Looking at this a bit closer and I don't see any benefit to having a `chisel` stage. It just adds more lines to the Dockerfile because it requires a copy to get the stuff into the `builder` stage.

mthalman:
> Since we seem to be okay pulling directly from `main` in the `canonical/chisel` repo, I followed the same pattern for `rocks-toolbox`. If we want to take more precautions with this repo I would consider adding a `rocks-toolbox` sha/version/branch entry to `manifest.versions.json`.

I'd like to have an implementation in place to pin to a stable version prior to making chiseled GA.

mthalman (commented):
Changes look great. I just have some comments of cleaning up some of the already existing content.

Before merging this, can you run a test build on the internal pipeline so that we can see whether our SBOM generation recognizes this as well?

lbussell:
> @mthalman: I'd like to have an implementation in place to pin to a stable version prior to making chiseled GA.

Both the chisel tool and rocks-toolbox or just rocks-toolbox? I can submit an issue for this.

> Before merging this, can you run a test build on the internal pipeline so that we can see whether our SBOM generation recognizes this as well?

Before submitting the PR I ran this nightly-playground run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2206441&view=results

Looking at the chiseled runtime-deps `manifest.spdx.json`, it has entries for `base-files`, `libssl3`, `zlib1g`, etc. The list is considerably shorter than regular Jammy of course but as far as I can tell everything is there.

mthalman:
> Both the chisel tool and rocks-toolbox or just rocks-toolbox? I can submit an issue for this.

Yes, both chisel and rocks-toolbox. The whole `chisel` stage can just be incorporated into the `builder` stage.

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
lbussell: My intention was to treat the text after the `--` as another command (it's what is passed to the chisel cut command by the wrapper) and thus have them indented one more stop. I thought that having the `chisel cut` arguments on the same tab stop as the slices made it more difficult to read:

```Dockerfile
RUN mkdir -p /rootfs/var/lib/dpkg/ \
    && chisel-wrapper --generate-dpkg-status /rootfs/var/lib/dpkg/status -- \
        --release "ubuntu-{{osVersionNumber}}" \
        --root /rootfs \
        base-files_base \
        base-files_release-info \
        ca-certificates_data \
        libc6_libs \
        libgcc-s1_libs \
        libssl3_libs \
        libstdc++6_libs \
        zlib1g_libs
```

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
```diff
@@ -4,18 +4,22 @@
     set username to "app" ^
     set uid to 64198 ^
     set gid to uid
-}}FROM {{ARCH_VERSIONED}}/golang:1.18 as chisel
+}}FROM {{ARCH_VERSIONED}}/golang:1.20 as chisel
```
lbussell, line 7: Needed to update the golang version here because:

1. 1.18 images are no longer supported
2. There are only Debian golang images, no Ubuntu golang images. `chisel-wrapper` depends on dpkg-deb to decompress some `.debs` to construct the dpkg status file. Problem is Ubuntu uses `zst` compression for its `.debs` and support for that was only added to Debian's `dpkg` in Bookworm. Our 1.18 image was on Bullseye. Hence the need to update the image in the first place.

MichaelSimons (approved):

mthalman (approved):

mthalman:
cc @cjdcordeiro for awareness

PR #4724, opened 2023-07-06
Fix errors when update-dependencies only updates Ubuntu Chisel tool

### LINKED ISSUE #4721: update-dependencies fails for chisel-only update
The [update-dependencies pipeline](https://github.com/dotnet/dotnet-docker/blob/main/eng/pipelines/update-dependencies.yml) fails when the only updates to the branch are an update to the commit SHA for the [chisel](https://github.com/canonical/chisel/) repo. It outputs the following error:

```
update-dependencies Information: 0 : git status --porcelain results:
 M manifest.versions.json
 M src/runtime-deps/6.0/jammy-chiseled/amd64/Dockerfile
 M src/runtime-deps/6.0/jammy-chiseled/arm32v7/Dockerfile
 M src/runtime-deps/6.0/jammy-chiseled/arm64v8/Dockerfile
 M src/runtime-deps/8.0/jammy-chiseled/amd64/Dockerfile
 M src/runtime-deps/8.0/jammy-chiseled/arm32v7/Dockerfile
 M src/runtime-deps/8.0/jammy-chiseled/arm64v8/Dockerfile

Failed to update dependencies:
System.Exception: 'git status' does not match DependencyInfo information. Git has modified files: True. DependencyInfo is updated: False.
   at Microsoft.DotNet.VersionTools.Automation.DependencyUpdateResults.ChangesDetected() in /_/src/Microsoft.DotNet.VersionTools/lib/src/Automation/DependencyUpdateResults.cs:line 46
   at Dotnet.Docker.UpdateDependencies.ExecuteAsync(Options options) in /update-dependencies/UpdateDependencies.cs:line 82
```

This is related to the changes that were made in https://github.com/dotnet/dotnet-docker/pull/4711.

Based on the output of the error message, the issue is that files in the repo were updated but the Updater class did not return any used DependencyInfos:
https://github.com/dotnet/dotnet-docker/blob/a0622ed5ee4f7341709cbc9161239fa084810641/eng/update-dependencies/BasicVariableUpdater.cs#L22

However, going beyond that particular issue, I think this whole scenario is something we want to avoid. My assertion is that we do not want to update the Dockerfiles solely for the purposes of updating the commit SHA for a dependent repo. A similar rule exists for the use of MinGit. We only update the MinGit version if we're also updating the SDK version:
https://github.com/dotnet/dotnet-docker/blob/a0622ed5ee4f7341709cbc9161239fa084810641/eng/update-dependencies/MinGitUpdater.cs#L23-L33

We should do a similar thing for commit SHAs like chisel.

lbussell:
Fixes #4721

I fixed this by adding the dependency info when we update the chisel and rocks-toolbox versions. I also tried to make it update only when the update-dependencies tool is called with an update to runtime. It still updates chisel when you run update-dependencies with the current runtime version that's in `manifest.versions.json` (I'm not sure if that's the desired behavior, I couldn't figure out how to get around that)


mthalman:
> It still updates chisel when you run update-dependencies with the current runtime version that's in manifest.versions.json

That's going to be the case when this is run in the update-dependencies pipeline. It queries to determine what the latest versions are and then passes them to update-dependencies. The caller doesn't know that the versions are the same, nor should it. It should no-op in that case, not update the commit SHA.

It should be possible for the updater to determine this information.

First, we're going to need a chisel/toolbox-specific updater since we only want this behavior for this scenario. The reasoning here is that we don't want to update the SHA by itself since it's only used for build purposes. Compare that to the MinGit version which has similar behavior in that it's tied to the SDK Dockerfile but, in that case, we _do_ want to update that version by itself because it ends up in final image and we want customers to have the latest.

To determine if the runtime version has changed, you can find the desired version within the dependency info. Then you can look up the current version by calling `ManifestHelper.GetVariableValue`. The variable name to use can be computed by reusing [`GetVersionVariableName`](https://github.com/dotnet/dotnet-docker/blob/5517616ffd9d5f163c89eb42cad27294ad1aaf8f/eng/update-dependencies/VersionUpdater.cs#L107). If the desired value is different than the current value then you know the runtime version is being changed. This requires that the [chisel updaters be listed before the version updaters](https://github.com/dotnet/dotnet-docker/blob/a0622ed5ee4f7341709cbc9161239fa084810641/eng/update-dependencies/UpdateDependencies.cs#L415-L437). They already are but adding a comment in the code in that spot will help to ensure that remains true. If the version updater gets run first, then the values will be the same before you get a chance to query if they're different.

eng/update-dependencies/ChiselToolUpdater.cs
```diff
@@ -0,0 +1,49 @@
...
+        _dockerfileVersion = dockerfileVersion;
+        _newValue = newValue;
+    }
+
+    protected sealed override string TryGetDesiredValue(IEnumerable<IDependencyInfo> dependencyInfos, out IEnumerable<IDependencyInfo> usedDependencyInfos)
+    {
+        IDependencyInfo? runtimeDependencyInfo = dependencyInfos.FirstOrDefault(info => info.SimpleName == "runtime");
+        usedDependencyInfos = Enumerable.Empty<IDependencyInfo>();
+
+        string currentChsielToolVersion = ManifestHelper.GetVariableValue(VariableName, ManifestVariables.Value);
```
mthalman, line 27: Typo in `currentChsielToolVersion`

eng/update-dependencies/ChiselToolUpdater.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            return currentChiselToolVersion;
+        }
+
+        // Avoid updating chisel tooling unless we already know we are
+        // rebuilding at least the runtime images, since changing the chisel
+        // tool shouldn't make a difference in the output image.
+        string runtimeVariableName = ManifestHelper.GetVersionVariableName(VersionType.Build, "runtime", _dockerfileVersion);
+        string currentRuntimeVersion = ManifestHelper.GetVariableValue(runtimeVariableName, ManifestVariables.Value);
+        if (/* runtimeDependencyInfo.SimpleVersion == currentRuntimeVersion */ false)
```
mthalman, line 45:
```suggestion
        if (runtimeDependencyInfo.SimpleVersion == currentRuntimeVersion)
```

eng/update-dependencies/ChiselUpdater.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#nullable enable
+namespace Dotnet.Docker;
+
+internal class ChiselUpdater : ChiselToolUpdater
```
mthalman, line 7: I think naming this `ChiselRefUpdater` would help clarify the purpose of this compared to the base class. And then similarly change `RocksToolboxUpdater` to `RocksToolboxRefUpdater`.

mthalman (approved):

PR #4747, opened 2023-07-12
Test .NET Monitor with samples regardless of mode and distrolessness

jander-msft:
The .NET Monitor tests were filtering .NET Monitor and sample image combinations based on the "distrolessness" of each image where only distroless was being tested with distroless and non-distroless tested with non-distroless. This doesn't cover the real possibility that customers are running their apps as a different user as compared to the .NET Monitor image; this is especially true of the .NET Monitor 8 images which are all distroless and exclusively use a non-root user.

I've removed this filter to allow all combinations and updated the volume creation to determine the least-privileged user of the two images to be used as the owner of the volumes and defer to the sample image user if there is a tie; this would be the guidance for our customers who have mixed user usage of containers in their deployments.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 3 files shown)


jander-msft:
cc @dotnet/dotnet-monitor

tests/Microsoft.DotNet.Docker.Tests/MonitorImageTests.cs
```diff
@@ -558,6 +535,54 @@ public Task VerifyListenMode(ProductImageData imageData, SampleImageData sampleD
...
+        private static int? AdjustMonitorUserAndCalculateVolumeOwner(bool listenMode, int sampleUid, int monitorUid, DockerRunArgsBuilder monitorArgsBuilder)
+        {
+            // Make sure volume is accessible by sample app without modifying the sample container
+            // and ensure monitor app can access it, even if needing to change its user. This is
+            // done by making the volume owned by the least privileged user; if they are the same user or
+            // are different non-root users, defer to the user of the sample image.
+
+            if (!IsRoot(sampleUid))
+            {
+                // If monitor has UDS to which sample will connect, change monitor to run as
```
mthalman, line 547: What is UDS?

jander-msft: Undid Domain Socket

mthalman (approved):

PR #4752, opened 2023-07-13
Use release tag to update rocks-toolbox

### LINKED ISSUE #4745: Consume release artifacts of rocks-toolbox repo
With the changes in https://github.com/dotnet/dotnet-docker/pull/4711, we pin to a specific commit of the https://github.com/canonical/rocks-toolbox repo when cloning it in the Dockerfile. This repo will now be creating releases (e.g. https://github.com/canonical/rocks-toolbox/releases/tag/v1.0.0) that we should consume instead of querying for the latest commit. This could reuse the [similar logic](https://github.com/dotnet/dotnet-docker/blob/6da64f31944bb16ecde5495b6a53fc170fbe100d/eng/update-dependencies/UpdateDependencies.cs#L411) that we do with MinGit.

lbussell:
Fixes #4745 

The next time the update-dependencies tool runs, it should update the rocks-toolbox ref to a tag, and then change the way the tag is cloned in the Dockerfile.

### DIFF
--- a/eng/dockerfile-templates/Dockerfile.git-clone
+++ b/eng/dockerfile-templates/Dockerfile.git-clone
@@ -8,7 +8,6 @@
         dir: The directory to clone the repository into ^
 
     set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
-
     set ref to token(ARGS["ref"], ":", 1) ^
     set isTag to token(ARGS["ref"], ":", 0) = "tag"
 
--- a/eng/dockerfile-templates/Dockerfile.git-clone
+++ b/eng/dockerfile-templates/Dockerfile.git-clone
@@ -7,7 +7,6 @@
              Formatted as "<ref type>:<ref>"
         dir: The directory to clone the repository into ^
 
-    set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set ref to token(ARGS["ref"], ":", 1) ^
     set isTag to token(ARGS["ref"], ":", 0) = "tag"
 
--- a/eng/dockerfile-templates/Dockerfile.git-clone
+++ b/eng/dockerfile-templates/Dockerfile.git-clone
@@ -4,8 +4,11 @@
     ARGS:
         url: The git repository URL
         ref: The commit hash, branch, or tag to checkout
-        dir: The directory to clone the repository into
+        dir: The directory to clone the repository into ^
 
-}}git clone --no-checkout {{ARGS["url"]}} {{ARGS["dir"]}} \
+    set is_tag to len(ARGS["ref"]) < 40
+
+}}{{if is_tag:git clone --branch {{ARGS["ref"]}} --depth 1 {{ARGS["url"]}} {{ARGS["dir"]}}^
+else:git clone --no-checkout {{ARGS["url"]}} {{ARGS["dir"]}} \
 && cd {{ARGS["dir"]}} \
-&& git checkout {{ARGS["ref"]}}
+&& git checkout {{ARGS["ref"]}}}}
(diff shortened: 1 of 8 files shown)


eng/dockerfile-templates/Dockerfile.git-clone
```diff
@@ -4,8 +4,11 @@
     ARGS:
         url: The git repository URL
         ref: The commit hash, branch, or tag to checkout
-        dir: The directory to clone the repository into
+        dir: The directory to clone the repository into ^
 
-}}git clone --no-checkout {{ARGS["url"]}} {{ARGS["dir"]}} \
+    set is_tag to len(ARGS["ref"]) < 40
```
mthalman, line 9: Given that it's possible for a tag name to be longer than 40 characters, we should consider other options. What if we included the ref type in the value itself and parse that out to determine what it is.

```json
"rocks-toolbox|ref": "commit:00f796f17323422704241517aa0386ded5e0fba1"
```

```json
"rocks-toolbox|ref": "tag:v1.0.0"
```

Or would it be too disruptive to the infrastructure if we did something like this?

```json
"rocks-toolbox|ref": {
  "commit": "00f796f17323422704241517aa0386ded5e0fba1"
}
```

```json
"rocks-toolbox|ref": {
  "tag": "v1.0.0"
}
```

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
```diff
@@ -2,12 +2,19 @@
     set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set osVersionBase to match(OS_VERSION, ".+(?=.*-)")[0] ^
     set osVersionNumber to split(OS_ARCH_HYPHENATED, "-")[1] ^
+
     set chiselDir to "/opt/chisel" ^
     set chiselUrl to VARIABLES[cat("chisel|", dotnetVersion, "|url")] ^
-    set chiselRef to VARIABLES[cat("chisel|", dotnetVersion, "|ref")] ^
+    set chiselRefVar to VARIABLES[cat("chisel|", dotnetVersion, "|ref")] ^
+    set chiselRef to token(chiselRefVar, ":", 1) ^
+    set chiselRefIsTag to token(chiselRefVar, ":", 0) = "tag" ^
```
mthalman, line 10: This pattern needs to be repeated for each repo to clone. Consider refactoring so that the Dockerfile.get-clone template handles the parsing instead. It does make the template's args less "pure" but given that the only usage we have for this template is for these types of prefixed variable values, I think that's ok.

eng/dockerfile-templates/Dockerfile.git-clone
```diff
@@ -4,8 +4,14 @@
     ARGS:
         url: The git repository URL
         ref: The commit hash, branch, or tag to checkout
-        dir: The directory to clone the repository into
+             Formatted as "<ref type>:<ref>"
+        dir: The directory to clone the repository into ^
 
-}}git clone --no-checkout {{ARGS["url"]}} {{ARGS["dir"]}} \
+    set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
```
mthalman, line 10: This doesn't seem needed.

mthalman (approved):

eng/dockerfile-templates/Dockerfile.git-clone
lbussell: And just to note here, I didn't choose the latter option because it didn't fit in with the rest of the file. It would make sense to me if the rest of the file were "json"-ified, like:
```json
"runtime": {
  "6.0": {
    "build-version": "6.0.20",
    "linux-musl": {
      "x64": {
        "sha": ...
      }
    }
},
"rocks-toolbox": {
  "commit": ...
}
```

..but it's not

mthalman (approved):

PR #4770, opened 2023-07-20
Remove CBL Mariner 1.0 Dockerfiles

### LINKED ISSUE #4558: Delete CBL-Mariner 1.0 Dockerfiles when it is EOL
Mariner 1.0 will be EOL on August 31, 2023. The .NET Dockerfiles for that version can be deleted after that date.

lbussell:
Fixes #4558


mthalman (commented):
There should be updates in the tests to remove this version as well.

Given that the EOL date is being moved to August, I would suggest we not merge this into nightly until after the August release.

PR #4773, opened 2023-07-20
Remove kerberos and ca-certificates from Alpine for .NET 8.0 images

### LINKED ISSUE #4751: Consider removing Kerberos and `ca-certificates` from Alpine `runtime-deps` image (.NET 8)
Alpine images come with `ca-certificates-bundle`. Basic testing suggests that this package satisfies our HTTPS need. Do we need  `ca-certificates` as well? We need to investigate the difference between these packages.

Kerberos is needed for niche scenarios. Also, it often relies on LDAP, which we don't install. We can likely skip installing Kerberos w/o negatively affecting many customers.

Context: https://learn.microsoft.com/aspnet/core/security/authentication/windowsauth?view=aspnetcore-7.0&tabs=visual-studio#kestrel

Related to #4748

lbussell:
Fixes #4751 

I forget if we decided if alpine's `ca-certificates-bundle` would work instead of `ca-certificates`, but it can be added back here if not.

cc @richlander


mthalman:
> I forget if we decided if alpine's `ca-certificates-bundle` would work instead of `ca-certificates`, but it can be added back here if not.

Part of this work is to determine the difference between them. That will inform what action we should be taking.

lbussell:
I did some research on this. [ca-certificates-bundle](https://pkgs.alpinelinux.org/package/v3.18/main/x86_64/ca-certificates-bundle) is simply all of the certificates from [ca-certificates](https://pkgs.alpinelinux.org/package/v3.18/main/x86_64/ca-certificates) bundled into one `.crt` file. You can see in [the APKBUILD](https://git.alpinelinux.org/aports/tree/main/ca-certificates/APKBUILD?h=3.18-stable) that it simply places the contents of all the `*.crt` files from `ca-certificates` into `/etc/ssl/certs/ca-certificates.crt`. We would also lose utilities like `/usr/sbin/update-ca-certificates`.

To me, it sounds safe to remove for Preview 7, and if it causes significant issues we can add it back for RC1.

richlander:
That sounds fine to me. I don't believe we make any active us of those tools. My read from what you've found is that the actual certificate aspect is equivalent. That works for us!

@dtivel for visibility.

mthalman:
I ran a quick script to verify that all the .cert and .pem files installed by `ca-certificates` are represented in `/etc/ssl/certs/ca-certificates.crt`. So that's good.

src/runtime-deps/8.0/alpine3.18/amd64/Dockerfile
```diff
@@ -11,10 +11,7 @@ ENV \
     DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=true
 
 RUN apk add --no-cache \
-        ca-certificates \
```
mthalman, line 14: I'm going to suggest that we explicitly install `ca-certificates-bundle`. As a general rule, we've always listed all our dependencies, regardless of whether they're defined in the base image. For example, `libssl3` is also in the base image but we list it here. This explicitly documents our dependencies for the distro version. It also ensures we install the latest version (which there's actually an existing configuration issue with now: https://github.com/dotnet/dotnet-docker/issues/4776).

mthalman (approved):

PR #4777, opened 2023-07-21
Force upgrade of .NET runtime deps in alpine images

### LINKED ISSUE #4776: Package installation in Alpine should be configured to upgrade
In the Alpine Dockerfiles, if a newer package is available than what is already installed in the container, the [installation command being used](https://github.com/dotnet/dotnet-docker/blob/9d6250ee6fa5200a0c300a6b10137d954110ead5/src/runtime-deps/6.0/alpine3.18/amd64/Dockerfile#L11-L20) will not cause that package to be upgraded. This differs from the default behavior of Debian, Ubuntu, and Mariner.

Let's take the example of the `zlib` package. That package exists in the base Alpine image. That package is also [listed for installation in the Dockerfile](https://github.com/dotnet/dotnet-docker/blob/9d6250ee6fa5200a0c300a6b10137d954110ead5/src/runtime-deps/6.0/alpine3.18/amd64/Dockerfile#L20). Let's say a newer version of `zlib` is available from the package feed than is contained in the base image and the newer version has a security fix. When the Dockerfile gets built, it should be installing the latest version. But `apk` is configured that way by default. It requires a `-u, --upgrade` option to cause existing packages to be upgraded. This option should be added to the Dockerfiles to ensure the latest version is installed.

lbussell:
fixes #4776 

At present (for runtime-deps\6.0\alpine3.18\amd64) this causes a size increase of 12.3 -> 17.4 MB because `libcrypto3` and `libssl3` are updated.

### DIFF
--- a/eng/dockerfile-templates/Dockerfile.linux.install-pkgs
+++ b/eng/dockerfile-templates/Dockerfile.linux.install-pkgs
@@ -18,7 +18,7 @@
     }}
 }}{{
 if isDnf:dnf install -y{{ARGS["pkg-mgr-opts"]}} \^
-elif isApk:apk add --no-cache{{ARGS["pkg-mgr-opts"]}} \^
+elif isApk:apk add --upgrade --no-cache{{ARGS["pkg-mgr-opts"]}} \^
 elif isTdnf:tdnf install -y{{ARGS["pkg-mgr-opts"]}} \^
 else:apt-get update \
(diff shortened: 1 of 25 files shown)


richlander:
The size increase is OK. It's a result of correctness.

All the more reason to do #4773 to do at the same time.

mthalman (approved):

PR #4786, opened 2023-07-31
.NET Monitor 7.3 images

jander-msft:
Add .NET Monitor 7.3 images and move major-only version 7 floating tags to 7.3 images. Changes exclude Ubuntu Chiseled images in order to facilitate easier merge to main branch.

cc @dotnet/dotnet-monitor


manifest.json
```diff
@@ -7014,6 +6992,202 @@
...
+              "variant": "v8"
+            }
+          ]
+        },
+        {
+          "productVersion": "$(monitor|7.3|product-version)",
+          "sharedTags": {
+            "$(monitor|7.3|product-version)-alpine": {},
+            "7.3-alpine": {},
+            "7-alpine": {},
```
mthalman, line 7006: When merging 7.3 Dockerfiles to main, will we also want the floating tags moved or will that happen at a later time? If it's later, it'd be better if the floating tag change was a separate commit or PR so that the 7.3 Dockerfiles can be isolated from floating tag change.

mthalman: @lbussell - Do you have a preference?

lbussell: The current workflow, while not fun, is just fine. If you want to take the time to do that it would probably make things a little simpler at servicing time though. no preference.

mthalman (approved):

PR #4806, opened 2023-08-09
Simplify Chisel tool installation

lbussell:
The newest chisel version lets you install with `go install ...`. This required a manual update to the chisel version but I'm happy to wait for that version to flow in naturally with dependency flow updates.


richlander:
That looks very nice!

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
```diff
@@ -21,23 +20,13 @@ RUN apt-get update \
         ca-certificates \
         file
 
+RUN go install {{chiselUrl}}@@{{chiselRef}}
```
mthalman, line 23: Would the syntax look the same regardless of whether it was a commit SHA or tag? If so, I think manifest.versions.json should be updated to not have the `commit:` syntax at all for the chisel repo.

src/runtime-deps/6.0/jammy-chiseled/arm32v7/Dockerfile
```diff
@@ -5,19 +5,12 @@ RUN apt-get update \
...
-    && git checkout e92d18b733647e77b6829968c91f8a16d25c1f2d
-
-WORKDIR /opt/chisel
+# Install chisel
+RUN go install github.com/canonical/chisel@50b1cb3cc6a3431407fadcaaf5a5c559220e30e1
 
-RUN go generate internal/deb/version.go \
-    && go build ./cmd/chisel \
-    && cp /opt/chisel/chisel /usr/bin/ \
+RUN git clone --no-checkout https://github.com/canonical/rocks-toolbox /opt/rocks-toolbox \
```
MichaelSimons, line 11: Since we are just grabbing a single file would it make sense to curl it instead?   This whole RUN instruction could be simplified to a single command.

MichaelSimons (commented):
Nice simplification.

src/runtime-deps/6.0/jammy-chiseled/arm32v7/Dockerfile
richlander: You know ... `ADD` supports this.

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
lbussell: Yep. I think it supports any kind of ref.

src/runtime-deps/8.0/jammy-chiseled/arm32v7/Dockerfile
```diff
@@ -5,11 +5,11 @@ RUN apt-get update \
...
         file
 
-# Install chisel
-RUN go install github.com/canonical/chisel@50b1cb3cc6a3431407fadcaaf5a5c559220e30e1
+RUN go install github.com/canonical/chisel/cmd/chisel@50b1cb3cc6a3431407fadcaaf5a5c559220e30e1
 
-RUN git clone --branch v1.0.0 --depth 1 https://github.com/canonical/rocks-toolbox /opt/rocks-toolbox \
-    && cp /opt/rocks-toolbox/chisel-wrapper /usr/bin/
+RUN curl -sSL --output /usr/bin/chisel-wrapper \
+    https://raw.githubusercontent.com/canonical/rocks-toolbox/v1.0.0/chisel-wrapper \
```
MichaelSimons, line 11: This should be indented or included in the previous line.

This is a bit orthogonal to your change, but it feels like this should be pinned for the same reasons the chisel tool is pinned.

MichaelSimons: Nevermind my pinned comment, I see this is a release reference.

lbussell: Not sure what you mean by pinned. The version/commit of rocks-toolbox is managed by the update-dependencies tool, if that's what you mean.

src/runtime-deps/6.0/jammy-chiseled/arm32v7/Dockerfile
lbussell: To humor @richlander I tried it with `ADD`... it can just as easily be reverted 🙂

MichaelSimons (approved):

lbussell:
The chisel tool has GitHub releases now, so I'm using that. However I think this PR will fail currently due to https://github.com/canonical/chisel/issues/96.

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
```diff
@@ -21,24 +19,11 @@ RUN apt-get update \
         ca-certificates \
         file
 
-RUN {{InsertTemplate("../Dockerfile.git-clone", [
```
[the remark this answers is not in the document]
mthalman: This is still relevant.

eng/update-dependencies/UpdateDependencies.cs
```diff
@@ -409,7 +409,7 @@ private static async Task<IEnumerable<IDependencyUpdater>> GetUpdatersAsync()
             // (e.g. sha updater requires the version numbers to be updated within the Dockerfiles)
 
             JObject minGitRelease = await GitHubHelper.GetLatestReleaseAsync("git-for-windows", "git");
-            string chiselRef = await GitHubHelper.GetLatestCommitAsync("canonical", "chisel", "main");
```
mthalman, line 412: `GetLatestCommitAsync` is no longer reference anymore. That method can be deleted.

mthalman (approved):

PR #4812, opened 2023-08-10
[Do not merge] Use experimental ImageBuilder to fix arm64 image build issues

lbussell:
So far, the issue in #4802 has only reproduced reliably in PR validation. I am trying a solution out here in this draft PR.


lbussell:
closing in favor of #4813

PR #4800, opened 2023-08-07
.NET Container Images Servicing August 2023

lbussell:

PR #4815, opened 2023-08-11
[Do not merge][nightly] Reproduce BuildKit caching issues with new ImageBuilder version

lbussell:
Trying to reproduce the same thing as in #4813, but this time after base images are updated to be arm64/ with no variant.

### DIFF
--- a/eng/common/templates/jobs/build-images.yml
+++ b/eng/common/templates/jobs/build-images.yml
@@ -13,10 +13,10 @@ parameters:
 
 jobs:
 - job: ${{ parameters.name }}
-  condition: and(${{ parameters.matrix }}, not(canceled()), in(dependencies.PreBuildValidation.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'))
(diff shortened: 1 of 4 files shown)


lbussell:
It looks like the issue[1] from https://github.com/dotnet/dotnet-docker/pull/4802, https://github.com/dotnet/dotnet-docker/pull/4800, https://github.com/dotnet/dotnet-docker/pull/4813, and https://github.com/dotnet/dotnet-docker/pull/4812 won't happen again after a forced no-cache build of all images once the new ImageBuilder is merged in https://github.com/dotnet/dotnet-docker/pull/4808.

[1] the issue where docker fails to use a locally tagged image with an architecture of `arm64/v8` in a build of an image targeting an `arm64/` architecture.

mthalman:
Does that mean we can close this PR and https://github.com/dotnet/dotnet-docker/pull/4813?

lbussell:
Closing this as we can re-open if more investigation is required.

Issue #4804, opened 2023-08-09
Consider an icu version of `aot-deps`


richlander:
This recent issue got me thinking that we should consider providing an icu version of `aot-deps` and maybe even `runtime-deps`.

https://github.com/dotnet/core/issues/8663

Today, we don't provide an ICU version of Alpine since we'd need two duplicate branches of images. This policy makes sense. When we think solely of self-contained, which all native aot apps are, then having an ICU image makes a lot more sense and avoids the need for the chisel tool. I see the cost/benefit tradeoff for ICU images being quite different (and more compelling) for `aot-deps` and consider same for `runtime-deps` (maybe later).

mthalman:
[Triage]
* The UX for dealing with the installation of additional packages in chiseled scenarios is not great. It's quite verbose, requires copying from some sort of sample, and is generally not easy to understand.
* Adding a new image type (repository) is not something we want to do without very good reason as it increases the burden of maintenance on the production side and adds to the concept count that users need to consider.
* There are two aspects of the UX of installing additional packages in chiseled scenarios that need to be improved:
  * Acquistion of the chisel tool
  * Installation of the package and subsequent cleanup
* One argument is that we should invest more in improving the UX of installing additional packages. That is a goal anyway as it will be applicable to more scenarios than just the need for `icu`. Ideally, the UX would be improved to the point where the need of an icu-specific image type goes away. In that case, we could always discontinue that image type for the next major .NET version.

jkotas:
ICU is written in C++ and brings in dependency  on C++ runtime.

Is the C++ runtime that only thing that differentiates between aot-deps and runtime-deps? If it is the case, aot-deps w/ ICU is going to be identical to runtime-deps w/ ICU.

richlander:
Great observation.

See runtime-deps:chiseled -> https://github.com/dotnet/dotnet-docker/blob/main/src/runtime-deps/8.0/jammy-chiseled/amd64/Dockerfile

You can see that we went the "Alpine route" and removed (didn't add) ICU.

I am now thinking that we should do the following:

- `aot-deps:chiseled` -- no ICU
- `runtime-deps:chiseled` has ICU

This will have a couple advantages:

- Much higher number of users can switch to chiseled (in CoreCLR land) from regular Ubuntu w/no fanfare.
- We have ICU and non-ICU variants covered.

[...]

Issue #4129, opened 2022-10-06
Container offering for supporting Native AOT scenarios


mthalman:
We need to determine if and how we should adjust the set of container images we provide to meet the needs of Native AOT scenarios. It looks like the set of package dependencies that are needed for those scenarios is different from standard deployments.

For example, there is work to remove a dependency on `libstdc++`: https://github.com/dotnet/runtime/issues/76655. That likely has no impact on the current full versions of Debian, Ubuntu, and Mariner since that package is installed by default. But it would impact Alpine, distroless Mariner, and chiseled Ubuntu since it's explicitly being added in those images.

We should avoid the inclusion of any unnecessary packages in our images. It leads to a larger image size and increased exposure to vulnerabilities.

mthalman:
If there were to be separate images only intended for Native AOT deployments, this would be limited to just `runtime-deps` images because it is, by definition, a trimmed application that has the runtime embedded in it.

mthalman:
[Triage]
First step is to investigate the scenario with an app building exercise. This will help to determine what the requirements are for the different Linux distros that are relevant. At the very least, we could then provide sample documentation to illustrate how to configure Dockerfiles for this scenario. We can then make a determination of whether it's worthwhile to actually provide official base images.

mthalman:
One interesting aspect of native AOT publishing is the following from https://learn.microsoft.com/dotnet/core/deploying/native-aot/:

> A native AOT binary produced on Linux machine is only going to work on same or newer Linux version. For example, native AOT binary produced on Ubuntu 20.04 is going to run on Ubuntu 20.04 and later, but it is not going to run on Ubuntu 18.04.

This could have implications on which OS versions we make the SDK available on. For 7.0, for example, we only provide Ubuntu 22.04 images.

xlievo:
sdk:
FROM mcr.microsoft.com/dotnet/sdk:7.0.103-alpine3.17-amd64 AS build
RUN apk update && apk upgrade
RUN apk add --no-cache clang build-base zlib-dev libssl1.1

runtime:
FROM alpine:3.17
RUN apk update && apk upgrade
RUN apk add --no-cache libstdc++

PublishAot=true

eerhardt:
> RUN apk add --no-cache libstdc++

Note that in .NET 8, `libstdc++` is no longer needed. This was fixed with https://github.com/dotnet/runtime/pull/79501.

mthalman:
[Triage]
[...]

PR #4822, opened 2023-08-15
Add native AOT runtime-deps and sdk variants

lbussell:
Please see https://github.com/dotnet/dotnet-docker/issues/4129 and https://github.com/dotnet/dotnet-docker/issues/4804

~~I am currently working on the tests for these images.~~ After discussion, tests will be included in a separate PR.


src/runtime-deps/8.0/cbl-mariner2.0-distroless/arm64v8/Dockerfile
```diff
@@ -13,7 +13,6 @@ RUN mkdir /staging \
         \
         # .NET dependencies
         glibc \
-        krb5 \
```
MichaelSimons, line 16: Is the reason this is being removed captured in an issue?

eng/dockerfile-templates/Dockerfile.linux.install-deps
```diff
@@ -1,30 +1,41 @@
 {{
     _ ARGS:
-        isSdk (optional): Indicates whether the dependencies are installed for an sdk Dockerfile
+        is-aot (optional): TODO: fill out
```
MichaelSimons, line 3: What are the plans to address the TODO comments here?

eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
```diff
@@ -11,6 +16,31 @@
...
+            "libgcc-s1_libs",
+            "libssl3_libs",
+            "zlib1g_libs"
+        ] ^
+
+    _ Arrays in cottle are actually key-value maps. Sorting the array sorts by value
+        but does not change the key value. In order to have the right index in the
+        array, we must call `cat` again to reset the indices (keys) in the map. ^
+    set standardPkgs to cat(sort(cat(aotPkgs, [ "libstdc++6_libs" ]))) ^
+    set extraPkgs to cat(sort(cat(standardPkgs,
```
MichaelSimons, line 33: I see support for extraPkgs but no invocation.  I understood this was going to be handled with a separate PR.  Is this a case of separating the two becomes difficult because they are touching the same template code?

src/sdk/8.0/jammy-aot/amd64/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports/ jammy-updates main restricted
+EOF
+
+RUN sed -i -e 's/deb http/deb [arch=amd64] http/g' /etc/apt/sources.list \
+    && dpkg --add-architecture arm64 \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        clang \
+        gcc-aarch64-linux-gnu \
+        llvm \
```
MichaelSimons, line 16: Why LLVM?

src/runtime-deps/6.0/jammy-chiseled/amd64/Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN groupadd \
 
 RUN mkdir -p /rootfs/var/lib/dpkg/ \
     && chisel-wrapper --generate-dpkg-status /rootfs/var/lib/dpkg/status -- \
-        --release "ubuntu-22.04" --root /rootfs \
+        --release ubuntu-22.04 --root /rootfs \
```
mthalman, line 37: Because this is making a change that would also be relevant to the main branch, this change should be made independently of this PR. It will get in the way of merging changes between the branches and dealing with diffs.

src/runtime-deps/8.0/cbl-mariner2.0-distroless/arm64v8/Dockerfile
mthalman: https://github.com/dotnet/dotnet-docker/issues/4751. IMO, this removal should be done separately from these changes. The same removal should apply to Debian and Ubuntu, right? https://github.com/dotnet/dotnet-docker/blob/a345a8f9085bd933c2850cbb4e1d203f5a23a987/src/runtime-deps/8.0/bookworm-slim/amd64/Dockerfile#L18

I understand it's installed by default there. We should have discussion of whether it should be listed in that case.

mthalman (commented):
My meta-review comment is "let's be sure to avoid including changes in this PR that we want in main" because AOT won't be getting merged to main.

src/sdk/8.0/jammy-aot/amd64/Dockerfile
lbussell: I referenced the package list from Rich's sample - https://github.com/dotnet/dotnet-docker/blob/58ee815e1a06614732a2c2726bec51a0c9b8ad67/samples/releasesapi-aot/Dockerfile.ubuntu-cross-x64-arm64#L14

@richlander is LLVM strictly necessary for cross-compilation scenarios?

src/runtime-deps/8.0/cbl-mariner2.0-distroless/arm64v8/Dockerfile
lbussell: I have removed this change and will include it in the `extra` PR when we'll have the full set of images available.

src/sdk/8.0/jammy-aot/amd64/Dockerfile
MichaelSimons: This feels related to a comment Matt made at https://github.com/dotnet/dotnet-docker/pull/4822#discussion_r1302102256.  If LLVM is required, should we be explicitly declaring it as a dependency in all distros even in scenarios where it is already provided in a base layer?

src/runtime-deps/8.0/cbl-mariner2.0-distroless/arm64v8/Dockerfile
richlander: We should have the discussion on whether we want to include this in `extra`. I don't think so. My understanding is that kerberos is often paired with ldap, which we have never installed.

src/sdk/8.0/jammy-aot/amd64/Dockerfile
richlander: `clang` depends on `llvm` so it will be pulled in anyway.

Clang is the front-end and LLVM is the backend.

mthalman: In that case, I lean towards only listing `clang` for now. Depending on CVE activity in dependent packages, we may want to include others but I think we can evaluate that over time.

src/sdk/8.0/jammy-aot/arm64v8/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# syntax=docker/dockerfile:1
+ARG REPO=mcr.microsoft.com/dotnet/sdk
+FROM $REPO:8.0.100-rc.1-jammy-arm64v8
+
+COPY <<EOF /etc/apt/sources.list.d/amd64.list
```
MichaelSimons, line 5: Do we have a general agreement that HereDocs should be used?  If so what is the thought on when it should be used?  I would expect consistent usage of it - e.g. in RUN instructions as well.  Making consistent use is beyond the scope of this PR but given the introduction of it here we should establish an agreed upon direction and log follow-ups tracking this work.

lbussell: Thanks for bringing this up. I tried heredocs for RUN instructions but they don't play nice with line endings with Git for Windows' default behavior (checkout as CRLF and commit as LF). It creates some really strange errors.

Because of that I'd probably lean towards switching this command to using `echo` instead, to keep everything consistent.

MichaelSimons: I agree that sounds best for now.

src/sdk/8.0/jammy-aot/arm64v8/Dockerfile
```diff
@@ -2,10 +2,8 @@
 ARG REPO=mcr.microsoft.com/dotnet/sdk
 FROM $REPO:8.0.100-rc.1-jammy-arm64v8
 
-COPY <<EOF /etc/apt/sources.list.d/amd64.list
-deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ jammy main restricted
-deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ jammy-updates main restricted
-EOF
+RUN echo "deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ jammy main restricted" > /etc/apt/sources.list.d/amd64.list \
```
MichaelSimons, line 5: Now that this utilizes echo, it doesn't feel necessary to keep this as a separate instruction especially given the first line of the next RUN instruction.

mthalman (approved):

MichaelSimons (approved):

src/sdk/8.0/jammy-aot/arm64v8/Dockerfile
richlander: I noticed the same thing. You have to do unix-style to use heredocs.

Line endings are the most annoying (at least obvious) difference between Windows and unix-style. More so than slash/backslash.

lbussell:
This PR and https://github.com/dotnet/dotnet-docker/pull/4847 have significant overlap in the templates. I'd like to wait until that is merged first before merging this one so that the merge of that PR to main is less complicated.

mthalman (approved):

PR #4832, opened 2023-08-18
Switch to new Docker public pools for ARM 

lbussell:
cc @JpratherMS @mthalman


mthalman:
Is there a reason we wouldn't just merge https://github.com/dotnet/docker-tools/pull/1175 instead?

lbussell:
> Is there a reason

Nope. I spaced out and forgot about that PR.

Issue #4834, opened 2023-08-19
ASP.NET Composite runtime and non-composite .NET runtime versions are out of sync for some preview versions


lbussell:
This is because the version of the composite runtime and the normal runtime are built from different locations. During preview dependency flows, the versions can get out of sync until the product is locked down for the next preview.

You can see this today:
```
> docker pull mcr.microsoft.com/dotnet/nightly/aspnet:8.0-alpine3.18-composite-amd64
digest: sha256:6a39b9c4d3e9e0c24ebb6432ff90e81f00b011ac6683f946eae78c179440eb71

> docker run --rm --entrypoint "/bin/sh" mcr.microsoft.com/dotnet/nightly/aspnet:8.0-alpine3.18-composite-amd64 -c 'echo $DOTNET_VERSION'
8.0.0-rc.1.23414.4

> docker run --rm --entrypoint "/bin/sh" mcr.microsoft.com/dotnet/nightly/aspnet:8.0-alpine3.18-composite-amd64 -c 'dotnet --list-runtimes | grep "Microsoft.NETCore.App"'
Microsoft.NETCore.App 8.0.0-rc.1.23414.1 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```

The current test infrastructure has bugs for composite images that don't correctly test this environment variable: https://github.com/dotnet/dotnet-docker/issues/4827

When this is fixed the test will always fail when this variable is incoherent between the runtime and aspnet composite runtime images. Ideally we would compute the correct DOTNET_VERSION environment variable to use for the composite images, so that even if they're different between different image versions, the environment variable correctly describes which runtime version is in the Docker image.

richlander:
> You can see this today:

The problem isn't obvious to me. Can you elaborate?

lbussell:
Our environment variable says that we have .NET Runtime version `8.0.0-rc.1.23414.4` but the version actually in the container is `8.0.0-rc.1.23414.1`.

lbussell:
This issue also causes apps built on preview SDKs to be unable to run on composite images, with the following error:

```bash
You must install or update .NET to run this application.

App: /app/app.dll
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '8.0.0-rc.1.23414.4' (x64)
.NET location: /usr/share/dotnet/

The following frameworks were found:
  8.0.0-rc.1.23414.1 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]

Learn more:
[...]

PR #4837, opened 2023-08-21
Refactor ASP.NET Composite tests to be more correct

### LINKED ISSUE #4827: Tests fail to run correctly outside of CI (without an image-info.json)
Since the introduction of composite images, since [Aspnet_Composite](https://github.com/dotnet/dotnet-docker/blob/507e96decf488287c0feb3b27162390f2910ef03/tests/Microsoft.DotNet.Docker.Tests/DotNetImageType.cs#L13) is in the DotNetImageType enum, the tests try to pull images from a non-existent `aspnet-composite` mcr repo. Some of the tests rely on this enum value so some refactoring is necessary for tests to run locally. This doesn't happen in CI since it always runs with an image-info.json file which informs exactly which built images to run tests on.

To reproduce this, clone the latest commit in nightly (https://github.com/dotnet/dotnet-docker/commit/746066ee2bac8b000e5a9771b1a873a904878602) and run the test script:
```pwsh
pwsh .\tests\run-tests.ps1 -TestCategories "runtime-deps" -paths @("*8.0*alpine3.18*")
```

lbussell:
Fixes #4827. Does not fix https://github.com/dotnet/dotnet-docker/issues/4834. Please see https://github.com/dotnet/dotnet-docker/issues/4834#issuecomment-1687117811 for more context.

A few notes:
- Added `ImageVariant` and `SupportedImageTypes` to `ProductImageData`
  - This allows the ability to express images that don't support the full product image hierarchy
    - ex. `composite` only supports aspnet, `extra` and `aot` will only support runtime-deps, etc.
  - `SupportedImageTypes` allows `ProductImageData` to return the special image variant only when it's supported.
- Removed `AspnetCompositeImageTests`. Now composite images will use the same tests as aspnet. There's no difference in how they should be validated, and they previously shared 99% of functionality except for the override of the `DotNetImageType` property.
- With that, removed the `AspnetComposite` `DotNetImageType`. The values of `DotNetImageType` represent our literal MCR repos. We don't have an `aspnet-composite` MCR repo so this has to go. The test category is removed as well and will fall under the aspnet test category since it uses the same test class.
- Let the caller of environment variable checking code pick which image to check the environment variables from. This prevents trying to pull a non-existent `mcr.microsoft.com/dotnet/runtime:*-composite` image when trying to get the aspnet composite runtime version.


tests/Microsoft.DotNet.Docker.Tests/TestData.cs
```diff
@@ -22,6 +22,7 @@ public static class TestData
             new ProductImageData { Version = V6_0, OS = OS.Alpine318,                    Arch = Arch.Amd64 },
             new ProductImageData { Version = V6_0, OS = OS.Mariner20,                    Arch = Arch.Amd64 },
             new ProductImageData { Version = V6_0, OS = OS.Mariner20Distroless,          Arch = Arch.Amd64,   SdkOS = OS.Mariner20 },
+
```
mthalman, line 25: I assume this blank line is mean to be a separator between architectures? In that case, it would be good to apply the same for Arm64 and Arm. I see those are mixed together right now so those could be sorted properly to make this happen.

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -79,7 +79,8 @@ public void VerifyNoSasToken(ProductImageData imageData)
         [MemberData(nameof(GetImageData))]
         public void VerifyEnvironmentVariables(ProductImageData imageData)
         {
-            string version = imageData.GetProductVersion(ImageType, DockerHelper);
+            string imageName = imageData.GetImage(ImageType, DockerHelper);
+            string version = imageData.GetProductVersion(imageName, ImageType, DockerHelper);
```
mthalman, line 83: It seems like this pattern could be wrapped into its own method. It's used in a couple spots in this class as well as RuntimeImageTests.

mthalman (commented):
I really like these changes. It provides a cleaner implementation and provides more flexibility with the upcoming AOT work.

tests/Microsoft.DotNet.Docker.Tests/DotNetImageVariant.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class DotNetImageVariant
```
MichaelSimons, line 7: The terms ImageType and ImageVariant could be difficult to differentiate.  Would it make sense to rename ImageType to ImageRepo?  Perhaps something to consider in a separate PR.

tests/Microsoft.DotNet.Docker.Tests/ProductImageData.cs
```diff
@@ -12,6 +13,15 @@ public class ProductImageData : ImageData
         private string _sdkOS;
         private string _osTag;
         private ImageVersion? _versionFamily;
+        private string _imageVariant;
+
+        private DotNetImageType[] _supportedImageTypes = new[]
```
MichaelSimons, line 18: Consider makine DotNetImageType a flag enum.  It would be a more efficient data structure here as well as a more performant way to check what types are supported.  Lastly it would make the TestData declarations simpler.

MichaelSimons (commented):
+1 to @mthalman's comments on this being a nice refactoring that is making the tests better!

tests/Microsoft.DotNet.Docker.Tests/ProductImageData.cs
```diff
@@ -165,5 +163,7 @@ private string GetTagName(DotNetImageType imageType)
 
             return GetTagName(imageVersion.GetTagName(), os, variant);
         }
+
+        public bool ImageTypeIsSupported(DotNetImageType imageType) => (SupportedImageTypes & imageType) == imageType;
```
mthalman, line 167:
```suggestion
        public bool ImageTypeIsSupported(DotNetImageType imageType) => SupportedImageTypes.HasFlag(imageType);
```

tests/Microsoft.DotNet.Docker.Tests/DotNetImageVariant.cs
lbussell: I agree with this so I made the change. We're already touching so many of these files anyways. If you think it's too much for this PR we can revert it.

MichaelSimons: I looked over the changes, I thought it would be more impactful.  +1 to including it with these changes.

MichaelSimons, line 7: I could see this potentially having value being a flag enum as more items are defined.  Variants could be stacked, IDK.  The type safety is nice versus a string regardless.

tests/Microsoft.DotNet.Docker.Tests/TestData.cs
```diff
@@ -22,63 +22,82 @@ public static class TestData
...
+            new ProductImageData { Version = V8_0, OS = OS.JammyChiseled,                Arch = Arch.Amd64,   SdkOS = OS.Jammy,
+                    ImageVariant = DotNetImageVariant.Composite, SupportedImageRepos = DotNetImageRepo.Aspnet },
             new ProductImageData { Version = V8_0, OS = OS.Alpine318,                    Arch = Arch.Amd64 },
-            new ProductImageData { Version = V8_0, OS = OS.Alpine318Composite,           Arch = Arch.Amd64 },
+            new ProductImageData { Version = V8_0, OS = OS.Alpine318,                    Arch = Arch.Amd64,   SdkOS = OS.Alpine318,
+                    ImageVariant = DotNetImageVariant.Composite, SupportedImageRepos = DotNetImageRepo.Aspnet },
             new ProductImageData { Version = V8_0, OS = OS.Mariner20,                    Arch = Arch.Amd64 },
             new ProductImageData { Version = V8_0, OS = OS.Mariner20Distroless,          Arch = Arch.Amd64,   SdkOS = OS.Mariner20 },
-            new ProductImageData { Version = V8_0, OS = OS.Mariner20DistrolessComposite, Arch = Arch.Amd64,   SdkOS = OS.Mariner20 },
```
MichaelSimons, line 66: With the refactoring of how Composite are handled, the Arch column appears like it could be moved to the left one tab.

MichaelSimons (approved):

tests/Microsoft.DotNet.Docker.Tests/ProductImageData.cs
```diff
@@ -82,13 +82,16 @@ public DotNetImageRepo SupportedImageRepos
...
-            string.IsNullOrEmpty(_imageVariant) ? "" : $"-{_imageVariant}";
+            ImageVariant == DotNetImageVariant.None ? "" : $"-{GetImageVariantName(ImageVariant)}";
 
         public override string GetIdentifier(string type) => $"{VersionString}-{base.GetIdentifier(type)}";
 
         public static string GetImageRepoName(DotNetImageRepo imageRepo) =>
             Enum.GetName(typeof(DotNetImageRepo), imageRepo).ToLowerInvariant().Replace('_', '-');
 
+        public static string GetImageVariantName(DotNetImageVariant imageVariant) =>
+            Enum.GetName(typeof(DotNetImageVariant), imageVariant).ToLowerInvariant().Replace('_', '-');
```
MichaelSimons, line 93: Why is the Replace necessary here?  It seems like None should be handled here.

lbussell: I just made it the same as `GetImageRepoName`, so it would work if any variant we wanted had hyphens in the tag name. I will remove it until necessary.

mthalman (approved):

Issue #4844, opened 2023-08-23
Remove Kerberos explicit installation from 8.0 Dockerfiles


mthalman:
Based on the package installation principles provided in https://github.com/dotnet/dotnet-docker/pull/4842, Kerberos should not be listed for installation in any of the 8.0 Dockerfiles. Currently it's only been removed for Alpine as part of https://github.com/dotnet/dotnet-docker/issues/4751.

Relevant points of the Dockerfile templates:

https://github.com/dotnet/dotnet-docker/blob/4484f4b5fcee3e21594b3886b8d6751200c85977/eng/dockerfile-templates/Dockerfile.linux.install-deps#L32

https://github.com/dotnet/dotnet-docker/blob/4484f4b5fcee3e21594b3886b8d6751200c85977/eng/dockerfile-templates/Dockerfile.linux.install-deps#L41

https://github.com/dotnet/dotnet-docker/blob/4484f4b5fcee3e21594b3886b8d6751200c85977/eng/dockerfile-templates/Dockerfile.linux.install-deps#L50

mthalman:
@richlander - You mentioned at https://github.com/dotnet/dotnet-docker/issues/4804#issuecomment-1674019454 that we could include Kerberos in the `extra` images. Still on board with that?

mthalman:
Actually, if we're including them in `extra`, then the removal from the "full" distro images shouldn't be done since we already decided that the packages added in `extra` would also be represented in the full images.

mthalman:
> @richlander - You mentioned at https://github.com/dotnet/dotnet-docker/issues/4804#issuecomment-1674019454 that we could include Kerberos in the `extra` images. Still on board with that?

After discussion, we settled on not including Kerberos in the `extra` images since it's such an edge case.

lbussell:
This is fixed in https://github.com/dotnet/dotnet-docker/pull/4860

PR #4847, opened 2023-08-24
Add runtime-deps `extra` Dockerfiles and conform to new packaging strategy

lbussell:
Following the guidelines laid out in https://github.com/dotnet/dotnet-docker/discussions/4821
- add `extra` runtime-deps images 
  - `extra` images contain `icu` and `tzdata`
  - distros that aren't focused on small deployment size also have `tzdata` added - Mariner 2.0, Ubuntu, and Debian
- Remove Kerberos in all 8.0 images according to https://github.com/dotnet/dotnet-docker/issues/4844


MichaelSimons (commented):
Readme need regenerating.

src/runtime-deps/8.0/alpine3.18-extra/amd64/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+ENV \
+    # UID of the non-root user 'app'
+    APP_UID=1654 \
+    # Configure web servers to bind to port 8080 when present
+    ASPNETCORE_HTTP_PORTS=8080 \
+    # Enable detection of running in a container
+    DOTNET_RUNNING_IN_CONTAINER=true \
+    # Set the invariant mode since ICU package isn't included (see https://github.com/dotnet/announcements/issues/20)
+    DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=true
```
MichaelSimons, line 11: This should be removed from the extra images since they add icu.

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -86,60 +86,72 @@ public void VerifyDistrolessPackages(ProductImageData imageData)
...
-                    "mariner-release",
-                    "openssl",
-                    "openssl-libs",
-                    "prebuilt-ca-certificates",
-                    "tzdata",
-                    "zlib"
-                };
-            }
-            else if (imageData.OS == OS.JammyChiseled)
+                { IsDistroless: true, OS: string os } when os.Contains(OS.Mariner) => new[]
```
MichaelSimons, line 98: What is the dependency on `IsDistroless: true` here?  And why is it not necessary for Jammy?

src/runtime-deps/8.0/alpine3.18-extra/amd64/Dockerfile
MichaelSimons: I would also expect to see some tests need refactoring once removed.

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
MichaelSimons, line 1: Given the extra Dockerfiles aren't based on the non-extra Docker images, Do you think there is value in authoring a test that validates the contents are identical minus the couple expected differences?

lbussell: I was thinking that VerifyDistrolessPackages could have a non-distroless counterpart or that it should just be called VerifyInstalledPackages and check all images including non-distroless.

However I don't like the duplication of logic between here and the templates.

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -86,60 +86,72 @@ public void VerifyDistrolessPackages(ProductImageData imageData)
...
-                    "ca-certificates",
-                    "libc6",
-                    "libgcc-s1",
-                    "libssl3",
-                    "libstdc++6",
-                    "zlib1g"
-                };
-            }
-            else
+                { IsDistroless: true, OS: string os } when os.Contains(OS.Mariner) => new[]
```
mthalman, line 126: This code makes me feel old. In my day, we used a bunch more statements to do this, and we liked it.

mthalman (commented):
The Alpine sdk image should be adding `tzdata`. It needs to have parity with the packages that are installed in the Alpine runtime-deps `extra` image.

eng/dockerfile-templates/sdk/Dockerfile.linux
```diff
@@ -11,12 +11,20 @@
...
+                    "icu-libs",
+                    "git"
+                ],
+                [
+                    "curl",
+                    "icu-data-full",
+                    "icu-libs",
+                    "git",
+                    "tzdata"
+                ]),
```
mthalman, line 27: This seems indented too far.

eng/dockerfile-templates/sdk/Dockerfile.linux
```diff
@@ -11,12 +11,20 @@
...
+                    "icu-data-full",
+                    "icu-libs",
+                    "git"
+                ],
+                [
+                    "curl",
+                    "icu-data-full",
+                    "icu-libs",
+                    "git",
+                    "tzdata"
```
mthalman, line 26: It'd be good if we had a test that could catch this for any future things we might add to `extra` images. Alpine is the outlier here. We want to ensure that `sdk` images contain the same (or more) packages that are contained in `extra` images.

lbussell: I am writing a test for this now.

src/sdk/8.0/alpine3.18/amd64/Dockerfile
```diff
@@ -21,7 +21,8 @@ RUN apk add --upgrade --no-cache \
         curl \
         icu-data-full \
         icu-libs \
-        git
+        git \
```
MichaelSimons, line 24: I see this was an existing issue, the package list should be sorted alphabetically.

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -73,80 +71,155 @@ public void VerifyDefaultUser(ProductImageData imageData)
...
                 return;
             }
+            Assert.NotEmpty(GetOSReleaseInfo(imageData));
+        }
 
-            const string SyftImage = "anchore/syft:v0.84.0";
-            DockerHelper.Pull(SyftImage);
+        [LinuxImageTheory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifyRuntimeDepsPackages(ProductImageData imageData)
```
MichaelSimons, line 87: The RuntimeDeps feels unncessary given the scope of the method.

```suggestion
        public void VerifyInstalledPackages(ProductImageData imageData)
```

tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
```diff
@@ -204,6 +206,62 @@ private IEnumerable<string> GetInstalledRpmPackages(ProductImageData imageData)
             }
         }
 
+        /// <summary>
+        /// Verifies that the packages installed are correct and scannable by security tools.
+        /// </summary>
+        protected void VerifyInstalledPackages(ProductImageData imageData, DotNetImageRepo imageRepo)
+        {
+            IEnumerable<string> basePackages = imageData.IsDistroless
+                ? RuntimeDepsImageTests.GetDistrolessBasePackages(imageData)
```
MichaelSimons, line 215: The circular dependency between ProductImageTests and RuntimeDepsImageTests smells.  Given the usage, it feels like this better belongs in the RuntimeDepsImageTests.

tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
```diff
@@ -204,6 +206,62 @@ private IEnumerable<string> GetInstalledRpmPackages(ProductImageData imageData)
...
+                        .Concat(RuntimeDepsImageTests.GetRuntimeDepsPackages(imageData))
+                : RuntimeDepsImageTests.GetRuntimeDepsPackages(imageData);
```
MichaelSimons, line 217: Consider refactoring, I think the following would be more readable.  I also think it would help if there was just one list of packages being built up versus basePackages and expectedPackages

IEnumerable<string> expectedPackages = RuntimeDepsImageTests.GetRuntimeDepsPackages(imageData);

If (imageData.IsDistroless)
    expectedPackages = expectedPackages.Concat(RuntimeDepsImageTests.GetDistrolessBasePackages(imageData));

tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
```diff
@@ -204,6 +206,62 @@ private IEnumerable<string> GetInstalledRpmPackages(ProductImageData imageData)
...
+            IEnumerable<string> actualPackages = GetInstalledPackages(imageData);
+
+            OutputHelper.WriteLine($"Expected Packages: [ {string.Join(", ", expectedPackages)} ]");
+
+            if (imageData.IsDistroless)
+            {
+                OutputHelper.WriteLine($"Actual Packages: [ {string.Join(", ", actualPackages)} ]");
+                Assert.Equal(expectedPackages, actualPackages);
+            }
+            else
```
MichaelSimons, line 233: Why is there different validation logic based on whether the image is distroless or not?

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -73,80 +71,155 @@ public void VerifyDefaultUser(ProductImageData imageData)
...
+        private string GetOSReleaseInfo(ProductImageData imageData)
+        {
+            JsonNode output = GetSyftOutput("os-release-info", imageData);
+            JsonObject distro = (JsonObject)output["distro"];
+            return (string)distro["version"];
+        }
 
-            string[] expectedPackages;
-            if (imageData.OS.Contains(OS.Mariner))
+        internal static IEnumerable<string> GetDistrolessBasePackages(ProductImageData imageData) => imageData switch
```
MichaelSimons, line 99: With the current design, it feels a bit odd to have to internal methods to separate out the base and declared package when the caller wants the full set.  Consider providing one method the dependent class calls into instead of two.  You could still have two private methods.  This provides a cleaner "public" contract IMO.

tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
lbussell: If we're inspecting a distroless image, we validate that the expected packages all installed and that they are the *only* packages installed.

If we're inspecting a non-distroless image, there are a lot more packages installed, so we just want to check that all of the deps packages are in the list of installed packages. This is also why we exclude the distroless base packages above, those packages don't exist in the non-distroless images.

tests/Microsoft.DotNet.Docker.Tests/ProductImageTests.cs
lbussell: I moved it back to RuntimeDepsImageTests.

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
lbussell: Since resolving https://github.com/dotnet/dotnet-docker/pull/4847#discussion_r1307411593, we can't have two methods with the same name but different signatures on RuntimeDepsImageTests. Would renaming the actual test logic method to something like `VerifyInstalledPackagesBase` work instead?

MichaelSimons: I'm not making the connection on why method overloads don't work.  Where does this constraint come from?

lbussell: It is because I made `VerifyInstalledPackages` static so that I can call it from SdkImageTests without creating a `RuntimeDepsImageTests` instance. Can't have static and instance methods with the same name.

MichaelSimons: Thanks, I overlooked static vs instance earlier.  How about VerifyInstalledPackagesExtra...j/k

I can't come up with anything better than VerifyInstalledPackagesBase

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -73,80 +74,232 @@ public void VerifyDefaultUser(ProductImageData imageData)
...
+                        "libicu70",
+                        "tzdata"
+                    },
+                { OS: string os } when os.Contains(OS.Alpine) => new[]
+                    {
+                        "icu-data-full",
+                        "icu-libs",
+                        "tzdata"
+                    },
+                _ => new string[0]
```
mthalman, line 301: nit
```suggestion
                _ => Array.Empty<string>()
```

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -136,76 +144,136 @@ private JsonNode GetSyftOutput(string name, ProductImageData imageData)
...
+                        "zlib1g"
+                    },
+                { OS: OS.Alpine318 } => new[]
+                    {
+                        "ca-certificates-bundle",
+                        "libgcc",
+                        "libssl3",
+                        "libstdc++",
+                        "zlib"
+                    },
```
mthalman, line 257: This is unlikely to change between Alpine versions. Can we generalize this condition for all Alpine versions?

eng/dockerfile-templates/Dockerfile.linux.install-deps
```diff
@@ -1,30 +1,76 @@
...
                 "libgcc",
-                "libintl",
                 cat("libssl", VARIABLES[cat("libssl|", OS_VERSION_BASE)]),
-                "libstdc++",
                 "zlib"
             ],
             [
-                "libgcc",
+                "libc6",
+                "libgcc-s1",
```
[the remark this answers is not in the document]
lbussell: Couple reasons,

1. The condition here was rewritten so the git diff makes it look like `libgcc` was changed to `libgcc-s1`. The `libgcc` package is valid for Alpine and Mariner only. The `libgcc-s1` in this condition is for Ubuntu/Debian now.

2. The `libgcc1` package we were using in Debian/Ubuntu is an alias to `libgcc-s1`, so I replaced it for its actual package in .NET 8.0:
```
root@163ec2e1e49c:/# apt-get install libgcc1
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'libgcc-s1' instead of 'libgcc1'
libgcc-s1 is already the newest version (12.2.0-14).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

Also see https://packages.ubuntu.com/search?keywords=libgcc1&searchon=names&suite=jammy&section=all.

mthalman (commented):
Looks like CI is having some issues with the `VerifyInstalledPackages` test.

lbussell:
> Looks like CI is having some issues with the `VerifyInstalledPackages` test.

Yeah, I'm investigating it. It doesn't repro locally so that's frustrating. The process output is cut off in the test in PR validation.

mthalman:
> Yeah, I'm investigating it. It doesn't repro locally so that's frustrating. The process output is cut off in the test in PR validation.

As a workaround, perhaps it needs to be output to a file and extracted from the container.

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -153,12 +155,26 @@ public void VerifyInstalledPackages(ProductImageData imageData)
...
+
             string output = dockerHelper.Run(
-                SyftImage, name, $"packages docker:{imageName} -o json",
-                useMountedDockerSocket: true);
+                SyftImage,
+                name,
+                $"packages docker:{imageName} -o template -t {containerTemplatePath}",
+                optionalRunArgs: $" -v {localTemplateDirectory}:{containerTemplateDirectory}",
+                useMountedDockerSocket: true
+            );
```
lbussell, line 175: Well this works great locally but it seems that bind mounting from docker-in-docker doesn't behave as expected... I'll try to write out to a file next maybe. It's also a little challenging since syft is distroless.

mthalman: > It's also a little challenging since syft is distroless.

You should need to use Bash though, right? It can directly output to a file: `-o json=<file>`

tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -22,7 +22,7 @@ public class DockerHelper
         public static bool IsLinuxContainerModeEnabled => string.Equals(DockerOS, "linux", StringComparison.OrdinalIgnoreCase);
         public static string TestArtifactsDir { get; } = Path.Combine(Directory.GetCurrentDirectory(), "TestAppArtifacts");
 
-        private ITestOutputHelper OutputHelper { get; set; }
+        public ITestOutputHelper OutputHelper { get; set; }
```
MichaelSimons, line 25: Why is this change needed?

lbussell: It's not. I was using it for debugging.

tests/Microsoft.DotNet.Docker.Tests/RuntimeDepsImageTests.cs
```diff
@@ -73,81 +74,253 @@ public void VerifyDefaultUser(ProductImageData imageData)
...
+            DotNetImageRepo imageRepo,
+            DockerHelper dockerHelper)
+        {
+            const string SyftImage = "anchore/syft:v0.87.1";
+            dockerHelper.Pull(SyftImage);
+
+            string imageToInspect = imageData.GetImage(imageRepo, dockerHelper);
+
+            string outputContainerFilePath = "/artifacts/output.json";
+            string tempDir = Directory.CreateDirectory(
```
mthalman, line 158: Should delete this directory after the file's contents are read.

mthalman (approved):

mthalman (approved):

PR #4857, opened 2023-09-06
Remove unnecessary custom build leg groups that were causing errors

lbussell:
This still fixes the root cause issue that prompted https://github.com/dotnet/dotnet-docker/pull/4856, but prevents the internal build (`PlatformDependencyGraph` matrix) from complaining that we have dependencies on multiple Dockerfile graphs.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -925,18 +925,10 @@
               },
               "customBuildLegGroups": [
                 {
-                  "name": "pr-build",
-                  "type": "Supplemental",
-                  "dependencies": [
-                    "$(Repo:sdk):8.0-alpine3.18-amd64"
-                  ]
-                },
-                {
                   "name": "test-dependencies",
                   "type": "Integral",
                   "dependencies": [
-                    "$(Repo:sdk):8.0-alpine3.18-aot-amd64",
-                    "$(Repo:runtime-deps):8.0-alpine3.18-amd64"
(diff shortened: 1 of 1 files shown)


mthalman (commented):
This runs into this error:
```
System.InvalidOperationException: Duplicate leg name(s) found in matrix 'linuxArm32': 8.0-alpine3.18-runtime-deps
```

Can this be fixed, for example, by including both `"$(Repo:sdk):8.0-jammy-arm32v7"` and `"$(Repo:sdk):8.0-jammy-aot-arm64v8"` for the arm32 Dockerfile?

lbussell:
> Can this be fixed, for example, by ...

That actually does work, mostly, I pushed that change. I also tested it against the other two scenarios that failed lately (generating internal build and internal test matrices).

I think I am seeing the bigger issue now, though. ImageBuilder is grouping legs by architecture before integral test dependencies. We want to build and test arm32 AOT images in the same leg as arm64 AOT SDK images. ImageBuilder will need changes to the GenerateBuildMatrix command to allow custom build leg dependencies to transcend architecture like that. It might even be lower effort to produce arm32 AOT SDKs.

Since we don't have AOT tests yet, for now we should just accept that the new arm32 runtime deps Dockerfiles are going to build in their own legs separate from arm64.

mthalman (approved):

PR #4859, opened 2023-09-08
Add native AOT image tests

lbussell:
Related: https://github.com/dotnet/dotnet-docker/issues/4129, https://github.com/dotnet/dotnet-docker/pull/4822

I added support for the `VerifyInstalledPackages` tests and added a new test, `VerifyAotAppScenario` that runs the new (with 8.0 Preview 7) `webapiaot` sample. I found that it was easier to just hit the `/todos` endpoint instead of trying to programatically edit the .csproj of some other sample to make it publish with AOT compilation.


tests/Microsoft.DotNet.Docker.Tests/DotNetImageVariant.cs
```diff
@@ -10,7 +10,8 @@ namespace Microsoft.DotNet.Docker.Tests
     public enum DotNetImageVariant
     {
         None        = 0,
-        Composite   = 1 << 0,
-        Extra       = 1 << 1
+        AOT       = 1 << 0,
```
MichaelSimons, line 13:
```suggestion
        AOT         = 1 << 0,
```

tests/Microsoft.DotNet.Docker.Tests/TestData.cs
```diff
@@ -14,196 +14,212 @@ public static class TestData
     {
         private static readonly ProductImageData[] s_linuxTestData =
         {
-            new ProductImageData { Version = V6_0, OS = OS.BullseyeSlim,        Arch = Arch.Amd64 },
```
MichaelSimons, line 17: The diff makes it difficult to tell if there are any changes aside from the new syntax.  Can you call out any functional changes made?

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/Dockerfile.linux
```diff
@@ -61,3 +62,29 @@ EXPOSE $port
...
+EXPOSE $port
+
+WORKDIR /source
+COPY NuGet.config .
+WORKDIR /source/app
+COPY app/*.csproj .
+RUN dotnet restore -r $rid
+
+COPY app/ .
+RUN dotnet publish -r $rid --no-restore -o /app
```
MichaelSimons, line 82: It looks like the build steps are the same as the existing build stage.  Could this be refactored to share the common build stage and then introduce a new publish_aot_app stage?  

Should the publish command include `-c Release`?

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -331,7 +346,12 @@ private async Task RunTestAppImage(string image, string user = null, string comm
 
                 if (_isWeb && !Config.IsHttpVerificationDisabled)
                 {
-                    await VerifyHttpResponseFromContainerAsync(containerName, _dockerHelper, _outputHelper, _imageData.DefaultPort);
+                    await VerifyHttpResponseFromContainerAsync(
+                        containerName,
+                        _dockerHelper,
+                        _outputHelper,
+                        _imageData.DefaultPort,
+                        pathAndQuery: _isAot ? "todos" : null);
```
MichaelSimons, line 354: Why is this `todos`?

tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs
```diff
@@ -38,11 +39,18 @@ public class ImageScenarioVerifier
...
+            _isAot = _imageData.ImageVariant.HasFlag(DotNetImageVariant.AOT);
         }
 
         public async Task Execute()
         {
-            string solutionDir = CreateTestSolutionWithSdkImage(_isWeb ? "web" : "console");
+            string appType = _isAot && _isWeb
+                ? "webapiaot"
+                : _isWeb
+                    ? "console"
```
MichaelSimons, line 50: Should `console` and `web` be inverted here?

tests/Microsoft.DotNet.Docker.Tests/TestData.cs
lbussell: I will go back and either remove non-functional changes or split them into different commits.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/Dockerfile.linux
mthalman: Are more changes coming for this, @lbussell?

tests/Microsoft.DotNet.Docker.Tests/ProductImageData.cs
```diff
@@ -97,8 +73,16 @@ public DotNetImageRepo SupportedImageRepos
...
         public string GetImage(DotNetImageRepo imageRepo, DockerHelper dockerHelper)
         {
-            // ASP.NET composite includes its own runtime that we want to test
-            if (ImageVariant == DotNetImageVariant.Composite && imageRepo == DotNetImageRepo.Runtime)
+            // Aspnet or Runtime images are irrelevant for the AOT image variant.
+            // AOT runtime-deps images should be able to run any AOT compiled console or web apps.
+            if (ImageVariant.HasFlag(DotNetImageVariant.AOT)
+                && (imageRepo == DotNetImageRepo.Aspnet || imageRepo == DotNetImageRepo.Runtime))
+            {
+                imageRepo = DotNetImageRepo.Runtime_Deps;
```
mthalman, line 81: I'm curious in what scenario would this code path be hit? Since the TestData specifies that AOT images are to be used for runtime-deps only, how would we ever request aspnet or runtime in an AOT scenario?

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/Dockerfile.linux
lbussell: Pushed the changes.

tests/Microsoft.DotNet.Docker.Tests/ProductImageData.cs
lbussell: This code path is called from BuildTestAppImage:
https://github.com/lbussell/dotnet-docker/blob/29d85b4e30e306cd8ddeeb0d32fe9ad6faa37497/tests/Microsoft.DotNet.Docker.Tests/ImageScenarioVerifier.cs#L218-L220

It will try to call GetImage with the Aspnet image type because we are building a web sample. If we don't have this logic, then we end up getting the regular aspnet version (for example, `aspnet:8.0-alpine3.18` instead of `runtime-deps:8.0-alpine3.18-aot`). It works silently because aspnet has all the requirements for running AOT apps. However, we're not testing the image we want.

Maybe we should error in this case and fix all of the calls to this method.

lbussell: I pushed a change like that. It could be rolled back just as easily.

mthalman: Yeah, I like that better. Looks like it needs to handle Monitor though.

mthalman (approved):

MichaelSimons (approved):

devnoname120:
@lbussell It has been available in nightly for a year already. Could you consider merging it to stable? Thanks!

PR #4860, opened 2023-09-08
.NET Container Images Servicing September 2023

lbussell:


README.monitor-base.md
```diff
@@ -45,13 +45,13 @@ The following Dockerfiles demonstrate how you can use this base image to build a
 ##### .NET Monitor 8 Preview Tags
 Tags | Dockerfile | OS Version
 -----------| -------------| -------------
-8.0.0-preview.7-ubuntu-chiseled-amd64, 8.0-preview-ubuntu-chiseled-amd64, 8-preview-ubuntu-chiseled-amd64, 8.0.0-preview.7-ubuntu-chiseled, 8.0-preview-ubuntu-chiseled, 8-preview-ubuntu-chiseled, 8.0.0-preview.7, 8.0-preview, 8-preview, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/main/src/monitor-base/8.0/ubuntu-chiseled/amd64/Dockerfile) | Ubuntu 22.04
+8.0.0-preview.7-ubuntu-chiseled-amd64, 8.0-ubuntu-chiseled-amd64, 8-ubuntu-chiseled-amd64, 8.0.0-preview.7-ubuntu-chiseled, 8.0-ubuntu-chiseled, 8-ubuntu-chiseled, 8.0.0-preview.7, 8.0, 8, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/main/src/monitor-base/8.0/ubuntu-chiseled/amd64/Dockerfile) | Ubuntu 22.04
```
jander-msft, line 48: If this goes in as-is, then the .NET Monitor 8.0 Preview 7 tags will point to an image that has .NET Monitor 8.0 Preview 7 sitting on top of .NET 8 RC 1. I don't know if this combination actually works, so this change might be destroying the .NET Monitor 8.0 Preview 7 images.

Is there some way that .NET Monitor could be completely excluded and then we'll follow up with a PR that takes care of all of the .NET Monitor changes? Or maybe that the actual build can exclude .NET Monitor until the .NET Monitor PR is merged? Let's briefly discuss our options here before this is merged in.

jander-msft: I believe that the .NET Monitor images will be filtered when running the build, but I won't if we want to "hedge a bet" and artificially bump these to RC.1 in this PR. This means if the official build accidentally includes the .NET Monitor images, then the Preview 7 tags aren't "poisoned" and the PR I opened will reconcile the RC 1 tags to the correct .NET Monitor binaries.

mthalman: We should exclude any changes that alter the Monitor tags. The tags should only be updated as part of https://github.com/dotnet/dotnet-docker/pull/4862.

@lbussell - Have you run a test build to verify we can build/test/publish everything except Monitor images?

src/monitor/7.1/alpine/amd64/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG REPO=mcr.microsoft.com/dotnet/aspnet
-FROM $REPO:7.0.10-alpine3.18-amd64
+FROM $REPO:7.0.11-alpine3.18-amd64
```
mthalman, line 2: We should avoid this change until https://github.com/dotnet/dotnet-docker/pull/4862.

src/monitor-base/8.0/cbl-mariner-distroless/amd64/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN dotnet_monitor_version=8.0.0-preview.7.23402.4 \
 
 
 # .NET Monitor Base image
-FROM $REPO:8.0.0-preview.7-cbl-mariner2.0-distroless-amd64
+FROM $REPO:8.0.0-rc.1-cbl-mariner2.0-distroless-amd64
```
mthalman, line 23: We should also avoid this change until https://github.com/dotnet/dotnet-docker/pull/4862.

README.monitor-base.md
lbussell: > Have you run a test build to verify we can build/test/publish everything except Monitor images?

I will run one. I'm working on reverting the changes to monitor Dockerfiles.

mthalman (approved):

mthalman (commented):
I guess there's an issue with matrix generation.

lbussell:
It looks like pinning the dotnet versions for monitor won't work with the matrix generation. However, some whitespace got trimmed when I edited the monitor templates so I'll see if removing that change causes us to skip the build..

PR #4867, opened 2023-09-12
Merge main to nightly for September 2023 servicing release.

lbussell:
Related: https://github.com/dotnet/dotnet-docker/pull/4860


manifest.json
```diff
@@ -1301,7 +1301,7 @@
                   "name": "test-dependencies",
                   "type": "Integral",
                   "dependencies": [
-                    "$(Repo:sdk):8.0-jammy-arm64v8"
+                    "$(Repo:sdk):8.0-jammy-arm32v7"
```
mthalman, line 1304: Is this right?

mthalman: If this is expected, feel free to merge.

lbussell: Yes, it's correct, it's extra not aot, so we want to test with the regular arm32 Jammy SDK. I got it wrong in nightly and corrected it with the PR to main. So now it shows up here.

lbussell:
Since .NET 8.0 RC1 is a security release, the bits aren't on `dotnetbuilds`. However the first RC2 builds will presumably be in `dotnetbuilds`, so I pinned the Dockerfiles to the `dotnetcli` URL until we get the first build so that update-dependencies can still download from the correct URL.

manifest.versions.json
```diff
@@ -51,6 +51,8 @@
     "base-url|7.0|nightly": "$(base-url|public|maintenance|nightly)",
     "base-url|8.0|main": "$(base-url|public|main)",
     "base-url|8.0|nightly": "$(base-url|public|nightly)",
```
mthalman, line 53: Can't this be fixed without all these template changes by changing this to `$(base-url|public|maintenance|nightly)`?

lbussell: Then update-dependencies will fail to hit the `dotnetbuilds` URL when there is a new build of .NET 8 RC2. Would that not be an issue?

lbussell: This complicates the SDK content tests as well. Maybe it'd be best to keep an eye on the update-dependencies pipeline and go with your solution.

mthalman (approved):

PR #4877, opened 2023-09-14
Update image sizes

lbussell:


mthalman (approved):
I assume these changes are due to the removal of `krb5` in https://github.com/dotnet/dotnet-docker/pull/4847.

lbussell:
Pushed some more changes from Sept. 2023 Servicing. Explanations:

- Jammy non-distroless increased due to installation of tzdata
- Jammy chiseled arm32 size increase due to change in base image size (file `/lib/arm-linux-gnueabihf/ld-linux-armhf.so.3` increased in size)

mthalman (approved):

PR #4882, opened 2023-09-14
Remove package install

richlander:


mthalman (commented):
Doing a search for `clang`, I see that Dockerfile.debian is installing it too. Shouldn't that be using the `aot` SDK image?

lbussell:
> Doing a search for `clang`, I see that Dockerfile.debian is installing it too. Shouldn't that be using the `aot` SDK image?

There aren't any AOT images for Debian.

mthalman:
> > Doing a search for `clang`, I see that Dockerfile.debian is installing it too. Shouldn't that be using the `aot` SDK image?
> 
> There aren't any AOT images for Debian.

Ugh. And yet we show an AOT sample for that distro. 🤔

richlander:
Sure. The Debian sample is to show people what to do if they want to use Debian.

mthalman (approved):

PR #4892, opened 2023-09-19
Update .NET Monitor 8 nightly builds url and use latest RC2 build

jander-msft:
- Update the base URL for .NET Monitor 8 to use dotnetbuilds instead of dotnetcli. This allows the nightly ingestion to pickup builds correctly instead of failing to get them from dotnetcli.
- Update .NET Monitor 8 to the latest RC2 build.

### DIFF
--- a/README.monitor-base.md
+++ b/README.monitor-base.md
@@ -51,13 +51,13 @@ The following Dockerfiles demonstrate how you can use this base image to build a
 ##### .NET Monitor 8 Preview Tags
 Tags | Dockerfile | OS Version
 -----------| -------------| -------------
-8.0.0-rc.1-ubuntu-chiseled-amd64, 8.0-ubuntu-chiseled-amd64, 8-ubuntu-chiseled-amd64, 8.0.0-rc.1-ubuntu-chiseled, 8.0-ubuntu-chiseled, 8-ubuntu-chiseled, 8.0.0-rc.1, 8.0, 8, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/monitor-base/8.0/ubuntu-chiseled/amd64/Dockerfile) | Ubuntu 22.04
+8.0.0-rc.2-ubuntu-chiseled-amd64, 8.0-ubuntu-chiseled-amd64, 8-ubuntu-chiseled-amd64, 8.0.0-rc.2-ubuntu-chiseled, 8.0-ubuntu-chiseled, 8-ubuntu-chiseled, 8.0.0-rc.2, 8.0, 8, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/monitor-base/8.0/ubuntu-chiseled/amd64/Dockerfile) | Ubuntu 22.04
 
 ## Linux arm64 Tags
 ##### .NET Monitor 8 Preview Tags
 Tags | Dockerfile | OS Version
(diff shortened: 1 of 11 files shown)


mthalman (approved):

jander-msft:
@lbussell, why did you push that commit? This PR shouldn't be regenerating Dockerfiles for non-.NET Monitor images. Also, it's suspicious that the runtime version was updated to the same exact value that I used for .NET Monitor. Is that a coincidence?

lbussell:
> @jander-msft why did you push that commit

The PR was failing because Dockerfiles were out of sync. In your original commit you changed the runtime version.

Would it be better to wait on that runtime version to flow naturally? It's even newer than what's in https://github.com/dotnet/dotnet-docker/pull/4886 I think. (8.0.0-rc.2.23466.4 vs. 8.0.0-rc.2.23468.7). If your version of monitor will run without the updated runtime we could just revert that too.

There may be some weird GitHub merging going on.

jander-msft:
> The PR was failing because Dockerfiles were out of sync. In your original commit you changed the runtime version.

Okay, I see that now. That shouldn't be happening though. I started with a clean `nightly` branch, fixed the base url for .NET Monitor, and executed `.\eng\Set-DotnetVersions.ps1 8.0 -MonitorVersion 8.0.0-rc.2.23468.7`. There might be something incorrect in the dependency update tool. I'll try to investigate that.

> If your version of monitor will run without the updated runtime we could just revert that too.

I had no intention of updating the runtime version.

lbussell:
@jander-msft I just tried to debug that situation and wasn't able to get it to change the runtime version. Please feel free to revert my commit and undo the runtime version change, or I can do it quickly.

lbussell (approved):

mthalman (approved):

eerhardt:
@mthalman @lbussell - is this why when I do:

`docker run --rm mcr.microsoft.com/dotnet/nightly/aspnet:8.0.0-rc.2 dotnet --info`

I am currently getting a really old (~20 days) build:

```
Host:
  Version:      8.0.0-rc.2.23431.9
  Architecture: arm64
  Commit:       3c48925a6c
  RID:          linux-arm64

.NET SDKs installed:
  No SDKs were found.

.NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.0-rc.2.23455.8 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 8.0.0-rc.2.23431.9 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```

cc @ellismg

MichaelSimons:
>@mthalman @lbussell - is this why when I do:

You likely have a cached image on your machine.  Looking at https://mcr.microsoft.com/en-us/product/dotnet/nightly/aspnet/tags, this tag was rebuilt on 9/15.  Try re-pulling the image.

```
docker run --rm --pull always mcr.microsoft.com/dotnet/nightly/aspnet:8.0.0-rc.2 dotnet --info
```

lbussell:
@MichaelSimons, Eric is right and that's the current build.

@eerhardt, you will probably want to watch for https://github.com/dotnet/dotnet-docker/pull/4886 to be merged, that should update to what was the latest installer build as of this morning. If you want even newer builds of ASP.NET/Runtime/etc. on a regular basis, please open an issue and we will discuss during triage how we could accomplish that.

PR #4895, opened 2023-09-20
Add DebugType flag

richlander:


MichalStrehovsky:
Note that this flag makes it impossible to investigate coredumps or do anything meaningful in a debugger attach scenario.

Setting `DebugType=none` on an app that runs in a .NET VM, one can still at least see the stack in the debugger. One might even see the values of method parameters. The VM can construct all of those without the PDB because the metadata is elsewhere.

Setting `DebugType=none` on a native AOT compiled app, the metadata is gone. Attaching a debugger will show stack that looks something like this:

```
app (offset 0xfeefeed)
app (offset 0xf00f00d)
app (offset 0xd00d00)
/system/lib64/libc.so (_start+32)
```

Locals and parameters will be unavailable in any frame.

MichaelSimons:
I assume this is done for size reasons.  Can you include the impact of this change?

mthalman:
Please post a description of the motivation for this change.

samples/releasesapi/Dockerfile
```diff
@@ -9,8 +9,7 @@ RUN dotnet restore -r linux-$TARGETARCH
 
 # copy and publish app and libraries
 COPY . .
-RUN dotnet publish -r linux-$TARGETARCH --no-restore -o /app releasesapi.csproj
-RUN rm /app/*.dbg /app/*.Development.json
```
eerhardt, line 13: Note that with this change you will get `/app/appsettings.Development.json` copied into your container

richlander:
The motivation for the change is to replace a whole line with a flag, as the PR does. I am assuming that they are equivalent w/rt not including symbols in the final container. Correct?

The `.json` file isn't worth keeping a line in the `Dockerfile` to delete it. If we think it is important to remove it, we should either do that automatically or offer a flag to do it.

MichalStrehovsky:
> The motivation for the change is to replace a whole line with a flag, as the PR does. I am assuming that they are equivalent w/rt not including symbols in the final container. Correct?

The compiler is deterministic - so if you compile something and delete debug symbols, but then compile again with _exactly the same inputs_, the debug info could be usable for the first compilation if push comes to shove.

Not generating debug info messes with flags in RyuJIT. It's unlikely a binary generated with DebugType=none would be identical to the one without that.

richlander:
I'll wait for the change from @agocke.

PR #4899, opened 2023-09-20
Switch AMD64 sample to composite

richlander:


mangod9 (approved):

mthalman (approved):
This puts amd64 in alignment with the Arm32/64 Dockerfiles.

mthalman (commented):
Dockerfiles need to be regenerated.

richlander:
I assume you mean that READMEs. I just regenerated them.

mthalman (approved):

lbussell:
Do we need to update the Dockerfile suffixes in the test data? https://github.com/dotnet/dotnet-docker/blob/7ce7e75a1699e6fdb6a6e84449c9595dbdd8a6bc/tests/Microsoft.DotNet.Docker.Tests/TestData.cs#L131-L133

@jander-msft

jander-msft:
> Do we need to update the Dockerfile suffixes in the test data?
> 
> https://github.com/dotnet/dotnet-docker/blob/7ce7e75a1699e6fdb6a6e84449c9595dbdd8a6bc/tests/Microsoft.DotNet.Docker.Tests/TestData.cs#L131-L133
> 
> @jander-msft

For consistency sake, yes, I believe that needs to be updated. Though, I don't think it impacts the .NET Monitor tests since they are pulling the sample images rather than building them as part of test runs.

lbussell:
It seems this change isn't applicable anymore. Should we close it @richlander, and re-evaluate later?

mangod9:
why is it not applicable? Was the change done as part of a different PR?

lbussell:
> @mangod9: why is it not applicable?

Due to unforseen circumstances we needed to change the samples to Alpine instead of Ubuntu Chiseled. We could probably still switch the samples to alpine3.18-composite.

richlander:
Ya. Let's close this and update the samples per our updated plans, which include using composite.

PR #4906, opened 2023-09-22
Change aspnetapp sample to be based on alpine

lbussell:


richlander (approved):

richlander:
We ran into a customer challenge. This is a temporary change.

PR #4917, opened 2023-09-28
Update for the v7.4.0-preview.6 release

daxian-dbw:
Update for the v7.4.0-preview.6 release


lbussell:
@daxian-dbw, does this fix https://github.com/PowerShell/PowerShell/issues/20177?

daxian-dbw:
No it doesn't. I just checked the size of "PowerShell.Linux.Alpine.7.4.0-preview.6.nupkg" and it's also ~34mb.
@adityapatwardhan, do you have any insight on this?

daxian-dbw:
I just compared the files from PowerShell.Linux.Alpine.7.4.0-preview.5 with PowerShell.Linux.Alpine.7.4.0-preview.3 and found the preview.5 package includes crossgen'ed PowerShell assemblies while the preview.3 package contains the IL assemblies.

lbussell (approved):
Thanks for the investigation @daxian-dbw. We would like to have that fixed well before .NET 8.0 GA (early november), ideally.

Issue #4845, opened 2023-08-23
Run the VerifyDotnetFolderContents test on all images


lbussell:
The [VerifyDotnetFolderContents test](https://github.com/dotnet/dotnet-docker/blob/e025c2591fa197bf1e8ca1d8d07b0a34c4fba4a7/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs#L135-L136) currently has incorrect conditions that may be causing it to not run on the majority of Linux platforms:

https://github.com/dotnet/dotnet-docker/blob/e025c2591fa197bf1e8ca1d8d07b0a34c4fba4a7/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs#L137-L153

In particular, Mariner hasn't relied on RPM installs since 6.0, and the `!imageData.SdkOS.StartsWith(OS.Alpine) && DockerHelper.IsLinuxContainerModeEnabled` condition seems completely incorrect. We should re-evaluate which platforms this test can be ran on and run it on as many as possible.

We should also run the test on (potentially) distroless images and platforms that don't support PowerShell.

Issue #4843, opened 2023-08-23
Improve the output of the VerifyDotnetFolderContents test


lbussell:
Currently, the output of the [VerifyDotnetFolderContents test](https://github.com/dotnet/dotnet-docker/blob/e025c2591fa197bf1e8ca1d8d07b0a34c4fba4a7/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs#L135-L136) looks something like this:
```
ACTUAL FILES:
<very long list of every single file and sha in the dotnet folder>

EXPECTED FILES:
<another very long list of every expected file and sha in the dotnet folder>
```

This requires the user to download the test output and check which files are extra/missing/different using an external tool. We should instead output a diff of the two directories when the test fails.

ellahathaway:
I assigned myself since I have some extra time between tasks to take this on.

I've reviewed Logan's PR with the changes, so I'll work on re-implementing some of those changes in a new PR. The changes in Logan's PR tackle two issues, but given the unpredictability of my time with the other work I'm being assigned, I plan on tackling just the portion of his changes that relate to this issue first.

PR #4926, opened 2023-10-04
Improve SDK content tests

lbussell:
This PR fixes two issues:
- https://github.com/dotnet/dotnet-docker/issues/4845
  - For Linux images, instead of using PowerShell to enumerate the SDK directory and file hashes, we export the image's file system to an archive and read it just like the official Microsoft SDK, just filtering to only the SDK directory. This allows us to run the test on Alpine arm64 and arm32, which don't include PowerShell.
- https://github.com/dotnet/dotnet-docker/issues/4843
  - To improve the test output, we use git to render the diff between the expected and actual file lists. This implementation is borrowed from Source Build.
    - Depends on https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/919

Also fixed an issue with test path filtering where images with variants would not be tested when specified via path (that only happens when running locally, I think).

I spent too much time trying to get rid of the SharpCompress library, but it started to get too complicated needing to extract arbitrary *.tar, *.zip, and *.tar.gz files, so I reverted to the existing implementation that works fine. The downside is that this implementation writes the entire image contents to the disk. If an image is too big, we could potentially fill up the disk on the agent. The plus to that is that Windows images can't be exported :)


tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -270,18 +216,32 @@ public void VerifyTarInstallation(ProductImageData imageData)
...
+
+        private static IEnumerable<SdkContentFileInfo> GetImageSdkContentsLinux(
+            ProductImageData imageData,
+            DotNetImageRepo imageRepo,
+            DockerHelper dockerHelper)
         {
-            string dotnetPath;
+            string dotnetPath = "/usr/share/dotnet";
+            string imageName = imageData.GetImage(imageRepo, dockerHelper);
+            string imageFsArchivePath = dockerHelper.Export(imageName, Path.GetTempFileName());
```
mthalman, line 234: Rather than exporting the entire image to a tarball, have you considered using a multi-stage helper Dockerfile, similar to what we have for distroless at https://github.com/dotnet/dotnet-docker/blob/main/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/Dockerfile.distroless? The Dockerfile could really use any base image. It doesn't even necessarily have to use PowerShell since this is Linux-specific anyway. It would copy the SDK contents of the image being tested into a local directory, and then run the necessary commands to output the hashes in the desired format.

tests/Microsoft.DotNet.Docker.Tests/ProductImageData.cs
```diff
@@ -64,29 +64,37 @@ public string GetDockerfilePath(DotNetImageRepo imageRepo)
...
-                _ => imageVariant == DotNetImageVariant.None ? "" : "-" + GetImageVariantName(imageVariant),
+                if (imageVariant == DotNetImageVariant.None)
+                {
+                    return string.Empty;
+                }
+
+                string variantName = Enum.GetName(typeof(DotNetImageVariant), imageVariant).ToLowerInvariant();
+                string suffix = isSuffix ? "-" : string.Empty;
+                return $"{suffix}{variantName}";
+            }
```
mthalman, line 84: This nested method has more lines than the method it's in. Perhaps just extract it out to be at class-level? IMO, it hurts readability.

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -270,18 +216,32 @@ public void VerifyTarInstallation(ProductImageData imageData)
...
-                dotnetPath = "/usr/share/dotnet";
-            }
-            else
-            {
-                dotnetPath = "Program Files\\dotnet";
-            }
+        private static IEnumerable<SdkContentFileInfo> GetImageSdkContentsWindows(
+            ProductImageData imageData,
+            DotNetImageRepo imageRepo,
+            DockerHelper dockerHelper)
```
mthalman, line 242: Maybe make this an instance method since it needs access to DockerHelper?

tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -198,6 +198,24 @@ private string ExecuteWithLogging(string args, bool ignoreErrors = false, bool a
...
+            // Include ":" as a command for the container (it does nothing in bash).
+            // Even though the container won't be run, Docker won't create the container unless it has a command.
+            string command = "\":\"";
+
+            // We need to create a container to export it, since Docker doesn't have a way to export the filesystem of
+            // an image by itself. `docker save` saves the layers of an image, but not the image's filesystem.
+            string containerName = ExecuteWithLogging($"create {imageName} {command}");
+            OutputHelper.WriteLine($"Created {containerName}");
+            ExecuteWithLogging($"export {containerName} --output=\"{destination}\"");
+            DeleteContainer(containerName);
```
mthalman, line 215: Put this in a `try..finally` to ensure the container gets cleaned up in case the export fails.

tests/Microsoft.DotNet.Docker.Tests/FileHelper.cs
```diff
@@ -17,31 +19,47 @@ public static TempFolderContext UseTempFolder()
...
+            string? message = null;
 
-            for (int i = 0; i < items.Count; i++)
+            if (baselineFileText != actualFileText)
             {
-                if (items[i] != expectedItems[i])
+                // Retrieve a diff in order to provide a UX which calls out the diffs.
+                string diff = DiffFiles(expectedFilePath, actualFilePath, outputHelper);
+                string prefix = warnOnDiffs ? "##vso[task.logissue type=warning;]" : string.Empty;
+                message = $"{Environment.NewLine}{prefix}Expected file '{expectedFilePath}' does not match actual file '{actualFilePath}`.  {Environment.NewLine}"
```
mthalman, line 38: Do you see there being value in storing the actual files that were diffed as pipeline artifacts? I'm just wondering whether that would help to provide additional context that would be useful in diagnosing the issue.

lbussell: I do think there's value in that. I'm trying to work through that now.

lbussell:
This work is not a high priority right now - the branch will stay open for future reference when I revisit this.

PR #4941, opened 2023-10-11
Manually update to .NET 8.0 latest build

lbussell:
.NET 8.0 SDK has the "RTM" version but doesn't have stable versioning yet, so that was causing issues with our tooling. When we get stable versioning again, the update-dependencies pipeline will fail again and I will update the `Get-IsStableBranding` script again.


mthalman (approved):

MichaelSimons (approved):

eng/Get-IsStableBranding.ps1
```diff
@@ -14,7 +14,7 @@ param(
 $ErrorActionPreference = 'Stop'
 Set-StrictMode -Version 2.0
 
-if ($Version.Contains("-servicing") -or $Version.Contains("-rtm")) {
+if ($Version.Contains("-servicing")) {
```
jander-msft, line 17: Can this somehow be parameterized? We're currently going through an effort to make .NET Monitor 8.0 final versioned because of how this is currently written, which this change will then force us to not be able to provide nightly updates until the other .NET packages are final versioned.

jander-msft: There are other parts of the code that assume final versions based on the version label e.g. https://github.com/dotnet/dotnet-docker/blob/nightly/eng/update-dependencies/VersionUpdater.cs#L90

jander-msft: If this PR goes in as-is, we will stop/back-out or final version kind work. So this request can be deferred.

lbussell: @jander-msft, I pushed a change that uses a different method to check for stable versions with updates from dotnet/installer. Monitor should be unaffected now.

mthalman (approved):

PR #4951, opened 2023-10-18
Update symbols handling

richlander:
The native AOT folks told me the new approach for stripping symbols.

I was able to validate that cache mounts save time. It's possible not all the mounts are needed. I didn't validate that. Perhaps we can take a more systematic approach with logs to decide which ones are providing value.

The `-a` bug is resolved, per https://github.com/dotnet/sdk/issues/34026

@baronfel  -- is there a good way for us to validate which cache mounts are providing benefit? We want to see which assets are downloaded fresh and then which are not on subsequent builds.


lbussell (approved):
Tested on Windows and these all work fine. Sorry for the delay.

samples/releasesapi/Dockerfile
```diff
@@ -2,15 +2,13 @@ FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/nightly/sdk:8.0-jammy-ao
...
-
-# copy and publish app and libraries
+# Copy all source assets
 COPY . .
-RUN dotnet publish -r linux-$TARGETARCH --no-restore -o /app releasesapi.csproj
-RUN rm /app/*.dbg /app/*.Development.json
+# Cache mounts caches runtime packs and NuGet packages across builds
+RUN --mount=type=cache,target=/root/.nuget \
+    --mount=type=cache,target=/source/bin \
+    --mount=type=cache,target=/source/obj \
```
mthalman, line 10: Why is this pattern only done for this sample?

lbussell: Seems like this should be split into 2 PRs - one for stripping symbols and one for cache mounting.

lbussell:
Closing so that we can re-visit this with #5057

samples/releasesapi/Dockerfile
richlander: I added the cache mounting because this sample downloads a lot of files not in the SDK. This is a question we should discuss. What's our approach for caching? The layer caching only works so well. Also, the `docker init` samples for .NET do exactly this.

PR #4958, opened 2023-10-20
Fix cross platform build on alpine sample

lbussell:
The `RUN --mount...` syntax with the NuGet cache, in its current form, is not compatible with our cross-platform builds. While https://github.com/dotnet/dotnet-docker/commit/21c85f0c9054e03e5e020d1a628801d40c3c7716 technically works to allow arm32 compilation on arm64 systems, it was a band-aid fix that we used to get the sample out on time. It doesn't make sense to remove `--no-restore` from our sample, as that defeats the entire point of splitting up the restore and publish steps.

@richlander, if you can figure out how to get the new `RUN --mount...` syntax to allow cross platform builds with the `publish --no-restore` option, then I'd be happy to amend this PR.


lbussell:
> @mthalman: Can you provide more detail on this? What's the behavior?

Sure. Using the full `RUN` instruction as it is:
```Dockerfile
RUN --mount=type=bind,source=aspnetapp/aspnetapp.csproj,target=aspnetapp.csproj \
    --mount=type=cache,target=/root/.nuget/ \
    dotnet restore -a $TARGETARCH
```

Results in the following error on the **publish** step:
```bash
5.628 /usr/share/dotnet/sdk/8.0.100-rc.2.23502.2/Microsoft.Common.CurrentVersion.targets(5189,5): error MSB3030: Could not copy the file "/source/obj/Release/net8.0/linux-musl-arm64/apphost" because it was not found. [/source/aspnetapp.csproj]
------
Dockerfile.alpine:14
--------------------
  12 |     # copy and publish app and libraries
  13 |     COPY aspnetapp/. .
  14 | >>> RUN dotnet publish --no-restore -a $TARGETARCH -o /app
  15 |
  16 |
--------------------
```

It's worth mentioning that this is with me building on Windows and I haven't tested it on Linux.
Removing the NuGet cache actually works, so maybe we should do that, like so: 
```Dockerfile
RUN --mount=type=bind,source=aspnetapp/aspnetapp.csproj,target=aspnetapp.csproj \
    dotnet restore -a $TARGETARCH
```

mthalman (approved):

Issue #4492, opened 2023-03-15
Switch multi-platform tags to Linux only


richlander:
# Switch multi-platform tags to Linux only

We have published multi-platform images, referencing both Linux and Windows images for many years. For example, the `6.0` tag pulls a Debian Arm64 image on an Arm64 Ubuntu machine and a Nano Server x64 image on a Windows x64 machine (if configured to Windows Containers). For Windows, both an OS and a version are selected, with a [matching algorithm](https://github.com/microsoft/Windows-Containers/issues/326), however it is known to have challenges. We believe that this matching algorithm is such a problem that we are planning on removing Windows images from our multi-platform tags, making them Linux only. This change will require the use of Windows-specific tags to use Windows images.

## Problem

The problem is that the [matching algorithm for containerd](https://github.com/containerd/containerd/issues/6508) doesn't work well when a Windows version is higher than any version in the manifest list. In this scenario, you get the oldest not newest container image. That's not what you'd expect. It also means the final image is often not optimal if you build images on a client environment.

For example, if you are running on Windows 11, you will [get a Windows Server 1809 era image](https://github.com/microsoft/Windows-Containers/issues/326), not a newer Windows Server 2022 era image.

Key point: Microsoft guides developers to adopt the latest client OS, but if they do that, then the Windows images they get via multi-platform tags (when using Docker on the client) will be the wrong ones.

This is partially due to the way that Windows Client and Server ship and also due to the containerd algorithm.

This issue doesn't affect Windows base images because those repos don't publish a tag that would include multiple Windows Server versions. In fact, this proposal would result in .NET images being the same as Windows base images in this respect. We're happy to reconsider this decision, but only if the Windows Server team publishes multi-platform tags, effectively expressing that they think that the scenario is important.

## .NET multi-platform tags

The following demonstrates (admittedly cryptically) which tags are being discussed/targeted.

```bash
$ docker manifest inspect mcr.microsoft.com/dotnet/runtime | grep windows
            "os": "windows",
            "os": "windows",
$ docker manifest inspect mcr.microsoft.com/dotnet/runtime:latest | grep windows
            "os": "windows",
            "os": "windows",
$ docker manifest inspect mcr.microsoft.com/dotnet/runtime:6.0 | grep windows
            "os": "windows",
            "os": "windows",
$ docker manifest inspect mcr.microsoft.com/dotnet/runtime:6.0.1 | grep windows
            "os": "windows",
            "os": "windows",
            "os": "windows",
[...]

PR #4975, opened 2023-10-31
Document Windows lack of multi-platform tags

lbussell:
Related to https://github.com/dotnet/dotnet-docker/issues/4492

### DIFF
--- a/documentation/supported-tags.md
+++ b/documentation/supported-tags.md
@@ -49,8 +49,17 @@ These tags reference images for [multiple platforms](https://docs.docker.com/bui
 They include:
 
 - Debian, unless specified (like `6.0-alpine`).
-- Each supported Nano Server version for OS-agnostic tags (like `7.0` and `latest`)
 - All [supported architectures](supported-platforms.md#architectures).
+
+**Note:** Since .NET 8, these multi-platform tags **specifically exclude all Windows versions** due to `containerd`'s platform matching algorithm for Windows hosts.
+Please see [#4492 (Switch multi-platform tags to Linux only)](https://github.com/dotnet/dotnet-docker/issues/4492) for more context.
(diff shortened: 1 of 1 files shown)


documentation/supported-tags.md
```diff
@@ -49,9 +49,18 @@ These tags reference images for [multiple platforms](https://docs.docker.com/bui
 They include:
 
 - Debian, unless specified (like `6.0-alpine`).
-- Each supported Nano Server version for OS-agnostic tags (like `7.0` and `latest`)
 - All [supported architectures](supported-platforms.md#architectures).
 
+**Note:** Since .NET 8, these multi-platform tags **specifically exclude all Windows versions** due to a bug in `containerd`'s platform matching algorithm.
```
mthalman, line 54: I don't think we should make a statement of whether it's a bug or not. It could very well be by design.

```suggestion
**Note:** Since .NET 8, these multi-platform tags **specifically exclude all Windows images** due to `containerd`'s platform matching algorithm.
```

documentation/supported-tags.md
```diff
@@ -51,7 +51,7 @@ They include:
 - Debian, unless specified (like `6.0-alpine`).
 - All [supported architectures](supported-platforms.md#architectures).
 
-**Note:** Since .NET 8, these multi-platform tags **specifically exclude all Windows versions** due to a bug in `containerd`'s platform matching algorithm.
+**Note:** Since .NET 8, these multi-platform tags **specifically exclude all Windows versions** due to `containerd`'s unintuitive platform matching algorithm for Windows hosts.
```
mthalman, line 54: The word `unintuitive` has a negative connotation. We should avoid using such language in our official docs, especially when describing 3rd party software.

PR #4979, opened 2023-11-01
Add documentation for image variants

lbussell:
TODO:
- Display all variants on main README.
- Clean up "Image Variants" section wording for repos with no variants (runtime, monitor?)

### DIFF
--- a/.mar/portal/README.aspnet.portal.md
+++ b/.mar/portal/README.aspnet.portal.md
@@ -62,6 +62,50 @@ After the application starts, navigate to `http://localhost:8000` in your web br
 
 See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotnet/dotnet-docker/blob/main/samples/host-aspnetcore-https.md) to use HTTPS with this image.
 
+### Tag Formatting
+
+#### .NET Versions
+
+All .NET container images have both "fixed version" and "floating version" tags.
+Floating version tags will always reference the latest version of a specific .NET major version, while fixed version tags will always only reference a specific patch version.
+For all tags below, `<.NET Version>` can be substituted for either `<Major.Minor>` or `<Major.Minor.Patch>`, for example: `7.0` or `7.0.12`.
+
+#### Single-platform tags
+
+These "fixed version" tags reference an image with a specific .NET version for a specific operating system and architecture.
+
+- `<.NET Version>-<OS>-<Architecture>`
+- `<.NET Version>-<OS>-<variant>-<Architecture>`
+- `<.NET Version>-<OS>-<Architecture>`
+- `<.NET Version>-<OS>-<variant>-<Architecture>`
+
+#### Multi-platform tags
+
+These tags reference images for [multiple platforms](https://docs.docker.com/build/building/multi-platform/).
+
+- `<.NET Version>`
+    - The version-only floating tag refers to the latest Debian version available at the .NET Major Version's release.
+- `<.NET Version>-<OS>`
+- `<.NET Version>-<OS>-<variant>`
+
+### Image Variants
+
+By default, Ubuntu and Debian images for .NET 8 will have both `icu` and `tzdata` installed.
+
+Our Alpine and Ubuntu Chiseled images are focused on size.
+These images do not and will not include `icu` or `tzdata`, meaning that these images only work iwth apps that are configured for [globalization-invariant mode](https://learn.microsoft.com/dotnet/core/runtime-config/globalization).
+Apps that require globalization support can use the `extra` image variant.
+
+#### `composite`
+
+Compared to the default ASP.NET images, ASP.NET Composite images provide a smaller image size on disk as well as performance improvements for framework-dependent ASP.NET apps by performing some cross-assembly optimizations and between the .NET and ASP.NET runtimes.
+However, this means that apps run on the ASP.NET Composite runtime cannot use handpicked custom versions of .NET or ASP.NET assemblies that are built into the image.
+
+Example tags:
+
+- `8.0.0-jammy-chiseled-composite`
+- `8.0-alpine3.18-composite` 
+
 ## Support
 
 ### Lifecycle
--- a/.mar/portal/README.monitor-base.portal.md
+++ b/.mar/portal/README.monitor-base.portal.md
@@ -45,6 +45,40 @@ The following Dockerfiles demonstrate how you can use this base image to build a
 * [Ubuntu Chiseled - amd64](https://github.com/dotnet/dotnet-docker/blob/main/src/monitor/8.0/ubuntu-chiseled/amd64/Dockerfile)
 * [Ubuntu Chiseled - arm64v8](https://github.com/dotnet/dotnet-docker/blob/main/src/monitor/8.0/ubuntu-chiseled/arm64v8/Dockerfile)
 
+### Tag Formatting
+
+#### .NET Versions
+
+All .NET container images have both "fixed version" and "floating version" tags.
+Floating version tags will always reference the latest version of a specific .NET major version, while fixed version tags will always only reference a specific patch version.
+For all tags below, `<.NET Version>` can be substituted for either `<Major.Minor>` or `<Major.Minor.Patch>`, for example: `7.0` or `7.0.12`.
+
+#### Single-platform tags
+
+These "fixed version" tags reference an image with a specific .NET version for a specific operating system and architecture.
+
+- `<.NET Version>-<OS>-<Architecture>`
+- `<.NET Version>-<OS>-<variant>-<Architecture>`
+- `<.NET Version>-<OS>-<Architecture>`
+- `<.NET Version>-<OS>-<variant>-<Architecture>`
+
+#### Multi-platform tags
+
+These tags reference images for [multiple platforms](https://docs.docker.com/build/building/multi-platform/).
+
+- `<.NET Version>`
+    - The version-only floating tag refers to the latest Debian version available at the .NET Major Version's release.
+- `<.NET Version>-<OS>`
+- `<.NET Version>-<OS>-<variant>`
+
+### Image Variants
+
+By default, Ubuntu and Debian images for .NET 8 will have both `icu` and `tzdata` installed.
+
+Our Alpine and Ubuntu Chiseled images are focused on size.
+These images do not and will not include `icu` or `tzdata`, meaning that these images only work iwth apps that are configured for [globalization-invariant mode](https://learn.microsoft.com/dotnet/core/runtime-config/globalization).
+Apps that require globalization support can use the `extra` image variant.
+
 ## Support
 
 ### Lifecycle
--- a/eng/readme-templates/About.variants.md
+++ b/eng/readme-templates/About.variants.md
@@ -1,0 +1,75 @@
+{{
+    _ ARGS:
+      top-header: The string to use as the top-level header.
+      readme-host: Moniker of the site that will host the readme ^
+    set repo to when(IS_PRODUCT_FAMILY,
+        "product-family",
+        when(PARENT_REPO = "monitor", cat("monitor-", SHORT_REPO), SHORT_REPO))
+}}{{ARGS["top-header"]}}# Tag Formatting
+
+{{ARGS["top-header"]}}## .NET Versions
+
+All .NET container images have both "fixed version" and "floating version" tags.
+Floating version tags will always reference the latest version of a specific .NET major version, while fixed version tags will always only reference a specific patch version.
+For all tags below, `<.NET Version>` can be substituted for either `<Major.Minor>` or `<Major.Minor.Patch>`, for example: `7.0` or `7.0.12`.
+
+{{ARGS["top-header"]}}## Single-platform tags
+
+These "fixed version" tags reference an image with a specific .NET version for a specific operating system and architecture.
+
+- `<.NET Version>-<OS>-<Architecture>`
+- `<.NET Version>-<OS>-<variant>-<Architecture>`
+- `<.NET Version>-<OS>-<Architecture>`
+- `<.NET Version>-<OS>-<variant>-<Architecture>`
+
+{{ARGS["top-header"]}}## Multi-platform tags
+
+These tags reference images for [multiple platforms](https://docs.docker.com/build/building/multi-platform/).
+
+- `<.NET Version>`
+    - The version-only floating tag refers to the latest Debian version available at the .NET Major Version's release.
+- `<.NET Version>-<OS>`
+- `<.NET Version>-<OS>-<variant>`
+
+{{ARGS["top-header"]}}# Image Variants
+
+By default, Ubuntu and Debian images for .NET 8 will have both `icu` and `tzdata` installed.
+
+Our Alpine and Ubuntu Chiseled images are focused on size.
+These images do not and will not include `icu` or `tzdata`, meaning that these images only work iwth apps that are configured for [globalization-invariant mode](https://learn.microsoft.com/dotnet/core/runtime-config/globalization).
+Apps that require globalization support can use the `extra` image variant.{{if repo = "runtime-deps":
+
+{{ARGS["top-header"]}}## `extra`
+
+The `extra` image variant is offered alongside our size-focused base images for self-contained or single file apps that depend on globalization functionality.
+Extra images contain everything that the default images do, plus `icu` and `tzdata`.
+
+Example tags:
+- `8.0-jammy-chiseled-extra`
+- `8.0.0-alpine3.18-extra`}}{{if repo = "aspnet":
+
+{{ARGS["top-header"]}}## `composite`
+
+Compared to the default ASP.NET images, ASP.NET Composite images provide a smaller image size on disk as well as performance improvements for framework-dependent ASP.NET apps by performing some cross-assembly optimizations and between the .NET and ASP.NET runtimes.
+However, this means that apps run on the ASP.NET Composite runtime cannot use handpicked custom versions of .NET or ASP.NET assemblies that are built into the image.
+
+Example tags:
+
+- `8.0.0-jammy-chiseled-composite`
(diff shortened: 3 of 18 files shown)


.mar/portal/README.monitor-base.portal.md
```diff
@@ -45,6 +45,46 @@ The following Dockerfiles demonstrate how you can use this base image to build a
 * [Ubuntu Chiseled - amd64](https://github.com/dotnet/dotnet-docker/blob/main/src/monitor/8.0/ubuntu-chiseled/amd64/Dockerfile)
 * [Ubuntu Chiseled - arm64v8](https://github.com/dotnet/dotnet-docker/blob/main/src/monitor/8.0/ubuntu-chiseled/arm64v8/Dockerfile)
 
+### Tag Formatting
```
jander-msft, line 48: Much of this information is not applicable to .NET Monitor, so maybe excluding it for now would make sense. I think .NET Monitor can have its own "tag formatting" section that has similar principles of describing its tagging scheme, which is different from the rest of the .NET products due to it being an appliance image with a different versioning scheme and lack of variant images.

eng/readme-templates/About.variants.md
```diff
@@ -0,0 +1,83 @@
+{{
+    _ ARGS:
+      top-header: The string to use as the top-level header.
+      readme-host: Moniker of the site that will host the readme ^
+    set repo to when(IS_PRODUCT_FAMILY,
+        "product-family",
+        when(PARENT_REPO = "monitor", cat("monitor-", SHORT_REPO), SHORT_REPO))
+}}{{ARGS["top-header"]}}# Tag Formatting
+
+{{ARGS["top-header"]}}## .NET Versions
```
mthalman, line 10: How does this section differ from our existing doc at https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-tags.md? Should we just reference that from here instead? I'm hesitant in duplicating that content for maintenance reasons. I'm also concerned about making the repo READMEs too large.

lbussell: It doesn't differ by very much. I tried to make it more concise in order to not make the repo readmes too long like you mentioned, but still provide the tag/variant information to customers without leaving MCR/DH. Do you think we should consolidate all this into `supported-tags.md` and link to it from repo readmes?

eng/readme-templates/About.variants.md
```diff
@@ -0,0 +1,83 @@
...
+{{ARGS["top-header"]}}## Multi-platform tags
+
+These tags reference images for [multiple platforms](https://docs.docker.com/build/building/multi-platform/).
+
+- `<.NET Version>`
+    - The version-only floating tag refers to the latest Debian version available at the .NET Major Version's release.
+- `<.NET Version>-<OS>`
+- `<.NET Version>-<OS>-<variant>`
+
+{{ARGS["top-header"]}}## Image Variants
```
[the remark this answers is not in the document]
lbussell: In our infrastructure at least, `jammy-chiseled` is an OS separate from `jammy`. I didn't want to have to explain that we can have multiple "variants" for one image and which order to put them in. I am fine with writing a short section on Chiseled either way though.

eng/readme-templates/About.variants.md
```diff
@@ -0,0 +1,83 @@
...
+
+By default, Ubuntu and Debian images for .NET 8 will have both `icu` and `tzdata` installed.
+These images are intended to satisfy the most common use cases of .NET developers.
+
+Our Alpine and Ubuntu Chiseled images are focused on size.
+These images do not and will not include `icu` or `tzdata`, meaning that these images only work iwth apps that are configured for [globalization-invariant mode](https://learn.microsoft.com/dotnet/core/runtime-config/globalization).
+Apps that require globalization support can use the `extra` image variant of the [dotnet/runtime-deps]({{InsertTemplate("Url.md", [ "readme-host": ARGS["readme-host"], "repo": "dotnet/runtime-deps" ])}}) images.
```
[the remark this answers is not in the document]
lbussell: My intention here was to provide context for users of `aspnet` and `runtime` alpine/chiseled images who want to use Globalization. The only supported way to do that is using self-contained in combination with `extra` images so I thought it was worth mentioning.

eng/readme-templates/About.variants.md
MichaelSimons: @mthalman's comment reminded me that Docker Hub in the past had a readme size limit which we have encountered numerous times.  The tags section is what causes us to hit this.  It would be good to see if there is a limit still and if these changes would cause us to hit it.

.mar/portal/README.aspnet.portal.md
```diff
@@ -62,6 +62,46 @@ After the application starts, navigate to `http://localhost:8000` in your web br
 
 See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotnet/dotnet-docker/blob/main/samples/host-aspnetcore-https.md) to use HTTPS with this image.
 
+### Tag Formatting
```
MichaelSimons, line 65: I know MAR has a set of allowed readme sections but putting this section under usage feels wrong/hidden to me.

eng/readme-templates/About.variants.md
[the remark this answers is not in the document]
mthalman: Ok, that makes sense. But it does seem odd that we have a Variants section in the README describing a variant that doesn't exist for the images that are relevant for the README.

Perhaps there should be a globalization section separate from the variants for everything except runtime-deps? I just feel like the Variants section should only be describing variants that are relevant to the images listed.

eng/readme-templates/About.variants.md
MichaelSimons: > Do you think we should consolidate all this into supported-tags.md and link to it from repo readmes?

I lean this way.

eng/readme-templates/README.mcr.md
```diff
@@ -8,4 +8,6 @@
 
 {{InsertTemplate("Use.md", commonArgs)}}
 
+{{InsertTemplate("About.variants.md", commonArgs)}}
```
jander-msft, line 11: Should this be excluded from the monitor readmes for MCR as it is for in this repo?

eng/readme-templates/README.md
```diff
@@ -25,7 +25,9 @@ if !IS_PRODUCT_FAMILY:{{InsertTemplate("FeaturedTags.md", commonArgs)}}
 }}
 {{InsertTemplate("About.md", commonArgs)}}
 
-{{InsertTemplate("Use.md", commonArgs)}}
+{{InsertTemplate("Use.md", commonArgs)}}{{if find(REPO, "monitor") < 0:
```
jander-msft, line 28: I think the readmes need to be regenerated for monitor. This condition should be excluding this section but it exists in the readmes.

lbussell: The monitor readmes should be unchanged now.

documentation/image-variants.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Ubuntu Chiseled (.NET 6.0+)
+
+[Ubuntu Chiseled](https://ubuntu.com/engage/chiselled-ubuntu-images-for-containers) is Ubuntu's take on "distroless" container images.
+The Ubuntu Chiseled .NET images include the minimum set of libraries necessary to run .NET applications.
+For more information, see the [Ubuntu Chiseled .NET Containers documentation](./ubuntu-chiseled.md).
+
+### `extra` (.NET 8.0)
+
+The `extra` image variant is offered alongside our size-focused base images for self-contained or single file apps that depend on globalization functionality.
```
mthalman, line 22: Why is `single file` called out here? That seems unrelated to what's being offered here. Single file publishing can be used for framework-dependent deployment in which case you could do it with the base Ubuntu image for example and still get globalization.

documentation/image-variants.md
```diff
@@ -0,0 +1,41 @@
...
+Extra images contain everything that the default images do, plus `icu` and `tzdata`.
+
+### `composite` (.NET 8.0)
+
+Compared to the default ASP.NET images, ASP.NET Composite images provide a smaller image size on disk as well as performance improvements for framework-dependent ASP.NET apps by performing some cross-assembly optimizations and between the .NET and ASP.NET runtimes.
+However, this means that apps run on the ASP.NET Composite runtime cannot use handpicked custom versions of .NET or ASP.NET assemblies that are built into the image.
```
mthalman, line 28:
```suggestion
Compared to the default ASP.NET Core images, ASP.NET Core composite images provide a smaller image size on disk as well as performance improvements for framework-dependent ASP.NET Core apps by performing some cross-assembly optimizations and between the .NET and ASP.NET Core runtimes.
However, this means that apps run on the ASP.NET Composite runtime cannot use handpicked custom versions of .NET or ASP.NET assemblies that are built into the image.
```

documentation/image-variants.md
```diff
@@ -0,0 +1,41 @@
...
+
+### (Preview) `aot` (.NET 8.0)
+
+`aot` images provide an optimized deployment size for [native AOT](https://learn.microsoft.com/dotnet/core/deploying/native-aot/) compiled .NET apps.
+Native AOT has the lowest size, startup time, and memory footprint of all .NET deployment models.
+Please see ["Limiatations of Native AOT deployment"](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot#limitations-of-native-aot-deployment) to see if your app might be compatible.
```
mthalman, line 34:
```suggestion
Please see ["Limitations of Native AOT deployment"](https://learn.microsoft.com/dotnet/core/deploying/native-aot#limitations-of-native-aot-deployment) to see if your app might be compatible.
```

documentation/supported-tags.md
```diff
@@ -14,105 +14,73 @@ Complete tag lists:
...
-- `7.0.2-alpine3.17-arm64v8`
-- `7.0.2-bullseye-slim-arm32v7`
+## Single-platform tags
 
-### `<Major.Minor .NET Version>-<OS>-<Architecture>`
+These "fixed version" tags reference an image with a specific .NET version for a specific operating system and architecture.
 
-These "floating version" tags reference an image with a specific `Major.Minor` (with latest patch) .NET version for a specific operating system and architecture.
+- `<.NET Version>-<OS>-<Architecture>`
+- `<.NET Version>-<OS>-<variant>-<Architecture>`
```
MichaelSimons, line 33: `variant` casing is inconsistent with the other parts.  This occurs throughout the document.

documentation/supported-tags.md
```diff
@@ -14,105 +14,73 @@ Complete tag lists:
...
+All .NET container images have both "fixed version" and "floating version" tags.
+Floating version tags will always reference the latest version of a specific .NET major version, while fixed version tags will always only reference a specific patch version.
+For all of the tags below, `<.NET Version>` can be substituted for either `<Major.Minor>` or `<Major.Minor.Patch>`, for example: `6.0` or `6.0.12`.
 
-These "fixed version" tags reference an image with a specific `Major.Minor.Patch` .NET version for a specific operating system and architecture.
+## Image Variants
 
-Examples:
+.NET Container Images have several variants that offer different combinations of flexibility and deployment size.
+The [Image Variants documentation](./image-variants.md) contains a summary of the image variants and their use-cases.
```
MichaelSimons, line 26: Even though the variant in-depth documentation is else where, I think it would be good to include the tagging pattern and examples here including what it means to have no <variant> section in the tag.

documentation/supported-tags.md
```diff
@@ -14,105 +14,73 @@ Complete tag lists:
...
-- `7.0.2-bullseye-slim-arm32v7`
+## Single-platform tags
 
-### `<Major.Minor .NET Version>-<OS>-<Architecture>`
+These "fixed version" tags reference an image with a specific .NET version for a specific operating system and architecture.
 
-These "floating version" tags reference an image with a specific `Major.Minor` (with latest patch) .NET version for a specific operating system and architecture.
+- `<.NET Version>-<OS>-<Architecture>`
+- `<.NET Version>-<OS>-<variant>-<Architecture>`
+- `<.NET Version>-<OS>-<Architecture>`
```
MichaelSimons, line 34: Unless I am missing something obvious, the patterns are duplicated here.

documentation/supported-tags.md
```diff
@@ -14,105 +14,73 @@ Complete tag lists:
...
 
-## Single-platform tags
-
-These tags reference an image for a single platform (e.g. "Linux Arm64" or "Windows x64").
+## .NET Versions
 
-### `<Major.Minor.Patch .NET Version>-<OS>-<Architecture>`
+All .NET container images have both "fixed version" and "floating version" tags.
+Floating version tags will always reference the latest version of a specific .NET major version, while fixed version tags will always only reference a specific patch version.
+For all of the tags below, `<.NET Version>` can be substituted for either `<Major.Minor>` or `<Major.Minor.Patch>`, for example: `6.0` or `6.0.12`.
```
MichaelSimons, line 21: This section no longer follows the same structure as the others.  It has been condensed into paragraph form rather than:

Description

Pattern

Examples

IMO this helps the ability to quickly browse this document.

documentation/supported-tags.md
```diff
@@ -14,105 +14,73 @@ Complete tag lists:
...
 
-These "fixed version" tags reference an image with a specific `Major.Minor.Patch` .NET version for a specific operating system and architecture.
+## Image Variants
```
MichaelSimons, line 23: I think this should be moved down later in the document as they are of lesser importance.

documentation/supported-tags.md
MichaelSimons, line 33: I would exclude variants from this section as it should be covered in the variants section.

documentation/image-variants.md
```diff
@@ -0,0 +1,41 @@
+# .NET Image Variants
+
+.NET Container images offer several image variants.
+
+.NET images for Ubuntu, Debian, Windows Server Core, and Nanoserver images are "batteries included" and include ICU libraries.
+As of .NET 8, we also guarantee these images include `tzdata`.
+These images are intended to satisfy the most common use cases of .NET developers.
+
+Alpine and [Ubuntu Chiseled](#ubuntu-chiseled-net-60) .NET images are focused on size.
+These images do not include `icu` or `tzdata`, meaning that these images only work with apps that are configured for [globalization-invariant mode](https://learn.microsoft.com/dotnet/core/runtime-config/globalization).
```
MichaelSimons, line 10: I think the wording needs to clarify that the "default" images do not include these packages.

documentation/image-variants.md
```diff
@@ -0,0 +1,41 @@
...
+Apps that require globalization support can use the `extra` image variant of the [dotnet/runtime-deps](https://hub.docker.com/_/microsoft-dotnet-runtime-deps/) images.
+Alpine and Chiseled images also come in `extra`, `composite`, and `aot` variants (see below).
+
+### Ubuntu Chiseled (.NET 6.0+)
```
MichaelSimons, line 14: Thoughts on formatting the title like the others sections where the variant name is used/called out?
```suggestion
### `chiseled` (.NET 6.0+)
```

documentation/image-variants.md
```diff
@@ -0,0 +1,41 @@
...
+For more information, see the [Ubuntu Chiseled .NET Containers documentation](./ubuntu-chiseled.md).
+
+### `extra` (.NET 8.0)
+
+The `extra` image variant is offered alongside our size-focused base images for self-contained or single file apps that depend on globalization functionality.
+Extra images contain everything that the default images do, plus `icu` and `tzdata`.
+
+### `composite` (.NET 8.0)
+
+Compared to the default ASP.NET images, ASP.NET Composite images provide a smaller image size on disk as well as performance improvements for framework-dependent ASP.NET apps by performing some cross-assembly optimizations and between the .NET and ASP.NET runtimes.
```
MichaelSimons, line 27: It would be good to link to further details/reading on what composites are.

documentation/image-variants.md
```diff
@@ -0,0 +1,41 @@
...
+`aot` images provide an optimized deployment size for [native AOT](https://learn.microsoft.com/dotnet/core/deploying/native-aot/) compiled .NET apps.
+Native AOT has the lowest size, startup time, and memory footprint of all .NET deployment models.
+Please see ["Limiatations of Native AOT deployment"](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot#limitations-of-native-aot-deployment) to see if your app might be compatible.
+`aot` image variants are only available for our size-focused `runtime-deps` images: Alpine and Ubuntu Chiseled.
+They also require the use of the `aot` SDK image which include extra libraries needed for Native AOT compilation.
+
+> [!NOTE]
+> `aot` images are only available as a preview in the [dotnet/nightly/sdk](https://hub.docker.com/_/microsoft-dotnet-nightly-sdk/) and [dotnet/nightly/runtime-deps](https://hub.docker.com/_/microsoft-dotnet-nightly-runtime-deps/) repos.
+> Native AOT compiled apps will function exactly the same on the existing `runtime-deps` (non-`aot`) images, but with a larger deployment size.
+> Please try out these new, smaller images and give us feedback!
```
MichaelSimons, line 41: Several of the variants tie in nicely with https://github.com/dotnet/dotnet-docker/pull/4992.  Consider linking from this document either in each relevant section or in an overview.

.mar/portal/README.aspnet.portal.md
```diff
@@ -62,6 +62,11 @@ After the application starts, navigate to `http://localhost:8000` in your web br
 
 See [Hosting ASP.NET Core Images with Docker over HTTPS](https://github.com/dotnet/dotnet-docker/blob/main/samples/host-aspnetcore-https.md) to use HTTPS with this image.
 
+## Image Variants
```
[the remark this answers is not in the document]
lbussell: I posted in the MCR Teams channel, I'll update this if I get a response.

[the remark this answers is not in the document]
lbussell: According to the MCR Team, they only check that the required headers are in place and in order, and adding additional headers should be OK.

MichaelSimons (approved):

mthalman (approved):

PR #4981, opened 2023-11-01
Enable .NET Monitor updates from dotnetstage storage account

jander-msft:
These changes allow generating an dependency update PR into the internal project based on an internal build of .NET Monitor.

Overview of changes:
- Move all common update-dependencies pipeline logic into the `eng/pipelines/steps/update-dependencies.yml` pipeline step template. This allows sharing of the aggregation and internal vs public logic between the SDK products and .NET Monitor.
- Update the `Get-*DropVersion.ps1` scripts to conform to the merged pipeline logic for updating dependencies.
- Update the `update-dependencies` tool to handle internal builds of .NET Monitor that have been staged to the dotnetstage storage account. The .NET Monitor binaries and checksums are staged into a different container with a different layout compared to the SDK files.

Example .NET Monitor layout:
- `https://dotnetstage.blob.core.windows.net/dotnet-monitor/<build_version>/BlobAssets/dotnet-monitor-base-8.0.0-linux-x64.tar.gz`
- `https://dotnetstage.blob.core.windows.net/dotnet-monitor/<build_version>/ChecksumAssets/dotnet-monitor-base-8.0.0-linux-x64.tar.gz.sha512`

cc @dotnet/dotnet-monitor

### DIFF
--- a/eng/pipelines/steps/update-dependencies.yml
+++ b/eng/pipelines/steps/update-dependencies.yml
@@ -1,12 +1,84 @@
 parameters:
-  # The customArgsArray parameter is used to specify the configuration for multiple Dockerfile versions.
+  # The versionInfosArray parameter is used to specify the configuration for multiple Dockerfile versions.
   # This allows for a single PR to be generated for different internal .NET build versions.
   # This is required in order to publish a single set of internal images when the PR is merged.
-  customArgsArray: ""
+  versionInfosArray: ""
 
   useInternalBuild: false
 
 steps:
+- powershell: |
+    $versionInfos = '${{ parameters.versionInfosArray }}' | ConvertFrom-Json
+
+    $index=0
+    foreach ($versionInfo in $versionInfos) {
+      $args = @{
+        ProductVersion = $versionInfo.DockerfileVersion
+        ComputeShas = $versionInfo.ComputeShas
+        AzdoVariableName = "updateDepsArgs-$index"
+        UseStableBranding = $versionInfo.StableBranding
+      }
+
+      if ($versionInfo.RuntimeVersion) {
+        $args += @{
+          RuntimeVersion = $versionInfo.RuntimeVersion
+        }
+      }
+
+      if ($versionInfo.AspnetVersion) {
+        $args += @{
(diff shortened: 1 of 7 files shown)


eng/pipelines/steps/update-dependencies.yml
```diff
@@ -1,12 +1,84 @@
...
+  # The versionInfosArray parameter is used to specify the configuration for multiple Dockerfile versions.
   # This allows for a single PR to be generated for different internal .NET build versions.
   # This is required in order to publish a single set of internal images when the PR is merged.
-  customArgsArray: ""
+  versionInfosArray: ""
 
   useInternalBuild: false
 
 steps:
+- powershell: |
```
jander-msft, line 10: I want to call out that this "new" code is practically the same as what was in `eng/pipelines/steps/update-dotnet-dependencies.yml` except for two differences:
- Each of the product versions is conditionally included based on version info that is passed. This avoids have some of the product version arguments set to null and causing the `Set-DotnetVersions.ps1` script to fail because it can't determine whether it's updating SDK products or the .NET Monitor product.
- The `ChecksumSasQueryString` is conditionally set based on whether .NET Monitor is being updated. The checksums for .NET Monitor are in the same storage account as the binaries.

mthalman (commented):
The one thing that gives me pause is that the PR generated for Monitor is independent of any PR for the base .NET images. That can be problematic because the Monitor image that gets built might not be based on the next-to-be-released ASP.NET Core version, depending on timing of the PRs. PR generation for the base images is currently triggered manually. But the Monitor workflow would be completely automated. I feel like we need a better coordinated solution if we want to truly test upcoming releases.

In addition to that concern, I'd also want some better protection for the PR that gets generated for Monitor so that it isn't accidentally merged since the intent is just to test the changes, not update the branch. Perhaps the PR can be marked as draft to better protect against that? That would certainly match the intent. That would likely require an update to Microsoft.DotNet.VersionTools, which handles the PR creation, in order to provide such a feature.

jander-msft:
Superseded by #4990

PR #4990, opened 2023-11-08
Allow using local checksums table file for looking up checksums

jander-msft:
These changes allow using a locally available checksums table file instead of requiring the checksums to be publicly available on a blob storage account. I refactored the existing code to maintain the caching and validation behavior but allow for supplying a separate source of checksum information (in this case, the local file system).

### DIFF
--- a/eng/update-dependencies/DockerfileShaUpdater.cs
+++ b/eng/update-dependencies/DockerfileShaUpdater.cs
@@ -466,7 +466,7 @@ namespace Dotnet.Docker
                 () =>
                 {
                     Trace.TraceInformation($"Opening '{_options.ChecksumsFile}'.");
-                    return Task.FromResult<string?>(File.ReadAllText(_options.ChecksumsFile));
+                    return Task.FromResult(File.ReadAllText(_options.ChecksumsFile));
                 });
         }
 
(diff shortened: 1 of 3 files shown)


eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -446,52 +459,79 @@ private static bool TryParsePreviewVersion(string version, out string? previewVe
...
+                        else
+                        {
+                            Trace.TraceInformation($"Failed to find dotnet release checksums");
+                            return null;
+                        }
+                    }
+                });
+        }
+
+        private async Task<IDictionary<string, string>> GetChecksums(string cacheId, Func<Task<string?>> getContentCallback)
```
mthalman, line 497: Instead of `cacheId`, how about something more semantically descriptive like `sourceUrlOrPath`.

eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -446,52 +459,79 @@ private static bool TryParsePreviewVersion(string version, out string? previewVe
...
 
             checksumEntries = new Dictionary<string, string>();
-            s_releaseChecksumCache.Add(uri, checksumEntries);
+            s_releaseChecksumCache.Add(cacheId, checksumEntries);
 
-            Trace.TraceInformation($"Downloading '{uri}'.");
-            using (HttpResponseMessage response = await s_httpClient.GetAsync(ApplySasQueryStringIfNecessary(uri, _options.BinarySasQueryString)))
+            string? content = await getContentCallback();
+
+            if (!string.IsNullOrEmpty(content))
```
mthalman, line 509: Instead of a big `if` block here, how about something like this:

```c#
if (string.IsNullOrEmpty(content))
{
    return checksumEntries;
}
```

mthalman (approved):

PR #4204, opened 2022-11-10
Add .NET 8 Dockerfiles

mthalman:
This adds Dockerfiles for .NET 8. The build branding for it is still currently labeled as "alpha" but I've defined the `-preview` suffix for floating tags to conform with https://github.com/dotnet/dotnet-docker/issues/4117.

Because PowerShell has no builds based on .NET 8, that's not included in the SDK images.

Fixes #4197

PR #5003, opened 2023-11-15
.NET Monitor 8.1

jander-msft:
- Add preliminary .NET Monitor 8.1 nightly images
- Move the `8-*`, `8`, and `latest` tags to target the 8.1 images
- Fix extensions template to use .NET Monitor version for base image

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -6983,10 +6983,10 @@
           "productVersion": "$(monitor|8.1|product-version)",
           "sharedTags": {
             "$(monitor|8.1|product-version)-ubuntu-chiseled": {},
-            "8.1-ubuntu-chiseled": {},
+            "8.1-preview-ubuntu-chiseled": {},
             "8-ubuntu-chiseled": {},
             "$(monitor|8.1|product-version)": {},
-            "8.1": {},
+            "8.1-preview": {},
             "8": {},
             "latest": {}
           },
@@ -7001,7 +7001,7 @@
               "osVersion": "jammy-chiseled",
               "tags": {
                 "$(monitor|8.1|product-version)-ubuntu-chiseled-amd64": {},
-                "8.1-ubuntu-chiseled-amd64": {},
+                "8.1-preview-ubuntu-chiseled-amd64": {},
                 "8-ubuntu-chiseled-amd64": {}
               }
             },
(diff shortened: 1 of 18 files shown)


manifest.json
```diff
@@ -6379,6 +6368,105 @@
...
+                }
+              },
+              "variant": "v8"
+            }
+          ]
+        },
+        {
+          "productVersion": "$(monitor|8.1|product-version)",
+          "sharedTags": {
+            "$(monitor|8.1|product-version)-ubuntu-chiseled": {},
```
mthalman, line 6380: All 8.1* tags should be 8.1-preview*

jander-msft: I'm not sure what to do with `latest` either.

jander-msft: I've moved the `latest` and `8-*` tags back to the 8.0 images. For latest, this seems to fall in line with how it was used on the .NET 7.0 images during .NET 8 previews until .NET 8 RC 1. I think `8-*` tags should follow the same pattern (sit on the shipped image until the next version becomes RC or later).

mthalman: In nightly, `latest` (and `8-*` is analagous to that) should point to the absolute latest available (i.e. preview version). For main, `latest` should continue to point to 8.0.

jander-msft: > In nightly, `latest` (and `8-*` is analagous to that) should point to the absolute latest available (i.e. preview version). For main, `latest` should continue to point to 8.0.

In the .NET 8 preview timeframe, the .NET 7 images remained tagged as `latest` in the nightly branch: https://github.com/dotnet/dotnet-docker/blob/f09007172b67f733e8172566d71d7ee753285793/manifest.json#L1655
In the .NET 7 preview timeframe, the .NET 6 images remained tagged as `latest` in the nightly branch: https://github.com/dotnet/dotnet-docker/blob/e2f756bba8741ad4ab8e4450c2a0d69488df623a/manifest.json#L1693

Historically, the pattern has been that `latest` remains on the latest GA/serviced version even in the nightly branch. However, the .NET Monitor images were the outlier here, where they have `latest` on literally the latest version; this has lead to inconsistencies when bringing changes from nightly to main, requiring manual fix ups, and merge conflicts when merging main back to nightly.

So I would contend that .NET Monitor should fall in line with how the rest of the .NET images have been treating `latest` for at least the last two years (and likewise apply that to the `8-*` tags) and only keep it on the latest GA/serviced regardless of the branch; this is currently how I have it in this PR.

If you say otherwise, then the rest of the .NET images should do the same as well in the future.

mthalman: I don't know where that branch that first file is pointing to but I don't think it's nightly. The `latest` tags were updated from 7.0 to 8.0 in the initial PR to add .NET 8 Dockerfiles in nightly: https://github.com/dotnet/dotnet-docker/pull/4204

jander-msft: I might have been looking at commits that were made in main but merged back to nightly. If I check commits that were initially created for nightly, you are correct. I've now moved the `latest` and `8-*` tags to the 8.1 images.

mthalman (approved):

PR #5013, opened 2023-11-17
Fix tagging for alpine arm images with PowerShell update

lbussell:
/cc @mthalman

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -4967,7 +4967,7 @@
               "os": "linux",
               "osVersion": "alpine3.18",
               "tags": {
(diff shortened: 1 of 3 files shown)


manifest.json
```diff
@@ -4952,7 +4952,7 @@
               "os": "linux",
               "osVersion": "alpine3.18",
               "tags": {
-                "$(sdk|6.0|product-version)-alpine3.18-amd64": {},
+                "$(sdk|6.0|alpine-arm-product-version)-alpine3.18-amd64": {},
```
mthalman, line 4955: This shouldn't be changed.

PR #5018, opened 2023-11-21
Document CBL Mariner 2.0 tags

### LINKED ISSUE #5004: Add Mariner .NET images to "Support tags only" section on MAR
### Describe the Problem

As it stands, the Mariner .NET images are hidden to external users on MAR.

### Describe the Solution

Change the metadata of the Mariner .NET images so that they appear in the "Supported tags only" section on MAR. Customers can expect the same level of support for these images as they do for .NET images built on Ubuntu.

lbussell:
Fixes #5004


MichaelSimons (approved):

MichaelSimons:
Friendly reminder as a follow-up to update the internal documentation for first parties to now reference mcr.microsoft.com

PR #5039, opened 2023-12-02
Add Dockerfiles for .NET 9

### LINKED ISSUE #5005: Add images for .NET 9
### Additional Context

- Previous issue: https://github.com/dotnet/dotnet-docker/issues/4197

- This should also be added to https://github.com/dotnet/dotnet-docker/issues/4574

lbussell:
Fixes #5005


manifest.json
```diff
@@ -1569,482 +1568,548 @@
...
-      ],
-      "mcrTagsMetadataTemplate": "eng/mcr-tags-metadata-templates/runtime-tags.yml",
-      "images": [
-        {
-          "productVersion": "$(dotnet|6.0|product-version)",
+          "productVersion": "$(dotnet|9.0|product-version)",
           "sharedTags": {
-            "$(dotnet|6.0|product-version)": {},
-            "6.0": {}
+            "$(dotnet|9.0|product-version)": {},
```
mthalman, line 1575: All these `$(dotnet|9.0|product-version)` and `9.0` tags should have `-preview`. I think this is the cause of the test failures. Perhaps there should be a variable defined for each of these that can be referenced throughout. So when we need to remove the `-preview`, it just needs to be done in one spot.

manifest.versions.json
```diff
@@ -63,12 +81,16 @@
 
     "chisel|6.0|url": "github.com/canonical/chisel",
     "chisel|8.0|url": "$(chisel|6.0|url)",
+    "chisel|9.0|url": "$(chisel|6.0|url)",
+
     "chisel|6.0|ref": "v0.8.0",
     "chisel|8.0|ref": "v0.8.0",
+    "chisel|9.0|ref": "v0.8.0",
```
mthalman, line 88: Since there's currently no need to have separate values. We might as well reuse the `$(chisel|6.0|ref)` variable. Same applies for rocks-toolbox.

lbussell: The reason they are separate is so that we don't rebuild all .NET versions in nightly when chisel has an update. But then again, it is nightly. Also, if we used the 6.0 ref as the base, then we would not pick up new chisel versions from the update-dependencies tool in 9.0 nightly builds (since we only update the chisel version when we update the runtime for a given .NET major version). I could see an argument to use 9.0 as the base variable so that we're testing the new chisel versions in old .NET versions before patch Tuesdays.

mthalman: > The reason they are separate is so that we don't rebuild all .NET versions in nightly when chisel has an update.

I don't see a problem with rebuilding in this case.

> I could see an argument to use 9.0 as the base variable so that we're testing the new chisel versions in old .NET versions before patch Tuesdays.

Yes, that would be my preference. In case there is any difference in the usage of chisel between versions, it'd be good catch any issues in the nightly repo as a result of a chisel version upgrade.

manifest.versions.json
```diff
@@ -63,12 +81,18 @@
 
     "chisel|6.0|url": "github.com/canonical/chisel",
     "chisel|8.0|url": "$(chisel|6.0|url)",
-    "chisel|6.0|ref": "v0.8.0",
-    "chisel|8.0|ref": "v0.8.0",
+    "chisel|9.0|url": "$(chisel|6.0|url)",
+
+    "chisel|6.0|ref": "$(chisel|9.0|ref)",
+    "chisel|8.0|ref": "$(chisel|9.0|ref)",
+    "chisel|9.0|ref": "v0.8.0",
```
mthalman, line 88: This needs to be defined before it can be referenced by other variables (top to bottom). Same applies to rocks-toolbox.

mthalman (approved):

lbussell:
There is an issue with internal build matrix generation that I will need to work out before this is merged.

manifest.json
```diff
@@ -1914,8 +1914,8 @@
               "variant": "v7",
               "customBuildLegGroups": [
                 {
-                  "name": "pr-build",
-                  "type": "Supplemental",
+                  "name": "test-dependencies",
+                  "type": "Integral",
```
mthalman, line 1918: Does this need to be Integral or Supplemental?

mthalman (approved):

manifest.json
lbussell: This was a typo. However, both appear to work in this scenario (I re-tested it), so I reverted it to the smaller of the two hammers (Supplemental).

PR #5066, opened 2023-12-19
Add libatomic to 8.0+ SDK images

### LINKED ISSUE #5033: Include `libatomic1` in the `dotnet/sdk` base images
### Describe the Problem

`libatomic1` is a dependency of nodejs, which is a bundled dependency of emsdk and the wasm-tools workload. This dependency is not present in the `dotnet/sdk:8.0` image, and so projects that install the wasm-tool workload fail to build. An example of what this looks like is https://github.com/dotnet/runtime/issues/95477.

### Describe the Solution

Ideally `libatomic1` would be added as a package dependency for the `sdk` images on Ubuntu, and matching dependencies would be added on other distros that we produce SDK images for (e.g. alpine).

cc @lewing @directhex

lbussell:
Fixes #5033

I was unable to verify that this fixes the scenario in https://github.com/dotnet/runtime/issues/95477. None of the wasm samples I tried (`wasmbrowser`, `blazorwasm`, `maui`) seemed to have issues with the SDK images as they are. However, I have prepared this PR with no tests in the meantime.

@baronfel, what would be the minimum possible repro for something like that runtime issue? Is there a `dotnet new` template I should be looking at?

### DIFF
--- a/eng/dockerfile-templates/sdk/Dockerfile.linux
+++ b/eng/dockerfile-templates/sdk/Dockerfile.linux
@@ -26,7 +26,6 @@
                 "git",
                 "icu-data-full",
                 "icu-libs",
-                "libatomic",
                 "tzdata",
             ]),
         when(isMariner,
--- a/eng/dockerfile-templates/sdk/Dockerfile.linux
+++ b/eng/dockerfile-templates/sdk/Dockerfile.linux
@@ -26,18 +26,34 @@
                 "git",
                 "icu-data-full",
                 "icu-libs",
-                "tzdata"
+                "libatomic",
+                "tzdata",
             ]),
         when(isMariner,
-            [
-                "git",
-                "tar"
-            ],
-            [
-                "curl",
-                "git",
-                "wget"
-            ])), len)
+            when (dotnetVersion = "6.0" || dotnetVersion = "7.0",
+                [
+                    "git",
+                    "tar",
+                ],
+                [
+                    "git",
+                    "libatomic_ops",
+                    "tar",
+                ]
+            ),
+            when (dotnetVersion = "6.0" || dotnetVersion = "7.0",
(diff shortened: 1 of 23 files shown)


baronfel:
@lbussell a user submitted a failing sample at https://github.com/dotnet/sdk/issues/37668 - can you try to replicate using that?

MichaelSimons (approved):
Approved with the conditions that a manual repro will be validated and that a follow-up PR will include an automated test scenario.

lbussell:
> Approved with the conditions that a manual repro will be validated and that a follow-up PR will include an automated test scenario.

I validated that the new images work for Debian, Ubuntu, and Mariner/Azure Linux. Alpine does not work due to https://github.com/dotnet/sdk/issues/32327, I added a comment here: https://github.com/dotnet/sdk/issues/32327#issuecomment-1874344759.

My test repo is https://github.com/lbussell/blazorwasm-repro, [Dockerfile](https://github.com/lbussell/blazorwasm-repro/blob/main/Dockerfile.alpine)

lbussell:
See the reply at https://github.com/dotnet/sdk/issues/32327#issuecomment-1874693353 - The musl fix for the wasm-tools workload appears to be in main/9.0 only. We should move forwards with the Debian/Ubuntu/Mariner fix and remove the changes to Alpine until it's fixed in 8.0 servicing (if it is fixed).

lbussell:
I tested this on Alpine with .NET 9 as well and there was another issue, so I removed it from both. I updated the SDK issue with my findings.

lbussell:
> Approved with the conditions that a manual repro will be validated and that a follow-up PR will include an automated test scenario.

Filed https://github.com/dotnet/dotnet-docker/issues/5076 for the test scenario.

PR #5072, opened 2024-01-02
Amend ASP.NET Core HTTPS documentation to avoid tripping CredScan

lbussell:
Azure DevOps credential scanning doesn't like our placeholder password in this documentation. I changed it to "placeholder" which hopefully it prefers and won't trip on in the future. I also added an empty commit with the special string to bypass CredScan for one push.

This PR should be merged, not squashed, to preserve that commit.

### DIFF
--- a/samples/host-aspnetcore-https.md
+++ b/samples/host-aspnetcore-https.md
@@ -32,18 +32,18 @@ Generate cert and configure local machine:
 > If you are using CMD instead of PowerShell, substitute `$env:USERPROFILE` with `%USERPROFILE%`.
 
 ```console
-dotnet dev-certs https -ep $env:USERPROFILE\.aspnet\https\aspnetapp.pfx -p crypticpassword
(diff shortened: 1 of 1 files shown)


mthalman:
The name of the special commit is being truncated:
image

Is that going to mess things up?

garath (approved):

lbussell:
> The name of the special commit is being truncated
> 
> Is that going to mess things up?

I think that is in the GitHub UI only. Git log shows:

```
PS C:\s\dotnet-docker> git log --oneline -1
8a45243b7 (HEAD -> credscan-fix, lbussell/credscan-fix) Bypass secret scanning false positive for code mirror **BYPASS_SECRET_SCANNING**
```

mthalman (approved):

lbussell:
I am going to try the rebase option as I'm worried about the latest commit being the merge commit.

PR #5083, opened 2024-01-04
Add tests for Blazor WASM

### LINKED ISSUE #5076: Add Blazor WASM test
https://github.com/dotnet/dotnet-docker/pull/5066 fixed #5033, enabling Blazor WASM apps to be compiled using the Debian, Ubuntu, and Mariner SDK images. We should add a test for this to ensure that this functionality doesn't regress in the future, and so we can verify the scenario on Alpine if it gets fixed.

https://github.com/lbussell/blazorwasm-repro is a repro repo that we should verify fails without the changes from #5066, when we create the test.

lbussell:
Fixes #5076

What's done
- Factored ImageScenarioVerifier into two classes - TestScenario and TestSolution
  - TestSolution manages creating the actual test code on disk
  - TestScenario manages building and running that code inside a Dockerfile
    - TestScenario is a basic console app by default, but can be inheirited from and overridden to customize functionality for different scenarios
    - Transformed all of the existing conditional logic into different classes inheriting from TestScenario.
  - Added a new scenario for the `blazorwasm` sample

TODO
- [x] - Install `wasm-tools` workload in BlazorWasm test
  - This is critical to reproduce the failure seen in https://github.com/dotnet/sdk/issues/37668, as the SDK can compile Blazor without the workload but doesn't run any optimizations, which is what causes the error
  - This scenario requires installing python3, which is going to be fun to work out how to do on every distro in one Dockerfile. Open to suggestions
- [x] - Implement IDisposable from `TestScenario` since it contains a `TestSolution`
- [x] - Nullability warnings in WebScenario.cs


lbussell:
Filed https://github.com/dotnet/aspnetcore/issues/53469 for Mariner .NET 6.0 issue.

mthalman (approved):

PR #5096, opened 2024-01-09
Fix incorrectly tagged Alpine SDK image (Remove PowerShell Alpine Arm tagging workaround)

lbussell:
The workaround from https://github.com/dotnet/dotnet-docker/pull/5013 caused some Alpine images to be tagged with the incorrect SDK patch version (seen in the README diff on this PR). This PR removes the workaround. Unfortunately it wasn't caught in https://github.com/dotnet/dotnet-docker/pull/5082.


mthalman (approved):
I assume this means these old tags were published yesterday targeting new versions which don't match the tags. We'll need to retag those images so they revert back to the original image. That will have to be a manual process.

PR #5099, opened 2024-01-10
Update Dockerfile.alpine

richlander:
Show how to enable for non-root

### DIFF
--- a/samples/aspnetapp/Dockerfile.alpine
+++ b/samples/aspnetapp/Dockerfile.alpine
@@ -20,4 +20,5 @@ FROM mcr.microsoft.com/dotnet/aspnet:8.0-alpine
 EXPOSE 8080
 WORKDIR /app
 COPY --from=build /app .
+# USER $APP_UID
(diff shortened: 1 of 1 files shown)


samples/aspnetapp/Dockerfile.alpine
```diff
@@ -20,4 +20,5 @@ FROM mcr.microsoft.com/dotnet/aspnet:8.0-alpine
 EXPOSE 8080
 WORKDIR /app
 COPY --from=build /app .
+# USER $APP_UID
```
mthalman, line 23: There should be a comment above this summarizing what this does. Otherwise, it looks like we commented it out while debugging stuff and pushed it to the repo accidentally or that we forgot to delete the line.

richlander:
```suggestion
# Uncomment to enable non-root user
# USER $APP_UID
```

lbussell (approved):

mthalman (approved):

PR #5108, opened 2024-01-17
Update Image Sizes

lbussell:
- Update Linux image sizes
  - Alpine 3.18 runtime-deps and AOT runtime-deps: All size decreases
  - Jammy Chiseled extra runtime and aspnet: Initial image sizes that were
    left out of the original PR
  - 6.0 Alpine 3.19 SDK: size decrease
  - 9.0 Alpine 3.19 SDK: my best guess is a package installation but I
    will check on this.

- Update Windows image sizes
  - Windows Server Core LTSC 2022 Images increased in size.


mthalman (approved):

lbussell:
> 9.0 Alpine 3.19 SDK

This size increase is actually all due to the 9.0 SDK's DotnetTools directory. dotnet format increased from 50 -> 82 MB and dotnet watch increased from 20 -> 51 MB.

Issue #5006, opened 2023-11-16
Enable nullable reference types in Microsoft.DotNet.Docker.Tests and update-dependencies projects


lbussell:
### Describe the Solution

Enabling nullable reference types by default for the project can help guard against dereferencing null objects. See https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

### Additional Context

- Also see https://github.com/dotnet/docker-tools/issues/327

PR #5115, opened 2024-01-19
Fix chisel update deps tool

lbussell:
https://github.com/dotnet/dotnet-docker/pull/5111 accidentally overwrites the 6.0 Chisel tool version in manifest.versions.json. That version should only be updated if we're updating .NET 9.0 Dockerfiles.

I also took the opportunity to add `#nullable enable` to the file I touched (#5006) and use some new .NET features where the IDE suggested. Also, the C# for VS Code extension requires a Solution file for any of its features, so I added one.


eng/update-dependencies/update-dependencies.sln
mthalman, line 1: What's the need for this?

eng/update-dependencies/UpdateDependencies.cs
```diff
@@ -412,18 +412,28 @@ private static async Task<IEnumerable<IDependencyUpdater>> GetUpdatersAsync()
...
                 new NuGetConfigUpdater(RepoRoot, Options),
                 new BaseUrlUpdater(RepoRoot, Options),
                 new MinGitUrlUpdater(RepoRoot, minGitRelease),
-                new MinGitShaUpdater(RepoRoot, minGitRelease),
+                new MinGitShaUpdater(RepoRoot, minGitRelease)
+            ];
+
+            // Chisel tooling versions are shared between versions, so we only want to update it with automatic
+            // dependency flows for preview versions.
+            if (Options.DockerfileVersion == "9.0")
```
mthalman, line 429: We should update the lifecycle checklist to update this value when setting up a new major version.

mthalman, line 429: Rather than explicitly checking for `9.0`, could a more maintainable solution be to not overwrite the JSON property if it's currently set to a variable reference? In other words, we should only set it if it's currently set to a literal value.

eng/update-dependencies/update-dependencies.sln
lbussell: It's mentioned in the PR body. VS Code C# Dev kit only works with solution files.

eng/update-dependencies/UpdateDependencies.cs
```diff
@@ -443,8 +443,8 @@ private static async Task<IEnumerable<IDependencyUpdater>> GetUpdatersAsync()
             updaters =
             [
                 ..updaters,
-                ScriptRunnerUpdater.GetDockerfileUpdater(RepoRoot),
-                ScriptRunnerUpdater.GetReadMeUpdater(RepoRoot)
+                // ScriptRunnerUpdater.GetDockerfileUpdater(RepoRoot),
+                // ScriptRunnerUpdater.GetReadMeUpdater(RepoRoot)
```
mthalman, line 447: revert

mthalman (approved):
Just one thing needed to revert.

PR #5120, opened 2024-01-20
Update Kubernetes samples

richlander:
These samples are embarrassingly broken in multiple ways. I'm not sure how some of this worked before. I think I was using Docker Desktop for Windows before and it must have been more lenient. The latest changes were made on Ubuntu 22.04.

I need to spend some more time with these samples, however, these changes are intended as an important stopgap.

Note that the port of the service cannot be `80` because that requires a `root` user. That doesn't work well with `ClusterIP`, at least when testing locally (and there is no obvious reason to need port `80` for `ClusterIP` when running remotely).


samples/kubernetes/non-root/README.md
```diff
@@ -57,16 +57,18 @@ This `securityContext` object enforces non-root hosting:
...
+        image: mcr.microsoft.com/dotnet/samples:aspnetapp
         securityContext:
           allowPrivilegeEscalation: false
           runAsNonRoot: true
+          runAsUser: 1654
 ```
 
 - [allowPrivilegeEscalation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) -- Prevents (if `false`) a process from gaining greater privileges than its parent process. This is a good setting, but not directly related to users.
 - `runAsNonRoot` -- Tests that the user (via uid) is a non-root user, otherwise fail.
+- `runAsUser` -- Sets the user; only needed if non set in the container image.
```
mthalman, line 69:
```suggestion
- `runAsUser` -- Sets the user; only needed if not set in the container image.
```

mthalman (approved):

lbussell (approved):

Issue #5084, opened 2024-01-04
Add `libatomic` in Alpine SDK images


lbussell:
Alpine images were left out of https://github.com/dotnet/dotnet-docker/issues/5033 due to [the wasm-tools workload not including musl binaries of the `mono-aot-cross` and `emcc` compilers](https://github.com/dotnet/sdk/issues/32327).

When https://github.com/dotnet/sdk/issues/32327 is fixed, we should verify it works with Alpine and add the `libatomic` library to the 9.0 SDK images (and 8.0 if it gets backported).

lbussell:
This is fixed in https://github.com/dotnet/dotnet-docker/pull/5123.

PR #5123, opened 2024-01-23
Add `libatomic` to .NET 9 Alpine SDK images

lbussell:
- Add libatomic to Alpine 9.0 SDK images
- Enable wasm-tools test for Alpine on .NET 9

Since https://github.com/dotnet/sdk/issues/32327 is fixed for .NET 9, this PR fixes https://github.com/dotnet/dotnet-docker/issues/5084. wasm-tools does not work on ARM in my experience. This would have been good to validate before we added libatomic to other distros' ARM SDK images. However, I'm not sure we can take that back now. That's why I've added it to the Alpine arm images as well.

We should investigate separately whether wasm-tools works for .NET 9 on ARM (it seems like [it should](https://github.com/dotnet/runtime/pull/89068))


tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -79,10 +79,13 @@ public async void VerifyBlazorWasmScenario(ProductImageData imageData)
                 return;
             }
 
-            // wasm-tools workload is not supported on Alpine or ARM, fall back to the unoptimized test
-            bool useWasmTools = !isAlpine && !imageData.IsArm;
+            // `wasm-tools` workload does not work on ARM
+            // `wasm-tools` is also not supported on Alpine for .NET < 9 due to https://github.com/dotnet/sdk/issues/32327
+            int[] unsupportedVersionsForAlpine = [6, 7, 8];
+            bool isSupportedVersionForAlpine = unsupportedVersionsForAlpine.Contains(imageData.Version.Major);
```
mthalman, line 85: I think the logic needs to be reversed here.

```suggestion
            bool isSupportedVersionForAlpine = !unsupportedVersionsForAlpine.Contains(imageData.Version.Major);
```

lbussell:
Filed https://github.com/dotnet/dotnet-docker/issues/5127 for unrelated Windows failures...

PR #5129, opened 2024-01-23
Add policy for rebuilding container images in response to CVEs

### LINKED ISSUE #2787: Define policy for when images get rebuilt to address package security vulnerabilities.
If there is a vulnerability in a package included in a .NET layer and there is a package patch available that addresses the CVE, when will the image get rebuilt to pick it up?  This should be captured in the [Image Update Policy](https://github.com/dotnet/dotnet-docker#image-update-policy).  The image update policy only discusses base image and .NET servicing.

* We update the supported .NET images within 12 hours of any updates to their base images (e.g. debian:buster-slim, windows/nanoserver:1909, buildpack-deps:bionic-scm, etc.).
* We publish .NET images as part of releasing new versions of .NET including major/minor and servicing.

lbussell:
Fixes #2787

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
 
 * We update supported .NET images within 12 hours of any updates to their base images (e.g. debian:bookworm-slim, windows/nanoserver:ltsc2022, etc.).
 * We re-build all .NET images as part of releasing new versions of .NET including new major/minor versions and servicing.
+* Distroless images such as Ubuntu Chiseled have no base image, and as such will only be updated with .NET releases and CVE fixes as described below.
 
 ### CVE Update Policy
 
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
 .NET container images are regularly monitored for the presence of CVEs. A given image will be rebuilt to pick up fixes for a CVE when:
 * We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
 * **AND** the CVE is in a package that is added in our Dockerfile layers (meaning the CVE is in a package we explicitly install or any transitive dependencies of those packages)
-* **AND** there is a fix for the package affected by the CVE available in the affected base image's package repository.
+* **AND** there is a CVE fix for the package available in the affected base image's package repository.
 
 ## Feedback
 
--- a/README.md
+++ b/README.md
@@ -93,8 +93,15 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
 
 ## Image Update Policy
 
-* We update the supported .NET images within 12 hours of any updates to their base images (e.g. debian:buster-slim, windows/nanoserver:ltsc2022, buildpack-deps:bionic-scm, etc.).
-* We publish .NET images as part of releasing new versions of .NET including major/minor and servicing.
+* We update supported .NET images within 12 hours of any updates to their base images (e.g. debian:bookworm-slim, windows/nanoserver:ltsc2022, etc.).
+* We re-build all .NET images as part of releasing new versions of .NET including new major/minor versions and servicing.
+
+### CVE Update Policy
+
+A given image will be rebuilt to pick up fixes for a CVE when:
+* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
+* **AND** the CVE is in a package that is added in our Dockerfile layers (meaning the CVE is in a package we explicitly install or any transitive dependencies of those packages)
+* **AND** there is a fix for the package affected by the CVE available in the affected base image's package feed.
 
 ## Feedback
 
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
 A given image will be rebuilt to pick up fixes for a CVE when:
 * We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
 * **AND** the CVE is in a package that is added in our Dockerfile layers (meaning the CVE is in a package we explicitly install or any transitive dependencies of those packages)
-* **AND** there is a fix for the package affected by the CVE available in the affected base image's package feed.
+* **AND** there is a fix for the package affected by the CVE available in the affected base image's package repository.
 
 ## Feedback
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
 
 ### CVE Update Policy
 
-A given image will be rebuilt to pick up fixes for a CVE when:
+.NET container images are regularly monitored for the presence of CVEs. A given image will be rebuilt to pick up fixes for a CVE when:
 * We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
 * **AND** the CVE is in a package that is added in our Dockerfile layers (meaning the CVE is in a package we explicitly install or any transitive dependencies of those packages)
 * **AND** there is a fix for the package affected by the CVE available in the affected base image's package repository.
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
 ### CVE Update Policy
 
 .NET container images are regularly monitored for the presence of CVEs. A given image will be rebuilt to pick up fixes for a CVE when:
-* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
+* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "Critical"
 * **AND** the CVE is in a package that is added in our Dockerfile layers (meaning the CVE is in a package we explicitly install or any transitive dependencies of those packages)
 * **AND** there is a CVE fix for the package available in the affected base image's package repository.
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,10 +6,12 @@ The .NET Core and ASP.NET Core support policy, including supported versions can 
 
 ## Reporting a Vulnerability
 
-Security issues and bugs should be reported privately to the Microsoft Security Response Center (MSRC), either by emailing secure@microsoft.com or via the portal at https://msrc.microsoft.com. 
-You should receive a response within 24 hours. If for some reason you do not, please follow up via email to ensure we received your 
+Security issues and bugs should be reported privately to the Microsoft Security Response Center (MSRC), either by emailing secure@microsoft.com or via the portal at https://msrc.microsoft.com.
+You should receive a response within 24 hours. If for some reason you do not, please follow up via email to ensure we received your
 original message. Further information, including the MSRC PGP key, can be found in the [MSRC Report an Issue FAQ](https://www.microsoft.com/en-us/msrc/faqs-report-an-issue).
 
 Reports via MSRC may qualify for the .NET Core Bug Bounty. Details of the .NET Core Bug Bounty including terms and conditions are at [https://aka.ms/corebounty](https://aka.ms/corebounty).
 
 Please do not open issues for anything you think might have a security implication.
+
+If you're reporting the presence of an already disclosed security vulnerability, such as a CVE, you must follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,4 +14,4 @@ Reports via MSRC may qualify for the .NET Core Bug Bounty. Details of the .NET C
 
 Please do not open issues for anything you think might have a security implication.
 
-If you're reporting the presence of an already disclosed security vulnerability, such as a CVE, you must follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).
+If you're reporting the presence of an already disclosed security vulnerability, such as a CVE, plaese follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,5 +14,5 @@ Reports via MSRC may qualify for the .NET Core Bug Bounty. Details of the .NET C
 
 Please do not open issues for anything you think might have a security implication.
 
-If you're reporting the presence of an already disclosed security vulnerability, such as a CVE, please follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).
+If you're reporting the presence of a disclosed security vulnerability, such as a CVE reported in one of our container images, please follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).
 
--- a/documentation/guiding-principles.md
+++ b/documentation/guiding-principles.md
@@ -37,17 +37,11 @@ See the [supported tags](supported-tags.md) for the tagging practices and polici
 
 1. Images will be included as part of the .NET release process. The Docker images will be released at the same time as the core product.
 
-1. Images will be rebuilt within 12 hours of any updates to their base images. For example, if a particular version of Alpine is patched, the .NET images based on this version of Alpine will be rebuilt with this new base image.
+1. Images will be rebuilt only as necessary in order to avoid downstream churn for consumers of .NET images.
+The [Image Update Policy](https://github.com/dotnet/dotnet-docker/blob/main/README.md#image-update-policy) section of the README contains the exact guidelines for when images will be updated.
 
 1. Images will never be deleted from the [official Docker repositories](https://hub.docker.com/_/microsoft-dotnet/). This does not apply to the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md).
 
 1. The [Dockerfiles](https://github.com/dotnet/dotnet-docker/search?q=filename%3ADockerfile) used to produce all of the images will be publicly available. Customers will be able to take the Dockerfiles and build them to produce their own equivalent images. No special build steps or permissions should be needed to build the Dockerfiles.
 
 1. No experimental Docker features will be utilized within the infrastructure used to produce the images. Utilizing experimental features can negatively affect the reliability of image production and introduces a risk to the integrity of the resulting artifacts.
-
-## CVE Update Policy
-
-A given image will be rebuilt to pick up fixes for a CVE when:
-* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
-* **AND** the CVE is in a package that is added in our Dockerfile layers (meaning the CVE is in a package we explicitly install or any transitive dependencies of those packages)
-* **AND** there is a fix for the package affected by the CVE available in the affected base image's package feed.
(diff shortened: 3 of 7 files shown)


documentation/guiding-principles.md
```diff
@@ -37,10 +37,23 @@ See the [supported tags](supported-tags.md) for the tagging practices and polici
...
+
+An image will be rebuilt to pick up fixes for CVEs when:
+
+- We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical".
+- **OR** when we detect the image contains a CVE that has not been assigned a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score for more than two weeks since the CVE was disclosed.
+- **OR** when we detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "None", "Low", or "Medium" and it has been more than two weeks since the CVE was disclosed.
+    - This is to give ample opportunity for distro maintainers to re-build their base images with package updates so that we avoid repeated re-builds of our images.
+- **AND** The CVE is in a package that is a direct dependency of .NET or is in a package that we explicitly install in our Dockerfiles.
+- **AND** There is a fix for the package affected by the CVE available in the affected base image's package feed.
```
lbussell, line 59: I understand that this could be a little hard to parse. I am amenable to creating a mermaid flowchart instead to represent this workflow if @MichaelSimons or @mthalman you think that would be more readable.

SECURITY.md
```diff
@@ -6,10 +6,12 @@ The .NET Core and ASP.NET Core support policy, including supported versions can
...
+You should receive a response within 24 hours. If for some reason you do not, please follow up via email to ensure we received your
 original message. Further information, including the MSRC PGP key, can be found in the [MSRC Report an Issue FAQ](https://www.microsoft.com/en-us/msrc/faqs-report-an-issue).
 
 Reports via MSRC may qualify for the .NET Core Bug Bounty. Details of the .NET Core Bug Bounty including terms and conditions are at [https://aka.ms/corebounty](https://aka.ms/corebounty).
 
-Please do not open issues for anything you think might have a security implication.
\ No newline at end of file
+Please do not open issues for anything you think might have a security implication.
+
+If you're reporting the presence of an already disclosed security vulnerability, such as a CVE, plaese follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).
```
MichaelSimons, line 17:
```suggestion
If you're reporting the presence of an already disclosed security vulnerability, such as a CVE, please follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).

```

documentation/guiding-principles.md
```diff
@@ -37,10 +37,23 @@ See the [supported tags](supported-tags.md) for the tagging practices and polici
...
+
+Since .NET container images are used as base images, excessive image rebuilds can cause immense usage of compute and network for our customers. We strive to rebuild our images only when absolutely necessary.
+
+Images will be rebuilt within hours of base image changes. For example, suppose a particular version of Alpine is patched. The .NET images based on this version of Alpine will be rebuilt with this new base image within hours of its release.
+
+An image will be rebuilt to pick up fixes for CVEs when:
+
+- We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical".
+- **OR** when we detect the image contains a CVE that has not been assigned a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score for more than two weeks since the CVE was disclosed.
+- **OR** when we detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "None", "Low", or "Medium" and it has been more than two weeks since the CVE was disclosed.
```
mthalman, line 56: Can't these two be combined?

`... the image contains a CVE with an unassigned CVSS score or a score of "None", "Low", or "Medium"...`

documentation/guiding-principles.md
```diff
@@ -37,10 +37,23 @@ See the [supported tags](supported-tags.md) for the tagging practices and polici
...
 
 1. The [Dockerfiles](https://github.com/dotnet/dotnet-docker/search?q=filename%3ADockerfile) used to produce all of the images will be publicly available. Customers will be able to take the Dockerfiles and build them to produce their own equivalent images. No special build steps or permissions should be needed to build the Dockerfiles.
 
 1. No experimental Docker features will be utilized within the infrastructure used to produce the images. Utilizing experimental features can negatively affect the reliability of image production and introduces a risk to the integrity of the resulting artifacts.
+
+## Image Rebuilds
+
+Since .NET container images are used as base images, excessive image rebuilds can cause immense usage of compute and network for our customers. We strive to rebuild our images only when absolutely necessary.
+
+Images will be rebuilt within hours of base image changes. For example, suppose a particular version of Alpine is patched. The .NET images based on this version of Alpine will be rebuilt with this new base image within hours of its release.
```
mthalman, line 50: This should be specific with the amount of hours.

documentation/guiding-principles.md
MichaelSimons: I think something is needed here to make the conditionals obvious.   The ORs and ANDs are confusing IMO.

documentation/guiding-principles.md
MichaelSimons, line 56: Can the previous OR be combined here?  e.g, with an unassigned score or a score of "None", "Low", or "Medium".

documentation/guiding-principles.md
```diff
@@ -37,10 +37,23 @@ See the [supported tags](supported-tags.md) for the tagging practices and polici
...
 
-1. Images will be rebuilt within hours of base image changes. For example, suppose a particular version of Alpine is patched. The .NET images based on this version of Alpine will be rebuilt with this new base image within hours of its release.
-
 1. Images will never be deleted from the [official Docker repositories](https://hub.docker.com/_/microsoft-dotnet/). This does not apply to the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md).
 
 1. The [Dockerfiles](https://github.com/dotnet/dotnet-docker/search?q=filename%3ADockerfile) used to produce all of the images will be publicly available. Customers will be able to take the Dockerfiles and build them to produce their own equivalent images. No special build steps or permissions should be needed to build the Dockerfiles.
 
 1. No experimental Docker features will be utilized within the infrastructure used to produce the images. Utilizing experimental features can negatively affect the reliability of image production and introduces a risk to the integrity of the resulting artifacts.
+
+## Image Rebuilds
```
mthalman, line 46: I see this section as more of a support policy statement that should belong in a separate doc. Since this is a principles doc, I would expect this to have a statement like "Rebuilding of images is kept at a minimum to avoid downstream churn". That's a principle. How we implement that principle is what should be in a policy doc (this section you've added).

documentation/guiding-principles.md
mthalman: I'd like to avoid a flowchart if possible.

documentation/guiding-principles.md
```diff
@@ -37,10 +37,23 @@ See the [supported tags](supported-tags.md) for the tagging practices and polici
...
+
+Images will be rebuilt within hours of base image changes. For example, suppose a particular version of Alpine is patched. The .NET images based on this version of Alpine will be rebuilt with this new base image within hours of its release.
+
+An image will be rebuilt to pick up fixes for CVEs when:
+
+- We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical".
+- **OR** when we detect the image contains a CVE that has not been assigned a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score for more than two weeks since the CVE was disclosed.
+- **OR** when we detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "None", "Low", or "Medium" and it has been more than two weeks since the CVE was disclosed.
+    - This is to give ample opportunity for distro maintainers to re-build their base images with package updates so that we avoid repeated re-builds of our images.
+- **AND** The CVE is in a package that is a direct dependency of .NET or is in a package that we explicitly install in our Dockerfiles.
```
MichaelSimons, line 58: What are indirect dependencies of .NET?  What I mean is a dependency of a direct dependency of .NET.

documentation/guiding-principles.md
lbussell: Do you see this separate policy doc being a part of the container vulnerability workflow? Or a separate doc, since that one is intended as a guide for users?

documentation/guiding-principles.md
richlander: I'm wondering if we can make this simpler.

> We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical".

Can we skip all those "ORs" and just do this? There is a cost to rebuilds and both distro maintainers and container users have a role to play here.

The key project here is to enable rebuilds. Once we have that, the policy should be easy to change. I propose we start with a very conservative policy and see how that goes.

mthalman: @richlander - We're needing to balance this with internal requirements (S360) that would, in theory, require us to rebuild more often than just "High" or "Critical".

documentation/guiding-principles.md
MichaelSimons, line 56: I know we had a discussion on this yesterday aimed at addressing s360 customers but I think it is wrong for us to rebuild outside of a monthly servicing cycle for anything but high and critical fixes.

documentation/guiding-principles.md
mthalman: I see it as part of a separate doc. As mentioned in the [original issue](https://github.com/dotnet/dotnet-docker/issues/2787), we should have this in https://github.com/dotnet/dotnet-docker#image-update-policy or a link from there to a standalone doc.

documentation/guiding-principles.md
```diff
@@ -37,10 +37,23 @@ See the [supported tags](supported-tags.md) for the tagging practices and polici
...
+
+## Image Rebuilds
+
+Since .NET container images are used as base images, excessive image rebuilds can cause immense usage of compute and network for our customers. We strive to rebuild our images only when absolutely necessary.
+
+Images will be rebuilt within hours of base image changes. For example, suppose a particular version of Alpine is patched. The .NET images based on this version of Alpine will be rebuilt with this new base image within hours of its release.
+
+An image will be rebuilt to pick up fixes for CVEs when:
+
+- We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical".
```
MichaelSimons, line 54: CVE scores can vary especially in the early stages after a CVE has been discovered.  How do we handle this?  Do we need to state what CVE source we will use?

MichaelSimons (commented):
Thanks for kicking off this conversation.

documentation/guiding-principles.md
richlander: I take your point.

My proposed (very simple) policy will still be an improvement. Let's start with the smallest change possible and measure the effect. If users (external or internal) complain, we can adapt.

SECURITY.md
```diff
@@ -14,4 +14,5 @@ Reports via MSRC may qualify for the .NET Core Bug Bounty. Details of the .NET C
 
 Please do not open issues for anything you think might have a security implication.
 
-If you're reporting the presence of an already disclosed security vulnerability, such as a CVE, plaese follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).
+If you're reporting the presence of an already disclosed security vulnerability, such as a CVE, please follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).
```
richlander, line 17: "already" is largely implied.

```suggestion
If you're reporting the presence of a disclosed security vulnerability, such as a CVE, please follow the documented guidance on vulnerability reporting in [vulnerability-reporting.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md).

```

documentation/guiding-principles.md
lbussell: Updated the wording there to refer to any packages that are added in any of our image layers.

documentation/guiding-principles.md
lbussell: Removed that condition for now.

documentation/guiding-principles.md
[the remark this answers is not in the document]
lbussell: Since I moved the content according to Matt's comment above, it's now located near another bullet that mentions servicing.

documentation/guiding-principles.md
lbussell: I removed the duplication between this doc and the main README. This now points to the README where we state our commitment of 12 hours for base image updates.

documentation/guiding-principles.md
```diff
@@ -37,23 +37,11 @@ See the [supported tags](supported-tags.md) for the tagging practices and polici
 
 1. Images will be included as part of the .NET release process. The Docker images will be released at the same time as the core product.
 
+1. Images will be rebuilt only as necessary in order to avoid downstream churn for consumers of .NET images.
```
richlander, line 40: I would be good to avoid a jargon word like "churn".

```suggestion
1. Images will be rebuilt only as necessary, to limit downstream image rebuild and redeployment activity.
```

README.md
```diff
@@ -93,8 +93,15 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
...
-* We publish .NET images as part of releasing new versions of .NET including major/minor and servicing.
+* We update supported .NET images within 12 hours of any updates to their base images (e.g. debian:bookworm-slim, windows/nanoserver:ltsc2022, etc.).
+* We re-build all .NET images as part of releasing new versions of .NET including new major/minor versions and servicing.
+
+### CVE Update Policy
+
+A given image will be rebuilt to pick up fixes for a CVE when:
+* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
+* **AND** the CVE is in a package that is added in our Dockerfile layers (meaning the CVE is in a package we explicitly install or any transitive dependencies of those packages)
+* **AND** there is a fix for the package affected by the CVE available in the affected base image's package feed.
```
mthalman, line 104: "Repository" or "archive" is the typical nomenclature.

```suggestion
* **AND** there is a fix for the package affected by the CVE available in the affected base image's package repository.
```

README.md
```diff
@@ -93,8 +93,15 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
 
 ## Image Update Policy
 
-* We update the supported .NET images within 12 hours of any updates to their base images (e.g. debian:buster-slim, windows/nanoserver:ltsc2022, buildpack-deps:bionic-scm, etc.).
-* We publish .NET images as part of releasing new versions of .NET including major/minor and servicing.
+* We update supported .NET images within 12 hours of any updates to their base images (e.g. debian:bookworm-slim, windows/nanoserver:ltsc2022, etc.).
+* We re-build all .NET images as part of releasing new versions of .NET including new major/minor versions and servicing.
+
+### CVE Update Policy
```
mthalman, line 99: Please also update https://github.com/dotnet/dotnet-docker/blob/main/documentation/vulnerability-reporting.md so that it refers to this.

documentation/vulnerability-reporting.md
```diff
@@ -415,7 +415,8 @@ These occur on Update Tuesday (the second Tuesday of each month).
 ### R. Log an issue at dotnet/dotnet-docker
 
 This vulnerability can be directly addressed by the .NET team.
-Log an issue at [dotnet/dotnet-docker](https://github.com/dotnet/dotnet-docker/issues/new/choose) and provide the vulnerability ID and the .NET image tag you've identified as being affected.
+Our [Image Update Policy](https://github.com/dotnet/dotnet-docker/blob/main/README.md#image-update-policy) defines when we will rebuild for vulnerabilities in .NET images.
```
mthalman, line 418: I think it's fine to keep this link here but I also think it should be listed up at the top of the doc because it provides a useful summary of when we update. That may be more convenient for someone to get the answer they need versus needing to wade through this whole workflow. So I think it'd be good to highlight it right away.

README.md
```diff
@@ -93,8 +93,15 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
...
+
+.NET container images are regularly monitored for the presence of CVEs. A given image will be rebuilt to pick up fixes for a CVE when:
+* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
+* **AND** the CVE is in a package that is added in our Dockerfile layers (meaning the CVE is in a package we explicitly install or any transitive dependencies of those packages)
+* **AND** there is a fix for the package affected by the CVE available in the affected base image's package repository.
```
MichaelSimons, line 104:
```suggestion
* **AND** there is a CVE fix for the package available in the affected base image's package repository.
```

README.md
```diff
@@ -93,8 +93,15 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
...
 
-* We update the supported .NET images within 12 hours of any updates to their base images (e.g. debian:buster-slim, windows/nanoserver:ltsc2022, buildpack-deps:bionic-scm, etc.).
-* We publish .NET images as part of releasing new versions of .NET including major/minor and servicing.
+* We update supported .NET images within 12 hours of any updates to their base images (e.g. debian:bookworm-slim, windows/nanoserver:ltsc2022, etc.).
+* We re-build all .NET images as part of releasing new versions of .NET including new major/minor versions and servicing.
+
+### CVE Update Policy
+
+.NET container images are regularly monitored for the presence of CVEs. A given image will be rebuilt to pick up fixes for a CVE when:
+* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
```
MichaelSimons, line 102: A comment I made earlier still applies - https://github.com/dotnet/dotnet-docker/pull/5129#discussion_r1464067746

MichaelSimons: I'm not sure what I think but I do want to raise the point.

documentation/guiding-principles.md
lbussell: No longer applies as we don't reference "None/low/medium" anymore.

lbussell: No longer applies as we don't reference "None/low/medium" anymore.

README.md
lbussell: Yes, I see the comment. One thought is we could have a "cooling-off" period for new CVEs, but that doesn't feel right especially if they are actually high or critical severity. Should we publicly state which scanning software we use? Could that leave our images open to attackers who are targeting vulnerabilities marked as medium severity by our software? I guess that could happen no matter which scanner we use. As a side note I feel like on average our reports over-estimate the severity of issues.

MichaelSimons: I agree that our reports tend to have a higher rating.  From a manual execution perspective, I think there is a human interpretation component.   For example the recent alpine openssl CVE was analyzed and realized to be much lower in priority than our scan results indicated.  The problem comes when wanting to automate the process.  Lets discuss at our next standup.

MichaelSimons (approved):
Signing off with the notion we will continue to discuss the scanner variance topic as make any adjustments that come from that conversation.

mthalman (approved):

richlander:
It occurred to me that we should consider switching Chiseled images to trigger on package updates only (+ .NET releases). Wouldn't that be better?

MichaelSimons:
> It occurred to me that we should consider switching Chiseled images to trigger on package updates only (+ .NET releases). Wouldn't that be better?

It is implied as the scratch base image never gets updated, but you are right in that it would be helpful to explicitly call this out.

README.md
```diff
@@ -100,7 +100,7 @@ The [Image Variants documentation](https://github.com/dotnet/dotnet-docker/blob/
 ### CVE Update Policy
 
 .NET container images are regularly monitored for the presence of CVEs. A given image will be rebuilt to pick up fixes for a CVE when:
-* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "High" or "Critical"
+* We detect the image contains a CVE with a [CVSS](https://nvd.nist.gov/vuln-metrics/cvss) score of "Critical"
```
mthalman, line 103: Please update these instances as well:

https://github.com/dotnet/dotnet-docker/blob/27ed26e141b943d8ae973ab7aed011f849c60b21/documentation/vulnerability-reporting.md?plain=1#L40

https://github.com/dotnet/dotnet-docker/blob/27ed26e141b943d8ae973ab7aed011f849c60b21/documentation/vulnerability-reporting.md?plain=1#L388

https://github.com/dotnet/dotnet-docker/blob/27ed26e141b943d8ae973ab7aed011f849c60b21/documentation/vulnerability-reporting.md?plain=1#L398-L403

https://github.com/dotnet/dotnet-docker/blob/27ed26e141b943d8ae973ab7aed011f849c60b21/documentation/vulnerability-reporting.md?plain=1#L412

mthalman (approved):

Issue #4932, opened 2023-10-09
Proposal for versioned samples images


lbussell:
### Describe the Problem

Some customers have expressed the desire for accessing old versions of samples in the dotnet/samples repo. We want .NET 8.0 samples to be compatible with older samples versions but still want to publish new samples with the latest .NET Docker features/best practices.

### Proposal

- Keep both `dotnetapp` and `aspnetapp` samples in the same `dotnet/samples` repo. This promotes visibility between different types of samples and also allows us to introduce more types of samples in the future with less overhead.
- Keep the floating tags published using traditional, rootful containers with a shell.
  - Consider moving to non-root or distroless for a future release.
- Add a new tag with a version suffix to all samples. This can be undocumented if necessary.
  - The new full tag structure would be `(sample app name)-(optional sample variant)-(.NET major version)`.
  - For each versioned tag, there should also be a floating versionless tag that we will point customers to. `(sample name)-(variant)`

#### Tags

I have provided several examples of the proposed tagging scheme below. The list is not exhaustive, the idea is just to provide some examples of how we could publish different types of samples with this naming scheme.

| Old Tags | New Tags | OS |
|---|---|---|
| `aspnetapp` | `aspnetapp`, `aspnetapp-8.0` | Alpine 3.18 |
| Not published | `aspnetapp-composite`, `aspnetapp-composite-8.0` | Jammy Chiseled |
| Not published | `aspnetapp-aot`, `aspnetapp-aot-8.0` | Jammy Chiseled |
| Not published | `aspnetapp-chiseled`, `aspnetapp-chiseled-8.0`<br>OR `aspnetapp-distroless`, `aspnetapp-distroless-8.0`... | Jammy Chiseled |
| | etc. | |
| `dotnetapp` | `dotnetapp`, `dotnetapp-8.0` | Alpine 3.18 |
| Not published | `dotnetapp-chiseled`, `dotnetapp-chiseled-8.0` | Jammy Chiseled |
| Not published | `dotnetapp-aot`, `dotnetapp-aot-8.0` | Jammy Chiseled |

lbussell:
Updated the tag listings to include OS versions.

lbussell:
Here is a greatly simplified proposal based on working with @richlander:


| Old Tags | New Tags | OS |
|---|---|---|
[...]

Issue #1619, opened 2020-01-23
PR build should build and test sample images


mthalman:
There is currently a separate [pipeline](https://github.com/dotnet/dotnet-docker/blob/master/eng/pipelines/dotnet-core-samples.yml) for running a sample build for PR validation.  This builds and tests the sample Dockerfiles but it ends up pulling the base .NET images the samples are based on.  So it's testing with already published base images rather than the content of the repo.

Instead the [dotnet-core](https://github.com/dotnet/dotnet-docker/blob/master/eng/pipelines/dotnet-core.yml) and [dotnet-core-nightly](https://github.com/dotnet/dotnet-docker/blob/master/eng/pipelines/dotnet-core-nightly.yml) pipelines should be updated to build the sample images in addition to the product images (but for PR builds only).

To make this work, the sample Dockerfiles for images that are published will need to be updated to have a `REPO` arg in the same way that the official product Dockerfiles do in order for a reference to a staging repo location to be passed when building.

MichaelSimons:
The official Dockerfiles make use of an ARG to be able to dynamically change the repo of the base image.  Using an ARG in the samples is undesirable.  This is not the standard pattern we would recommend to customers.

There is functionality in Image-Builder to [rewrite the from statements](https://github.com/dotnet/docker-tools/blob/8642cbbfda83b7e5df32a6a0094c4bed9ad322fc/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs#L286).  This code may be useful for this scenario.

MichaelSimons:
We don't think this has a high ROI.  The samples don't add significant test coverage over the existing Dockerfile unit tests.

PR #5135, opened 2024-01-24
Add versioned sample tags and publish Chiseled samples

lbussell:
This PR addresses #4932.

Samples aren't published from this (nightly) branch, but I'm not sure if we want to publish additional samples in the middle of a servicing release. I guess it doesn't matter that much if we are only adding tags, I'm happy to re-target this PR to `main`.


manifest.samples.json
```diff
@@ -81,13 +86,54 @@
...
+            "dotnetapp-chiseled-8.0": {},
+            "dotnetapp-chiseled": {}
+          },
+          "platforms": [
+            {
+              "dockerfile": "samples/dotnetapp/Dockerfile.chiseled",
+              "os": "linux",
+              "osVersion": "ubuntu",
+              "tags": {
+                "dotnetapp-8.0-chiseled-amd64": {},
```
mthalman, line 106: The sample tests will need to be updated to include these new image tags once these images get published. Please log an issue for that.

README.samples.md
mthalman, line 1: Do we want to include `*-chiseled` tags in the "Featured Tags` section of this doc? I feel like we should.

That's an open question for everyone. /cc @MichaelSimons, @richlander

manifest.samples.json
lbussell: Oh, thanks for the reminder. Why shouldn't we do that now?

mthalman: Because I would expect the tests to fail if you did that here. The sample tests don't actually test the images that get built. They are testing the images that are currently live on MAR. See https://github.com/dotnet/dotnet-docker/issues/1619

mthalman: Note this is specific to PR validation

README.samples.md
richlander: The featured tags are intended to be a subset of the tags. One option is to only include the chiseled images there. Including both the regular and non-chiseled images seems like a lot.

manifest.samples.json
lbussell: Filed https://github.com/dotnet/dotnet-docker/issues/5138 and also added chiseled samples to the existing non-published test data because they were missing.

README.samples.md
MichaelSimons: I lean towards including the chiseled images in the featured tags section.  

> The featured tags are intended to be a subset of the tags. 

Rest assured it would still a subset 😄 4 tags versus 69 for the full set if I counted right.

MichaelSimons (approved):

mthalman (approved):

PR #5147, opened 2024-01-31
[nightly] Update dotnet-monitor 6.3 and 7.3. Fix storage location.

wiktork:


jander-msft (approved):

jander-msft (commented):
You might have to regenerate the Dockerfiles from the templates so that the blob storage URLs are updated in the Dockerfiles.

jander-msft (approved):

jander-msft (approved):

PR #5149, opened 2024-01-31
Add dotnet monitor 9-preview1

wiktork:


manifest.json
```diff
@@ -10018,6 +10018,90 @@
             }
           ]
         },
+        {
```
jander-msft, line 10021: The 9.0 images should be place after the 8.1 images

manifest.json
```diff
@@ -10018,6 +10018,90 @@
...
+          "productVersion": "$(monitor|9.0|product-version)",
+          "sharedTags": {
+            "$(monitor|9.0|product-version)-ubuntu-chiseled": {},
+            "9.0-preview-ubuntu-chiseled": {},
+            "9.0-ubuntu-chiseled": {},
```
jander-msft, line 10026: The non-preview tags (e.g. `9.0-ubuntu-chiseled`) should not exist until RC.

manifest.json
```diff
@@ -10018,6 +10018,90 @@
...
+            "$(monitor|9.0|product-version)": {},
+            "9.0-preview": {},
+            "9": {},
```
jander-msft, line 10029:
```suggestion
            "9-preview": {},
```

src/monitor/9.0/cbl-mariner-distroless/arm64v8/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    \
+    && mkdir -p /app \
+    && tar -oxzf dotnet-monitor-egress-azureblobstorage.tar.gz -C /app \
+    && rm dotnet-monitor-egress-azureblobstorage.tar.gz \
+    && tar -oxzf dotnet-monitor-egress-s3storage.tar.gz -C /app \
+    && rm dotnet-monitor-egress-s3storage.tar.gz
+
+
+# .NET Monitor image
+FROM $REPO:9.1.0-alpha.1-cbl-mariner-distroless-arm64v8
```
jander-msft, line 30: Was this accidentally changed after Dockerfile update? It seems strange only one of the Dockerfiles would have this incorrect version.

wiktork: I missed one 8.1->9.0 update in the manifest.

jander-msft:
Please add entries in the tests for the 9.0 images as well as image size baselines, even if they are approximate.

tests/performance/Validate-ImageSize.ps1
```diff
@@ -30,7 +30,11 @@ try {
         $branch = ".nightly"
     }
     
+    #https://github.com/docker/compose/issues/8186
```
wiktork, line 33: I can revert this, but I found it necessary with the latest docker version.

jander-msft (approved):

tests/performance/Validate-ImageSize.ps1
lbussell: Was this just due to an issue with the new Docker version's output?

It looks like it should be, but I want to make sure this is backwards-compatible with the old version of Docker. I don't really use the this script for updating the image size baseline anymore, I usually just build the images and run `docker image inspect <image> --format "{{.Size}}"`. But this script is used by the pipeline agents whose Docker version is usually a few versions behind Docker Desktop.

wiktork: Yep, the newest versions have trouble with docker -f outputs. However, I have not really tested this with older versions, I will revert this for now to avoid any issues with the pipeline build.

jander-msft (approved):

Issue #5154, opened 2024-02-01
Automatically update dependencies for Aspire Dashboard


lbussell:
Follow-up to https://github.com/dotnet/dotnet-docker/pull/5148. Support for Aspire Dashboard should be added to the update-dependencies project so we can automatically keep the Aspire Dashboard version up-to-date.

lbussell:
@joperezr Would you be up for this? I'm happy to help if you would like to get the automated dependency flows working. Especially for Aspire 9.0 Previews once 8.0 GA is out. I won't be able to get to this any time soon.

lbussell:
(I closed it by accident, oops)

joperezr:
It's fine if you want to assign this to me, but I don't have the bandwidth to work on this in the short term. We are also likely not going to ship a 9.0 image for now and just stay with 8.0 one which means this would really only need updating once a month at most

lbussell:
No problem, if there are no plans for 9.0 preview images at the moment then there's not much point in implementing this any time soon.

lbussell:
Completed with https://github.com/dotnet/dotnet-docker/pull/6757

PR #5179, opened 2024-02-07
Add Aspire Dashboard support to update-dependencies tooling

lbussell:
First half of #5154.

What's done:
- Added Get-AspireDropVersions.ps1
- Factored out some common functions from Get-AspireDropVersions and Get-DropVersions into a PowerShell module that can be shared between multiple scripts. The only scripts left in eng/ are ones meant to be directly called.
- Added Aspire support to update-dependencies tool and used the tool to calculate shas for existing Aspire dashboard version
- Added the shas to the dashboard Dockerfiles

The next step will be to author a pipeline that runs the Get-AspireDropVersions script and Set-DotnetVersions scripts together on a schedule to take automatic dependency flows from the aspire repo. This will be similar to the way we get dependency updates from installer, but simpler.


eng/Get-Branch.ps1
mthalman, line 1: Looks like the run-tests.ps1 script depends on this file.

eng/DependencyManagement.psm1
```diff
@@ -3,17 +3,6 @@ Set-StrictMode -Version 2.0
 
 # Common functions for .NET Docker dependency management
 
-function Get-Branch() {
```
mthalman: Was it not possible for run-tests.ps1 to import this module like the other scripts?

lbussell: Yes but since that script isn't related to "DependencyManagement" I decided not to do that. I can go the other way though and make the module more general.

mthalman (approved):

PR #5202, opened 2024-02-15
Fix update-dependencies Docker build based on new Alpine version

lbussell:
Alpine 3.19 doesn't have OpenSSL 1 compat libraries. This requires an update to libgit2sharp. I updated a couple other package versions at the same time. I also took another look at updating S.CL but it was not trivial.

### DIFF
--- a/eng/update-dependencies/Dockerfile
+++ b/eng/update-dependencies/Dockerfile
@@ -24,8 +24,7 @@ FROM mcr.microsoft.com/dotnet/sdk:8.0-alpine
 
 # install Docker
 RUN apk add --no-cache \
-        docker \
-        openssl1.1-compat
+        docker
 
 # copy update-dependencies
 WORKDIR /update-dependencies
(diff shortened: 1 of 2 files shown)


eng/update-dependencies/Dockerfile
mthalman, line 1: Update this line too: https://github.com/dotnet/dotnet-docker/blob/4893ce28227d2817193a5b2638d80302266b587b/eng/update-dependencies/Dockerfile#L5

richlander: Does the package we're using not offer `linux-musl` artifacts?

lbussell: It looks like they do with this latest update: 

```
PS> get-childitem -r -name -filter "*libgit*" bin/debug/net8.0/runtimes/
linux-arm\native\libgit2-a2bde63.so
linux-arm64\native\libgit2-a2bde63.so
linux-musl-arm\native\libgit2-a2bde63.so
linux-musl-arm64\native\libgit2-a2bde63.so
linux-musl-x64\native\libgit2-a2bde63.so
linux-x64\native\libgit2-a2bde63.so
osx-arm64\native\libgit2-a2bde63.dylib
osx-x64\native\libgit2-a2bde63.dylib
```

mthalman (approved):

PR #5226, opened 2024-02-28
Add Aspire Dashboard image unit tests

### LINKED ISSUE #5155: Add unit tests for Aspire Dashboard images
Follow-up to https://github.com/dotnet/dotnet-docker/pull/5148. We should have unit tests for Aspire Dashboard images that match the other .NET images. These should tests things like environment variables, default user, Aspire or .NET Version, etc.

lbussell:
Fixes #5155


tests/Microsoft.DotNet.Docker.Tests/DotNetImageRepo.cs
```diff
@@ -0,0 +1,18 @@
...
+
+[Flags]
+public enum DotNetImageRepo
+{
+    SDK             = 1 << 0,
+    Runtime         = 1 << 1,
+    Runtime_Deps    = 1 << 2,
+    Aspnet          = 1 << 3,
+    Monitor         = 1 << 4,
+    Aspire_Dashboard = 1 << 5,
```
MichaelSimons, line 17: It looks like the intent is for the values to be column aligned; the addition of dashboard warrants changes to the other values.

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -142,8 +142,23 @@ public static string GetImageName(string tag, string repoName, string repoNameMo
             return $"{registry}{repo}:{tag}";
         }
 
-        protected string GetTagName(string tagPrefix, string os, string tagPostfix = null) =>
-            $"{tagPrefix}-{os}{(string.IsNullOrEmpty(tagPostfix) ? "" : $"-{tagPostfix}")}{GetArchTagSuffix()}";
+        // protected string GetTagName =>
```
MichaelSimons: Why this change/commented out code?  The new implementation doesn't look equivalent because the arch suffix is missing.

tests/Microsoft.DotNet.Docker.Tests/AspireDashboardImageTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        IEnumerable<EnvironmentVariableInfo> expectedVariables =
+        [
+            // Unset ASPNETCORE_HTTP_PORTS from base image
+            new EnvironmentVariableInfo("ASPNETCORE_HTTP_PORTS", string.Empty),
+            // These two URL environment variables should be in the more compact format, i.e. "http://+:18888", but need
+            // to have a base URL of 0.0.0.0 due to a bug in the Aspire Dashboard.
+            // Change the format when https://github.com/dotnet/dotnet-docker/issues/5190 is closed.
+            new EnvironmentVariableInfo("ASPNETCORE_URLS", "http://0.0.0.0:18888"),
+            new EnvironmentVariableInfo("DOTNET_DASHBOARD_OTLP_ENDPOINT_URL", "http://0.0.0.0:18889"),
+            new EnvironmentVariableInfo("DOTNET_gcServer", "1"),
```
joperezr, line 36: mm this doesn't look to be one we set in our image, is this one .NET platform related?

joperezr: Just for reference, here are the ones we look for: https://github.com/dotnet/aspire/blob/main/src/Aspire.Dashboard/README.md

joperezr (commented):
Not very familiar with the rest of the repo, but at least the tests for Aspire look good.

tests/Microsoft.DotNet.Docker.Tests/AspireDashboardImageTests.cs
lbussell: I set it here:
https://github.com/dotnet/dotnet-docker/blob/6c04222ed53397dc9e3fe605e8471eec45218536/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile#L33-L34

because I was referencing .NET Monitor. It seemed like a good idea at the time since it'll be running on servers, although if you don't think it's appropriate, we could leave it up to users to set.

https://github.com/dotnet/dotnet-docker/blob/6c04222ed53397dc9e3fe605e8471eec45218536/src/monitor-base/8.1/cbl-mariner-distroless/amd64/Dockerfile#L38-L39

lbussell: /cc @jander-msft

jander-msft: I checked the `Aspire.Dashboard.runtimeconfig.json` file of the current version that is the nightly branch and it's GC mode is Server already:

```
{
  "runtimeOptions": {
    ...
    "configProperties": {
      "System.GC.Server": true
      ...
    }
  }
}
```

Setting this environment variable for the image should not be necessary.

tests/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -79,16 +79,23 @@ public string BuildDistrolessHelper(DotNetImageRepo imageRepo, ProductImageData
             // version of the runtime-deps image get the correct image.
             ProductImageData runtimeDepsImageData = new()
             {
-                Version = imageData.VersionFamily,
+                // Special case for .NET 8.0 Aspire Dashboard images - the Dashboard is in preview even though
+                // .NET 8.0 is not. The distroless helper image should not be built with the preview version.
+                Version = imageData.Version == ImageVersion.V8_0_Preview ? ImageVersion.V8_0 : imageData.Version,
```
mthalman, line 84: Shouldn't it be like this to match the original logic?
```suggestion
                Version = imageData.Version == ImageVersion.V8_0_Preview ? ImageVersion.V8_0 : imageData.VersionFamily,
```

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -142,8 +142,20 @@ public static string GetImageName(string tag, string repoName, string repoNameMo
...
+            foreach (string tagPart in new string[] { os, tagPostfix })
+            {
+                if (!string.IsNullOrEmpty(tagPart))
+                {
+                    tagParts.Add(tagPart);
+                }
+            }
+
+            return string.Join('-', tagParts);
+        }
```
mthalman, line 158: We're missing `GetArchTagSuffix` in this new logic.

tests/Microsoft.DotNet.Docker.Tests/TestScenarios/AspireDashboardBasicScenario.cs
```diff
@@ -32,25 +34,33 @@ public class AspireDashboardBasicScenario : ITestScenario
...
         {
             _dockerHelper.DeleteContainer(containerName);
         }
     }
+
+    private static int GetDashboardPort(string imageTag, DockerHelper dockerHelper)
+    {
+        IDictionary<string, string> envVars = dockerHelper.GetEnvironmentVariables(imageTag);
+        string portString = envVars[DashboardPortEnvVar].Split(':')[^1];
+        return int.Parse(portString);
```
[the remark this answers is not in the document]
lbussell: I had that in [a previous version ](https://github.com/dotnet/dotnet-docker/pull/5226/commits/1e6eb0d975c448e20c1a456c280af2b93d9e2517)but I figured reading from the image would be better so we don't have to define the port in two places. I am fine with switching back to a const if this is over-engineering.

tests/Microsoft.DotNet.Docker.Tests/TestScenarios/SampleTestScenario.cs
```diff
@@ -7,7 +11,7 @@
 
 namespace Microsoft.DotNet.Docker.Tests;
 
-public abstract class TestScenario : IDisposable
+public abstract class SampleTestScenario : ITestScenario, IDisposable
```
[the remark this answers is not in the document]
lbussell: I wanted to make it clearer that all `TestScenarios/SampleTestScenarios` are based on specific `dotnet new ...` [templates](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-new). Maybe `DotnetTemplateTestScenario` would be a better name?

tests/Microsoft.DotNet.Docker.Tests/TestScenarios/AspireDashboardBasicScenario.cs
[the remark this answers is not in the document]
lbussell: I switched to using consts and made sure to only define the ports once.

mthalman (approved):

PR #5231, opened 2024-02-29
Update Syft version to `v1.0.0`

lbussell:
Release notes: https://github.com/anchore/syft/releases/tag/v1.0.0

At some point `syft packages` was renamed to `syft scan`. Not sure which release that was in. I also moved the Syft version to manifest.versions.json so that maybe we can automate this some day.


mthalman (commented):
There are some diffs in the output that the tests are complaining about.

mthalman (approved):

Issue #5181, opened 2024-02-07
Produce Ubuntu 24.04 images (Noble Numbat)


richlander:
The next Ubuntu LTS is practically on our doorstep.

We should do the following:

- Do nothing for .NET 6/7
- Add Noble to .NET 8
- Ship only Noble (for Ubuntu) for .NET 9

lbussell:
[Triage] We discussed if 9.0-preview.3 images should include Noble in the main branch before it is release. Since 9.0 is already a preview, it should be acceptable to ship Noble in the main branch as a preview for what we will actually ship. We don't want to maintain Jammy support for 9.0 since it's not in our plans to ship it.

richlander:
Sounds great!

lbussell:
These images are now available for .NET 8: https://github.com/dotnet/dotnet-docker/discussions/5466

.NET 9 Noble images are available in the nightly repos. When Preview 4 is released, they will be available in the non-nightly repos.

PR #5241, opened 2024-03-05
Add Ubuntu 24.04 "noble" images

lbussell:
- Added Noble images to 8.0
- Replaced Jammy with Noble for 9.0 preview

We also needed a new solution for the AOT cross Dockerfiles since Ubuntu 24.04 is changing to the DEB822 standard for package source lists: https://manpages.debian.org/stretch/apt/sources.list.5.en.html

Related: https://github.com/dotnet/dotnet-docker/issues/5181


eng/dockerfile-templates/sdk/Dockerfile.linux.aot
```diff
@@ -5,6 +5,7 @@
     set isAlpine to find(OS_VERSION, "alpine") >= 0 ^
     set isMariner to find(OS_VERSION, "cbl-mariner") >= 0 ^
     set isUbuntu to find(OS_ARCH_HYPHENATED, "Ubuntu") >= 0 ^
+    set ubuntuVersion to when(find(OS_VERSION, "jammy") >= 0, "jammy", "noble") ^
```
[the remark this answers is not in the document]
lbussell: It actually doesn't save any lines of code to have this variable, so I removed it.

richlander:
FYI @baronfel

lbussell:
I was not able to reproduce the previous arm32 permissions issues locally today so we'll see if it passes PR validation this time.

lbussell:
Removed Noble arm32 images due to https://bugs.launchpad.net/ubuntu/+source/tar/+bug/2059734.

lbussell:
Filed https://github.com/dotnet/dotnet-docker/issues/5319 to follow-up on the arm32 issue.

lbussell:
Current issue: Noble package repos are undergoing a huge re-build since the compromised versions of `liblzma` and `xz-utils` were in the build environment of every package built since February 26. Not all packages are re-built yet so that's why we're getting unresolvable dependencies. We should continue retrying a couple times a day. [Source](https://discourse.ubuntu.com/t/whats-happening-in-noble-repositories/43729/10)

lbussell:
According to the [Noble release schedule](https://discourse.ubuntu.com/t/noble-numbat-release-schedule/35649), it's probably best to wait until the Beta release date (2024-04-11) to merge these images.

lbussell:
Current issue - the Chisel tool isn't able to install some packages from the archive: 
- https://github.com/canonical/chisel-releases/issues/216

richlander:
I'm glad we're catching these issues early. Our support for Ubuntu has become more complex, so the early work is definitely warranted.

lbussell:
The Noble arm32 SDK image is going to take additional work to work around the `tar` arm32 issue, since it isn't yet multi-stage. https://github.com/dotnet/dotnet-docker/pull/5241/files#diff-9f4ce0ee04bfb6f01b0ba93e25a9b6a5eaef983ab2b50d7791dc1c7dafa29a85

mthalman:
@lbussell - Looks like Dockerfiles need to be regenerated.

eng/dockerfile-templates/sdk/Dockerfile.linux
```diff
@@ -11,9 +11,13 @@
         VARIABLES[cat("dotnet|", dotnetVersion, "|product-version")]
         VARIABLES[cat("dotnet|", dotnetVersion, "|fixed-tag")]) ^
     set baseImageTag to cat("$REPO:", tagVersion, "-", OS_VERSION, ARCH_TAG_SUFFIX) ^
+
+    _ Workaround for https://bugs.launchpad.net/ubuntu/+source/tar/+bug/2059734. Remove when fixed. ^
+    set osVersionBase to when(OS_VERSION_BASE = "noble" && ARCH_VERSIONED = "arm32v7", "jammy", OS_VERSION_BASE) ^
```
mthalman, line 16: The `when` condition should just use `useNobleArm32Workaround` instead.

eng/dockerfile-templates/sdk/Dockerfile.linux
```diff
@@ -78,8 +83,9 @@ RUN {{InsertTemplate("../Dockerfile.linux.install-pkgs",
     [
         "install-method": when(isRpmInstall, "download", "download-and-install"),
         "is-internal": isInternal,
-        "url-suffix": "$SAS_QUERY_STRING",
-        "is-rpm-install": isRpmInstall
+        "url-suffix": when(useNobleArm32Workaround, "", "$SAS_QUERY_STRING"),
```
mthalman, line 86: Why would the SAS query string not be used when this workaround is enabled?

lbussell: Switched the condition to check `isInternal` instead.

mthalman (approved):

lbussell:
The issue is that we can't establish an SSL connection to our NuGet feeds due to a time issue. This is probably due to the t64 migration. I am doing some additional investigation to determine where/if we need to respond to this in .NET.

```
 #10 2.041 info : X.509 certificate chain validation will use the fallback certificate bundle at '/usr/share/dotnet/sdk/9.0.100-preview.4.24221.5/trustedroots/codesignctl.pem'.
 #10 2.129 info : X.509 certificate chain validation will use the fallback certificate bundle at '/usr/share/dotnet/sdk/9.0.100-preview.4.24221.5/trustedroots/timestampctl.pem'.
 #10 2.133 info : Adding PackageReference for package 'foo' into project '/source/app/app.csproj'.
 #10 8.469 error: Unable to load the service index for source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json.
 #10 8.469 error:   The SSL connection could not be established, see inner exception.
 #10 8.469 error:   The remote certificate is invalid because of errors in the certificate chain: NotTimeValid
```

lbussell:
Filed https://github.com/dotnet/runtime/issues/101444 for the arm32 issue.

mthalman (approved):

richlander:
Woohoo!

PR #5251, opened 2024-03-08
Add `DOTNET_RUNTIME_ID` environment variable to .NET 6 Alpine 3.19 images

lbussell:
Context: https://github.com/dotnet/dotnet-docker/issues/5243#issuecomment-1986096650

This is a draft PR so that we can review and talk about this change. It is up in the air whether or not we will take this change yet, depending on if we decide to address the issue in the .NET 6 Runtime.

I also have a work-in-progress test for this issue, but it's not complete yet.

Related: https://github.com/dotnet/runtime/issues/96155


richlander:
You know ... if we can just do this and this is the only real requirement to add a RID to .NET 6, then maybe this is good enough for the remainder of .NET 6 support.

@agocke

PR #5255, opened 2024-03-11
Update Json Serialization with releasesapi sample

richlander:
This sample was failing, as reported by @shanselman.

I also edited `.gitconfig` to allow committing `launchSettings.json` to source control. I believe this file is required for correctness. The sample fails w/o it. It is a template-provided file, so I assume it is intended for source control.


lbussell:
Just tried this out and I needed to use HTTPS to access these endpoints, HTTP wouldn't work as written in the README. Can we change these?

https://github.com/richlander/dotnet-docker/blob/460b88928b3cdc61fc20f381bcdd39d8df6c8090/samples/releasesapi/README.md?plain=1#L20-L21

richlander:
That's why I added `launchSettings.json`. DId you try that from my branch or main?

lbussell (approved):
I tried again and it worked with HTTP. I was definitely on your branch. Not sure what I did differently. Seems good to me.

richlander:
Odd. Was just going through the motions of building this. You replied about 10s after I started `docker build` on the sample! Just tried again and it works.

Issue #5153, opened 2024-02-01
Add README content for Aspire Dashboard images


lbussell:
The README content for the Aspire Dashboard images should detail how exactly they should be used and/or point to official Aspire documentation on how to use the Dashboard project.

richlander:
The Aspire team should own this content and make their signoff on this PR a pre-req of shipping their images.

@joperezr

joperezr:
cc @kvenkatrajan @drewnoakes @davidfowl

davidfowl:
Sounds good, where does it go? In the readme?

joperezr:
No, that will live in this repo. I'm assuming next to the Dockerfile @lbussell? Which means once #5148 goes in it would be in `src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/README`

kvenkatrajan:
Thanks @joperezr . Can we get a sample of what this should like for a container image?  Sorry haven’t done one if this before.

MichaelSimons:
> Sounds good, where does it go? In the readme?

The readme's are generated from a template.  Once https://github.com/dotnet/dotnet-docker/pull/5148 is merged you will see the generated readme at the root of the repo.  One of the main pieces that needs your content is here - https://github.com/dotnet/dotnet-docker/pull/5148/files#diff-5550790524c4ed6cc5505b1ce1767c913bfa0b75fc74a5028ecf5bd0f2547044

> Can we get a sample of what this should like for a container image? Sorry haven’t done one if this before.

I would suggest looking at the readme's for the other images - examples include

https://github.com/dotnet/dotnet-docker/blob/main/README.runtime.md
https://github.com/dotnet/dotnet-docker/blob/main/README.aspnet.md
https://github.com/dotnet/dotnet-docker/blob/main/README.monitor.md

After https://github.com/dotnet/dotnet-docker/pull/5148 is merged, the aspire readme will look like: https://github.com/dotnet/dotnet-docker/blob/a0384a3d75794221fa132e1b90eeb24fb1c1d167/README.aspire-dashboard.md

MichaelSimons:
The readme content appears now.

MichaelSimons:
Is there any work remaining here or can this be closed?

lbussell:
Completed with https://github.com/dotnet/dotnet-docker/pull/5326

PR #5307, opened 2024-03-26
Update Aspire dashboard version to latest preview 5

joperezr:
cc: @lbussell 

FYI: @davidfowl @kvenkatrajan @drewnoakes

### DIFF
--- a/manifest.versions.json
+++ b/manifest.versions.json
@@ -1,11 +1,11 @@
 {
   "variables": {
-    "aspire-dashboard|8.0|build-version": "8.0.0-preview.5.24157.3",
+    "aspire-dashboard|8.0|build-version": "8.0.0-preview.5.24172.8",
     "aspire-dashboard|8.0|product-version": "8.0.0-preview.5",
     "aspire-dashboard|8.0|fixed-tag": "$(aspire-dashboard|8.0|product-version)",
     "aspire-dashboard|8.0|floating-tag": "8.0-preview",
-    "aspire-dashboard|8.0|linux|x64|sha": "686ff138636e589126f9d33ca01a71a336704bac5a74dbf3f3de6e108f0739e647154a73db8ce419148f24447670884531d78613bf07563d01d1199887cf6e78",
-    "aspire-dashboard|8.0|linux|arm64|sha": "695ce3d6440ec0942fcf0e5b13e696beee1074189957f8f679128850db6e754afe2999974560d3c0f2b8dc2b41a5389a71a4beb5c086c0efc2d534572fe0b54f",
+    "aspire-dashboard|8.0|linux|x64|sha": "3bfb8d6b97a48d0265674fa5f3acd2bc9871fbf2ea7389eee9b4d5cc58392d28c210fbb149e6509081112267d40208d8e5b8eb3729996b6aead37895a5d2198b",
+    "aspire-dashboard|8.0|linux|arm64|sha": "b4ddfc655e4417fd11506ee72c10d6d265e451e4487e56a0dd419bb7b2c4d42fa719c61bfdaae6550f7d8719738dd0c543093402f6332a1099b322253eb673ca",
 
     "aspnet|6.0|build-version": "6.0.28",
     "aspnet|6.0|targeting-pack-version": "$(aspnet|6.0|build-version)",
--- a/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile
+++ b/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile
@@ -9,9 +9,9 @@ RUN tdnf install -y \
     && tdnf clean all
 
 # Retrieve Aspire Dashboard
-RUN dotnet_aspire_version=8.0.0-preview.5.24157.3 \
+RUN dotnet_aspire_version=8.0.0-preview.5.24172.8 \
     && curl -fSL --output aspire_dashboard.zip https://dotnetbuilds.azureedge.net/public/aspire/$dotnet_aspire_version/aspire-dashboard-linux-x64.zip \
-    && aspire_dashboard_sha512='686ff138636e589126f9d33ca01a71a336704bac5a74dbf3f3de6e108f0739e647154a73db8ce419148f24447670884531d78613bf07563d01d1199887cf6e78' \
+    && aspire_dashboard_sha512='3bfb8d6b97a48d0265674fa5f3acd2bc9871fbf2ea7389eee9b4d5cc58392d28c210fbb149e6509081112267d40208d8e5b8eb3729996b6aead37895a5d2198b' \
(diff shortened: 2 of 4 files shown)


lbussell (approved):

lbussell:
@joperezr did something with the port change? Looks like the test is failing to hit the endpoint.

joperezr:
cc: @kvenkatrajan @drewnoakes can you help answering @lbussell's question?

joperezr:
I tested the generated image and the issue seems to be related to our new changes for adding Auth to OTLP endpoints. cc @JamesNK, @davidfowl .

This is the output I get when trying to use the image:
```
Unhandled exception. System.InvalidOperationException: Configuration of OTLP endpoint authentication is required. Either specify DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS with a value of true, or specify DOTNET_DASHBOARD_OTLP_AUTH_MODE. Possible values: None, ApiKey, ClientCertificate
   at Aspire.Dashboard.DashboardWebApplication.LoadDashboardConfig(IConfiguration configuration) in /_/src/Aspire.Dashboard/DashboardWebApplication.cs:line 403
   at Aspire.Dashboard.DashboardWebApplication..ctor(Action`1 configureBuilder) in /_/src/Aspire.Dashboard/DashboardWebApplication.cs:line 66
   at Program.<Main>$(String[] args) in /_/src/Aspire.Dashboard/Program.cs:line 7
```

I'm assuming we'd not want to set any of those two and instead just provide docs for users so that they set these variables as appropriate, and for the tests in this repo we should probably just launch the container with the extra env variable for Allow anonymous. Thoughts?

lbussell:
> I'm assuming we'd not want to set any of those two and instead just provide docs for users so that they set these variables as appropriate, and for the tests in this repo we should probably just launch the container with the extra env variable for Allow anonymous. Thoughts?

Thanks for the investigation @joperezr as I've been pretty busy. We can definitely launch the test containers with `DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS=true`

joperezr:
I pushed a change which will fix the test, but it would be good to check if we also want to change Dockerfile to set some defaults.

davidfowl:
> I pushed a change which will fix the test, but it would be good to check if we also want to change Dockerfile to set some defaults.

Which dockerfile?

JamesNK:
The intent of `DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS` is for people to make an explicit decision to run without auth. It's not a default.

JamesNK:
btw I think we might rename this setting to `DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS` (insecure -> unsecured).

See https://github.com/dotnet/aspire/pull/3233

joperezr:
> > I pushed a change which will fix the test, but it would be good to check if we also want to change Dockerfile to set some defaults.
> 
> Which dockerfile?

Aspire dashboard [dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile). So essentially, before the addition of this check, one could just do `docker run` on the image and it would run the dashboard container. Now, you will get the exception message I pointed out above, unless you either set DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS to true, or alternatively (and preferably) you set DOTNET_DASHBOARD_OTLP_AUTH_MODE to the appropriate mode. My question is, what do we want the default experience is for someone using our Docker image? Should we expect them to pass one of these environment variables always to their docker run commands (or compose files or whatever), or do we want to provide a default as part of the image which can then later be overriden?

JamesNK:
We want it to be required. Specifying the env var is a user gesture to indicate that they intentionally want to be unsecured.

davidfowl:
Will we show this as the instruction to run in the docs?

joperezr:
> Will we show this as the instruction to run in the docs?

We should, yes. We have a README for our docker image here: https://github.com/dotnet/dotnet-docker/blob/nightly/README.aspire-dashboard.md and there is an open issue tracking filling the content of that before we move to the prod registry: https://github.com/dotnet/dotnet-docker/issues/5153

I've been holding off on filling that README as we still haven't finalized how we expect the interaction of the image should be (meaning we are still adding env vars and config switches) but once we are settled we should fill that in.

MichaelSimons:
> The intent of `DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS` is for people to make an explicit decision to run without auth. It's not a default.

From a general UX perspective of the Docker aspire image this gives me an allergic reaction.  People are going to pull this image, run it, and not expect an exception.  I can see this as a source of frustration for customers and therefore be source of issues/discussions/questions.  Even if users have to read docs to know how to use the image, I think having the run experience fail is not ideal.

joperezr:
Also, to state the obvious, we don't want users to use `DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS`, as we instead want them to configure their container with the right parameters such that they run with auth enabled and in a secure manner. The INSECURE env variable should only be set if as a consumer, you understand the risks, and want to proceed anyway, but it is by no means our recommendation. We of course plan to document all of this in the README.

MichaelSimons:
I get all that but I know from experience users are going to pull these images and try running them before reading any documentation.  This was the perspective that prompted my earlier comments.  Having the container do something successful even if the user don't know what to do next is better than throwing an exception IMO.

joperezr:
So given our stance that I described above, what would be your suggestion? Should we change our Dockerfile so that instead of having the entrypoint launch the dashboard to instead launch a script that checks if the required configuration is met, and if not, to print a helpful message? (and of course if it finds the right configs to just launch the app)

jander-msft:
Can the dashboard be changed to run without additional configuration but then display some kind of diagnostic in the shell itself that reports that it hasn't been configured to do anything and provides an fwlink/aka.ms link to docs on how to configure it?

joperezr:
Yes, if you set the environment variable to "opt-in" for insecure mode, the dashboard will run, but since we don't know if people will ready the output and check the note about the fwlink, We'd instead want to make sure that users are explicitly opting-in for this and that it is not a default in the image. This is essentially what @JamesNK said above about: `We want it to be required. Specifying the env var is a user gesture to indicate that they intentionally want to be unsecured.`

joperezr:
This is one of the ones were we want to force the consumer to read the docs 😃

mthalman:
If this is the desired default experience and we want to force the user to read the docs, the error message should probably direct them to the docs with a URL.

MichaelSimons:
> So given our stance that I described above, what would be your suggestion? Should we change our Dockerfile so that instead of having the entrypoint launch the dashboard to instead launch a script that checks if the required configuration is met, and if not, to print a helpful message? (and of course if it finds the right configs to just launch the app)

I lean towards making the exception message as helpful as possible.  e.g. linking to the appropriate documentation.

PR #5326, opened 2024-04-02
Adding Readme contents for the Aspire container image.

### LINKED ISSUE #5153: Add README content for Aspire Dashboard images

joperezr:
Fixes #5153 

@drewnoakes @kvenkatrajan @JamesNK @davidfowl Grabbed most of the content from https://github.com/dotnet/aspire/blob/main/src/Aspire.Dashboard/README.md and just adapted variable names by swapping `:` for `__`. Please let me know if there is anything that should be changed here, or if there is any other feedback around the documentation structure.

@lbussell I tried several times but wasn't able to run the generate script to evaluate all the templates and generate the actual READMEs. Happy to follow any additional instructions on how to do this.

### DIFF
--- a/README.aspire-dashboard.md
+++ b/README.aspire-dashboard.md
@@ -17,8 +17,6 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
 
 # Usage
 
-The [.NET Docker samples](https://github.com/dotnet/dotnet-docker/blob/main/samples/README.md) show various ways to use .NET and Docker together. See [Building Docker Images for .NET Applications](https://docs.microsoft.com/dotnet/core/docker/building-net-docker-images) to learn more.
-
 The .NET Aspire Dashboard is a browser-based app to view run-time information about your distributed application.
 
 The dashboard shows:
--- a/README.aspire-dashboard.md
+++ b/README.aspire-dashboard.md
@@ -19,12 +19,80 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
 
 The [.NET Docker samples](https://github.com/dotnet/dotnet-docker/blob/main/samples/README.md) show various ways to use .NET and Docker together. See [Building Docker Images for .NET Applications](https://docs.microsoft.com/dotnet/core/docker/building-net-docker-images) to learn more.
 
+The .NET Aspire Dashboard is a browser-based app to view run-time information about your distributed application.
 
+The dashboard shows:
 
-# Image Variants
+- Resources that make up your app, such as .NET projects, executables and containers.
+- Live console logs of resources.
+- Live telemetry, such as structured logs, traces and metrics.
 
-.NET container images have several variants that offer different combinations of flexibility and deployment size.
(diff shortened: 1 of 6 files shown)


README.aspire-dashboard.md
```diff
@@ -19,12 +19,80 @@ Watch [discussions](https://github.com/dotnet/dotnet-docker/discussions/categori
 
 The [.NET Docker samples](https://github.com/dotnet/dotnet-docker/blob/main/samples/README.md) show various ways to use .NET and Docker together. See [Building Docker Images for .NET Applications](https://docs.microsoft.com/dotnet/core/docker/building-net-docker-images) to learn more.
 
+The .NET Aspire Dashboard is a browser-based app to view run-time information about your distributed application.
 
+The dashboard shows:
```
lbussell, line 24: This section here ends up reading a little awkward with the samples line that gets automatically injected. Maybe we should remove it while there aren't any Aspire Dashboard samples.

MichaelSimons:
My general question is if it makes sense to keep the aspire repo as the home for the core documentation and link to it from the readme?  I would like to avoid introducing a maintenance issue and having the doc be out of date.  The monitor usage text simply references their primary documentation.

joperezr:
> My general question is if it makes sense to keep the aspire repo as the home for the core documentation and link to it from the readme? I would like to avoid introducing a maintenance issue and having the doc be out of date. As a reference, the monitor usage text simply references their primary documentation.

We could totally do that, the only 2 main drawbacks I see with that are:

- It will require an extra click navigation to get to the info that the user's want/need to set their container up
- How can we make sure that the link it points to are the docs that match the version of the container image? For instance, if we rename an environment variable, or add new ones, people using old images might be miss-leaded. Curious on how dotnet/monitor solves this issue.

MichaelSimons:
Ultimately this is your call.  As I eluded to earlier the amount and details in the documentation raised concerns that it could easily become stale.

> It will require an extra click navigation to get to the info that the user's want/need to set their container up

Generally the readme's are an entry point.  This does not concern me.

> How can we make sure that the link it points to are the docs that match the version of the container image? For instance, if we rename an environment variable, or add new ones, people using old images might be miss-leaded. Curious on how dotnet/monitor solves this issue.

I would expect the documentation link would align with the version of Aspire being served up in the containers.  If you get to a place where there are multiple versions served via containers, then I would expect the documentation would also support the various versions.

lbussell:
> * How can we make sure that the link it points to are the docs that match the version of the container image? For instance, if we rename an environment variable, or add new ones, people using old images might be miss-leaded. Curious on how dotnet/monitor solves this issue.

We have had no user complaints about having all of our samples and documentation reference .NET 8 only, while we still service both .NET 7 and 6 with the same image repos, for example. We will want to make sure the documentation (in `main`) refers to the latest stable release.

lbussell (approved):

lbussell:
Merging this to include in the servicing PR. Any further edits can be included in the servicing PR or a follow-up PR to `main` branch and ported to `nightly`.

PR #5396, opened 2024-04-24
Use manifest version variables for .NET Monitor tags

jander-msft:
Similar to the .NET 9 image entries, use manifest version variables for the floating .NET Monitor tags to ease updating from using preview tags to non-preview tags in the future.

One additional minor change was made to move all of the .NET Monitor 9 versions after the 8.1 versions to correctly order the variables.

The generated Dockerfiles and readmes are not impacted by this change since none of the versions have changed.


manifest.versions.json
```diff
@@ -143,40 +143,34 @@
...
     "monitor-ext-azureblobstorage|8.1|linux|arm64|sha": "ed8089cb3c5c7fef56aa82cc65505beff074e76b9324aff51edc35f559ed962f6b8ca4d03ef969cd72e246fa1370136ffb9419f0fecba307b50e9c8d1d6e880b",
-
     "monitor-ext-s3storage|8.1|linux|x64|sha": "c60b4b73c205c6aa1c04c2163e6e542b83b5e1f0a797b403f36e34e6e48c46e868497c2c977b7e26b91d18d62180fd0f3e7e44e242622663379e0ed35370ee60",
     "monitor-ext-s3storage|8.1|linux|arm64|sha": "77966b231a527f38e8d6657a9b4b55687aefc0366f3f1e55dd0c2ce5448007f10e2ebce21dff5feed6ba03bd33b40f16bea9fac0dddc3aadf85a88cf313004c8",
 
+    "monitor|9.0|build-version": "9.0.0-preview.4.24222.6",
+    "monitor|9.0|product-version": "9.0.0-preview.4",
+    "monitor|9.0|tag|fixed": "$(monitor|9.0|product-version)",
+    "monitor|9.0|tag|majorminor": "9.0-preview",
+    "monitor|9.0|tag|major": "9-preview",
```
mthalman, line 166: There should be a consistent naming convention for these variables across the repos. The other repos are using `fixed-tag` and `floating-tag`. So either these changes should adopt that convention or the others should be updated to this pattern.

I understand that you need to account for a "major-only" version tag. If you adopt the existing pattern, one way to do this is to have `monitor|9.0|tag|floating-tag|major-only` or something like that.

manifest.json
```diff
@@ -11446,11 +11446,11 @@
...
-            "$(monitor|9.0|product-version)-ubuntu-chiseled": {},
-            "9.0-preview-ubuntu-chiseled": {},
-            "$(monitor|9.0|product-version)": {},
-            "9.0-preview": {},
-            "9-preview": {},
+            "$(monitor|9.0|fixed-tag)-ubuntu-chiseled": {},
+            "$(monitor|9.0|floating-tag|majorminor)-ubuntu-chiseled": {},
+            "$(monitor|9.0|fixed-tag)": {},
+            "$(monitor|9.0|floating-tag|majorminor)": {},
+            "$(monitor|9.0|floating-tag|major)": {},
```
lbussell, line 11453: What about just `major-tag`, `minor-tag`, and `fixed-tag`?

lbussell (approved):
@mthalman thoughts on applying this new pattern to other images (.NET 9+)? 

Or do you have any preference on e.g. `major-tag` vs. `floating-major-tag`?

mthalman (commented):
> @mthalman thoughts on applying this new pattern to other images (.NET 9+)?
> 
> Or do you have any preference on e.g. `major-tag` vs. `floating-major-tag`?

I'm fine with the new tags. I'd just like that pattern applied to the other existing variables so things are consistent.

lbussell (approved):

Issue #5094, opened 2024-01-09
Move .NET 6 Alpine floating tags to Alpine 3.19


lbussell:
This is blocked on https://github.com/PowerShell/PowerShell/issues/20945.

For reference: https://github.com/dotnet/dotnet-docker/discussions/5091

lbussell:
https://github.com/PowerShell/PowerShell/issues/20945 is fixed and the floating tags for .NET 6.0 Alpine now point to Alpine 3.19: https://github.com/dotnet/dotnet-docker/blob/39f6e74e3a44ffd4d19346ed92bd60221a15179e/README.sdk.md?plain=1#L80-L81

PR #5428, opened 2024-05-03
Add PowerShell to .NET 6 Alpine 3.19 image and move floating tag

lbussell:
Part of #5094 

I tested PowerShell interactively and it is working!

TBD if we should move the .NET 6 Alpine 3.19 floating tag in the `main` branch. If not, then we can create a merge commit for this PR and pick only the PowerShell change for May servicing.


mthalman (approved):

lbussell:
Not sure why the merge image infos step is failing.

Issue #5432, opened 2024-05-03
Not all Alpine floating tags point to the correct Alpine versions


lbussell:
There was an issue in https://github.com/dotnet/dotnet-docker/pull/5106 where not all of the floating tags were updated to Alpine 3.19.

For example: https://github.com/dotnet/dotnet-docker/blob/8dd75378bcf016743198be01cfb0c87fbcdc1830/manifest.json#L629-L639

Related:
- https://github.com/dotnet/dotnet-docker/pull/5427
- https://github.com/dotnet/dotnet-docker/pull/5428
- This is also something else #4704 would have helped with.

lbussell:
This affects all .NET 7.0 and 8.0 architecture specific floating tags, e.g. tags matching this regex:

`(7|8)\.0-alpine(-(composite|extra|aot))?-(amd64|arm64v8|arm32v7)` 

I have prepared a draft PR to revert https://github.com/dotnet/dotnet-docker/pull/5426 and help illustrate which tags are affected: https://github.com/dotnet/dotnet-docker/pull/5433

lbussell:
There's an existing issue for tests for this scenario: https://github.com/dotnet/dotnet-docker/issues/2316

lbussell:
[Triage] This is fixed in nightly by https://github.com/dotnet/dotnet-docker/pull/5433 and will roll out with the May servicing release.

lbussell:
This is fixed now and an announcement has been posted:
- https://github.com/dotnet/dotnet-docker/discussions/5465

PR #5433, opened 2024-05-03
Revert Alpine 3.18 removal and move arch-specific floating tags to Alpine 3.19

lbussell:
See #5432


lbussell:
Per discussion with the .NET Docker team, we'll re-add the Alpine 3.18 images and move all floating tags to Alpine 3.19. Alpine 3.18 tags will stay around a little longer to give arch-specific floating tag users time to address any breaking changes in Alpine.

mthalman (approved):

Issue #5020, opened 2023-11-21
Provide Officially-Supported .NET AOT Images


simonkurtz-MSFT:
We just published container optimizations to the [Azure Container Apps .NET Workshop](https://azure.github.io/aca-dotnet-workshop/aca/12-optimize-containers/#31-table-of-improvements). This workshop provides guidance to our customers using ACA & .NET. Since ACA is Linux only, building for a specific platform with AOT is very suitable and would provide great advantages in terms of storage, transfer, start-up (no JIT compiler) and, likely, memory footpring. 

Presently, the AOT images are *nightly* images, which doesn't instill much confidence with customers looking for a stable version. Please consider providing periodic AOT images given all the benefits they provide.

Thank you!

lbussell:
Hi @simonkurtz-MSFT, thanks for the feedback!

If you are looking for a short-term solution for publishing Native AOT apps, I'd encourage you to look at the [releasesapi Debian sample](https://github.com/dotnet/dotnet-docker/blob/26ac86660467e64de0cf7890fd688374a5b9a0c0/samples/releasesapi/Dockerfile.debian).

In that Dockerfile, you can replace the SDK tag with the Jammy SDK and the runtime deps tag with the Jammy Chiseled runtime deps image to get almost all of the size reduction benefits. We cover this in the [.NET sample image size report](https://github.com/dotnet/dotnet-docker/blob/main/documentation/sample-image-size-report.md#self-contained--trimming-deployment). The only difference between the runtime-deps and the runtime-deps AOT image is the presence of the `libstdc++` library, which ends up being about a 1 MB (compressed) image size savings.

lbussell:
@zeroskyx we currently plan to provide some form of officially-supported Native AOT images for .NET 10. They may not be exactly the same as the AOT images that we currently have in nightly, but there will an easy way to build and host Native AOT apps without installing extra packages for most use-cases. We are still working out what that will look like. You can watch this issue for the latest info.

lbussell:
Here's a new proposal for AOT images. 

1. Create AOT SDK images. These images won't support cross-platform AOT compilation (previously, only the Ubuntu image supported cross-compilation, and it came at an image size and complexity cost). 

3. Change nothing about the OS/Distro offerings for the AOT SDK images, but provide an `aot` convenience tag which points to Ubuntu. I am intentionally choosing to leave Debian out of the AOT image offering because we have not seen requests for it.

[...]

PR #5473, opened 2024-05-16
Remove preview designation from Native AOT doc

mthalman:


lbussell:
This is still true though. AOT images are not in main yet because we determined there was too short of a runway from their inception to .NET 8 GA for them to be fully supported. Are you suggesting that we ship them in main now?

Related: https://github.com/dotnet/dotnet-docker/issues/5020

mthalman:
That's what I thought. But then I saw that we have the Dockerfiles in main (e.g. https://github.com/dotnet/dotnet-docker/blob/main/src/runtime-deps/8.0/alpine3.18-aot/amd64/Dockerfile). But I see they are not in the manifest.

lbussell:
> That's what I thought. But then I saw that we have the Dockerfiles in main (e.g. https://github.com/dotnet/dotnet-docker/blob/main/src/runtime-deps/8.0/alpine3.18-aot/amd64/Dockerfile). But I see they are not in the manifest.

That's the result of a bad merge then. I'll get that removed.

PR #5481, opened 2024-05-21
9.0 Preview 4 Updates

mthalman:


tests/performance/ImageSize.nightly.linux.json
```diff
@@ -159,22 +159,20 @@
...
-    "src/runtime/9.0/cbl-mariner2.0-distroless/arm64v8": 98344674,
-    "src/runtime/9.0/cbl-mariner2.0-distroless-extra/amd64": 131038401,
-    "src/runtime/9.0/cbl-mariner2.0-distroless-extra/arm64v8": 134248990
+    "src/runtime/9.0/noble/amd64": 193045485,
+    "src/runtime/9.0/noble/arm64v8": 190484812,
+    "src/runtime/9.0/noble-chiseled/amd64": 86613767,
+    "src/runtime/9.0/noble-chiseled/arm64v8": 90106887,
+    "src/runtime/9.0/noble-chiseled-extra/amd64": 124118557,
+    "src/runtime/9.0/noble-chiseled-extra/arm64v8": 127442477,
+    "src/runtime/9.0/azurelinux3.0/amd64": 182018231,
```
jander-msft, line 170: These should be `cbl-mariner2.0`, correct? If so, please fix for each image repository.

jander-msft (approved):

tests/performance/ImageSize.nightly.linux.json
lbussell: Those are for the nightly branch, so we just accept all incoming changes.

MichaelSimons (approved):

tests/performance/ImageSize.nightly.linux.json
jander-msft: Actually, these might only apply to nightly, so this change would be fine if that's correct.

Issue #5479, opened 2024-05-20
Latest Azure Linux 3.0 drop is getting TLS error during `dotnet restore`


mthalman:
A new drop of Azure Linux 3.0 occurred last Friday which has caused AutoBuilder to attempt to rebuild the corresponding Dockerfiles in the nightly branch. The tests are failing there with this error:

```
#6 [build 6/8] RUN dotnet restore -r linux-x64
 #6 0.987   Determining projects to restore...
 #6 1.356 [ERROR] error:41080106:SCOSSL::passed invalid argument:set_iv_fixed only works with TLS IV length at /usr/src/azl/BUILD/SymCrypt-OpenSSL-1.4.2/ScosslCommon/src/scossl_aes_aead.c, line 305
 #6 2.483 [ERROR] error:41080106:SCOSSL::passed invalid argument:set_iv_fixed only works with TLS IV length at /usr/src/azl/BUILD/SymCrypt-OpenSSL-1.4.2/ScosslCommon/src/scossl_aes_aead.c, line 305
 #6 3.538 [ERROR] error:41080106:SCOSSL::passed invalid argument:set_iv_fixed only works with TLS IV length at /usr/src/azl/BUILD/SymCrypt-OpenSSL-1.4.2/ScosslCommon/src/scossl_aes_aead.c, line 305
 #6 4.633 [ERROR] error:41080106:SCOSSL::passed invalid argument:set_iv_fixed only works with TLS IV length at /usr/src/azl/BUILD/SymCrypt-OpenSSL-1.4.2/ScosslCommon/src/scossl_aes_aead.c, line 305
 #6 5.715 [ERROR] error:41080106:SCOSSL::passed invalid argument:set_iv_fixed only works with TLS IV length at /usr/src/azl/BUILD/SymCrypt-OpenSSL-1.4.2/ScosslCommon/src/scossl_aes_aead.c, line 305
 #6 6.807 [ERROR] error:41080106:SCOSSL::passed invalid argument:set_iv_fixed only works with TLS IV length at /usr/src/azl/BUILD/SymCrypt-OpenSSL-1.4.2/ScosslCommon/src/scossl_aes_aead.c, line 305
 #6 6.811 /source/app/app.csproj : error NU1301: Unable to load the service index for source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json.
[...]

PR #5485, opened 2024-05-21
Workaround symcrypt openssl issue in Azure Linux 3.0 base image

lbussell:
Workaround https://github.com/dotnet/dotnet-docker/issues/5479 by removing the symcrypt package with the bug. I tested by running the following:

Before:
```pwsh
PS> docker run --rm azurelinuxpreview.azurecr.io/public/azurelinux/base/core:3.0 /bin/sh -c "curl https://bing.com"

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0[ERROR] error:41080106:SCOSSL::passed invalid argument:set_iv_fixed only works with TLS IV length at /usr/src/azl/BUILD/SymCrypt-OpenSSL-1.4.2/ScosslCommon/src/scossl_aes_aead.c, line 305
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (35) OpenSSL/3.3.0: error:41080106:SCOSSL::passed invalid argument
```

After:
```pwsh
PS> docker run --rm azurelinuxpreview.azurecr.io/public/azurelinux/base/core:3.0 /bin/sh -c "tdnf remove -y symcrypt-openssl && curl https://bing.com"

Loaded plugin: tdnfrepogpgcheck
<snip>

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0<html><head><title>Object moved</title></head><body>
<h2>Object moved to <a href="https://www.bing.com:443/?toWww=1&amp;redig=2B67BF9265244C288D9C018DF26669DC">here</a>.</h2>
</body></html>
100   193  100   193    0     0   3076      0 --:--:-- --:--:-- --:--:--  3112
```

### DIFF
--- a/eng/dockerfile-templates/Dockerfile.linux.install-deps
+++ b/eng/dockerfile-templates/Dockerfile.linux.install-deps
@@ -10,6 +10,7 @@
     set isDebian to find(OS_ARCH_HYPHENATED, "Debian") >= 0 ^
     set isUbuntu to find(OS_ARCH_HYPHENATED, "Ubuntu") >= 0 ^
     set isAzureLinux to find(OS_VERSION, "cbl-mariner") >= 0 || find(OS_VERSION, "azurelinux") >= 0 ^
+    set isAzureLinux3 to isAzureLinux && find(OS_VERSION, "3.0") >= 0 ^
     set isDistrolessAzureLinux to defined(match(OS_VERSION, "^cbl-mariner\d+\.\d+-distroless$")) || defined(match(OS_VERSION, "^azurelinux\d+\.\d+-distroless$")) ^
     set isFullAzureLinux to isAzureLinux && !isDistrolessAzureLinux ^
     set useLegacyPackageLayout to dotnetVersion = "6.0" ^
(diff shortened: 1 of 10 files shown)


mthalman (approved):

lbussell:
@mthalman should we revert the package version instead? Are we concerned that other functionality will break due to not using the same crypto algorithms? Is OpenSSL just going to use its own/another functionally equivalent implementation?

mthalman:
> @mthalman should we revert the package version instead? Are we concerned that other functionality will break due to not using the same crypto algorithms? Is OpenSSL just going to use its own/another functionally equivalent implementation?

This approach is fine since the Azure Linux folks recommended that as an option and it does seem to work. This is just temporary anyway. If our test scenarios work, then I think that's good enough.

PR #5488, opened 2024-05-22
Update scripts to compensate for installer->sdk migration

mthalman:
Builds of 9.0 SDK now come from dotnet/sdk instead of dotnet/installer. This updates the scripts to compensate for this.


lbussell (approved):

lbussell:
@mthalman does this mean that the changes in https://github.com/dotnet/dotnet-docker/pull/5486 are invalid?

mthalman:
> @mthalman does this mean that the changes in #5486 are invalid?

No. It just means that the update-dependencies pipeline isn't able to successfully resolve _newer_ builds because it's looking for the commit in the installer repo instead of sdk.

PR #5496, opened 2024-05-23
Add BuildId option to Get-DropVersions script

### LINKED ISSUE #5295: Allow Set-DotNetVersions script to read from config.json for .NET releases
### Describe the Problem

Manually looking up the product versions to release and typing them in/copying them to the commandline takes time and is error-prone.

### Describe the Solution

The .NET Release staging pipeline outputs a file called `config.json` that contains the .NET versions in the build. It looks like this:

```
{
  "Channel": "8.0",
  "Release": "8.0.3",
  "Runtime": "8.0.3",
  "Asp": "8.0.3",
  "Sdks": [
    "8.0.202",
    "8.0.103"
  ],
  ...
  "Runtime_Build": "8.0.3-servicing.24114.23",
  "Asp_Build": "8.0.3-servicing.24116.15",
  "Sdk_Builds": [
    "8.0.202-servicing.24116.49",
    "8.0.103-servicing.24116.48"
  ],
  ...
}
```

We should be able to read from this file in order to set the correct .NET versions for a release.

Ideally, we could also connect directly to Azure DevOps. I would like to be able to give Set-DotNetVersions.ps1 a build number and have it update everything automatically.

lbussell:
Fixes #5295

### DIFF
--- a/eng/Get-DropVersions.ps1
+++ b/eng/Get-DropVersions.ps1
@@ -11,7 +11,7 @@ param(
     [string]
     $Channel,
 
-    # The release channel to use for determining the latest .NET build.
+    # The pipeline run ID of the .NET release staging build.
     [Parameter(ParameterSetName = "BuildId")]
     [string]
     $BuildId,
@@ -146,8 +146,8 @@ function GetDependencyVersion([string]$dependencyName, [xml]$versionDetails) {
 }
 
 function GetVersionInfoFromBuildId([string]$buildId) {
-    $configPath = "$tempDir/config.json";
-    $setVersionsScript = "$tempDir/Set-DotnetVersions.ps1"
+    $configPath = Join-Path $tempDir "config.json"
+    $setVersionsScript = Join-Path $tempDir "Set-DotnetVersions.ps1"
 
     try {
         az pipelines runs artifact download --organization https://dev.azure.com/dnceng/ --project internal --run-id $buildId --path $tempDir --artifact-name drop
@@ -158,11 +158,16 @@ function GetVersionInfoFromBuildId([string]$buildId) {
 
         return [PSCustomObject]@{
             DockerfileVersion = $config.Channel
-            SdkVersion = $config.Sdks[0]
+            SdkVersion = ($config.Sdks | Sort-Object -Descending)[0]
             RuntimeVersion = $config.Runtime
             AspnetVersion = $config.Asp
             StableBranding = $isStableVersion
         }
+    }
+    catch [System.Management.Automation.CommandNotFoundException] {
+        Write-Error "Azure CLI is not installed. Please visit https://learn.microsoft.com/cli/azure/install-azure-cli."
+        Write-Host "Original Exception: $_"
+        exit 1
     }
     finally {
         Remove-Item -Force $configPath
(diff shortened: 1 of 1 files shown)


eng/Get-DropVersions.ps1
```diff
@@ -11,6 +11,11 @@ param(
     [string]
     $Channel,
 
+    # The release channel to use for determining the latest .NET build.
```
mthalman, line 14: This comment needs updating.

eng/Get-DropVersions.ps1
```diff
@@ -140,6 +145,30 @@ function GetDependencyVersion([string]$dependencyName, [xml]$versionDetails) {
     return $result.Node.Value
 }
 
+function GetVersionInfoFromBuildId([string]$buildId) {
+    $configPath = "$tempDir/config.json";
+    $setVersionsScript = "$tempDir/Set-DotnetVersions.ps1"
+
+    try {
+        az pipelines runs artifact download --organization https://dev.azure.com/dnceng/ --project internal --run-id $buildId --path $tempDir --artifact-name drop
```
mthalman, line 153: This presumes the Azure CLI is installed. It may not exist when running from a dev machine. It would be nice to provide a better experience here if it's not installed. Otherwise, the user would just get an error saying `The term 'az' is not recognized ...`.

eng/Get-DropVersions.ps1
```diff
@@ -140,6 +145,30 @@ function GetDependencyVersion([string]$dependencyName, [xml]$versionDetails) {
     return $result.Node.Value
 }
 
+function GetVersionInfoFromBuildId([string]$buildId) {
+    $configPath = "$tempDir/config.json";
```
mthalman, line 149: It'd be better to get a random file name to avoid overwriting this file if it already exists: `[System.IO.Path]::GetRandomFileName()`.

eng/Get-DropVersions.ps1
```diff
@@ -140,6 +145,30 @@ function GetDependencyVersion([string]$dependencyName, [xml]$versionDetails) {
...
+    try {
+        az pipelines runs artifact download --organization https://dev.azure.com/dnceng/ --project internal --run-id $buildId --path $tempDir --artifact-name drop
+
+        $config = $(Get-Content -Path $configPath | Out-String) | ConvertFrom-Json
+
+        $isStableVersion = Get-IsStableBranding -Version $config.Sdk_Builds[0]
+
+        return [PSCustomObject]@{
+            DockerfileVersion = $config.Channel
+            SdkVersion = $config.Sdks[0]
```
mthalman, line 161: This presumes the SDKs are sorted in descending order. Might be good to explicitly do that to be sure.

eng/Get-DropVersions.ps1
lbussell: I fixed this by adding a GUID to `$tempDir`. The Azure CLI always downloads files as the artifact name without renaming them, so we can't get away from config.json here.

lbussell:
Ready for review - I added an additional option to allow triggering the update dependencies tooling from Get-DropVersions.ps1. To update all product versions for one .NET major version release, you'd run something like -

```pwsh
.\eng\Get-DropVersions.ps1 -BuildId 2410899 -UpdateDependencies
```

And at the end it will also print a commit message that can be copied (the same message I've been using for months now - like https://github.com/dotnet/dotnet-docker/commit/c79e1a6c43807044f5ad0b0669abb7327c994913).

PR #5538, opened 2024-06-05
Simplify Azure Linux 3.0 Runtime deps build

### LINKED ISSUE #5484: Simplify AzureLinux 3.0 multi-stage build with non-root user
AzureLinux 3.0's version of `tdnf` has the `autoremove` feature now. We can revert most changes from https://github.com/dotnet/dotnet-docker/pull/4652 and just use `autoremove` instead of a multi-stage build.

- Related: https://github.com/dotnet/dotnet-docker/issues/4478

lbussell:
Fixes #5484 

These changes make the build single-stage since tndf is version > 3.4.0 in AzureLinux 3.0. This means we have access to tdnf autoremove, which can remove packages along with dangling dependencies.

This change actually adds about 2.3-2.4 MB to the image (uncompressed). Some additional files are left behind after user creation, unrelated to the package manager. The biggest change is a ~400kb increase in `/var/lib/rpm/rpmdb.sqlite`.

Some of the files, like `/home/app/.*` (dotfiles) are the same but just moved locations.

Here is a screenshot from `dive` showing the diff of the final image layer:

image

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.azlinux
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.azlinux
@@ -9,7 +9,6 @@
     set isDebian to find(OS_ARCH_HYPHENATED, "Debian") >= 0 ^
     set isUbuntu to find(OS_ARCH_HYPHENATED, "Ubuntu") >= 0 ^
     set isAzureLinux to defined(match(OS_VERSION, "^cbl-mariner\d+\.\d+$")) || defined(match(OS_VERSION, "^azurelinux\d+\.\d+$")) ^
-    set isCblMariner2 to isAzureLinux && find(OS_VERSION_NUMBER, "2.0") >= 0 ^
 
     set baseUrl to VARIABLES[cat("base-url|", dotnetVersion, "|", VARIABLES["branch"])] ^
     set isInternal to find(baseUrl, "msrc") >= 0 || find(baseUrl, "internal") >= 0 ^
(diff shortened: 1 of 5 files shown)


eng/dockerfile-templates/runtime-deps/Dockerfile
```diff
@@ -9,6 +9,7 @@
     set isDebian to find(OS_ARCH_HYPHENATED, "Debian") >= 0 ^
     set isUbuntu to find(OS_ARCH_HYPHENATED, "Ubuntu") >= 0 ^
     set isAzureLinux to defined(match(OS_VERSION, "^cbl-mariner\d+\.\d+$")) || defined(match(OS_VERSION, "^azurelinux\d+\.\d+$")) ^
+    set isCblMariner2 to isAzureLinux && find(OS_VERSION_NUMBER, "2.0") >= 0 ^
```
mthalman, line 12: This seems more logical to me:

```suggestion
    set isCblMariner2 to defined(match(OS_VERSION, "^cbl-mariner\d+\.\d+$")) ^
    set isAzureLinux to isCblMariner2 || defined(match(OS_VERSION, "^azurelinux\d+\.\d+$")) ^
```

Checking the Mariner version is unnecessary since there's only one version we provide.

eng/dockerfile-templates/runtime-deps/Dockerfile.azlinux
mthalman, line 1: It looks like this template is unnecessary?

mthalman (approved):

PR #5552, opened 2024-06-11
Merge main to nightly

lbussell:


src/runtime-deps/8.0/alpine3.19/alpine3.19-aot/amd64/Dockerfile
```diff
@@ -1,4 +1,8 @@
+<<<<<<<< HEAD:src/runtime-deps/8.0/alpine3.19/alpine3.19-aot/amd64/Dockerfile
```
mthalman, line 1: There are some merge issues here.

src/runtime-deps/9.0/azurelinux3.0-distroless-extra/amd64/Dockerfile
mthalman, line 1: These 9.0 runtime-deps Dockerfiles shouldn't be here, right? They should be shared with 8.0.

src/sdk/6.0/alpine3.18/arm32v7/Dockerfile
mthalman, line 1: Alpine 3.18 should be deleted.

tests/Microsoft.DotNet.Docker.Tests/TestData.cs
```diff
@@ -159,6 +159,7 @@ public static class TestData
             new ProductImageData { Version = V8_0, OS = OS.Alpine320,           Arch = Arch.Arm64,   SdkOS = OS.Alpine320, SdkImageVariant = DotNetImageVariant.AOT,
                     ImageVariant = DotNetImageVariant.AOT, SupportedImageRepos = DotNetImageRepo.Runtime_Deps },
 
+
```
mthalman, line 162: Extra lines in this file are unnecessary

src/sdk/8.0/alpine3.19-aot/arm32v7/Dockerfile
mthalman, line 1: Why is this just showing up in nightly now?

lbussell: We have never produced Arm32 AOT images for any distro. I think I may have accidentally copied this file over when resolving a "Dockerfile does not exist" issue with the templates. That's why the content has nothing to do with AOT, and it wasn't regenerated so that's why the merge markers are still there. I'll delete it.

mthalman (approved):

Issue #5558, opened 2024-06-12
ExecutionPolicy breaking change in 6.0.423


perosb:
### Describe the Bug

A breaking change was discovered when trying to run our images using the .423 version of sdk:6.0
It seem the ExecutionPolicy has changed between 422 and 423.
Possible comes with the change of pwsh v7.2.20?

```
Status: Downloaded newer image for mcr.microsoft.com/dotnet/sdk:6.0.423
PowerShell 7.2.20

PS C:\> Get-ExecutionPolicy -List

        Scope ExecutionPolicy
        ----- ---------------
MachinePolicy       Undefined
   UserPolicy       Undefined
      Process       Undefined
  CurrentUser       Undefined
 LocalMachine       Undefined
```

### Other Information

```
Status: Downloaded newer image for mcr.microsoft.com/dotnet/sdk:6.0.422
PowerShell 7.2.19

PS C:\> Get-ExecutionPolicy -List

        Scope ExecutionPolicy
        ----- ---------------
MachinePolicy       Undefined
   UserPolicy       Undefined
      Process       Undefined
  CurrentUser       Undefined
 LocalMachine    RemoteSigned
```

Looking at latest 8.0:
```
Status: Downloaded newer image for mcr.microsoft.com/dotnet/sdk:8.0.302-nanoserver-ltsc2022
PowerShell 7.4.2
PS C:\> Get-ExecutionPolicy -List

        Scope ExecutionPolicy
        ----- ---------------
MachinePolicy       Undefined
   UserPolicy       Undefined
      Process       Undefined
  CurrentUser       Undefined
 LocalMachine    RemoteSigned
```

Latest 6.0-alpine:
```
11:17 $ docker run -ti --entrypoint pwsh mcr.microsoft.com/dotnet/sdk:6.0-alpine
PowerShell 7.2.20

PS /> Get-ExecutionPolicy -List

        Scope ExecutionPolicy
        ----- ---------------
MachinePolicy    Unrestricted
   UserPolicy    Unrestricted
      Process    Unrestricted
  CurrentUser    Unrestricted
 LocalMachine    Unrestricted
```

### Output of `docker info`

```
Client:
 Version:    26.1.4
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.3
    Path:     C:\Users\Administrator\.docker\cli-plugins\docker-buildx.exe
[...]

PR #5579, opened 2024-06-19
Update .NET 6 and .NET 8 sdk images for PowerShell hotfix

mthalman:
Fixes https://github.com/dotnet/dotnet-docker/issues/5558


manifest.json
```diff
@@ -6725,7 +6725,7 @@
         {
           "productVersion": "$(sdk|6.0|product-version)",
           "sharedTags": {
-            "$(sdk|6.0|product-version)": {},
+            "$(sdk|6.0|product-version)-1": {},
```
MichaelSimons, line 6728: This is a follow-up comment.  Would it make sense to create an encapsulation mechanism for the tag version in our manifest.  It would default to $(sdk|6.0|product-version) but would allow us to easily add the dash suffix version for cases like this.  It would allow this to be easily done with confidence.

MichaelSimons (approved):

Issue #5327, opened 2024-04-02
Image tags on DockerHub are out of sync


lbussell:
See the Aspire Dashboard (Preview) page: https://hub.docker.com/_/microsoft-dotnet-nightly-aspire-dashboard

<img width="775" alt="image" src="https://github.com/dotnet/dotnet-docker/assets/36081148/47970d9b-172a-43ce-9ec9-fc14527feaa8">

The "Last Modified" column is out of date as well as the Dockerfile link - https://github.com/dotnet/dotnet-docker/blob/598ce3d47883dcaae202041ee898ebf4874267cc/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile

The latest build was actually only a few hours ago on 2024-04-02 - https://github.com/dotnet/versions/blob/cab82d6d3f56be045618a092121f2f8fc914e5c8/build-info/docker/image-info.dotnet-dotnet-docker-nightly.json#L4-L72

lbussell:
MAR also has an incorrect last modified time - 
<img width="976" alt="image" src="https://github.com/dotnet/dotnet-docker/assets/36081148/944dca22-257c-4998-8959-96562d6c4c5c">

The digests also don't match except for the topmost one.

lbussell:
MCR issue is mitigated. Opening another ticket for the DH issue.

lbussell:
[Triage] We now know that this extends to all of our image repos and none of the README content is being updated (not just the "Last Modified" info). The MAR team is working with Docker to get this resolved.

William-Froelich:
Is there a recommended work-around for this in the meanwhile? I want to be able to build an ASP.Net based container without all the vulnerabilities.

lbussell:
> Is there a recommended work-around for this in the meanwhile? I want to be able to build an ASP.Net based container without all the vulnerabilities.

@William-Froelich, the images are available - they are not actually hosted on DockerHub, but rather on Microsoft Asset Registry. Our DockerHub repo just contains a list of tags for images on MAR. This issue is about that list of tags not being replicated from MAR to DockerHub properly.

For now, you can see the latest supported tags either at the [MAR portal](https://mcr.microsoft.com/product/dotnet/aspnet/tags) or [here in this repo](https://github.com/dotnet/dotnet-docker/blob/main/README.aspnet.md).

William-Froelich:
[...]

PR #5585, opened 2024-06-20
Define new READMEs for Docker Hub

mthalman:
In order to address the character limit issue with #5327, these changes define a separate set of READMEs that are specific to Docker Hub. These READMEs are configured to have the tags tables removed and instead include links to the MAR portal and this repo.

Contributes to #5327


.portal-docs/README.aspire-dashboard.md
```diff
@@ -0,0 +1,267 @@
...
+* [dotnet/nightly/runtime](https://hub.docker.com/_/microsoft-dotnet-nightly-runtime/): .NET Runtime (Preview)
+* [dotnet/nightly/runtime-deps](https://hub.docker.com/_/microsoft-dotnet-nightly-runtime-deps/): .NET Runtime Dependencies (Preview)
+* [dotnet/nightly/monitor](https://hub.docker.com/_/microsoft-dotnet-nightly-monitor/): .NET Monitor Tool (Preview)
+
+.NET Framework:
+
+* [dotnet/framework](https://hub.docker.com/_/microsoft-dotnet-framework/): .NET Framework, ASP.NET and WCF
+* [dotnet/framework/samples](https://hub.docker.com/_/microsoft-dotnet-framework-samples/): .NET Framework, ASP.NET and WCF Samples
+
+# Full Tag Listing
```
MichaelSimons, line 111: From the PR description, I thought this was being removed?

MichaelSimons (commented):
Is there a way we can make the portal-docs apparent what portal (mcr/dh) they are for?  Either sub-folder or a filename prefix/suffix?

.portal-docs/README.aspire-dashboard.md
MichaelSimons: > The table is being removed.

But the table is still present.  I was under the impression it is being removed with this PR.

MichaelSimons: see https://github.com/dotnet/dotnet-docker/pull/5585/files#diff-eebee3282380281b727711c740ab97298dc872b487bd13e547490bd78e5d2898R113

mthalman: Oh, what the heck. There's another `Full Tag Listing` header further below that is configured correctly: https://github.com/dotnet/dotnet-docker/pull/5585/files#diff-eebee3282380281b727711c740ab97298dc872b487bd13e547490bd78e5d2898R233. I need to figure out what's going on.

mthalman:
> Is there a way we can make the portal-docs apparent what portal (mcr/dh) they are for? Either sub-folder or a filename prefix/suffix?

The filenames need to be exactly what they are. That's what MAR requires. Yes, technically they could be in separate directories but that would require a different implementation than what was provided by https://github.com/dotnet/docker-tools/pull/1345.

MichaelSimons (approved):
Approving to unblock the readme issue in DH.  I am not a fan of the readme file layout.  It is difficult for outsiders/new contributors to understand the differences in the two types of readmes.  I think this is something that should be fixed in a follow-up.

.portal-docs/README.runtime.md
```diff
@@ -0,0 +1,100 @@
...
+* [dotnet/nightly/aspire-dashboard](https://hub.docker.com/_/microsoft-dotnet-nightly-aspire-dashboard/): .NET Aspire Dashboard (Preview)
+
+.NET Framework:
+
+* [dotnet/framework](https://hub.docker.com/_/microsoft-dotnet-framework/): .NET Framework, ASP.NET and WCF
+* [dotnet/framework/samples](https://hub.docker.com/_/microsoft-dotnet-framework-samples/): .NET Framework, ASP.NET and WCF Samples
+
+# Full Tag Listing
+
+View the current tags at the [Microsoft Artifact Registry portal](https://mcr.microsoft.com/product/$dotnet/nightly/runtime/tags) or on [Docker Hub](https://github.com/dotnet/dotnet-docker/blob/nightly).
```
MichaelSimons, line 66: Shouldn't this say GitHub?
```suggestion
View the current tags at the [Microsoft Artifact Registry portal](https://mcr.microsoft.com/product/$dotnet/nightly/runtime/tags) or on [GitHub](https://github.com/dotnet/dotnet-docker/blob/nightly).
```

MichaelSimons: Also, Can we should deep link to the appropriate readme's Full Tag Listing section.

MichaelSimons: One more thing - the MAR link appears broken because of the `$dotnet`.

PR #5623, opened 2024-06-26
Compare VerifyDotnetFolderContents using Git Diff

ellahathaway:
Closes https://github.com/dotnet/dotnet-docker/issues/4843

This PR primarily focuses on changing the `VerifyDotnetFolderContents` test to compare the folder contents using `git diff`. These changes involve the introduction of new helper methods for file comparison, the creation of temporary files, and the simplification of the test assertions.

- FileHelper.cs:
  * Removed unused imports and added new ones for `System.Diagnostics` and `Xunit.Abstractions`.
  * Replaced the `CompareLists` method with a new `CompareFiles` method that compares two files using `git diff` via the `DiffFiles` method. Also, added a new `UseTempFile` method that creates a temporary file.
  * Introduced a new `TempFileContext` class that creates a temporary file in its constructor and deletes it in its `Dispose` method.

- SdkImageTest.cs
  * Removed an unused import for `System.Data.Common`.
  * Simplified the `VerifyDotnetFolderContents` test method by using the new `FileHelper` methods. The test now writes the expected and actual contents to temporary files and compares them using `FileHelper.CompareFiles`. The assertion has been simplified to check whether the files contained any differences.

[Example run of the tests](https://dev.azure.com/dnceng/internal/_build/results?buildId=2483215&view=results) (internal Microsoft link)

Note - I'd like to do a test run that produces failing test results, but I'm unsure of how to do this. If anyone has information on how to do this, please let me know.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
@@ -4,7 +4,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Data.Common;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


mthalman:
> Note - I'd like to do a test run that produces failing test results, but I'm unsure of how to do this. If anyone has information on how to do this, please let me know.

You could change the SDK Dockerfile so that it modifies the contents of the extracted SDK tarball. That would then produce diffs when compared to the official SDK tarball. That would be done after this part of the Dockerfile: https://github.com/dotnet/dotnet-docker/blob/49969b2509003bab5c43cd953dd60dc574ab51df/src/sdk/6.0/alpine3.19/amd64/Dockerfile#L33

You could add new files, for example.

If you directly modify the Dockerfile, you'll get a pre-build validation error because it will be out of sync with the templates. To disable that check, comment this line out: https://github.com/dotnet/dotnet-docker/blob/49969b2509003bab5c43cd953dd60dc574ab51df/tests/Microsoft.DotNet.Docker.Tests/GeneratedArtifactTests.cs#L23

tests/Microsoft.DotNet.Docker.Tests/SdkImageTests.cs
```diff
@@ -178,41 +177,15 @@ public async Task VerifyDotnetFolderContents(ProductImageData imageData)
             IEnumerable<SdkContentFileInfo> actualDotnetFiles = GetActualSdkContents(imageData);
             IEnumerable<SdkContentFileInfo> expectedDotnetFiles = await GetExpectedSdkContentsAsync(imageData);
 
-            bool hasCountDifference = expectedDotnetFiles.Count() != actualDotnetFiles.Count();
+            using TempFileContext actualFiles = FileHelper.UseTempFile();
+            using TempFileContext expectedFiles = FileHelper.UseTempFile();
```
mthalman, line 181: Renaming these variables would provide better clarity.

```suggestion
            using TempFileContext actualFilesContext = FileHelper.UseTempFile();
            using TempFileContext expectedFilesContext = FileHelper.UseTempFile();
```

mthalman (approved):

ellahathaway:
The output upon failure was printing all of the files as part of the `docker run` command. To reduce the amount of output, I made some changes as part of https://github.com/dotnet/dotnet-docker/pull/5623/commits/8d7554f3c4fbd6fe909af655196d64b56fe93eac so I'm re-requesting approval.

[This run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2484093&view=results) shows what the output looks like on failure now... it's just the diff and the assertion message.

mthalman (approved):

PR #5650, opened 2024-07-03
Add documentation for Azure Linux

### LINKED ISSUE #5031: Migrate internal CBL Mariner documentation to this repo
Because [Azure Linux (CBL Mariner) .NET container images are now publicly supported](https://github.com/dotnet/dotnet-docker/discussions/5022), we should migrate as much information from the internal-only Mariner documentation to this repo as possible.

lbussell:
Fixes #5031


documentation/azure-linux-distroless.md
mthalman, line 1: There's a mix of "Azure Linux" and "Mariner" terms used throughout the doc. If possible, these should be consolidated to just one name.

documentation/azure-linux-distroless.md
```diff
@@ -0,0 +1,132 @@
+# Azure Linux Distroless Support
+
+Distroless containers contain only the minimal set of packages your application needs, with anything extra removed (i.e. package manager, libraries and shell).
+Both Qualys and Trivy support vulnerability scanning in Mariner distroless containers.
```
mthalman, line 4: This sentence should be moved lower, probably at the end of the 2nd paragraph.

documentation/azure-linux-distroless.md
```diff
@@ -0,0 +1,132 @@
...
+You can use the following image tags (SDK is not available for distroless):
+
+- `mcr.microsoft.com/dotnet/<image-type>:6.0-cbl-mariner2.0-distroless`
+- `mcr.microsoft.com/dotnet/<image-type>:8.0-cbl-mariner2.0-distroless`
+
+## Installing Additional Packages
+
+If you require additional packages to be installed in your Azure Linux Distroless image, you can follow the same pattern that .NET uses to install the .NET runtime dependencies.
+
+```Dockerfile
```
mthalman, line 20: There should be another example for Azure Linux 3.0 since that will have slightly different content (e.g. tag names and `releasever`).

documentation/azure-linux-distroless.md
```diff
@@ -0,0 +1,132 @@
...
+    && find /staging2/var/log -type f -size +0 -delete \
+    && find /staging2 -type d -empty -delete
+
+
+FROM base
+
+COPY --from=installer /staging2/ /
+```
+
+## Debugging with Container Fast Mode
```
mthalman, line 73: This section isn't specific to distroless.

documentation/image-artifact-details.md
```diff
@@ -82,3 +83,44 @@ $ docker run --rm mcr.microsoft.com/dotnet/sdk:6.0-bullseye-slim /bin/sh -c "fin
 ./usr/share/dotnet/sdk/6.0.100/Sdks/Microsoft.NET.Sdk.WindowsDesktop/THIRD-PARTY-NOTICES.TXT
 ./usr/share/powershell/.store/powershell.linux.x64/7.2.0-preview.10/powershell.linux.x64/7.2.0-preview.10/tools/net6.0/any/ThirdPartyNotices.txt
 ```
+
+## Distroless Images
```
mthalman, line 87: The content in this section should be generalized so that it can be used for chiseled images as well.

documentation/azure-linux-distroless.md
lbussell: Then it seems to me like we should make an additional sample under `samples/aspnetapp` for container fast mode debugging. Thoughts? @richlander?

samples/container-fast-mode.md
```diff
@@ -0,0 +1,64 @@
+# Debugging with Container Fast Mode
+
+In order to reduce image size and attack surface, .NET Runtime images do not contain the tools required to debug .NET apps using Visual Studio.
```
mthalman, line 3: This makes it sound like it affects all runtime-based images. I don't think that's the case. It affects Mariner, I know.

documentation/azure-linux-distroless.md
```diff
@@ -1,21 +1,37 @@
...
 You can use the following image tags (SDK is not available for distroless):
 
-- `mcr.microsoft.com/dotnet/<image-type>:6.0-cbl-mariner2.0-distroless`
-- `mcr.microsoft.com/dotnet/<image-type>:8.0-cbl-mariner2.0-distroless`
+- `6.0-cbl-mariner2.0-distroless`
+- `8.0-cbl-mariner2.0-distroless`
+
+## Vulnerability Scanning
+
+Azure Linux Distroless images maintain a detailed list of packages installed in the `/var/lib/rpmmanifest/container-manifest-1` and `/var/lib/rpmmanifest/container-manifest-2` files, which is supported by major image scanners like Qualys, Trivy, and Syft.
```
mthalman, line 20: I don't think we need to mention the manifest files. That's an unnecessary implementation detail for the purposes of this doc. I don't want to have to maintain that if the implementation changes.

documentation/distroless.md
```diff
@@ -0,0 +1,151 @@
...
+We’re not offering distroless SDK images at the moment as there hasn't been a strong need for one, and because they may be difficult to use to use for some scenarios due to the lack of a shell.
+You can use the existing SDK images to build your apps to run on distroless base images.
+If you have a compelling use case for a distroless SDK image, please leave a comment on [this issue](https://github.com/dotnet/dotnet-docker/issues/4942) and we’ll be happy to reconsider.
+
+## FAQs
+
+### How do I use globalization with distroless images?
+
+Our distroless images are focused on size. That means the default distroless images do not include the `icu` or `tzdata`
+libraries from Ubuntu. However, we offer an `extra` image variant that includes `tzdata` and `icu` by default.
```
mthalman, line 28:
```suggestion
libraries. However, we offer an `extra` image variant that includes `tzdata` and `icu` by default.
```

mthalman (approved):

PR #5669, opened 2024-07-09
Synchronize Dockerfile templates with nightly branch

lbussell:
This PR resets the templates in the main branch to the state that they're currently in in the nightly branch.

The Dockerfile templates are carefully architected to prevent the addition of new features from affecting existing Dockerfiles. As such, including new templates (like AOT) in the main branch does not affect the Dockerfiles in the main branch. To simplify merges from nightly to main and vice versa, the templates should stay the same between the two branches.

I'm submitting this PR now so that we can sync up the templates before there are any additional changes added to nightly for the August release.

### DIFF
--- a/eng/dockerfile-templates/Dockerfile.linux.install-deps
+++ b/eng/dockerfile-templates/Dockerfile.linux.install-deps
@@ -1,5 +1,6 @@
 {{
     _ ARGS:
+        is-aot (optional): Whether to use the smallest set of packages that is necessary to run native AOT apps
         is-extra (optional): Whether to include extra packages that are not needed for every app,
             for example globalization support
         staging-dir (optional): Location of the staging directory for distroless installation ^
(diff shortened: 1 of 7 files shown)


eng/dockerfile-templates/sdk/Dockerfile.ubuntu.aot
mthalman, line 1: This looks unused unless I'm missing something. I see the same file exists in nightly.

mthalman: Ah, I see that's addressed here: https://github.com/dotnet/dotnet-docker/pull/5670. Are you planning to merge that first and then just exclude it from this PR?

lbussell: I could go either way. If you think it's better to remove now then we can do that. If I don't merge it before this sync then it'd get removed in next month's servicing.

mthalman: I'll leave it up to you since it impacts the merging workflow. Either way is fine.

mthalman (approved):

eng/dockerfile-templates/sdk/Dockerfile.ubuntu.aot
lbussell: I'll leave it here and then pull #5670 in for servicing.

PR #5671, opened 2024-07-09
Synchronize samples with main branch

### LINKED ISSUE #5659: Clean up diffs in samples between main and nightly branches
There are a variety of minor diffs in the samples directory between main and nightly. These should be resolved so they're in sync.

lbussell:
This PR resets the state of the samples to what's currently in the `main` branch. Fixes #5659


samples/releasesapi/Dockerfile.ubuntu-chiseled
```diff
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.6
 # Learn about building .NET container images:
 # https://github.com/dotnet/dotnet-docker/blob/main/samples/README.md
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/nightly/sdk:8.0-jammy-aot AS build
```
mthalman, line 4: This change was specifically made here: https://github.com/dotnet/dotnet-docker/commit/5ffbeeffb66b08f13a9a00369a5c282e05863442. Was that not needed?

lbussell: OK, I had naively just reset the nightly branch to the state of the main branch assuming that it was the most up-to-date since that's where we publish samples from. I do think we need to keep this change in nightly and port it to main. I'll fix it up here and submit a follow-up PR to main.

mthalman: What follow up is needed in main?

mthalman (approved):

samples/releasesapi/Dockerfile.ubuntu-chiseled
lbussell: The `--platform` option we're discussing here is missing from main. We should add it: https://github.com/dotnet/dotnet-docker/blob/042f01ff1fafed068b34927e5d36133dc2b644ce/samples/releasesapi/Dockerfile.ubuntu-chiseled#L4

PR #5672, opened 2024-07-09
Update Aspire Dashboard to 8.0.2

lbussell:
Port of #5651 to main.


joperezr:
Thanks for doing this @lbussell ! I was OOF for the last two weeks which is why I didn't take care of it 😃

PR #5677, opened 2024-07-10
Use UID environment variable for non-root user in Azure Linux 3.0 images

### LINKED ISSUE #5649: Azure Linux 3.0 and CBL Mariner 2.0 distroless images have hardcoded USER declaration with name instead of UID
### Azure Linux Distroless images:

https://github.com/dotnet/dotnet-docker/blob/163ebc7ee3a1a2b52aeb0c79ea095a523e6084d3/src/runtime-deps/8.0/cbl-mariner2.0-distroless/amd64/Dockerfile#L73

https://github.com/dotnet/dotnet-docker/blob/163ebc7ee3a1a2b52aeb0c79ea095a523e6084d3/src/runtime-deps/8.0/azurelinux3.0-distroless/amd64/Dockerfile#L73

### Ubuntu Chiseled images:

https://github.com/dotnet/dotnet-docker/blob/163ebc7ee3a1a2b52aeb0c79ea095a523e6084d3/src/runtime-deps/8.0/noble-chiseled/amd64/Dockerfile#L56

https://github.com/dotnet/dotnet-docker/blob/163ebc7ee3a1a2b52aeb0c79ea095a523e6084d3/src/runtime-deps/8.0/jammy-chiseled/amd64/Dockerfile#L56

These should all be the same, all using $APP_UID in order to behave well with K8s.

lbussell:
Fixes #5649

### DIFF
--- a/eng/dockerfile-templates/runtime-deps/Dockerfile.distroless-mariner
+++ b/eng/dockerfile-templates/runtime-deps/Dockerfile.distroless-mariner
@@ -4,11 +4,13 @@
         is-extra (optional): Whether to include extra packages that are not needed for every app,
             for example globalization support ^
 
+    set isAzureLinux3 to find(OS_VERSION_NUMBER, "3.0") >= 0 ^ 
     set distrolessStagingDir to "/staging" ^
-    set marinerRepo to when(find(OS_VERSION_NUMBER, "3.0") >= 0, "azurelinuxpreview.azurecr.io/public/azurelinux", "mcr.microsoft.com/cbl-mariner") ^
+    set marinerRepo to when(isAzureLinux3, "azurelinuxpreview.azurecr.io/public/azurelinux", "mcr.microsoft.com/cbl-mariner") ^
     set dotnetVersion to join(slice(split(PRODUCT_VERSION, "."), 0, 2), ".") ^
     set baseImage to cat(marinerRepo, "/distroless/minimal:", OS_VERSION_NUMBER) ^
     set username to "app" ^
+    set nonRootUser to when(isAzureLinux3, "$APP_UID", username) ^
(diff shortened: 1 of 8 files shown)


mthalman (commented):
Looks like the tests need to be updated for this too.

MichaelSimons:
I noticed main has 9.0 variants but nightly doesn't. e.g. https://github.com/dotnet/dotnet-docker/blob/main/src/runtime-deps/9.0/azurelinux3.0-distroless/amd64/Dockerfile.  Why is that?

lbussell:
@MichaelSimons, there are no .NET 8 images based on Azure Linux 3.0 in `main`. Only .NET 9 Preview images. When Azure Linux 3.0 is released, then we can move the runtime-deps Dockerfiles from the 9.0 to 8.0 folder because we will add the 8.0 images to main, and they can share the runtime-deps Dockerfiles like they do in `nightly` today. But for now there's no reason to have Azure Linux 3.0 Dockerfiles under the 8.0 folder in `main`.

MichaelSimons:
> @MichaelSimons, there are no .NET 8 images based on Azure Linux 3.0 in `main`. Only .NET 9 Preview images. When Azure Linux 3.0 is released, then we can move the runtime-deps Dockerfiles from the 9.0 to 8.0 folder because we will add the 8.0 images to main, and they can share the runtime-deps Dockerfiles like they do in `nightly` today. But for now there's no reason to have Azure Linux 3.0 Dockerfiles under the 8.0 folder in `main`.

Thanks for the explanation.  So when this change flows to main, those images will get updated because of the template changes.

lbussell:
> So when this change flows to main, those images will get updated because of the template changes.

Correct, they use the same template and the Dockerfiles are identical anyways besides the tag.

mthalman (approved):

PR #5705, opened 2024-07-17
Fix improper FROM-AS casing in Dockerfiles to eliminate build warnings

lbussell:
All of these Dockerfiles gave the following warning during builds: https://docs.docker.com/reference/build-checks/from-as-casing/

The `FROM` and `AS` keywords are supposed to match case in Dockerfiles.


mthalman (approved):

lbussell:
Superseded by https://github.com/dotnet/dotnet-docker/pull/5706

PR #5706, opened 2024-07-17
Add openssl to Noble Chiseled expected packages list

lbussell:
Our tests are failing because the Noble ca-certificates slice added a dependency on a file from the openssl slice. See https://github.com/canonical/chisel-releases/pull/266 for more context.

PR #5713, opened 2024-07-18
Use Chisel binary instead of building from source

### LINKED ISSUE #5675: Install Chisel tool from binaries instead of compiling from source
Chisel is currently being installed by compiling it to produce the binaries:

https://github.com/dotnet/dotnet-docker/blob/dd53688fef338a92c9860840788a9e4ebbc78f9a/src/runtime-deps/8.0/jammy-chiseled/amd64/Dockerfile#L8

Instead, we should just download the install tarball directly from the release page (e.g. [0.10.0](https://github.com/canonical/chisel/releases/tag/v0.10.0)).

This would prevent issues like https://github.com/dotnet/dotnet-docker/pull/5663. It also would make https://github.com/dotnet/dotnet-docker/issues/5666 irrelevant.

lbussell:
Fixes #5675


eng/dockerfile-templates/runtime-deps/Dockerfile.chiseled-ubuntu
```diff
@@ -53,14 +56,10 @@
...
-}}FROM {{ARCH_VERSIONED}}/golang:{{golangVersion}} AS chisel
-
-RUN apt-get update \
-    && apt-get install -y \
-        ca-certificates \
-        file
+}}FROM {{installerImageTag}} AS chisel
 
-RUN go install {{chiselUrl}}/cmd/chisel@@{{chiselRef}} \
+RUN wget -O chisel.tar.gz https://{{chiselUrl}}/releases/download/{{chiselRef}}/chisel_{{chiselRef}}_linux_{{chiselArch}}.tar.gz \
```
mthalman, line 61: Instead of hardcoding the URL, it should be defined in manifest.versions.json. This is already done today for MinGit, for example. Then you can also do checksum validation by having the SHA value stored there too.

lbussell:
I manually imported the missing image.

eng/update-dependencies/GitHubHelper.cs
```diff
@@ -36,11 +37,29 @@ public static async Task<JObject> GetLatestReleaseAsync(string owner, string rep
         return response;
     }
 
+    public static Task<Release> GetLatestReleaseWithOctokitAsync(string owner, string repo)
```
mthalman, line 40: If we're going to use OctoKit, it should be used consistently. That would mean getting rid of the `GetLatestReleaseAsync` method. And in that case, this method should just be renamed to `GetLatestReleaseAsync`.

lbussell: Yes, makes sense. I had thought about updating all of the MinGit code to not use raw JObjects in this PR as well but didn't want to get too far into the weeds. I can adapt it now that I've got Chisel figured out though.

lbussell:
Similar to https://github.com/dotnet/dotnet-docker/pull/5706, it looks like the Chisel slices have been updated for Jammy to include copyright information. We'll need to change the package tests here.

PR #5719, opened 2024-07-22
Fix errors in update-dependencies tool, always try to update both latest and specific Chisel versions

lbussell:
Fixes some errors that were left in https://github.com/dotnet/dotnet-docker/pull/5713

1. update-dependencies tool failed to build.
2. update-dependencies tool would never update version-specific Chisel versions, only the `latest` version. Fix this by running an updater for both the latest and current Chisel version variables.


eng/update-dependencies/UpdateDependencies.cs
```diff
@@ -425,23 +425,23 @@ private static async Task<IEnumerable<IDependencyUpdater>> GetUpdatersAsync()
...
+            //         updaters.Add(shaUpdater);
+            //     }
+            // }
+
+            // updaters =
+            // [
+            //     ..updaters,
+            //     ScriptRunnerUpdater.GetDockerfileUpdater(RepoRoot),
+            //     ScriptRunnerUpdater.GetReadMeUpdater(RepoRoot)
+            // ];
```
mthalman, line 444: Revert

mthalman (approved):

PR #5720, opened 2024-07-22
Update Aspire to 8.1

joperezr:
cc: @lbussell 

This version will be published tomorrow.


lbussell:
@joperezr you'll need to merge in the latest changes from the nightly branch and potentially regenerate the Dockerfiles/readmes again.

lbussell (approved):

manifest.versions.json
```diff
@@ -1,11 +1,12 @@
...
-    "aspire-dashboard|8.0|minor-tag": "8.0",
-    "aspire-dashboard|8.0|linux|x64|sha": "c6e0c976af3a1d203588f46f63161b9a6202ecfbf3e3c76a88601369757164085c5e8dd7a7f1adc78b493e6fb92c2cf9766f523cbe2bebc162f3f5b013111640",
-    "aspire-dashboard|8.0|linux|arm64|sha": "857a0a6c4b1ac3d96f19735e9c25b8ec18bc979b1c2df4919f4c6b2b40918618f902ef6a2be1368c35941dbc8ebc2f012def74eec0a8be4b30d76a28cc50bb45",
+    "aspire-dashboard|8|build-version": "8.1.0",
+    "aspire-dashboard|8|product-version": "8.1.0",
+    "aspire-dashboard|8|fixed-tag": "$(aspire-dashboard|8|product-version)",
+    "aspire-dashboard|8|minor-tag": "8.1",
+    "aspire-dashboard|8|major-tag": "8",
+    "aspire-dashboard|8|linux|x64|sha": "d823c3177dc1f1acc87cff7f252b90f9bc9b215522749b61e540c747801f697ba60a0defd94dccad7b144fa165fd736c4c6ed97f9c5a123170bb46200e122831",
+    "aspire-dashboard|8|linux|arm64|sha": "96e78fc6ab0da4fce82f01f725633820db36fd04fb1f20ab923afe68bb0f155e3d547b4102ad8f95cfbc0fb2d3d7b9103a68f439801aedac6a5973cb34272699",
```
mthalman, line 9: So this variable name change from `8.0` to `8` will require that `8` be passed to the `ProductVersion` parameter of [Set-DotnetVersions.ps1](https://github.com/dotnet/dotnet-docker/blob/main/eng/Set-DotnetVersions.ps1), instead of `8.0`. Is that right?

README.aspire-dashboard.md
```diff
@@ -114,12 +114,12 @@ Limits are per-resource. For example, a `MaxLogCount` value of 10,000 configures
 ## Linux amd64 Tags
 Tags | Dockerfile | OS Version
 -----------| -------------| -------------
-8.0.2-amd64, 8.0-amd64, 8.0.2, 8.0, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile) | CBL-Mariner 2.0
+8.1.0-amd64, 8.1-amd64, 8-amd64, 8.1.0, 8.1, 8, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile) | CBL-Mariner 2.0
```
MichaelSimons, line 117: Should there be an 8-<arch> tag added as well?

lbussell: @MichaelSimons what specific tag is missing? We have all permutations of [major-minor-patch, major-minor, major-only] * [amd64, arm64v8, no-arch].

MichaelSimons: oops - the MD rendering didn't like my format.

That being said, I see the tag exists I was asking about.  No is action needed.

manifest.versions.json
lbussell: I just checked and made sure that this works with some additional changes. I thought about allowing any variant of `8.X` for updating aspire-dashboard but I figured it would make more sense to use exactly what's in `manifest.versions.json`.

eng/update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -49,6 +49,13 @@ public class DockerfileShaUpdater : FileRegexUpdater
             string productName, string dockerfileVersion, string? buildVersion, string arch, string os, string versions, Options options)
         {
             _productName = productName;
+
+            // Allow major-version only strings
+            if (dockerfileVersion.Split('.').Length == 1)
+            {
+                dockerfileVersion += ".0";
+            }
```
mthalman, line 57: Instead of doing this, can we just pass `8.1` to the `ProductVersion` parameter instead? It really represents, the Dockerfile version, so we should just pass that instead. That way we don't need this change. I don't think that has any other impact if we pass in `8.1`, right?

lbussell: >  I don't think that has any other impact if we pass in 8.1, right?
@mthalman The issue is that when you pass 8.1 to update-dependencies, it's going to look for variables like "aspire-dashboard|8.1|build-version". Whereas we have `aspire-dashboard|8|build-version` in `manifest.versions.json`. If we wanted to update from say, 8.1 to 8.2, then we couldn't do that while using `8.1` in the manifest versions file. We would need a special case somewhere else to look for a major-only version variable in `manifest.versions.json`.

jander-msft: Why not use `aspire-dashboard|8.1|` variables instead of `aspire-dashboard|8|` variables? That way it works with the existing updater without having to add more specializations into it.

lbussell: @jander-msft that would break updating between minor versions, right? We don't have a way to set the left-hand-side of variables in `manifest.versions.json`. So updating from 8.0 to 8.1 as this PR did wouldn't have worked.

jander-msft: Yes, it would require updating the variable name when transitioning between minor versions. That's a one-time change that occurs when introducing a new minor version, regardless of it being an addition or full replacement.

We already do this for .NET Monitor when adding new minor versions and it hasn't been a problem from my perspective. My proposal here would be maintain status-quo and don't add more specializations for situations that already have solutions that aren't costly or painful.

lbussell: I could get behind that. @mthalman any concerns with that approach? Keeping the variables in manifest.versions.json as major-minor versions. 

That could get in the way of any automation we may do in the future, but minor version bumps seem infrequent. It's been two months since 8.0 GA.

src/aspire-dashboard/8/cbl-mariner-distroless/amd64/Dockerfile
```diff
@@ -9,9 +9,9 @@ RUN tdnf install -y \
     && tdnf clean all
 
 # Retrieve Aspire Dashboard
```
jander-msft, line 11: Shouldn't the Dockerfiles still be under the <major>.<minor> folder (`8.1`) instead of major-only folder (`8`)? I thought that the `ProductVersion` parameter of the Set-DotnetVersions.ps1 script (and related update tooling) required this.

lbussell: The tooling does not require it, since the update-dependencies step is completely separate from the Dockerfile generation step.

We'll never carry two major-minor versions of Aspire Dashboard, so I think it makes sense to have these in an `8` folder to avoid unnecessary renames later on.

mthalman: @lbussell - I understand your desire to avoid needing to mess with Dockerfiles when updating to a new major.minor version. But if part of the reason to use major.minor variable names in the versions manifest is to maintain consistency, why wouldn't we apply that to the Dockerfile path as well? In addition, I like the extract clarity using a major.minor directory name provides when just browsing the directories.

lbussell: Sure, I suppose that'd be more consistent overall then. I'll change it.

mthalman: The Dockerfiles still need the `8` directory renamed.

mthalman (approved):

lbussell:
Tests are failing because they are looking for 8.1 tags. `System.Version` doesn't support major-version-only versions, and it wouldn't be good to hardcode the tests to a specific patch version. For consistency's sake should we add the major-minor tags back, or introduce a special case for Aspire?

@jander-msft @mthalman

JamesNK (changes requested):
@joperezr I want to make a change to the image for 8.1. We added support for OTLP HTTP. This should be configured as an endpoint by default in the image. This is done by adding `DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL` as an argument to the dashboard.

```diff
-    DOTNET_DASHBOARD_OTLP_ENDPOINT_URL=http://0.0.0.0:18889
+    DOTNET_DASHBOARD_OTLP_ENDPOINT_URL=http://0.0.0.0:18889 \
+    DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL=http://0.0.0.0:18890
```

README.aspire-dashboard.md
```diff
@@ -114,12 +114,12 @@ Limits are per-resource. For example, a `MaxLogCount` value of 10,000 configures
 ## Linux amd64 Tags
 Tags | Dockerfile | OS Version
 -----------| -------------| -------------
-8.0.2-amd64, 8.0-amd64, 8.0.2, 8.0, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile) | CBL-Mariner 2.0
+8.1.0-amd64, 8-amd64, 8.1.0, 8, latest | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/aspire-dashboard/8.1/cbl-mariner-distroless/amd64/Dockerfile) | CBL-Mariner 2.0
```
JamesNK, line 117: Line 7 of this file, should 8.1 be added as a featured tag?

jander-msft: IMO, the minor version tags (e.g. `8.1-*`) don't make sense because the 8.1 image is full replacing the 8.0 image without overlapping lifecycle support. A customer using the `8.0-*` tags would immediately be unsupported once the 8.1 image is available; they won't get .NET base image updates nor any .NET Aspire bug fixes and there is no time for customers to evaluate the update path from 8.0 to 8.1. With this type of lifecycle, the minor tags do not provide customer value. Only major version tags (e.g. `8-*`) and patch version tags (e.g. `8.1.0*`) should be available so that customers have to ability to pin to an exact version should there be an issue with the next update and they need to rollback.

If there were overlapping support between a new minor version and the previous minor version, even for a short time (we do this for .NET Monitor, which has about a 3 month overlap before removing the prior minor version), then there is value in having the minor version tags.

jander-msft: Additionally, if the minor tags are being added back just to resolve https://github.com/dotnet/dotnet-docker/pull/5720#issuecomment-2248917336, then I would propose to make them undocumented given my argument that they do not provide customer value.

MichaelSimons: > With this type of lifecycle, the minor tags do not provide customer value

The <major>.<minor> tags do provide the ability to receive patch updates yet isolate yourself from any functional/breaking changes that may occur across minor version updates.  They do provide value for that reason IMO regardless if there is support overlap in minor versions.  Because of this, I think they should be offered.

lbussell:
Waiting on a final review from @MichaelSimons and an Aspire team member (@joperezr or @JamesNK) to merge.

README.aspire-dashboard.md
MichaelSimons: > Line 7 of this file, should 8.1 be added as a featured tag?

I don't see that this issue has been addressed yet.

joperezr:
LGTM, would be good to merge when we can as 8.1 has been out for a couple days now and people are looking for the update in docker image 😃

README.aspire-dashboard.md
lbussell: Done in https://github.com/dotnet/dotnet-docker/pull/5720/commits/c12d483f0011ad0ecfd929ffe2e06b6a2592710a

MichaelSimons (approved):

PR #5726, opened 2024-07-24
Remove workaround for Mariner ARM container images not setting architecture variant metadata

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/3520


eng/pipelines/steps/set-imagebuilder-build-args-var.yml
mthalman, line 1: This whole file should be able to be deleted, plus updating referrers to this file.

mthalman (approved):

mthalman (approved):

PR #5738, opened 2024-07-29
Factor test cache ACR argument out of common pipeline infrastructure

lbussell:
This argument was unnecessarily placed in the shared templates under eng/common. It would be overwritten by https://github.com/dotnet/dotnet-docker/pull/5736. Instead, we should change the common template to allow additional arguments to be set beforehand by consuming repos, in case some pipelines may not have the variable set. Then, in this repo, we'll set the argument in the `build-test-publish` variables.

### DIFF
--- a/eng/common/templates/steps/test-images-linux-client.yml
+++ b/eng/common/templates/steps/test-images-linux-client.yml
@@ -17,7 +17,7 @@ steps:
 - script: |
     echo "##vso[task.setvariable variable=testRunner.container]testrunner-$(Build.BuildId)-$(System.JobId)"
 
-    optionalTestArgs=" -CacheRegistry $(public-mirror.server)"
+    optionalTestArgs="$OPTIONALTESTARGS"
     if [ "${{ parameters.preBuildValidation }}" == "true" ]; then
       optionalTestArgs="$optionalTestArgs -TestCategories pre-build"
     else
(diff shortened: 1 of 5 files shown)


eng/common/templates/steps/test-images-linux-client.yml
```diff
@@ -17,7 +17,7 @@ steps:
 - script: |
     echo "##vso[task.setvariable variable=testRunner.container]testrunner-$(Build.BuildId)-$(System.JobId)"
 
-    optionalTestArgs=" -CacheRegistry $(public-mirror.server)"
+    optionalTestArgs="$OPTIONALTESTARGS"
```
mthalman, line 20: This is kind of a duplication of the already existing `repoTestArgs` variable that gets consumed further below. We should consolidate these as there should only be one means to customize test args.

I think the only usage of `repoTestArgs` is here:

https://github.com/dotnet/dotnet-docker/blob/b64fbd3c0914c1dd1ef1b5476872be3a2028571d/eng/pipelines/variables/samples.yml#L5-L6

lbussell: I had noticed that but I decided against using it since it seemed to be repo-specific. However I agree, it would be simpler to just build up one single list of additional test arguments. I'll make that change.

mthalman (approved):

Issue #5113, opened 2024-01-17
Some sample commands still assume the in-container user will be root


amcasey:
### Describe the Bug

[Here](https://github.com/dotnet/dotnet-docker/blob/7bca20cb06e1f912fc2e7fa8ce04dda606277537/samples/run-aspnetcore-https-development.md?plain=1#L92), for example, there's a command for running a docker image and mounting `%appdata%\ASP.NET\https` as `/root/.aspnet/https`.  However, if the container runs as user `app`, AspNetCore will actually look in `/home/app/.aspnet/https`.

The VS tooling appears to handle this by mounting it in both locations.
```
-v "C:\Users\amcasey\AppData\Roaming\ASP.NET\Https:/root/.aspnet/https:ro" -v "C:\Users\amcasey\AppData\Roaming\ASP.NET\Https:/home/app/.aspnet/https:ro"
```

There may also be a clever way to consume a docker or environment variable to pick the right one.

### Steps to Reproduce

We heard about it [here](https://github.com/dotnet/aspnetcore/issues/53187).  I assume following the steps verbatim will result in the https certificate not being found, but I haven't personally verified.

### Other Information

I believe running as non-root is new in net 8.0, so earlier versions should be unaffected (but, ideally, should still work with the revised instructions).

lbussell:
@amcasey We'll be doing a once-over of our samples for RC1, so we can try to improve the documentation for this scenario then. I put it in the queue for September.

PR #5770, opened 2024-08-08
Drop caveats about dev cert trust on Linux

amcasey:
As of 9.0 RC1 and 8.0.9, `--trust` should work on Linux as well (though it is only officially supported on Ubuntu and Fedora).

### DIFF
--- a/samples/host-aspnetcore-https.md
+++ b/samples/host-aspnetcore-https.md
@@ -59,9 +59,6 @@ dotnet dev-certs https --trust
 ```
 
 > [!NOTE]
-> `dotnet dev-certs https --trust` is only supported on macOS and Windows. You need to trust certs on Linux in the way that is supported by your distro. It is likely that you need to trust the certificate in your browser.
-
-> [!NOTE]
 > `<CREDENTIAL_PLACEHOLDER>` is used as a stand-in for a password of your own choosing.
 
 Run the container image with ASP.NET Core configured for HTTPS:
--- a/samples/host-aspnetcore-https.md
+++ b/samples/host-aspnetcore-https.md
@@ -51,6 +51,12 @@ docker run --rm -it -p 8001:8001 -e ASPNETCORE_HTTPS_PORTS=8001 -e ASPNETCORE_Ke
 
 ### macOS or Linux
 
+Create a certificate directory with appropriate permissions:
+
+```console
+mkdir -p -m 700 ${HOME}/.aspnet/https
+```
+
 Generate cert and configure local machine:
 
 ```console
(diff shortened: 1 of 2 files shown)


amcasey:
FYI @richlander

richlander:
What about Alpine? It is very relevant for this repo.

amcasey:
TBH, I expect most linux distros to work, but we've only committed to Ubuntu and Fedora.  If you think Alpine needs to be on the list, we can certainly discuss it.  Are you set up to be able to test that?

Also, to confirm, we're talking about the docker host, not the container OS, right?  You're running docker server and your browser in Alpine?

amcasey:
I made an Alpine VM and HttpClient and curl worked.  I didn't try Edge.

richlander:
Right, container OS.

Alpine is one of the most popular container OSes for us, so it follows that we should support it. It is very much not niche.

amcasey:
The instructions I updated seem to describe commands to be run on the host machine, not on the container.  Am I reading that correctly?

lbussell (changes requested):
Since this does not apply to 8.0.8 and 9.0 Preview 7, please re-target this to the nightly branch.

mthalman:
> Since this does not apply to 8.0.8 and 9.0 Preview 7, please re-target this to the nightly branch.

@lbussell - What's the reasoning for targeting nightly? This is a doc change only which we've historically applied directly to main branch usually.

mthalman:
> > Since this does not apply to 8.0.8 and 9.0 Preview 7, please re-target this to the nightly branch.
> 
> @lbussell - What's the reasoning for targeting nightly? This is a doc change only which we've historically applied directly to main branch usually.

Oh, never mind. I understand now what you're saying about the version.

amcasey:
Back to `main` then?

amcasey:
Also, I'm having some trouble following the instructions and I think I might need some help from a docker expert.  The instructions mount `~/.aspnet/https` from the host machine as `/root/.aspnet/https` in the container, but we've started (as of 8.0?) running as `app` in the container, so it seems like it should actually be mounted as `/home/app/.aspnet/https`.  I made that change locally, but it's still not working because the directory in the container has the same owning uid as the folder in the host OS (i.e. it's not owned by `app`).  The certificate isn't (and shouldn't be) world-readable in the host OS, so it's not accessible to `app` when the app launches in the container.  How can I change the folder owner inside the container?

mthalman:
> Back to `main` then?

No, nightly branch is correct.

amcasey:
Here's an alpine host (dotnet 9.0 RC1) accessing an alpine container (dotnet 8.0.7) using firefox in gnome
image

amcasey:
@lbussell @mthalman Who would I talk to about the mounted volume permissions issues?

amcasey:
Oh, right: https://github.com/dotnet/dotnet-docker/issues/5113

amcasey:
@lbussell If you're going to review user permissions later, I think this PR is ready to go.  Anything else you'd like to revisit?

lbussell:
For context I added this comment: https://github.com/dotnet/dotnet-docker/issues/5113#issuecomment-2284485765.

This PR looks OK to me.

lbussell (approved):

mthalman (approved):

Issue #5375, opened 2024-04-16
Update 8.* aspire-dashboard and monitor mariner images to 3.0


MichaelSimons:
Once Azure Linux 3.0 is released, the 8.* aspire-dashboard and monitor mariner images should be updated to it.  These images are appliance images that have either no distro tags (aspire) or floating distro tags (monitor) which means updating the OS version is intended.  Given there the branding change between cbl-mariner and azurelinux, the recommendation is to add azurelinux tags for monitor and undocument the existing mariner tags (they will still be updated/produced, just not documented).

lbussell:
[Triage] Related: https://github.com/dotnet/dotnet-docker/issues/5467 

The distro used in these images can be updated since they're "Appliance" style images and don't guarantee a specific distro version. We do this for Alpine Linux and .NET Monitor already. We should capture this policy decision in our documentation for future reference.

Since [Mariner 2.0 may be EOL](https://microsoft.github.io/azurelinux/docs/) before .NET 8.0 is EOL, we will likely have to do this. It's just a matter of when - whether it's immediately or closer to the EOL of Mariner 2.0. We should work with the Aspire and Monitor teams to ensure that we have CI coverage for Azure Linux 3.0 before its release.

jander-msft:
.NET Monitor runs tests on Helix and would require https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1005 to be completed first before adding Azure Linux 3.0 to its test matrix.

lbussell:
This was blocked on FedRAMP approval of Azure Linux 3.0: https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-platforms.md#fedramp-compliance

Azure Linux 3.0 is now FedRAMP certified, so we can implement this any time.

Also, Azure Linux 2.0 reaches end of life in July 2025, so we need to complete this migration well before then.

PR #5781, opened 2024-08-09
Update Monitor 8.0 and 8.1 to Azure Linux 3.0

lbussell:
2/2 of #5375

According to the recommendation in that issue, this adds new tags for Azure Linux and makes the existing `cbl-mariner*` tags undocumented. This way, users referencing the existing floating tags will still get updates, but new users will only see the Azure Linux tags.

I chose to remove tags with the format `8.1*-cbl-mariner-distroless*` since .NET Monitor 8.1 is still in preview, and thus those tags were never published as supported. The `8-cbl-mariner-distroless` tag will need to stick around, however, since that tag is currently supported in the main branch.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -9548,21 +9548,25 @@
         {
           "productVersion": "$(monitor|8.0|product-version)",
           "sharedTags": {
-            "$(monitor|8.0|fixed-tag)-cbl-mariner-distroless": {},
-            "$(monitor|8.0|minor-tag)-cbl-mariner-distroless": {}
+            "$(monitor|8.0|fixed-tag)-azurelinux-distroless": {},
+            "$(monitor|8.0|minor-tag)-azurelinux-distroless": {},
+            "$(monitor|8.0|fixed-tag)-cbl-mariner-distroless": { "docType": "Undocumented" },
+            "$(monitor|8.0|minor-tag)-cbl-mariner-distroless": { "docType": "Undocumented" }
           },
           "platforms": [
             {
               "buildArgs": {
                 "REPO": "$(Repo:aspnet)"
               },
-              "dockerfile": "src/monitor-base/8.0/cbl-mariner-distroless/amd64",
+              "dockerfile": "src/monitor-base/8.0/azurelinux-distroless/amd64",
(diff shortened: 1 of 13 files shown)


manifest.json
```diff
@@ -9628,38 +9634,47 @@
         {
           "productVersion": "$(monitor|8.1|product-version)",
           "sharedTags": {
-            "$(monitor|8.1|fixed-tag)-cbl-mariner-distroless": {},
-            "$(monitor|8.1|minor-tag)-cbl-mariner-distroless": {},
-            "$(monitor|8|major-tag)-cbl-mariner-distroless": {}
+            "$(monitor|8.1|fixed-tag)-azurelinux-distroless": {},
+            "$(monitor|8.1|minor-tag)-azurelinux-distroless": {},
+            "$(monitor|8|major-tag)-azurelinux-distroless": {},
+            "$(monitor|8.1|fixed-tag)-cbl-mariner-distroless": { "docType": "Undocumented" },
```
jander-msft, line 9640: Similar to the decision for the `dotnet/monitor` images, should the `$(monitor|8.1|fixed-tag)-cbl-mariner-distroless`* and `$(monitor|8.1|minor-tag)-cbl-mariner-distroless`* tags be removed for `dotnet/monitor/base` images?

src/monitor-base/8.0/azurelinux-distroless/amd64/Dockerfile
mthalman, line 1: Looks like the size validation baseline file needs to be updated to reflect the new paths.

manifest.json
lbussell: Yes - that's an oversight on my part.

manifest.json
```diff
@@ -10151,4 +10205,4 @@
       ]
     }
   ]
-}
+}
```
mthalman, line 10208: nit: add back line ending
```suggestion
}

```

mthalman (approved):

manifest.json
lbussell: OK, but for the record I *did* just "apply file formatting in VS Code" :)

lbussell:
Closing in order to wait for FedRAMP approval for Azure Linux 3.0.

Issue #5316, opened 2024-03-28
Use a Markdown linter in PR validation


lbussell:
We should add a test or pre-build validation step that runs a markdown linter to check for dead links and proper markdown formatting across all of our docs (not just generated readmes)

Related: #5297 and https://github.com/dotnet/docker-tools/issues/1172

richlander:
The team is trying to put my creative efforts in a straightjacket. "I cannot create under these conditions!"

MichaelSimons:
FYI, the markdownlint tool I have integrated into CI only validates intra document links that use the # anchor ID syntax.  There are requests for it.  It appears a lot of project use https://github.com/tcort/markdown-link-check.

MichaelSimons:
This work is now complete.  The dotnet-docker and dotnet-framework have a markdown linter and markdown link validator GH action.  All markdown issues within the repos have also been addressed

PR #5782, opened 2024-08-12
Configure duplicate headers markdown rule and address violations

MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/issues/5316


samples/run-in-sdk-container.md
mthalman, line 1: This section above needs to be the headers updated to use `###`: https://github.com/dotnet/dotnet-docker/pull/5782/files#diff-f8ce3051596340203e9f0d647e514f333583e4bb2a35dba4ba85bc3d2b0f0b9fR56-R70.

MichaelSimons: Good catch - fixed in https://github.com/dotnet/dotnet-docker/pull/5782/commits/39c6ad7bdd0f1b009d59afc47e3752a693dc2084

Issue #4517, opened 2023-03-27
Onboard .NET Docker Bot to Secret Manager


lbussell:
Instructions here: https://github.com/dotnet/arcade-services/tree/main/src/Microsoft.DncEng.SecretManager#onboarding-a-new-repo

PR #5797, opened 2024-08-14
Onboard .NET Docker Bot to secret manager

ellahathaway:
Related to https://github.com/dotnet/dotnet-docker/issues/4517

Changes were made by following guidance in [this documentation](https://github.com/dotnet/dnceng/tree/main/src/SecretManager/Microsoft.DncEng.SecretManager#onboarding-a-new-repo) and what is currently implemented in some repos (https://github.com/dotnet/arcade/blob/main/azure-pipelines-daily.yaml for example).

Additionally, to my knowledge we don't use any secrets that should go into a json.settings file, so I did not include the `Verify Secret Usages` steps as part of the changes in this PR.

- [Test build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2516748&view=results) for validate-secrets.yml in pre-publish steps

Steps to do before publishing this PR:
- Validate new secret-management-weekly.yml


eng/pipelines/stages/build-test-publish-repo.yml
```diff
@@ -36,5 +36,6 @@ stages:
     customTestInitSteps:
     - template: /eng/pipelines/steps/set-custom-test-variables.yml@self
     customPublishInitSteps:
+    - template: /eng/pipelines/steps/validate-secrets.yml@self
```
mthalman, line 39: I don't understand the need for this. If the secret management pipeline runs weekly, it ensures the secrets are up to date. So what's the point of validating? If a secret for some reason was not up-to-date, wouldn't it naturally fail during its usage? So I don't understand the need for a separate validation step, especially at this point in the pipeline.

ellahathaway: I agree. I added it because the documentation said to, but truthfully it feels redundant, and I had trouble finding places in different repos that closely followed the documentation. I'm going to submit an issue to dnceng about updating the documentation.

ellahathaway:
Ported to docker-tools: https://github.com/dotnet/docker-tools/pull/1405.

PR #5809, opened 2024-08-16
Switch markdownlint GH action to use super-linter

MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/issues/5316

Switched the MarkDown linter GH action to use the super-linter.  This gives us the advantage of enabling other linters via a "flip of a switch".  The configuration will be more standardized with the super-linter versus having separate GH actions/linters.  The super-linter used the markdownlint tool we are already using so there should be no functional changes.


.github/workflows/lint-code-base.yml
```diff
@@ -0,0 +1,31 @@
+name: Lint Code Base
+
+on:
+  pull_request: null
```
mthalman, line 4: Don't we want to keep the path filter on this?

MichaelSimons: I kept it off with the idea that we would enable other linters.  Trying to manage the paths becomes unfeasible when you start doing this.

mthalman (approved):

PR #5822, opened 2024-08-20
Normalize base url variable names

jander-msft:
The following changes are intended to make the base URL variables more consistent in structure and naming so that future changes can more easily enable automated updates of the base URL variables.

- Renamed the common base URL variables to adhere to the following pattern so they have the same number of elements in each of the URL variations: `base-url|public|{daily|maintenance}|<branch>`. The following changes were made:
  - `base-url|public|main` -> `base-url|public|daily|main`
  - `base-url|public|nightly` -> `base-url|public|daily|nightly`
- Renamed product base URL variables to adhere to the following pattern so they have the same number of elements in each of the URL variations: `<product>|<version>|base-url|<branch>`. The following changes were made:
  - `base-url|<version>|<branch>` -> `dotnet|<version>|base-url|<branch>`
  - `base-url|8.1-aspire-dashboard|<branch>` -> `aspire-dashboard|8.1|base-url|<branch>`
  - `base-url|<version>-monitor|<branch>` -> `monitor|<version>|<branch>`
- Moved all product base URL variables to the same group as the product variables.


manifest.versions.json
```diff
@@ -1,12 +1,19 @@
 {
   "variables": {
+    "base-url|public|daily|main": "https://dotnetcli.azureedge.net/dotnet",
```
jander-msft, line 3: I'm not set on using the term `daily` here; I just needed something that would contrast with the `maintenance` term. If there is a better term to use here with a better connotation, totally open to it.

mthalman: How about `preview`? I'd prefer something other daily simply to avoid having a variable with both "daily" and "nightly" in the name (`base-url|public|daily|nightly`), which is just weird. 😄

manifest.versions.json
```diff
@@ -1,12 +1,19 @@
 {
   "variables": {
+    "base-url|public|preview|main": "https://dotnetcli.azureedge.net/dotnet",
+    "base-url|public|preview|nightly": "https://dotnetbuilds.azureedge.net/public",
+    "base-url|public|maintenance|main": "https://dotnetcli.azureedge.net/dotnet",
+    "base-url|public|maintenance|nightly": "https://dotnetcli.azureedge.net/dotnet",
```
lbussell, line 6: Consider swapping the order of `preview` and `maintenance` here? Alphabetical order, and maintenance before preview. Then you could also consider setting `base-url|public|preview|main` to `$(base-url|public|maintenance|main)` - but maybe you've chosen to duplicate the strings here for a reason.

mthalman (approved):

PR #5832, opened 2024-08-23
Dynamically generate Dockerfiles for test scenarios

### LINKED ISSUE #5787: Use default NuGet configuration for scenario tests
This is the preferred solution to https://github.com/dotnet/dotnet-docker/issues/5704. In order to more accurately validate user experience, we should use the default NuGet configuration for scenario tests in both main and nightly (except for pre-release versions of .NET, which require a special NuGet feed).

The relevant NuGet configs that we use today are here:

- nightly: https://github.com/dotnet/dotnet-docker/blob/5492718d4ddb6949b3853aac148b784a6572c8fe/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config.nightly
- main: https://github.com/dotnet/dotnet-docker/blob/5492718d4ddb6949b3853aac148b784a6572c8fe/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config

Once this is complete, the workaround from https://github.com/dotnet/dotnet-docker/pull/5786 should be removed.

lbussell:
Done:
- [x] Remove static test Dockerfiles and replace with dynamically generated ones
- [x] Split framework-dependent, self-contained, AOT, and Xunit project into separate test scenarios

Todo:
- [x] Re-implement BlazorWASM test dockerfile
- [x] Allow disabling use of NuGet.config - fixes #5787


tests/Microsoft.DotNet.Docker.Tests/TestScenarios/TestDockerfile.cs
```diff
@@ -0,0 +1,156 @@
...
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.DotNet.Docker.Tests;
+
+public class TestDockerfile(IEnumerable<string> args, IEnumerable<string> layers)
```
mthalman, line 14: This would be more accurately described as `stages` instead of `layers`. I think this probably applies to all the usages of `layer` throughout this PR.

mthalman (approved):

PR #5833, opened 2024-08-23
Allow update-dependencies to update product base URL

### LINKED ISSUE #4939: Update-dependencies tool can't handle switching between security releases and daily/public builds
### Describe the Bug

The update dependencies tool uses the base URL from `manifest.versions.json` to get the checksums for new builds:
https://github.com/dotnet/dotnet-docker/blob/57e9112e172fa162018c92c57e1910c020786a3a/manifest.versions.json#L42-L53

In the nightly branch, when previews/release candidates have security releases, the URL needs to be the maintenance version for exactly one release: the merge from main to nightly after a release. However, when the URL is changed for the merge from main to nightly, it stops the update-dependencies tool from picking up the checksums for the latest public builds:

https://dev.azure.com/dnceng/internal/_build/results?buildId=2288455&view=results
 
### Solution

The update-dependencies tool should have a separate URL defined in `manifest.versions.json` just for checking public builds that doesn't change. Then, we should have some switch like "preview-is-security-release" that update-dependencies can flip off when it gets a new public build after a release. We would read that variable from the Dockerfile template to get the correct URL. Potentially similar to .NET Framework's solution: 

https://github.com/microsoft/dotnet-framework-docker/blob/6a3c9d048f75c1e69c0e1059564cd56a90bf9e3c/manifest.versions.json#L20

jander-msft:
The intent of this change is to allow the update-dependencies tool to set the product base URL as it is updating the product version. This allows automation to automatically fix up the product base URL to correspond with where the product version assets are staged at the time of updating.

An example situation where this would be useful is when a security release is inserted directly into the main branch (which means its only the dotnetcli storage account at time of publishing) but the assets were never available on the dotnetbuilds storage account. When those changes are merged back to the nightly branch, automated updates for prerelease versions are stuck until the base URL is manually updated (along with a new product version) in the nightly branch.

With the proposed changes, the update-dependencies tool that is invoked by the update pipelines will set the base URL that corresponds with the product version. The pipelines assume that if targeting the nightly branch, then the product assets are in a prerelease mode, whereas if targeting the main branch, then the product assets are already released.

If need be, the `Set-DotnetVersions.ps1` script can be manually invoked with the `-ReleaseState` parameter to indicate whether the product assets have been released or not, which then informs the update-dependency tool which common base URL should be used for the product version. This parameter is not required; if not specified, the base URL for the product version is not updated.

closes #4939

### DIFF
--- a/eng/DependencyManagement.psm1
+++ b/eng/DependencyManagement.psm1
@@ -27,5 +27,9 @@ function Resolve-DotnetProductUrl([string] $akaMsUrl) {
 }
 
 function Get-ProductReleaseState() {
-    return $(Get-Branch) -ieq 'main' ? 'Release' : 'Prerelease'
+    if ($(Get-Branch) -ieq 'main') {
+        return 'Release'
+    } else {
+        return 'Prerelease'
+    }
 }
--- a/eng/DependencyManagement.psm1
+++ b/eng/DependencyManagement.psm1
@@ -25,3 +25,7 @@ function Resolve-DotnetProductUrl([string] $akaMsUrl) {
     Write-Host "Resolved URL: $resolvedUrl"
     return $resolvedUrl
 }
+
+function Get-ProductReleaseState() {
+    return $(Get-Branch) -ieq 'main' ? 'Release' : 'Prerelease'
+}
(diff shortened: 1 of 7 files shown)


jander-msft:
cc @dotnet/dotnet-monitor

mthalman (approved):

mthalman (approved):

PR #5835, opened 2024-08-26
Simplify appliance product base URL calculation

jander-msft:
The appliance-style image templates use a non-trivial base URL lookup based on version patterns and equality of product and build version. This was born out of necessity when there were only one set of base URLs for a given version and branch in the [manifest.versions.json](https://github.com/dotnet/dotnet-docker/blob/d5943b12928c24b8d16da6c0a2d805823c3f436a/manifest.versions.json) file that dictated the base URL for all products. [Subsequent changes](https://github.com/dotnet/dotnet-docker/commit/6a957822418dfdfd78a4816154dfdd3926b51e29#diff-f657b02176508554b475067f99d956c3c0fb1c08be55990212cb9179813b4aca) further split the URLs by product, thus each product can independently declare its base URLs for each version and branch combination.

The complicated logic for calculating the base URL for the appliance-style images is no longer necessary due to this split. While the proposed change in this PR largely has no effect on the existing dockerfiles nor an effect on how versions are updated, the removal of the unnecessary logic simplifies the templates a small amount and can prevent confusion to those who attempt to reason about why this logic exists.


mthalman (approved):
Nice cleanup.

PR #5837, opened 2024-08-27
Enable EOL annotation publishing

mthalman:
Contributes to https://github.com/dotnet/docker-tools/issues/1201

This enables EOL annotation publishing for the pipelines related to the nightly branch. This also manually syncs the eng/common directory from docker-tools due to https://github.com/dotnet/docker-tools/issues/1410.


lbussell:
@mthalman Looks like we will need to skip the EOL annotation data generation for public pipelines.

mthalman:
> @mthalman Looks like we will need to skip the EOL annotation data generation for public pipelines?

We don't want to skip it. But it should be doing a dry run that doesn't require auth. I'll make a fix for that.

mthalman:
Need some additional fixes in docker-tools: https://github.com/dotnet/docker-tools/pull/1423

mthalman:
@dotnet-docker-reviewers - This is ready for review now.

lbussell (approved):

PR #5848, opened 2024-08-29
Update Aspire to 8.2.0 version

joperezr:
cc: @lbussell


joperezr:
We found a bug in 8.2.0 dashboard when using insecure mode, so we can't merge this yet, we'll want to take a fix before we update the image

joperezr:
Closing this for now until @JamesNK pushes a fix and we have a build ready.

joperezr:
@lbussell this is now ready to review and get merged. cc: @JamesNK

joperezr:
@lbussell do you know how to fix the build issues? I think they are related to the fact that the product version itself is marked as 8.2.0 stable (which it should) but the url is preview (which had to be done in order to avoid blob clashes).

lbussell (approved):

lbussell:
PR validation failure is https://github.com/dotnet/dotnet-docker/issues/5871. We can merge around that check. This should wait on https://github.com/dotnet/dotnet-docker/pull/5870 though.

joperezr:
> @joperezr @lbussell
> 
> seemed the docker image is still not pushed, the latest tag is still `8.1`

That is correct @WeihanLi. This got merged into the `nightly` branch, where is were we push new Aspire builds each time we have just shipped. That 8.2 tag is ready for our nightly images: https://mcr.microsoft.com/en-us/product/dotnet/nightly/aspire-dashboard/about

For the one from main, we typically have to wait till the next merge happens from `nightly` to `main` branch, which is scheduled for the first two weeks of October. Typically there won't be such a lag between shipping Aspire and getting a non-nightly image, but this time around it was mostly due to timing since Aspire shipped 8.2.0 right after a the merge from `nightly` to `main` happened.

joperezr:
Apologies for the delay and the trouble.

WeihanLi:
@joperezr thanks very much for the detailed reply, from my understanding, when a stable release is published, stable tags should also be pushed, the nightly image tags should be used for test before the stable release published

Issue #5838, opened 2024-08-27
Static Analysis Tests for Tag Validation


ellahathaway:
### Problem to solve

We need to implement static tests (based on image-builder manifest/README) for the following scenarios to ensure the correctness of our tag schema and adherence to our tagging rules:

**Tag Schema Validation**
- Ensure the tag schema matches the description in [supported-tags.md](https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-tags.md). Specifically:
   - Verify that the Alpine floating tag references the correct version of Alpine.
   - Confirm that appliance images use the correct tags, as they follow a different schema.

**"Latest" Tag Rules**
-  https://github.com/dotnet/dotnet-docker/pull/5852
- Ensure the following rules are adhered to for the latest tags:
  - There is only one latest tag per repository.
  - The latest tag points to the expected version:
      - For main, it should point to the latest non-preview/release candidate major release.
      - For nightly, it should point to the latest unreleased version.

### Additional implementation details
My initial thoughts are that we should use the tag tables/READMEs for this validation. Upon initial examination of this issue, parsing the manifest is difficult as it requires deserialization and replacement of variables. This is something that is already handled by the image builder, and rewriting this logic is tedious and prevents us from making changes to the manifest without large pieces of the test code becoming invalid. Since we already validate the readmes against the manifest, the readmes are another truthful place to validate tags that doesn't require immense parsing and duplicate image-builder logic. That said, another option for these tests, other than the manifest and the readme, is to make the image builder output some sort of artifact that contains the values we wish to validate.

lbussell:
@mthalman @ellahathaway I do like the idea of an imagebuilder command that resolves all the variable values in the manifest. However that still means we'd need to parse the manifest. Parsing the manifest would be useful for other reasons such as https://github.com/dotnet/dotnet-docker/issues/5337.

Another option could be to re-use the MCR tags metadata yaml that we already generate. That would prevent needing to read the manifest from the dotnet-docker repo. However we'd still be stuck parsing another new file format from this repo, just yaml this time instead of json.

This would require a new command to output the tags metadata files to the disk, and would also require an option to include undocumented tags in the output.

Example:
```yaml
  - tags: [ 9.0.0-preview.7-bookworm-slim-amd64, 9.0-preview-bookworm-slim-amd64, 9.0.0-preview.7-bookworm-slim, 9.0-preview-bookworm-slim, 9.0.0-preview.7, 9.0-preview ]
[...]

Issue #5337, opened 2024-04-05
Use manifest.json/image-info.json to determine which tests to run.


lbussell:
### Describe the Problem

Maintaining [TestData.cs](https://github.com/dotnet/dotnet-docker/blob/main/tests/Microsoft.DotNet.Docker.Tests/TestData.cs) causes merge conflicts during servicing and allows for human error. Most of this data is already duplicated in some way by the manifest.json file. We should consider using that file to drive our tests.

### Describe the Solution

[There is already a model for the manifest in the ImageBuilder tool](https://github.com/dotnet/docker-tools/tree/main/src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest). We could either duplicate that model, create a simpler model in this repo just for reading the manifest, or factor out the model from ImageBuilder and publish it as a package.

Including a model to read the manifest file in this repo would also help with https://github.com/dotnet/dotnet-docker/issues/4704.

### Additional Context

- Should still allow for disabling specific image tests for certain cases.

lbussell:
[Triage] 

1. We discussed generating the TestData in the Test project, but decided against it since it might be easy to miss special cases that exist in the manifest if the test cases aren't explicitly enumerated.

It would be a good idea to create a prototype of this, probably by creating a simple model in the Test project.

lbussell:
[Triage] If it's possible to share images between 9.0 and 10.0, then we should aim to have a solution for this issue in place before .NET 10 is released.

PR #5852, opened 2024-09-03
Latest Tag Static Validation Tests

ellahathaway:
This PR is part of the logic needed to complete https://github.com/dotnet/dotnet-docker/issues/5838.

**Tests for Latest Tags**
In `StaticTagTests.cs`, this PR introduces tests to ensure adherence to the latest tag rules:

- `LatestTag_OnePerRepo`: Verifies that there is only one latest tag per repository.
- `LatestTag_IsNotPlatformSpecific`: Ensures the latest tag is a shared-tag and is not specified in any platform's tags.
- `LatestTag_OnCorrectMajorVersion`: Confirms that the latest tag points to the expected major version for each repo.

**Parsing Logic for Image-Builder Manifest**
In `ManifestHelper.cs`, this PR adds logic to parse the image-builder manifest into a format suitable for testing.


tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        private void LatestTag_OnePerRepo(ManifestHelper.Repo repo)
+        {
+            var latestTagImages = repo.Images
+                .Where(image => image.GetSharedTags().Contains(LatestTagValue))
+                .ToList();
+
+            Assert.True(
+                latestTagImages.Count == 1,
+                $"Expected a single image with the shared latest tag, but found {latestTagImages.Count}.");
```
mthalman, line 42:
```suggestion
            Assert.Single(latestTagImages);
```

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            foreach (var image in repo.Images)
+            {
+                var latestTagPlatforms = image.Platforms
+                    .Where(platform => platform.GetTags().Contains(LatestTagValue))
+                    .ToList();
+
+                Assert.True(
+                    latestTagPlatforms.Count == 0,
+                    $"Expected no platforms with a specific latest tag, but found {latestTagPlatforms.Count}: \n" +
+                    $"  {string.Join("\n", latestTagPlatforms.Select(platform => platform.Dockerfile))}");
```
mthalman, line 56:
```suggestion
                Assert.Empty(latestTagPlatforms);
```

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            var actualMajorVersion = latestImage.ProductVersion.Split('.')[0];
+
+            Assert.Equal(expectedMajorVersion, actualMajorVersion);
+        }
+
+        private string GetExpectedMajorVersion(ManifestHelper.Repo repo)
+        {
+            List<string> productVersions = repo.Images.Select(image => image.ProductVersion.Split('.')[0]).Distinct().ToList();
+
+            Assert.True(productVersions.Count > 0, "No product versions found for repo.");
```
mthalman, line 76:
```suggestion
            Assert.NotEmpty(productVersions);
```

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                return productVersions[0];
+            }
+            else if (Config.IsNightlyRepo)
+            {
+                return productVersions.Max()!;
+            }
+            else
+            {
+                return productVersions.OrderByDescending(version => version).ElementAt(1);
```
mthalman, line 88: This will fail on the GA release because we will move `latest` to point to 9.0 at that time.

ellahathaway: Ah you're right. I was thinking that we'd add 10.0 alpha at the same time that we GA for 9.0, but we don't add anything for 10.0 until it's in preview, correct?

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public static IEnumerable<object[]> GetRepoObjects() => ManifestHelper.GetManifest().Repos.Select(repo => new object[] { repo }).ToArray();
+
+        [Theory]
+        [MemberData(nameof(GetRepoObjects))]
+        public void ValidateLatestTag(ManifestHelper.Repo repo)
+        {
+            LatestTag_OnePerRepo(repo);
+            LatestTag_IsNotPlatformSpecific(repo);
+            LatestTag_OnCorrectMajorVersion(repo);
```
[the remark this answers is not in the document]
ellahathaway: This was my original implementation, but as I added the tests for the tag schema validation, the number of tests being run became quite large. If we're okay that larger number, then I'm alright with splitting these into separate tests.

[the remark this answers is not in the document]
mthalman: I'm not concerned about a large number of tests. The checks are still being run either way. Benefits of having these as separate tests:
* They get run in parallel
* If one fails, it won't block seeing the results of others

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
mthalman: Correct, 10.0 wouldn't exist in the main branch until its Preview 1 release next year.

tests/Microsoft.DotNet.Docker.Tests/ManifestHelper.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Text.RegularExpressions;
+
+#nullable enable
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public static class ManifestHelper
+    {
+        public static SimplifiedManifest GetManifest() =>
+            JsonConvert.DeserializeObject<SimplifiedManifest>(Config.Manifest.Value.ToString())!;
```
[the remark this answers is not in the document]
lbussell: Wouldn't deserializing the manifest get us most of the way towards https://github.com/dotnet/dotnet-docker/issues/5337 as well? It's not like deserialization code can't be re-used in this project. The JsonConverter approach may be more complicated than necessary for this use case but I don't see a big problem with deserializing in general. Do you think all future scenarios involving the manifest in these test should use Json node API?

IF we were deserializing here, I would expect just plain old classes and resolve the variables as necessary using the existing helper methods.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
lbussell: We can add it in Alpha, but these tests should not assume that there is a preview version.

ellahathaway: [This new logic](https://github.com/dotnet/dotnet-docker/pull/5852/files#diff-ba5f21c7428ca6e5f48f4fc38a102bb329c13fba9e07684b94a6d5b29f2c5e50R84) should work; for nightly it takes the greatest version listed, and for main it takes the greatest GA version (any version without a `-`). I validated the logic in main by changing the latest 9.0 sdk product version from 9.0.100-rc.1 to 9.0.0, and the test failed as expected.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -71,21 +63,30 @@ private void LatestTag_OnCorrectMajorVersion(ManifestHelper.Repo repo)
...
+                    .Select(version => version.Split('.')[0])
+                    .Distinct()
+                    .ToList();
+                return majorVersions.Max()!;
             }
             else
             {
-                return productVersions.OrderByDescending(version => version).ElementAt(1);
+                List<string> gaVersions = productVersions
+                    .Where(version => !version.Contains("-"))
```
mthalman, line 89: This would benefit from a comment explaining what the goal is here.

mthalman (approved):

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        public StaticTagTests(ITestOutputHelper outputHelper) => OutputHelper = outputHelper;
+
+        public static IEnumerable<object[]> GetRepoObjects() =>
+            ManifestHelper.GetManifest().Repos.Select(repo => new object[] { repo }).ToArray();
+
+        [Theory]
+        [MemberData(nameof(GetRepoObjects))]
+        private void LatestTag_OnePerRepo(ManifestHelper.Repo repo)
+        {
+            var latestTagImages = repo.Images
```
lbussell, line 30: Nit: use explicit types where it's not immediately obvious what the type is (I have to hover to figure out if this is strings, or Images, etc.). Also applies to some other tests.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,97 @@
...
+                .Where(image => ManifestHelper.GetResolvedSharedTags(image).Contains(LatestTagValue))
+                .ToList();
```
lbussell, line 32: The `ToList()` call here is unnecessary since Xunit's `Assert.Single` takes an `IEnumerable`. Applies to other tests as well.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        [Theory]
+        [MemberData(nameof(GetRepoObjects))]
+        private void LatestTag_OnCorrectMajorVersion(ManifestHelper.Repo repo)
+        {
+            var latestImage = repo.Images
+                .Where(image => ManifestHelper.GetResolvedSharedTags(image).Contains(LatestTagValue))
+                .First();
+
+            var expectedMajorVersion = GetExpectedMajorVersion(repo);
```
lbussell, line 59: Nit: use explicit types for built-in types (primitives). This reads as ambiguous between int or string.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            var actualMajorVersion = ManifestHelper.GetResolvedProductVersion(latestImage).Split('.')[0];
+
+            Assert.Equal(expectedMajorVersion, actualMajorVersion);
+        }
+
+        private string GetExpectedMajorVersion(ManifestHelper.Repo repo)
+        {
+            List<string> productVersions = repo.Images
+                .Select(image => ManifestHelper.GetResolvedProductVersion(image))
```
lbussell, line 68: My IDE points out that this is an unnecessary lambda:

```suggestion
                .Select(ManifestHelper.GetResolvedProductVersion)
```

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            else if (Config.IsNightlyRepo)
+            {
+                List<string> majorVersions = productVersions
+                    .Select(version => version.Split('.')[0])
+                    .Distinct()
+                    .ToList();
+                return majorVersions.Max()!;
+            }
+            else
+            {
```
lbussell, line 87: Avoid cascading if statements/unnecessary else clauses. Since each case here returns from the method, this could be formatted as:

```
if foo:
    return thing1;

if bar:
    return thing2;

return thing3;
```

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
lbussell: Agree, the three different string-split operations could each have explanations here.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,97 @@
...
+                .Where(image => ManifestHelper.GetResolvedSharedTags(image).Contains(LatestTagValue))
+                .First();
+
+            var expectedMajorVersion = GetExpectedMajorVersion(repo);
+            var actualMajorVersion = ManifestHelper.GetResolvedProductVersion(latestImage).Split('.')[0];
+
+            Assert.Equal(expectedMajorVersion, actualMajorVersion);
+        }
+
+        private string GetExpectedMajorVersion(ManifestHelper.Repo repo)
```
lbussell, line 65:
```suggestion
        private static string GetExpectedMajorVersion(ManifestHelper.Repo repo)
```

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            {
+                return productVersions[0].Split('.')[0];
+            }
+            else if (Config.IsNightlyRepo)
+            {
+                List<string> majorVersions = productVersions
+                    .Select(version => version.Split('.')[0])
+                    .Distinct()
+                    .ToList();
+                return majorVersions.Max()!;
```
lbussell, line 84: Prefer not to use the null-forgiving operator. If the method can return null then change the method signature to have nullable annotation. Otherwise if the method shouldn't return null, throw an exception with a detailed comment (or, use an assertion here since this is a test after all).

ellahathaway: I understand your point; however, I have already asserted that the product versions are not empty using `Assert.NotEmpty(productVersions)`;. Given this assumption, `majorVersions.Max()` is guaranteed to return a value since it only parses the major version numbers from the product versions. While I could assert that `version.Split('.')[0]` is not null, it seems excessive because, in my opinion, it is an axiom that the version number will contain at least one period.

lbussell: OK, that's fine then. I'm good with the null-forgiving operator in this case.

tests/Microsoft.DotNet.Docker.Tests/ManifestHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (!match.Success)
+            {
+                return input;
+            }
+            string variableName = match.Value.Replace("$(", string.Empty).Replace(")", string.Empty);
+            string remainingString = input.Split(match.Value)[1];
+            return Config.GetVariableValue(variableName) + remainingString;
+        }
+
+        public class SimplifiedManifest
```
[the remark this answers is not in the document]
ellahathaway: Sure, that makes sense.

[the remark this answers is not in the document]
ellahathaway: I created [this separate file](https://github.com/dotnet/dotnet-docker/pull/5852/files#diff-90167091f8a14aab39fa3a5793ec468b92725a4bb9485284c4bc157a17c2d885) for the classes, but I feel a little odd about all of the classes being in this single file. What are your thoughts on either nesting the classes, leaving this file as is, creating separate files for each of the classes?

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,103 @@
...
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    [Trait("Category", "pre-build")]
+    public class StaticTagTests
+    {
+        private const string LatestTagValue = "latest";
+        private ITestOutputHelper OutputHelper { get; }
+
+        public StaticTagTests(ITestOutputHelper outputHelper) => OutputHelper = outputHelper;
```
lbussell, line 22: OutputHelper seems unused here? Can it be removed?

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            
+            if (Config.IsNightlyRepo)
+            {
+                // Use the latest major version on the nightly branch
+                List<string> majorVersions = productVersions
+                    .Select(GetMajorVersion)
+                    .Distinct()
+                    .ToList();
+
+                return majorVersions.Max() ?? throw new Exception("No latest product versions found.");
```
lbussell, line 84: This calling `IEnumerable<string>.Max()`. That means it'll return `"9"` as max when we have .NET versions `["8", "9", "10"]`. Instead, `GetMajorVersion` should probably parse the string as an int before returning.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            // Use the latest GA major version on the main branch
+            List<string> gaVersions = productVersions
+                .Where(version => !version.Contains('-'))
+                .Select(GetMajorVersion)
+                .Distinct()
+                .ToList();
+
+            return gaVersions.Max() ?? throw new Exception("No GA product versions found.");
```
lbussell, line 94: This could have a more descriptive error message.
```suggestion
            return gaVersions.Max()
                ?? throw new Exception($"Did not find any GA product versions for repo ${repo}");
```

tests/Microsoft.DotNet.Docker.Tests/Manifest.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Microsoft.DotNet.Docker.Tests
+{
+    // Simplified version of the ImageBuilder manifest model
+    // See https://github.com/dotnet/docker-tools/tree/main/src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest
+    public class Manifest
+    {
+        public List<Repo> Repos { get; set; } = new List<Repo>();
+    }
+
+    public class Repo
```
lbussell, line 16: Consider using Records for these model classes, as they have automatically implemented ToString() methods - helps with logging and debugging.

```suggestion
    public record Repo
```

tests/Microsoft.DotNet.Docker.Tests/ManifestHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            string variablePattern = @"\$\(.+\)";
+            var match = Regex.Match(input, variablePattern);
+            if (!match.Success)
+            {
+                return input;
+            }
+            string variableName = match.Value.Replace("$(", string.Empty).Replace(")", string.Empty);
+            string remainingString = input.Split(match.Value)[1];
+            return Config.GetVariableValue(variableName) + remainingString;
```
lbussell, line 40: This only matches tags of the pattern `$(foo)*`. We should be free to create tags using variables at any point in the string, such as `$(foo)-$(bar)-$(baz)-extra`. Consider iterating over each variable regex match when replacing values here.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -0,0 +1,100 @@
...
+                    .Select(GetMajorVersion)
+                    .Distinct()
+                    .ToList();
+
+                return majorVersions.Max() ?? throw new Exception("No latest product versions found.");
+            }
+
+            // Use the latest GA major version on the main branch
+            List<string> gaVersions = productVersions
+                .Where(version => !version.Contains('-'))
```
lbussell, line 89: I think the comment should point out that this is just a heuristic for finding the latest GA major version. We are assuming that GA versions will look like `9.0` and non-GA versions will look like `9.0-preview.7` or `9.0-rc.1`. That assumption is Good Enough but it should be noted here.

lbussell (changes requested):

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
ellahathaway: The exception is no longer needed after converting the function to return an int.

tests/Microsoft.DotNet.Docker.Tests/Manifest.cs
ellahathaway: I tried this, and Json deserialization did not work as expected. After doing a little research, I learned that I'd have to write additional constructors in order for the deserialization to work. That feels a bit unnecessary, so I'm choosing to leave them as classes.

PR #5860, opened 2024-09-06
[nightly] Dynamically generate Dockerfiles for test scenarios

lbussell:
Port of #5832 to nightly


tests/Microsoft.DotNet.Docker.Tests/TestScenarios/WebApiAotScenario.cs
mthalman, line 1: I missed this on the original review. This file should be deleted.

mthalman (approved):

PR #5878, opened 2024-09-11
Schema Tag Tests

ellahathaway:
Introduces the second half of the static tag tests

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
@@ -5,7 +5,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
+using FluentAssertions;
 using Xunit;
 
 #nullable enable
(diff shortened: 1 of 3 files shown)


tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -15,37 +17,131 @@ namespace Microsoft.DotNet.Docker.Tests
...
+            Alpine,
+            Version
+        }
+        private const string AlpineOs = "alpine";
         private const string LatestTagValue = "latest";
+        private const string SingleNumberRegex = @"\d+";
+        private const string DoubleNumberRegex = @$"{SingleNumberRegex}\.{SingleNumberRegex}";
+        private static readonly string[] ApplianceRepos = { "monitor", "monitor-base", "aspire-dashboard" };
+
+        public static IEnumerable<object[]> GetTagTestObjects(TestType testType)
```
ellahathaway, line 33: I may have over engineered this.

For context, I wanted the Platform, Alpine, and Version tag tests to be run on each type of tag format that was being tested so that they functioned as "unit tests" for each type of tag format. This allows for easy diagnostics of what causes the test(s) to fail.

In an effort to accomplish this "unit test" type structure, I parsed the input logic out into individual functions that created the objects for the tests, and when I realized how similar the logic was between the different types of tests, I created the TestType enum and the functions that switch based on the type of test being run.

I feel that there is a tradeoff between the maintainability of the tests vs the dev experience when the tests fail, and I am unsure if I have teetered too far in the direction of favoring the dev experience. Input on this perspective would be helpful.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -5,7 +5,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
+using FluentAssertions;
```
ellahathaway, line 10: Added this package. I wanted the feedback when tests fail to be clear, and I found that this package helped achieve this.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -15,37 +17,131 @@ namespace Microsoft.DotNet.Docker.Tests
...
+                                GetTagTestInput(testType, repo, 1, true, false, IsApplianceVersionUsingNewSchema),    // <Major>-<os>
+                            });
 
-        public static IEnumerable<object[]> GetRepoObjects() =>
-            ManifestHelper.GetManifest().Repos.Select(repo => new object[] { repo }).ToArray();
+                            if (!repo.Name.Contains("monitor"))
+                            {
+                                // Monitor repos don't have these tags
+                                testObjects.Add(GetTagTestInput(testType, repo, 1, false, true, IsApplianceVersionUsingOldSchema)); // <Major>-<architecture>
+                            }
```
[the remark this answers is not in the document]
lbussell: Yes - IIRC .NET Monitor wants to explicitly make moving between minor versions intentional.

/cc @jander-msft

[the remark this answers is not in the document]
lbussell: Aspire on the other hand has only one `major.minor` version supported at a time.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
ellahathaway: > Can you explain what you mean by "when tests fail to be clear"?

Yes. The platform, alpine, and version tag tests iterate of a series of dockerfiles in a given repo, so when a test fails, it will fail for a specific dockerfile. I found that when using xunit assertions (eg `Assert.Single(tag)`), the test would fail but it was unclear which dockerfile it failed for.

> Is it just the extra output messages when the assertion fails?

Yes. An alternative was to use `Assert.True(<check>, "error message")`, but based on comments left on https://github.com/dotnet/dotnet-docker/pull/5852, I was under the assumption that it would be better to use `Assert.Single()` over `Assert.True()`. Given this constraint, I found that fluent assertions allowed me to check for the existence of a single tag while still outputting a descriptive error message.

lbussell: It sits atop xUnit (and other test libraries) and provides more flexible assertions and better failure messages. e.g.

```cs
List<int> numbers = [1, 2, 3];
numbers.Should().Contain(item => item > 3, $"at least one item should be larger than 3");
// Expected numbers to have an item matching (item > 3) because at least 1 item should be larger than 3
```

```cs
string accountNumber = "1234567890";
accountNumber.Should().Be("0987654321");
// Expected accountNumber to be “0987654321”, but “1234567890” differs near “123” (index 0).
```

[source](https://fluentassertions.com/about/#why)

Maybe not strictly the most useful for testing equality and single items in collections like in these cases. But seems to have better semantics for multi-item collections that could be useful elsewhere - https://fluentassertions.com/collections/

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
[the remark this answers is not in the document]
mthalman: This is specifically related to the `<major>-<arch>` pattern. Monitor already has a `9` tag but no `9-amd64` tag. But Aspire has both `8` and `8-amd64`.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -310,12 +259,288 @@ public void VersionTag_SameOsAndVersion(Repo repo, int versionParts)
...
+                        if (ApplianceRepos.Any(repo.Name.Contains))
+                        {
+                            // Only appliance repos have major version tags
+                            // Only appliance repos have floating distro tags (<cref="IsApplianceVersionUsingOldSchema"/>)
+                            testObjects.AddRange(new[]
+                            {
+                                GetTagTestInput(
+                                    testType, 
+                                    repo, 
+                                    versionType: VersionType.MajorMinorPatch,
```
mthalman, line 283: To reduce some of the verbosity, `versionType` isn't really necessary now since the change from `int` to `VersionType`. The `VersionType` enum name already indicates what is being passed here.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -310,12 +259,288 @@ public void VersionTag_SameOsAndVersion(Repo repo, int versionParts)
...
+        {
+            switch (testType)
+            {
+                case TestType.Latest:
+                    return [ repo ];
+
+                case TestType.Platform:
+                    if (versionType == null || checkOs == null || checkArchitecture == null)
+                    {
+                        throw new ArgumentException("'versionType', 'checkOs' and 'checkArchitecture' must be specified", nameof(testType));
```
mthalman, line 510: Follow this pattern for the other exceptions below too.

```suggestion
                        throw new ArgumentException($"'{nameof(versionType)}', '{nameof(checkOs)}' and '{nameof(checkArchitecture)}' must be specified", nameof(testType));
```

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -344,40 +569,45 @@ private static bool IsExpectedMajorMinorVersion(Repo repo, string version)
                     // Use the latest major version on the nightly branch
                     return group;
                 })
-                .Select(group => group.Select(GetMajorMinorVersion).Max())
-                .Distinct()
+                .Select(group => group.Select(version => 
+                {
+                    Version parsedVersion = GetVersion(version);
+                    return double.Parse($"{parsedVersion.Major}.{parsedVersion.Minor}");
```
mthalman, line 575: We should be able to avoid the use of `double` here, right? Just return `Version` instead?

mthalman (approved):

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -15,87 +17,640 @@ namespace Microsoft.DotNet.Docker.Tests
...
+                    .Select(dockerfile => dockerfile.Os)
+                    .Distinct()
+                    .ToList();
+
+                dockerfileOses.Should().ContainSingle(
+                    "all dockerfiles for tag " + tag + " should have the same os");
             }
+        }
+
+        private static IEnumerable<object[]> GetTagTestObjects(TestType testType)
```
lbussell, line 262: If possible, I think that I would prefer having a separate `Get*TestData` method per each TestType that you have here - rather than having two separate methods that both have a switch on the TestType. Is there any benefit to having them combined as they are now?

i.e.
```cs
private static IEnumerable<object[]> GetLatestTagsTestData() => [ ... ];
private static IEnumerable<object[]> GetPlatformTagsTestData() => [ ... ];
etc.
```

ellahathaway: > Is there any benefit to having them combined as they are now?

The main benefit is primarily not having to have duplicate code for the loops iterating over the repos. If we're okay with more code duplication, then I can adjust this. Let me know.

lbussell: Chatted offline about some other options for this - the alternatives didn't seem to be different enough to warrant re-organizing these immediately, and the tests provide value as-is - so looks good to me.

lbussell (approved):

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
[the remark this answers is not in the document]
jander-msft: It looks like `dotnet/monitor` and `dotnet/monitor-base` is missing the following tag patterns:
- `<major>.<minor>-<arch>` (for everything before 9)
- `<major>-<arch>` (for all versions)

I think it makes sense to have these added, at least from a consistency standpoint.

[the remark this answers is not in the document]
jander-msft: On the other hand, the introduction of the `<major>.<minor>-<arch>` and `<major>-<arch>` tag schemes was largely only done for the Aspire dashboard. Take the `dotnet/runtime` repo as an example; while it doesn't contain major-only tags, there are no `<major>.<minor>-<arch>` tags. However, there are `<major>.<minor>-<distro>-<arch>` tags. The .NET Monitor images are, by-in-large, consistent with these tag schemes in the base .NET images whereas the Aspire images have introduced a new pattern.

We'll have to investigate the reasoning as to why the `<major>.<minor>-<arch>` were never added to the base .NET images and see if the same reasoning should apply to the appliance-style images. I'll open an issue to document the inconsistency and allow discussion separate from this PR on what should be done to rectify the inconsistency.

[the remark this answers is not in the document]
ellahathaway: >I'll open an issue to document the inconsistency and allow discussion separate from this PR on what should be done to rectify the inconsistency.

This is a great path forward. Thank you for filing the issue and starting this discussion.

I spoke with the team offline, and the decision is to go ahead with merging the changes in this PR. These tests can then be adjusted later given the future discussion on the issue.

I'll leave this comment thread open as well.

Issue #5850, opened 2024-08-30
Support shipping multiple versions of .NET Samples


lbussell:
We should have support for shipping multiple versions of .NET Samples at the same time. This would allow us to develop samples for new .NET features farther in advance while still providing stable documentation for supported/LTS versions of .NET.

Related: #4932

lbussell:
Completed with https://github.com/dotnet/dotnet-docker/pull/5882

Issue #5798, opened 2024-08-15
.NET 9 Documentation and Samples Updates


lbussell:
This issue is an epic for tracking the major work to documentation and samples before .NET 9 GA.

**Out of scope for this epic**: 

- https://github.com/dotnet/dotnet-docker/issues/5113
- #5126

lbussell:
Closing in favor of https://github.com/dotnet/dotnet-docker/issues/5978 and https://github.com/dotnet/dotnet-docker/issues/5979 since the two issues aren't really related.

Issue #5702, opened 2024-07-17
Add sample Dockerfiles for Azure Linux


lbussell:
Azure Linux images are [publicly supported](https://github.com/dotnet/dotnet-docker/discussions/5022) and now much of the [documentation has been moved to this repo](https://github.com/dotnet/dotnet-docker/issues/5031). One of the last things missing for Azure Linux support is example Dockerfiles for our various samples.

lbussell:
Completed with https://github.com/dotnet/dotnet-docker/pull/5882

Issue #5402, opened 2024-04-25
Heredocs (<<EOF) syntax in samples doesn't work on Windows when using default Git line ending settings


lbussell:
https://github.com/dotnet/dotnet-docker/blob/3e8c396639c30387237fc898d9152858b479a458/samples/releasesapi/Dockerfile.debian-cross-arm64-x64#L6-L10

1. This syntax is only used in the releasesapi samples and isn't consistent with our other samples.
2. This syntax does not work on Windows when using Git's default "checkout with CRLF, commit LF" setting.

Proposal: Remove heredocs in samples or provide a note about this in the samples README.

Workaround: Switch the file to LF line endings using your editor before building the sample.

richlander:
I'm happy with any option. I noticed that this worked poorly on Windows when I made it was too stubborn to relent. Such surprising behavior for me.

lbussell:
Here's the issue for that - https://github.com/moby/buildkit/issues/4282

polarathene:
<details>
<summary>Collapsed as doesn't contribute much to the discussion</summary>

> 2\. This syntax does not work on Windows when using Git's default "checkout with CRLF, commit LF" setting.

> Workaround: Switch the file to LF line endings using your editor before building the sample.

> Another option is to specify the line ending for these files specifically with the `.gitattributes` file
> 
> However, this could blow up if a user copies this to their repo and doesn't include the gitattributes. Since this doesn't provide a ton of value, we should just remove the Heredocs.

FWIW, [`.gitattributes` has been used on projects I maintain](https://github.com/docker-mailserver/docker-mailserver/blob/94751e00c9787c5223e39bd0cd5b139a0c1f6255/.gitattributes#L23) to avoid the issue and it's been a non-problem.

That said:
- I'm probably one of the few that builds images for such projects on Windows, whereas your demographic may very much be more likely to encounter the concern.
- Copy/pasting examples (_where editors like VSCode would default to CRLF on Windows_) since there isn't as much reason to clone the repo for such, I can see why you'd want to drop the syntax (_as nice as it is_).

</details>

lbussell:
As you mentioned, @polarathene, the primary purpose of the samples is to be easy to run and understand. A syntax that depends on a specific line ending to work is a no-go.

These have been removed from the samples with https://github.com/dotnet/dotnet-docker/pull/5882, so I'm closing the issue as complete.

PR #5882, opened 2024-09-12
Update samples to .NET 9

lbussell:
### Related
- #5798 

### Updates
- [x] https://github.com/dotnet/dotnet-docker/issues/5057
  - [x] dotnetapp
  - [x] aspnetapp
  - [x] complexapp
  - [x] releasesapi
  - [x] releasesapp
  - [x] globalapp
- [x] Added sample for self-contained publish
- [x] Clean up non-shipping Dockerfiles and documentation under `samples/8.0`
- [x] https://github.com/dotnet/dotnet-docker/issues/5702
- [x] https://github.com/dotnet/dotnet-docker/issues/5850
- [x] https://github.com/dotnet/dotnet-docker/issues/5402


lbussell:
I'll defer #5113 to a follow-up PR since this PR is big enough already - this is ready for review.

samples/aspnetapp/Dockerfile.self-contained
```diff
@@ -0,0 +1,23 @@
...
+ARG TARGETARCH
+WORKDIR /source
+
+# Copy project file and restore as distinct layers
+COPY --link aspnetapp/*.csproj .
+RUN dotnet restore -r linux-$TARGETARCH
+
+# Copy source code and publish app
+COPY --link aspnetapp/. .
+RUN dotnet publish --self-contained -r linux-$TARGETARCH --no-restore -o /app
```
mthalman, line 13: It still works like this, doesn't it? Otherwise, can we use `--use-current-runtime` instead?

```suggestion
RUN dotnet publish --self-contained -a $TARGETARCH --no-restore -o /app
```

samples/aspnetapp/README.md
```diff
@@ -90,7 +130,7 @@ dotnet publish /p:PublishProfile=DefaultContainer
 That command can be further customized to use a different base image and publish to a container registry. You must first use `docker login` to login to the registry.
 
 ```console
-dotnet publish /p:PublishProfile=DefaultContainer /p:ContainerBaseImage=mcr.microsoft.com/dotnet/aspnet:8.0-jammy-chiseled /p:ContainerRegistry=docker.io /p:ContainerRepository=youraccount/aspnetapp
+dotnet publish /p:PublishProfile=DefaultContainer /p:ContainerBaseImage=mcr.microsoft.com/dotnet/nightly/aspnet:9.0-noble-chiseled /p:ContainerRegistry=docker.io /p:ContainerRepository=youraccount/aspnetapp
```
mthalman, line 133:
```suggestion
dotnet publish /p:PublishProfile=DefaultContainer /p:ContainerBaseImage=mcr.microsoft.com/dotnet/aspnet:9.0-noble-chiseled /p:ContainerRegistry=docker.io /p:ContainerRepository=youraccount/aspnetapp
```

samples/dotnetapp/Dockerfile.nanoserver
```diff
@@ -2,19 +2,19 @@
...
-FROM mcr.microsoft.com/dotnet/sdk:8.0-nanoserver-$TAG AS build
+FROM mcr.microsoft.com/dotnet/sdk:9.0-nanoserver-$TAG AS build
 WORKDIR /source
 
-# copy csproj and restore as distinct layers
-COPY *.csproj .
-RUN dotnet restore --ucr
+# Copy project file and restore as distinct layers
+COPY --link *.csproj .
+RUN dotnet restore --use-current-runtime
```
mthalman, line 10: Why is `--use-current-runtime` needed?

samples/globalapp/README.md
```diff
@@ -42,66 +42,66 @@ Etc/UTC
...
-$ docker run --rm -it -e TZ="America/Los_Angeles" app
+$ docker build --pull -t globalapp .
+$ docker run --rm -it -e TZ="America/Los_Angeles" globalapp
 Hello, World!
 
 ****Print baseline timezones**
-Utc: (UTC) Coordinated Universal Time; 2/14/2023 12:09:26AM
-Local: (UTC-08:00) Pacific Time (Los Angeles); 2/13/2023 4:09:26PM
+Utc: (UTC) Coordinated Universal Time; 09/06/2024 21:55:49
+Local: (UTC) Coordinated Universal Time; 09/06/2024 21:55:49
```
mthalman, line 51: Since the container was configured to run with Pacific time, shouldn't this output reflect that time zone?

That's what this comment suggests:

https://github.com/dotnet/dotnet-docker/blob/a4a40f1519b6bbf8008dab3dff94a36dd18698db/samples/globalapp/Program.cs#L13-L15

samples/releasesapi/Dockerfile.debian
```diff
@@ -1,28 +1,27 @@
...
+FROM mcr.microsoft.com/dotnet/sdk:9.0 AS build
 ARG TARGETARCH
-RUN <<EOF
-apt-get update
-apt-get install -y clang zlib1g-dev
-EOF
+RUN apt-get update \
+    && apt-get install -y \
+        clang \
+        zlib1g-dev
```
mthalman, line 9:
```suggestion
        zlib1g-dev
    && rm -rf /var/lib/apt/lists/*
```

samples/releasesapi/Dockerfile.debian-cross-arm64-x64
```diff
@@ -1,29 +1,32 @@
...
-apt-get install -y clang zlib1g-dev zlib1g-dev:amd64 gcc-x86-64-linux-gnu llvm
-EOF
+RUN dpkg --add-architecture amd64 \
+    && apt-get update \
+    && apt-get install -y \
+        clang \
+        gcc-x86-64-linux-gnu \
+        llvm \
+        zlib1g-dev \
+        zlib1g-dev:amd64
```
mthalman, line 12:
```suggestion
        zlib1g-dev:amd64
    && rm -rf /var/lib/apt/lists/*
```

samples/releasesapi/Dockerfile.debian-cross-x64-arm64
```diff
@@ -1,29 +1,31 @@
...
-apt-get install -y clang zlib1g-dev zlib1g-dev:arm64 gcc-aarch64-linux-gnu llvm
-EOF
+RUN dpkg --add-architecture arm64 \
+    && apt-get update \
+    && apt-get install -y \
+        clang \
+        gcc-aarch64-linux-gnu \
+        llvm \
+        zlib1g-dev \
+        zlib1g-dev:arm64
```
mthalman, line 13:
```suggestion
        zlib1g-dev:arm64
    && rm -rf /var/lib/apt/lists/*
```

samples/releasesapi/nuget.config
```diff
@@ -1,7 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <clear />
```
mthalman, line 3: Removing this will trigger an alert. All NuGet.config files need to have a `<clear />` element to avoid implicit feed to nuget.org.

samples/aspnetapp/Dockerfile.self-contained
lbussell: Oh, right. So using `-a` doesn't work with --self-contained, unless you enable `PublishReadyToRun`. This is https://github.com/dotnet/sdk/issues/36334.

This is why I put the commented-out properties in the aspnetapp.csproj file. https://github.com/dotnet/dotnet-docker/blob/dddd8f0a53cda994b91cb50a2dfc0a9651399c12/samples/aspnetapp/aspnetapp/aspnetapp.csproj#L9-L14

If I add a comment here referring to those, it should be OK to use the `-a` option instead of `-r`.

Another option is to set the properties directly in the Dockerfile (just tested this):

```Dockerfile
# Copy project file and restore as distinct layers
COPY --link aspnetapp/*.csproj .
RUN dotnet restore -a $TARGETARCH \
    -p:PublishReadyToRun=true \
    -p:PublishSingleFile=true

# Copy source code and publish app
COPY --link aspnetapp/. .
RUN dotnet publish -a $TARGETARCH --no-restore -o /app \
    --self-contained \
    -p:PublishReadyToRun=true \
    -p:PublishSingleFile=true
```

I think if we did that, we'd need to add a comment recommending putting these in the project file instead. But that would let the Dockerfile build without modifying the project file.

mthalman: I think this was why @richlander had removed the self-contained sample in the first place. Because we wanted to have publishing configuration defined in the project file. But that made it not easily reusable between self-contained and framework-dependent.

samples/releasesapi/nuget.config
lbussell: I removed it because for the purposes of this sample, users probably won't want to use our nuget.org mirror - since it doesn't have all packages and others can be several versions behind.

lbussell: We determined with https://github.com/dotnet/dotnet-docker/issues/5787 that using NuGet.org for the purpose of validating user scenarios is acceptable.

samples/dotnetapp/Dockerfile.nanoserver
lbussell: Turns out it's not needed.

It was changed from `-r win-x64` to `--ucr` in https://github.com/dotnet/dotnet-docker/pull/4742. I needed to look up what `--ucr` meant, so I changed it to the longform version so it was easier to read.

But I just tested the Dockerfile without any RID arguments and it works fine. So I'll remove them from Windows Dockerfiles.

samples/globalapp/README.md
lbussell: Good catch. I forgot to run the container with the env variable. I updated all of the outputs.

samples/aspnetapp/Dockerfile.self-contained
mthalman: I see you re-requested review but this isn't resolved yet. Using `-r linux` isn't preferred because it wouldn't translate to Alpine without modification. Is `--use-current-runtime` an option?

lbussell: It seems like `--ucr` works in the same scenarios as `-a $TARGETARCH` - it still needs `ReadyToRun` to be set to true. I'm OK with dropping the self-contained sample since we do still have `releasesapp` for that.

mthalman: Yeah, I think that's probably the way to go for now.

lbussell:
I'm working on getting the manifest.samples.json to work with the build matrix generation now. It's creating some duplicate legs since we've got 8.0 and 9.0 samples now.

@mthalman, @richlander, @MichaelSimons, any other feedback on the actual samples content?

samples/releasesapp/releasesapp.csproj
```diff
@@ -2,14 +2,21 @@
...
-    <Nullable>enable</Nullable>
+    <EnableSdkContainerSupport>true</EnableSdkContainerSupport>
+
+    <!--
+      Learn more about publishing configuration:
+      https://learn.microsoft.com/dotnet/core/deploying/#publish-self-contained
+      https://learn.microsoft.com/dotnet/core/deploying/single-file
+      https://learn.microsoft.com/dotnet/core/deploying/trimming/trim-self-contained
+      https://learn.microsoft.com/dotnet/core/deploying/ready-to-run
+    -->
```
mthalman, line 15: I think it would provide better clarity if these links were placed next to their corresponding MSBuild property.

lbussell: Done. I kept the "learn more about..." text for each line. But I could remove it since it does look a little repetitive.

lbussell:
The markdown link failures are due to https://github.com/dotnet/dotnet-docker/issues/5872#issuecomment-2369097600 (not an actual issue with this PR). They would be fixed by https://github.com/dotnet/docker-tools/pull/1446.

mthalman (approved):

PR #5921, opened 2024-09-26
Update documentation to reduce the number of hub.docker.com references

MichaelSimons:
This is being done as a result of https://github.com/dotnet/dotnet-docker/pull/5585 where the Docker Hub readmes were exceeding the size limit and as a result the tags tables were removed.  The GitHub readmes are now the preferred location to view the tags table therefore the various hub.docker.com references were updated to GitHub.  There are still around 6 references to Docker Hub that should remain - e.g. the example that illustrates pushing images to https://hub.docker.com

This is related to https://github.com/dotnet/docker-tools/pull/1446


lbussell (approved):
Thank you!

PR #5924, opened 2024-09-26
Update supported OS documentation

### LINKED ISSUE #5756: Document policy on when support is added for new OS versions
The [Supported Platforms documentation](https://github.com/dotnet/dotnet-docker/blob/main/documentation/supported-platforms.md) describes what Linux distros and Windows versions are supported as well as what new versions will have support added for.  The document is not specific however on **when** the new versions will appear.  For example the document states the following for Ubuntu:

> As new LTS versions are released, support is added to the latest .NET version and latest LTS (if they differ).

Is support added during the next .NET servicing cycle or within "n" days of the OS release date?  What is the SLA?

Capturing these specifics will guide the expectations of customers and will be the basis for them to provide feedback on.  It will also instruct the development team on how to operate.

lbussell:
Fixes #5756

### DIFF
--- a/documentation/supported-platforms.md
+++ b/documentation/supported-platforms.md
@@ -6,6 +6,11 @@ This document describes the platforms (OS and architectures) supported by the of
 
 .NET supports [a broad set of operating systems and versions](https://github.com/dotnet/core/blob/master/os-lifecycle-policy.md). When producing container images, it’s impractical to support the full matrix of OS, arch, and .NET version combinations. In practice, images are produced for a select set of operating systems and versions. If official .NET container images aren't provided for your preferred OS, [let us know by opening a discussion](https://github.com/dotnet/dotnet-docker/discussions). Alternatively, you can [author your own .NET images](scenarios/installing-dotnet.md).
 
+New versions of operating systems that we support are added on a best-effort basis, typically within one month.
+New OSes are always avialable in the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md) first, and then are usually added to the officially supported repos with the next .NET servicing release.
+Occasionally, depending on the specific timing of OS releases and .NET releases, we may add new OSes to the officially supported repos in the middle of a servicing cycle.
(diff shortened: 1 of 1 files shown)


documentation/supported-platforms.md
```diff
@@ -6,6 +6,11 @@ This document describes the platforms (OS and architectures) supported by the of
 
 .NET supports [a broad set of operating systems and versions](https://github.com/dotnet/core/blob/master/os-lifecycle-policy.md). When producing container images, it’s impractical to support the full matrix of OS, arch, and .NET version combinations. In practice, images are produced for a select set of operating systems and versions. If official .NET container images aren't provided for your preferred OS, [let us know by opening a discussion](https://github.com/dotnet/dotnet-docker/discussions). Alternatively, you can [author your own .NET images](scenarios/installing-dotnet.md).
 
+New versions of operating systems that we support are added on a best-effort basis, typically within one month.
+New OSes are always avialable in the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md) first, and then are usually added to the officially supported repos with the next .NET servicing release.
+Occasionally, depending on the specific timing of OS releases and .NET releases, we may add new OSes to the officially supported repos in the middle of a servicing cycle.
```
mthalman, line 11: I'd prefer it if we could have a crisp definition of this policy rather than make ad hoc decisions.

lbussell: If we remove this clause, are we OK with committing to only releasing new OSes on Patch Tuesdays?

Another thought - what if .NET servicing skips a month as well? Would we want to be up to 2 months behind (the worst-case scenario) for a new Ubuntu or Alpine version? If the intent is to reduce "out of band" work by our team then perhaps the current wording should be updated to "Patch Tuesdays" instead of ".NET Servicing releases.

/cc @richlander

richlander: I'm not sure we need this level of detail. Out-of-band updates are important/impactful for existing tags. They are not important for new tags.

mthalman (approved):

mthalman (approved):

PR #5925, opened 2024-09-27
Switch to using linkspector for validating markdown links

MichaelSimons:


documentation/vulnerability-reporting.md
```diff
@@ -120,7 +120,18 @@ Rerun the scan of your image using your scanning tool. Ensure you get the latest
 When .NET drops support for an image tag, it means it will no longer be updated, even when there is a new base OS image available.
 This means that vulnerabilities will be reported for that image over time if it continues to be used.
 Our [supported tag policy](supported-tags.md) provides detailed information about when these tags are no longer supported.
-The simple rule to follow: only the tags shown in our [tag listing](supported-tags.md#tag-listing) are supported.
+The simple rule to follow: only the tags shown in our tag listing are supported:
+
+Complete tag lists:
```
MichaelSimons, line 125: The supported-tags opening description includes these links but I felt including the tag listing links here provides a better ux.

lbussell: If this is a "complete" list then should it include monitor-base?

MichaelSimons: It should.  I also updated the vulnerability document where this list was copied from.

mthalman (approved):

.github/linters/.linkspector.yml
```diff
@@ -0,0 +1,12 @@
+aliveStatusCodes:
+  - 200
+  - 406
+dirs:
+  - .
```
lbussell, line 5: Does this recursively scan markdown files in e.g. the documentation directory? If so, then why is the additional `.github` dir needed below?

MichaelSimons: It does scan recursively but hidden directories are ignored by default.  It seemed valuable to scan .github which is why I added it.  I couldn't find documentation for this, I found this by the behavior I saw and then checked the implementation to confirm this appears to be expected.

lbussell (approved):

Issue #5057, opened 2023-12-12
Update samples to .NET 9


lbussell:
- [ ] Consider creating new samples to feature new .NET or .NET Docker features

lbussell:
This should include https://github.com/dotnet/dotnet-docker/issues/5702

lbussell:
Completed with https://github.com/dotnet/dotnet-docker/pull/5882

PR #5926, opened 2024-09-27
Rewrite complexapp test documentation

lbussell:
What I did:

- Consolidated test documentation into run-tests-in-container.md instead of splitting it between there and complexapp/README.md
- Added section on running tests with Docker build using BuildKit's new(-ish) build output feature
- Re-ran example commands to get new output for .NET 9
- Removed the strict recommendation for running tests as an executable stage over running tests as part of Docker build. With BuildKit, they seem about equal to me.

Potential for improvement:

- I have not played with getting test output from the Docker build without buildkit. Theoretically you could build the test-output stage and then start the container and copy the test results out. However, if you don't have BuildKit you probably don't want to use this approach anyways. Since without BuildKit, the Docker builder won't skip stages in a multi-stage build.


samples/complexapp/Dockerfile
mthalman, line 1: I would like to see just one testing pattern shown in this Dockerfile. Stuffing multiple patterns into it makes it overly complex and I think will cause confusion to people who are just trying to find a simple way to enable testing. They'd have to figure out what is relevant and trim out the other pattern that's not needed.

samples/complexapp/tests/StringLibraryTests.cs
```diff
@@ -12,4 +12,10 @@ public void InputString_ReturnsCorrectString()
         var returnedString = StringLibrary.GetString();
         Assert.True(inputString == returnedString, "The input string was not correct.");
     }
+
+    // [Fact]
+    // public void MyTest()
+    // {
+    //     Assert.True(false);
+    // }
```
mthalman, line 20: I assume this was inadvertent?

samples/run-tests-in-sdk-container.md
mthalman, line 1: I honestly think we should pick one testing pattern and document that. Yes, there are different ways to do things but none of them are specific to .NET and so I don't see them as being relevant to enumerate here.

mthalman (approved):

mthalman (commented):
Didn't mean to approve yet. Changing my status.

samples/run-tests-in-sdk-container.md
lbussell: I think I generally agree. My intention here was to "replace" the old documentation that no longer worked with BuildKit, which was removed with #5057. If we want to have just one example, it should be the most compatible one which I think would be the "Entrypoint" pattern again.

lbussell: Attempted to clean this up. The document is much less complex now.

Issue #5635, opened 2024-07-01
Add support for Server 2025


MichaelSimons:
# New Windows Release

Windows version: Server 2025

Related to https://github.com/dotnet/core/issues/9374

## Main Branch Tasks

1. - [x] Ensure a ["New Windows Release" issue](https://github.com/dotnet/docker-tools/blob/main/.github/ISSUE_TEMPLATE/releases/new-windows-release.md) exists for docker-tools repo
1. - [x] Copy the Dockerfiles of the most recent published Windows version for all supported .NET versions and place them in a version-specific folder under their respective variants (runtime, aspnet, sdk)
1. - [x] Modify the Dockerfile templates if there are any specific changes needed that are related to the new Windows version
1. - [x] Update [manifest.json](https://github.com/dotnet/dotnet-docker/blob/nightly/manifest.json) to reference the new set of Dockerfiles with the appropriate tags
1. - [x] Update the [test data](https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/TestData.cs) to include the new Windows version
1. - [x] Update the [tags metadata templates](https://github.com/dotnet/dotnet-docker/tree/main/eng/mcr-tags-metadata-templates) to include the new Windows version
1. - [x] Run the command to update the READMEs: `.\eng\readme-templates\Get-GeneratedReadmes.ps1`
1. - [x] Run the command to update the image size baseline file: `.\tests\performance\Validate-ImageSize.ps1 -UpdateBaselines`
1. - [x] Inspect generated changes for correctness
1. - [ ] Test the images
      1. - [ ] Create a local VM of the new Windows version
      1. - [ ] Clone this repo with the above changes onto the VM
      1. - [ ] Run `.\build-and-test.ps1 -OS nanoserver-<VERSION>` to build and test your changes for Nano Server
      1. - [ ] Run `.\build-and-test.ps1 -OS windowsservercore-<VERSION>` to build and test your changes for Windows Server Core
1. - [x] Create PR
1. - [ ] Get PR signoff
1. - [ ] Merge PR and build/publish as part of the main branch [release process](dotnet-release.md) for the next .NET release
1. - [ ] Create an announcement (example: [Nano Server, version 20H2](https://github.com/dotnet/dotnet-docker/issues/2322))
1. - [ ] Update the samples to reference the new Windows version:
[...]

PR #5944, opened 2024-10-02
Add Windows Server 2025 Dockerfiles

lbussell:
This is the first part of #5635.

Since I'm not running Server 2025 on my desktop, running in CI is the easiest way to test these images.


lbussell:
It looks like globalization related functions are working in Nano Server 2025 without explicitly installing the FoD. I'll investigate this some more.

lbussell:
@dotnet/dotnet-docker-reviewers, since this is for `nightly` and all tests are passing, why don't we ship these images now? That may make it easier to validate different scenarios for https://github.com/dotnet/dotnet-docker/issues/6034.

eng/dockerfile-templates/aspnet/Dockerfile.windows
```diff
@@ -1,20 +1,28 @@
...
-    set tagVersion to when(dotnetVersion = "6.0" || dotnetVersion = "8.0",
-        VARIABLES[cat("dotnet|", dotnetVersion, "|product-version")]
-        VARIABLES[cat("dotnet|", dotnetVersion, "|fixed-tag")]) ^
-    set runtimeBaseTag to cat("$REPO:", tagVersion, "-", OS_VERSION)
+
+    set repoBaseUrl to "mcr.microsoft.com/windows" ^
+    set installerStageRepo to cat(repoBaseUrl, "/servercore") ^
+    set installerStageTag to cat(OS_VERSION_NUMBER, "-amd64") ^
+
+    set installerStageFromImage to cat(installerStageRepo, ":", installerStageTag) ^
```
mthalman, line 13: Most of these variables don't seem to provide any value. This can be simplified:

```suggestion
    set installerStageFromImage to cat("mcr.microsoft.com/windows/servercore:", OS_VERSION_NUMBER, "-amd64") ^
```

This comment applies to the other image type templates as well.

eng/pipelines/stages/build-test-publish-repo.yml
```diff
@@ -27,6 +27,11 @@ stages:
     - template: /eng/pipelines/steps/set-public-source-branch-var.yml@self
     customBuildInitSteps:
     - template: /eng/pipelines/steps/set-public-source-branch-var.yml@self
+    - powershell: |
+        if ($PSVersionTable.PSEdition -ne "Core") {
+          systeminfo
+        }
+      displayName: Check Windows system info
```
mthalman, line 34: I assume this was for debugging and not needed?

eng/dockerfile-templates/aspnet/Dockerfile.windows
lbussell: Yes these were used when Server 2025 was insider-only and had a different repo/tag. Fixed in https://github.com/dotnet/dotnet-docker/pull/5944/commits/9ba744d6a66be79e76c3c99e538a1049a1c22892.

eng/pipelines/stages/build-test-publish-repo.yml
lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/5944/commits/bf2afe10fe3fc4996262c729e62f75dc9d26e486

mthalman (approved):

Issue #5946, opened 2024-10-03
Undocument arch-specific tags on .NET appliance images


jander-msft:
### Describe the Problem

The appliance-style (`dotnet/monitor` and `dotnet/aspire-dashboard`) and sample (`dotnet/samples`) images have various tags for describing the product version, the architecture, and distro of the image. While distro information has been removed in newer releases, the architecture and product version information remains. The tag listing for these images can be intimidating due to the matrix of possibilities and may lead to choice paralysis or confusion as to which tag should be used.

Upon further reflection, the tag matrix can be greatly simplified if the arch-specific tags are removed or undocumented. An example simplification in this manner is described for the sample images in #5918. Using `dotnet/monitor:8` Ubuntu Chiseled as another example, the tag listing would be simplified from:

```
8.0.4-ubuntu-chiseled-amd64, 8.0-ubuntu-chiseled-amd64, 8-ubuntu-chiseled-amd64, 8.0.4-ubuntu-chiseled, 8.0-ubuntu-chiseled, 8-ubuntu-chiseled, 8.0.4, 8.0, 8, latest
```
to
```
8.0.4-ubuntu-chiseled, 8.0-ubuntu-chiseled, 8-ubuntu-chiseled, 8.0.4, 8.0, 8, latest
```

### Describe the Solution

- [x] For `dotnet/aspire-dashboard:8*` images, all arch-specific tags are to be undocumented to prevent breaking changes in current usage.
- [x] For `dotnet/monitor:9*` and lower version images, all arch-specific tags are to be undocumented to prevent breaking changes in current usage.
- [ ] For `dotnet/samples` images, all arch-specific tags are to be removed.

For `dotnet/monitor/base`, all arch-specific tags are to remain since this is a base image used for adding extensions upon it, which may require native dependencies or may have native entrypoints.

### Additional Context

The arch-specific tags are largely useful for images that are meant to be used as base images upon which derivative images are produced, especially if native dependencies are added in the deriving image. Given that the appliance-style and sample images are intended to be used as-is (not expected to be derived), the arch-specific tags do not provide much value. The assumption is that the vast majority use case is that users will use the manifest tags as-is without needing to further specify the platform. If needed, users can specify platform information on container tooling (such as `docker run --platform linux/amd64` or `FROM --platform linux/amd64` within Dockerfiles) to specify which architecture should be used.

jander-msft:
[...]

PR #5948, opened 2024-10-03
Undocument .NET Monitor 9 arch-specific tags

jander-msft:
As described in #5946, we should strive to remove the arch-specific tags for .NET Monitor 9. Due to technical limitations at this time, we must have at least one arch-specific tag for each image. Instead of removing them, these tags will be undocumented for .NET Monitor 9.


tests/performance/ImageSize.nightly.linux.json
```diff
@@ -379,9 +379,7 @@
     "src/monitor/8.1/ubuntu-chiseled/amd64": 127821834,
     "src/monitor/8.1/ubuntu-chiseled/arm64v8": 134958622,
     "src/monitor/8.1/cbl-mariner-distroless/amd64": 141200979,
-    "src/monitor/8.1/cbl-mariner-distroless/arm64v8": 148117492,
-    "src/monitor/9.0/azurelinux-distroless/amd64": 143287803,
```
jander-msft, line 383: @lbussell, we may need to make some type of adjustment to the image size tester to understand when an image doesn't have a tag pointing directly to it. The error from the test is:
```
-- Extra baseline images not defined in manifest:
src/monitor/9.0/azurelinux-distroless/amd64
    Expected:       143,287,803
    Variation Allowed: 136,123,413 - 150,452,193

src/monitor/9.0/azurelinux-distroless/arm64v8
    Expected:       149,490,964
    Variation Allowed: 142,016,416 - 156,965,512
```
I've removed these from the baseline for now.

lbussell: Ah... unforeseen consequences of this change. Per [Ingest image/layer size in Kusto (dotnet/docker-tools#745)](https://github.com/dotnet/docker-tools/issues/745) and [Remove Image Size commands and pipeline code (dotnet/docker-tools#1390)](https://github.com/dotnet/docker-tools/issues/1390), I propose we turn this check off (can be a separate issue/PR). It is not currently providing any value.

lbussell:
This clause in `PublishImageInfoAsync` is causing issues when there are no published platform-specific tags for an image: https://github.com/dotnet/docker-tools/blob/e36a91c00081ee416f4c0c1e5889172f7a53889f/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs#L194-L210

mthalman:
> This clause in `PublishImageInfoAsync` is causing issues when there are no published platform-specific tags for an image: https://github.com/dotnet/docker-tools/blob/e36a91c00081ee416f4c0c1e5889172f7a53889f/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs#L194-L210

This actually doesn't surprise me now that I think about it. We have no test coverage in any of the commands for a scenario without concrete tags. I'm willing to bet there will be other commands that will run into issues with this. For example, this line right here will fail too: https://github.com/dotnet/docker-tools/blob/e36a91c00081ee416f4c0c1e5889172f7a53889f/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs#L141. Before we move forward with this, we need more extensive testing in Image Builder. And the changes necessary to fix these issues are not small either.

lbussell:
> Before we move forward with this, we need more extensive testing in Image Builder.

To me that sounds too risky for October servicing. So what should we do in the meantime? Undocument these tags and re-evaluate if it's feasible for GA?

mthalman:
> > Before we move forward with this, we need more extensive testing in Image Builder.
> 
> To me that sounds too risky for October servicing. So what should we do in the meantime? Undocument these tags and re-evaluate if removing them completely is feasible for GA?

Yes, it makes sense to undocument for now. And really, we only require one arch-specific tag. My opinion is to get rid of `$(monitor|9.0|minor-tag)-<arch>` and undocument the other one. Then we'll have to figure out next steps. I'm not optimistic of having infra support by GA.

lbussell:
> My opinion is to get rid of `$(monitor|9.0|minor-tag)-<arch>` and undocument the other one. Then we'll have to figure out next steps. I'm not optimistic of having infra support by GA.

If we may not have infra support by GA, then perhaps we should undocument for 9.0 and remove for 10.0. That may be easier to explain to users than trying to explain why we only removed some tags but not all.

mthalman:
> > My opinion is to get rid of `$(monitor|9.0|minor-tag)-<arch>` and undocument the other one. Then we'll have to figure out next steps. I'm not optimistic of having infra support by GA.
> 
> If we may not have infra support by GA, then perhaps we should undocument for 9.0 and remove for 10.0. That may be easier to explain to users than trying to explain why we only removed some tags but not all.

Yeah, that's fair. So it comes down to how willing we are to invest in getting the infra to support this by GA. I would argue we've got better things to do during this period. And so simply undocumenting for 9.0 and target the removal for 10.0 would be the way to go.

jander-msft:
> > > Before we move forward with this, we need more extensive testing in Image Builder.
> > 
> > 
> > To me that sounds too risky for October servicing. So what should we do in the meantime? Undocument these tags and re-evaluate if removing them completely is feasible for GA?
> 
> Yes, it makes sense to undocument for now. And really, we only require one arch-specific tag. My opinion is to get rid of `$(monitor|9.0|minor-tag)-<arch>` and undocument the other one. Then we'll have to figure out next steps. I'm not optimistic of having infra support by GA.

I'll get to work on that right now. I am also running into similar issues for the samples as well as other issues. While it was good to try to squeeze this in, I think we need some more bake time to work through the issues.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -533,16 +533,24 @@ private static object[] GetTagTestInput(
...
+            (dockerfileInfo.Repo.Contains("monitor") && GetVersion(dockerfileInfo.MajorMinor).Major <= 8) ||
+            (dockerfileInfo.Repo.Contains("aspire") && GetVersion(dockerfileInfo.MajorMinor).Major <= 8);
+
+        private static bool IsApplianceVersionWithoutArchTags(ManifestHelper.DockerfileInfo dockerfileInfo) =>
+            !IsApplianceVersionWithArchTags(dockerfileInfo);
+
         // Certain versions of appliance repos use a new tag schema.
         // This new schema excludes the OS from all tags.
         // The aspire-dashboard repo uses this schema for all versions.
         // The monitor and monitor-base repos use this schema for versions 9 and above.
```
ellahathaway, line 547:
```suggestion
        // Certain versions of appliance-style images do not have distro-specific tags
        // For aspire-dashboard, this starts with 8.0
        // For monitor and monitor-base, this starts with 9.0
```

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -533,16 +533,24 @@ private static object[] GetTagTestInput(
         private static bool IsWindows(ManifestHelper.DockerfileInfo dockerfileInfo) =>
             dockerfileInfo.Os.Contains("windowsservercore") || dockerfileInfo.Os.Contains("nanoserver");
 
+        // All appliance-style images do not have arch-specific tags starting with version 9
+        private static bool IsApplianceVersionWithArchTags(ManifestHelper.DockerfileInfo dockerfileInfo) =>
+            (dockerfileInfo.Repo.Contains("monitor") && GetVersion(dockerfileInfo.MajorMinor).Major <= 8) ||
```
ellahathaway, line 538: Looking at the monitor-base readme, and it still lists arch-specific tags for .NET 9.0: `9.0.0-rc.2-amd64, 9.0-amd64, 9.0.0-rc.2, 9.0, 9, latest`.

Is monitor-base following this new schema as well, or just monitor?

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -359,7 +359,7 @@ private static IEnumerable<object[]> GetTagTestObjects(TestType testType)
                                     VersionType.Major,
                                     checkOs: false,
                                     checkArchitecture: true,
-                                    skipDockerfileOn: IsApplianceVersionUsingOldSchema
+                                    skipDockerfileOn: IsApplianceVersionWithDistroTags
```
ellahathaway, line 362: Wouldn't we skip on `IsApplianceVersionWithoutArchTags`?

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
ellahathaway: I'm a bit surprised that the tests pass given this.

jander-msft: `dotnet/monitor/base` should still have arch-specific tags. I'll update this condition to check specifically that it ends with `monitor` instead of contains.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
```diff
@@ -301,7 +301,7 @@ private static IEnumerable<object[]> GetTagTestObjects(TestType testType)
                                     VersionType.MajorMinorPatch, 
                                     checkOs: false, 
                                     checkArchitecture: true, 
-                                    skipDockerfileOn: IsApplianceVersionUsingOldSchema
+                                    skipDockerfileOn: info => IsApplianceVersionWithDistroTags(info) || IsApplianceVersionWithoutArchTags(info)
```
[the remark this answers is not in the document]
jander-msft: If this GetTagTestInput invocation is trying to gather everything that looks like "<Major.Minor.Patch>-<architecture>", then it needs to skip for versions that contain a distro in their tag, correct? That's how I'm interpreting what each of the GetTagTestInput invocations are supposed to do.

The `IsApplianceVersionWithDistroTags` is just a rename of `IsApplianceVersionUsingOldSchema` to better indicate what its checking for without changing the logic of the check.

manifest.json
```diff
@@ -9419,8 +9419,9 @@
               "os": "linux",
               "osVersion": "azurelinux3.0-distroless",
               "tags": {
-                "$(monitor|9.0|fixed-tag)-amd64": {},
-                "$(monitor|9.0|minor-tag)-amd64": {}
```
mthalman, line 9423: The suggestion from @lbussell [here](https://github.com/dotnet/dotnet-docker/pull/5948#issuecomment-2393920591) is to not remove any tags, which I think is the correct approach. That way we don't need to announce anything and can have everything consistent.

jander-msft: Okay, that makes things easier for now.

tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs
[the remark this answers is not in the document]
jander-msft: We're going to just undocument the tags for .NET Monitor 9 and attempt to fix this update at a later date. So I've reverted all of the test changes (I will double check that is the case) since all of the existing tags still remain.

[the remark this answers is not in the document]
ellahathaway: Ah I see. `GetTagTestInput` is not filtering for tags that match the given pattern; it's only creating an object of input data to pass to the [PlatformTag_TagExists](https://github.com/dotnet/dotnet-docker/blob/fafd58e3b4f31f98a490b5ed31698722ea713b31/tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs#L106) test.

It's the `PlatformTag_TagExists` test that [skips any dockerfiles matching the given condition](https://github.com/dotnet/dotnet-docker/blob/fafd58e3b4f31f98a490b5ed31698722ea713b31/tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs#L124) before [checking each remaining dockerfiles for a tag that matches the given input pattern](https://github.com/dotnet/dotnet-docker/blob/fafd58e3b4f31f98a490b5ed31698722ea713b31/tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs#L129-L146).

Therefore, given a collection of dockerfiles, we do not want to check dockerfiles without arch-specific tags for the `<Major.Minor.Patch>-<arch>` tag. I don't see a reason to skip dockerfiles with distro-specific tags. I see your point that it's simply a rename though, so I'm confused. I'll parse through the skip logic for a bit to verify my assertions.

lbussell (commented):
@jander-msft, the changes look good to me. Given that we are only undocumenting, should we undocument for all Monitor versions now?

And does any documentation need to be updated?

lbussell (approved):

Issue #5190, opened 2024-02-12
Revert Port environment variable workaround in Aspire Dashboard images


lbussell:
Original issue: https://github.com/dotnet/dotnet-docker/issues/5164

The Aspire Dashboard Dockerfiles currently contain a workaround for a port environment variable issue with the Aspire Dashboard:

https://github.com/dotnet/dotnet-docker/blob/02e6fbd795bb2637042de91092a1194f52ca825c/src/aspire-dashboard/8.0/cbl-mariner-distroless/amd64/Dockerfile#L31-L32

I did not find a tracking issue for this on the Aspire repo but presumably it will be fixed eventually. We want all port/URL environment variables to have the shorthand format:

https://github.com/dotnet/dotnet-docker/blob/32c6a5c755ac3316ca552a55b5a8d5bc927c2d66/src/runtime-deps/6.0/alpine3.18/amd64/Dockerfile#L5

lbussell:
[Triage] @joperezr, can you make this change when the correct fix is in Aspire? I didn't find a tracking issue for it.

joperezr:
Yup, will do.

lbussell:
@joperezr is this fixed in Aspire Dashboard yet?

JamesNK:
Hi, I don't know of a tracking issue for this in dotnet/aspire, but I'm fixing it here: https://github.com/dotnet/aspire/pull/5941

It should be fixed for Aspire 9.0.

PR #5980, opened 2024-10-14
Add Aspire Dashboard 9.0 RC1 images

joperezr:
.NET Aspire 9.0 RC1 will be released soon, so we'll want a nightly image with its contents to be published as well.


joperezr:
I think there are still some changes needed here @lbussell, as we don't want 9.0 image to replace 8.2 one just yet as 9.0 one won't be stable. So the idea would be to have 9.0 prerelease and 8.2.1 side by side, and then when 9.0 ships stable we can remove the 8.2 one.

manifest.json
```diff
@@ -9585,6 +9585,39 @@
...
             }
           ]
+        },
+        {
+          "productVersion": "$(aspire-dashboard|9.0|product-version)",
+          "sharedTags": {
+            "$(aspire-dashboard|9.0-preview|fixed-tag)": {},
+            "$(aspire-dashboard|9.0-preview|minor-tag)": {},
+            "$(aspire-dashboard|9-preview|major-tag)": {},
+            "latest": {}
```
lbussell, line 9595: You will need to remove `latest` from the 8.0 images - you can only have one `latest` per repo.

src/aspire-dashboard/9.0/cbl-mariner-distroless/amd64/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+WORKDIR /app
+COPY --from=installer /app .
+
+ENV \
+    # Unset ASPNETCORE_HTTP_PORTS from base image
+    ASPNETCORE_HTTP_PORTS= \
+    # Aspire Dashboard environment variables
+    ASPNETCORE_URLS=http://0.0.0.0:18888 \
+    DOTNET_DASHBOARD_OTLP_ENDPOINT_URL=http://0.0.0.0:18889 \
+    DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL=http://0.0.0.0:18890
```
lbussell, line 33: Please include the fix for https://github.com/dotnet/dotnet-docker/issues/5190 since it should be fixed in 9.0

tests/Microsoft.DotNet.Docker.Tests/TestData.cs
```diff
@@ -343,6 +343,24 @@ public static class TestData
                 Arch = Arch.Arm64,
                 SupportedImageRepos = DotNetImageRepo.Aspire_Dashboard
             },
+            new() {
+                Version = V9_0,
+                VersionFamily = V9_0,
```
lbussell, line 348: Add a new static version for `V9_0_Preview`, otherwise the tests won't be able to find the image tags with the preview suffix.

https://github.com/dotnet/dotnet-docker/blob/126e8ad647ffb352605519db930eb684f782ab0c/tests/Microsoft.DotNet.Docker.Tests/ImageVersion.cs#L16-L19

lbussell:
There was an issue with the Build command when there are no arch-specific tags - 
```
Unhandled exception: System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.ThrowHelper.ThrowNoMatchException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.PublishImageInfoAsync() in /image-builder/src/Commands/BuildCommand.cs:line 200
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.<ExecuteAsync>b__18_0() in /image-builder/src/Commands/BuildCommand.cs:line 101
```

This is something we should sort out separately from the Aspire Dashboard update. To unblock the PR, the arch-specific tags can be added back for now.

lbussell (approved):

joperezr:
Awesome! thank you so much for all the hard work on getting this through @lbussell !! really appreciate it!

PR #6005, opened 2024-10-24
Pipeline for updating dependencies for testing of internal builds

NikolaMilosavljevic:
Contributes to: https://github.com/dotnet/dotnet-docker-internal/issues/5952

This is the second part of the work that allows automated testing of internal servicing builds. Includes the following feature work:
- Modifies existing pipeline: `update-dependencies-internal.yml`
- Updates templates to propagate `isInternalServicingValidation` to recently refactored steps - this was already fixed in `docker-tools` with https://github.com/dotnet/docker-tools/pull/1473 and https://github.com/dotnet/docker-tools/pull/1475
- Modifies `eng/Get-DropVersions.ps1` and `eng/Set-DotnetVersions.ps1` so they can be invoked individually, and to not update products out-of-scope for this work (i.e. rpm-based and 6.0-related products)
- Modifies UpdateDependencies tool's internal scenario, to not create a PR, but instead commit changes to a branch
- Adds correct trigger for `/testing/*` branches to `dotnet-core-internal-testing.yml` pipeline, so it can be auto-triggered whenever a branch with updated dependencies is created/updated.

Sample pipeline run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2564157&view=results
Corresponding run of the testing pipeline, that was triggered: https://dev.azure.com/dnceng/internal/_build/results?buildId=2564162&view=results


eng/common/templates/stages/build-test-publish-repo.yml
mthalman, line 1: Friendly reminder that changes to eng/common need to be backported to docker-tools.

eng/Set-DotnetVersions.ps1
```diff
@@ -75,11 +75,21 @@ if ($SdkVersion) {
 }
 
 if ($AspnetVersion) {
-    $updateDepsArgs += @("--product-version", "aspnet=$AspnetVersion", "--product-version", "aspnet-runtime-targeting-pack=$AspnetVersion", "--product-version", "aspnet-composite=$AspnetVersion")
+    $updateDepsArgs += @("--product-version", "aspnet=$AspnetVersion", "--product-version", "aspnet-composite=$AspnetVersion")
+
+    if (!$InternalBaseUrl) {
+        # rpm packages are only needed for 6.0 which isn't supported for internal testing scenarios
+        $updateDepsArgs += @("aspnet-runtime-targeting-pack=$AspnetVersion")
```
mthalman, line 82:
```suggestion
        $updateDepsArgs += @("--product-version", "aspnet-runtime-targeting-pack=$AspnetVersion")
```

eng/pipelines/update-dependencies-internal.yml
```diff
@@ -13,15 +13,7 @@ stages:
...
       vmImage: $(defaultLinuxAmd64PoolImage)
     steps:
-      # Merge the internal release branch into the current branch but ignoring all those changes in the internal release branch
-      # This gives us a clean branch to apply updates to but also allows the PR to be merged conflict-free into the internal
-      # release branch.
-    - script: |
-        git config user.email "$(dotnetDockerBot.userName)"
-        git config user.name "$(dotnetDockerBot.email)"
-        git merge -s ours origin/internal/release/$(Build.SourceBranchName)
-      displayName: Configure Branch
```
mthalman, line 23: By getting rid of this, you're no longer keeping the branch in sync with nightly. Otherwise, this branch will continue to get more out of date as time goes on.

mthalman (commented):
1. I assume there is more work planned here since this currently requires explicitly defining the build ID as a pipeline variable rather than have it determined dynamically?
2. How is this intended to work for testing multiple .NET versions (e.g. 8.0, 9.0)? Based on the currently implementation, it doesn't seem this would work since it's always targeting the same branch which would mean changes to the branch would get stomped between the two versions. In the original implementation, this was handled by gathering all versions to be tested into one PR.

eng/pipelines/update-dependencies-internal.yml
mthalman: I'm talking about the target branch (internal/release/...) that the generated files are being merged into. That needs to be up-to-date with the current state of the nightly branch.

eng/common/templates/stages/build-test-publish-repo.yml
mthalman: Oh, I see they haven't been merged in yet: https://github.com/dotnet/dotnet-docker/pull/5985. Which is why they show up in this PR.

eng/pipelines/update-dependencies-internal.yml
NikolaMilosavljevic: When would this merge into `internal/release/...` happen? We create a new branch `testing/...` that is being used for running tests, and that branch is created off of the current branch running the pipeline, i.e. `nightly`.

I might need to tweak the branch name so it has the build# in its name, to avoid stomping over existing ones.

mthalman: Ah, ok. I need to get my eyes checked. When I looked at the test build, I completely misinterpreted the branch being used. Never mind!

NikolaMilosavljevic:
> 1. I assume there is more work planned here since this currently requires explicitly defining the build ID as a pipeline variable rather than have it determined dynamically?
> 2. How is this intended to work for testing multiple .NET versions (e.g. 8.0, 9.0)? Based on the currently implementation, it doesn't seem this would work since it's always targeting the same branch which would mean changes to the branch would get stomped between the two versions. In the original implementation, this was handled by gathering all versions to be tested into one PR.

New testing model uses build ID as input, so it can only refer to one product version. I'm not sure if current/pending code meets the expectation. My impression was that we want to kick off these tests based on build availability. Considering that builds can't be easily identified or used as triggers, I thought that manual start of this validation is sufficient and meets the goals.

mthalman:
> > 1. I assume there is more work planned here since this currently requires explicitly defining the build ID as a pipeline variable rather than have it determined dynamically?
> > 2. How is this intended to work for testing multiple .NET versions (e.g. 8.0, 9.0)? Based on the currently implementation, it doesn't seem this would work since it's always targeting the same branch which would mean changes to the branch would get stomped between the two versions. In the original implementation, this was handled by gathering all versions to be tested into one PR.
> 
> New testing model uses build ID as input, so it can only refer to one product version. I'm not sure if current/pending code meets the expectation. My impression was that we want to kick off these tests based on build availability. Considering that builds can't be easily identified or used as triggers, I thought that manual start of this validation is sufficient and meets the goals.

Question 2 was based on my misunderstanding that there would only be one target branch for all versions. So as long as we can push 8.0 Dockerfile changes to a branch that is separate from 9.0 Dockerfile changes, then we should be good. Is that the case?

Question 1 still stands though. What is the plan for determining what build ID to use? My understanding is that this would all be automated to get the "latest" build. So is there more work planned here?

lbussell:
> What is the plan for determining what build ID to use? My understanding is that this would all be automated to get the "latest" build. So is there more work planned here?

I think that we would test on a per-build basis. So, we'd get an email that a new build is available, then copy the build ID for that build (e.g. `2567915`), then queue the pipeline. If there are failures, then report back to the servicing/release team. That should be all the interaction necessary.

NikolaMilosavljevic:
> > > 1. I assume there is more work planned here since this currently requires explicitly defining the build ID as a pipeline variable rather than have it determined dynamically?
> > > 2. How is this intended to work for testing multiple .NET versions (e.g. 8.0, 9.0)? Based on the currently implementation, it doesn't seem this would work since it's always targeting the same branch which would mean changes to the branch would get stomped between the two versions. In the original implementation, this was handled by gathering all versions to be tested into one PR.
> > 
> > 
> > New testing model uses build ID as input, so it can only refer to one product version. I'm not sure if current/pending code meets the expectation. My impression was that we want to kick off these tests based on build availability. Considering that builds can't be easily identified or used as triggers, I thought that manual start of this validation is sufficient and meets the goals.
> 
> Question 2 was based on my misunderstanding that there would only be one target branch for all versions. So as long as we can push 8.0 Dockerfile changes to a branch that is separate from 9.0 Dockerfile changes, then we should be good. Is that the case?
> 
> Question 1 still stands though. What is the plan for determining what build ID to use? My understanding is that this would all be automated to get the "latest" build. So is there more work planned here?

I'll tweak the branch naming to include the build ID to make them unique.

lbussell:
> I'll tweak the branch naming to include the build ID to make them unique.

For the branch name, are you planning to use the update-dependencies-internal build ID, or the staging pipeline build ID, or both? 

Using only the staging pipeline build ID could run into issues if we need to run multiple test builds for a single SDK build - for example if we ran into issues and needed test changes. I am trying to think through what the dev experience will be like for this.

If a build requires changes to our infrastructure, should we be able to modify the testing branch in-place or is a new update-dependencies pipeline run needed every time?

NikolaMilosavljevic:
> > I'll tweak the branch naming to include the build ID to make them unique.
> 
> For the branch name, are you planning to use the update-dependencies-internal build ID, or the staging pipeline build ID, or both?
> 
> Using only the staging pipeline build ID could run into issues if we need to run multiple test builds for a single SDK build - for example if we ran into issues and needed test changes. I am trying to think through what the dev experience will be like for this.
> 
> If a build requires changes to our infrastructure, should we be able to modify the testing branch in-place or is a new update-dependencies pipeline run needed every time?

If needed we could modify testing branch in place. But, to support multiple runs of the same staging build, we could add both staging and pipeline build IDs, i.e. `testing/<staging-buildID>/<pipeline-buildID>` (i.e. `testing/12345/67890`, or more explicit `testing/staging-12345/pipeline-67890`) Any preference, or any other proposals?

mthalman:
> > > I'll tweak the branch naming to include the build ID to make them unique.
> > 
> > 
> > For the branch name, are you planning to use the update-dependencies-internal build ID, or the staging pipeline build ID, or both?
> > Using only the staging pipeline build ID could run into issues if we need to run multiple test builds for a single SDK build - for example if we ran into issues and needed test changes. I am trying to think through what the dev experience will be like for this.
> > If a build requires changes to our infrastructure, should we be able to modify the testing branch in-place or is a new update-dependencies pipeline run needed every time?
> 
> If needed we could modify testing branch in place. But, to support multiple runs of the same staging build, we could add both staging and pipeline build IDs, i.e. `testing/<staging-buildID>/<pipeline-buildID>` (i.e. `testing/12345/67890`, or more explicit `testing/staging-12345/pipeline-67890`) Any preference, or any other proposals?

Ideally the branch name would also include the .NET version, just for reference/troubleshooting purposes.

MichaelSimons:
> New testing model uses build ID as input, so it can only refer to one product version. I'm not sure if current/pending code meets the expectation. My impression was that we want to kick off these tests based on build availability. Considering that builds can't be easily identified or used as triggers, I thought that manual start of this validation is sufficient and meets the goals.

Builds can be used as triggers.  This is how we triggered the SB builds back when we had tarballs - [example](https://github.com/dotnet/installer/blob/release/6.0.1xx/eng/source-build-tarball-build-official.yml#L4).

We should strive to have this automated so that we can surface issues ASAP.  If we trigger this manually it is going to get overlooked and just add more to our workload.

lbussell:
> Builds can be used as triggers. This is how we triggered the SB builds back when we had tarballs - [example](https://github.com/dotnet/installer/blob/release/6.0.1xx/eng/source-build-tarball-build-official.yml#L4).
> 
> We should strive to have this automated so that we can surface issues ASAP. If we trigger this manually it is going to get overlooked and just add more to our workload.

If we use triggers, will we inadvertently trigger build updates based on builds that we don't service or are out of scope for this work? For example .NET 6.

Also, are there concerns about missing builds due to the success/failure status of the pipeline? I don't see any considerations about this in the pipeline trigger documentation: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/pipeline-triggers?view=azure-devops&tabs=yaml

[Stage-DotNet-Prepare-Artifacts pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=1300)

MichaelSimons:
> If we use triggers, will we inadvertently trigger build updates based on builds that we don't service or are out of scope for this work? For example .NET 6.

Pipeline triggers support branch filters

> Also, are there concerns about missing builds due to the success/failure status of the pipeline?

You can trigger on stages which may be helpful here.  Another option is we have to modified the prepare artifacts pipeline to meet our needs.

lbussell:
> Pipeline triggers support branch filters

All staging pipeline runs seem to be on the main branch of that repo

NikolaMilosavljevic:
Proposal for branch name for testing pipeline run: `testing/<version>-staging-<staging-BuildId>-pipeline-<BuildNumber>`, which becomes this, as example: `testing/9.0-staging-2546092-pipeline-20241018.2`, or alternatively: `testing/9.0-2546092-20241018.2`

NikolaMilosavljevic:
Updated the PR with a tweak to branch name, to include relevant information.

Originally, this PR was reusing the same branch no matter the inputs - using the existing code: `testing/internal-release-nightly-UpdateDependencies-internal-release-nightly-From-dotnet-sdk`

With this update the name is unique: `testing/9.0-UpdateDependencies-staging-2546092-pipeline-20241025.2-From-dotnet-sdk`

This can be tweaked further. Looking for suggestions.

lbussell:
My 2c on the pipeline name:

- `UpdateDependencies` and `From-dotnet-sdk` are extra information and not necessary
- Could we be consistent in the usage of build ID/pipeline build number? Since build IDs are universally unique in the AzDO org I feel that we should use that over the build number.

So, perhaps something like: `testing/9.0-internal/staging-2546092-pipeline-2549999`

That being said, I don't see anything wrong with the code as it is, so if we want to merge and file follow up issue for branch name discussion that's fine.

lbussell (approved):

lbussell:
> so if we want to merge

But it's probably a good idea to wait until Monday at this point 😄

NikolaMilosavljevic:
I've created 2 follow-up issues: https://github.com/dotnet/dotnet-docker-internal/issues/6740, https://github.com/dotnet/dotnet-docker-internal/issues/6741

NikolaMilosavljevic:
> My 2c on the pipeline name:
> 
> * `UpdateDependencies` and `From-dotnet-sdk` are extra information and not necessary
> * Could we be consistent in the usage of build ID/pipeline build number? Since build IDs are universally unique in the AzDO org I feel that we should use that over the build number.
> 
> So, perhaps something like: `testing/9.0-internal/staging-2546092-pipeline-2549999`
> 
> That being said, I don't see anything wrong with the code as it is, so if we want to merge and file follow up issue for branch name discussion that's fine.

I've updated the branch name to use pipeline's build ID instead of build number and to closely match the suggested naming pattern - https://github.com/dotnet/dotnet-docker/pull/6005/commits/0cdc9262b065dc82abd1e1a0dd1eda4ef495b140

lbussell (approved):

mthalman (approved):

PR #6012, opened 2024-10-29
Add globalization scenario test

### LINKED ISSUE #6011: Add scenario test for globalization
Currently we have the following test that is ran inside the SDK images when testing `xunit` functionality: https://github.com/dotnet/dotnet-docker/blob/4f48d36a98187a6e350d54167ef5b568ccd3882f/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/UnitTests.cs#L12-L19

This is not sufficient to validate Globalization functionality in all images. We should add a test that validates all aspects of globalization are working correctly:

- The test should detect if globalization is enabled in .NET.
  - If globalization is enabled:
    - It should validate that the environment variable is set. 
    - It should validate that localization works (currency formatting for example).
    - It should validate that using time zone related classes works.
  - If globalization is disabled:
    - It should validate that calling globalization-related methods fails.

lbussell:
Fixes #6011


tests/Microsoft.DotNet.Docker.Tests/TestScenarios/GlobalizationScenario.cs
```diff
@@ -0,0 +1,77 @@
...
+    RUN dotnet new console -n App -o /src --no-restore
+    WORKDIR /src
+    COPY Program.cs /src/Program.cs
+    RUN dotnet restore
+    RUN dotnet publish --no-restore -o /app
+
+    FROM ${runtime_image} AS runtime
+    COPY --from=sdk /app /app/
+    ENTRYPOINT ["/app/App"]
+    """;
```
mthalman, line 30: nit: please indent

mthalman (approved):

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/GlobalizationTest.cs
```diff
@@ -0,0 +1,63 @@
...
+        TestTimeZoneFunctionality();
+    }
+    catch (CultureNotFoundException)
+    {
+        WriteLine("Successfully caught a CultureNotFoundException, invariant mode is working as expected.");
+    }
+}
+else
+{
+    WriteLine("Globalization invariant mode is disabled. The following should not throw any exceptions:");
```
MichaelSimons, line 28: Should we be validating the envVar is not set in this path?

lbussell:
I tested internally in lieu of PR validation. Some 6.0 legs are failing ([example](https://dev.azure.com/dnceng/internal/_build/results?buildId=2570764&view=logs&j=072873f0-5b27-5c4a-a12b-516992b381a4&t=89337686-2dab-52ee-516f-d082701d77d6)), so more changes are needed. This is because before .NET 8 we weren't consistent with including `tzdata` and `icu` together. I plan to disable the test for .NET 6.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/GlobalizationTest.cs
```diff
@@ -0,0 +1,66 @@
+using System.Globalization;
+using static System.Console;
+
+string envVarName = "DOTNET_SYSTEM_GLOBALIZATION_INVARIANT";
```
mthalman, line 4: nit
```suggestion
const string envVarName = "DOTNET_SYSTEM_GLOBALIZATION_INVARIANT";
```

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/GlobalizationTest.cs
```diff
@@ -0,0 +1,66 @@
...
+
+if (invariantModeEnabled != invariantModeEnvVarSetting)
+{
+    throw new Exception("Environment variable mis-match: " + isEnabledString
+        + $", but {envVarName} is set to `{envVarValue}`, which evaluates to {invariantModeEnvVarSetting}.");
+}
+
+try
+{
+    WriteLine($"The following should {(invariantModeEnabled ? "" : "not ")}produce an exception:");
```
mthalman, line 20: There's no logic below that checks the case where an exception was expected but no exception occurred.

mthalman (approved):

MichaelSimons (approved):

PR #6018, opened 2024-10-30
Use issue form for bug report template

lbussell:
### Why

I wanted to add field to the issue template for which image/digest is causing a bug. I saw that there have been issue forms for a while now, so I figured why not translate our current template into a form at the same time.

### Results

What does the form look like? I can't capture the whole thing in one screenshot, but this should give you the gist:

image

What a submitted issue looks like:

image


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,94 @@
...
+      required: true
+
+  - type: input
+    id: image
+    attributes:
+      label: Which .NET image are you using?
+      description: Please specify the exact digest if possible.
+      placeholder: mcr.microsoft.com/dotnet/sdk:8.0@sha256:00ab6163f4c8fdf8ea427b1fa9d9d2312ff623d5810577ad73605a07ba67397a
+    validations:
+      required: true
```
mthalman, line 40: I'm wondering whether this should be optional. It's possible that a bug applies to multiple images.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,94 @@
...
+    attributes:
+      label: Describe the bug
+    validations:
+      required: true
+
+  - type: input
+    id: image
+    attributes:
+      label: Which .NET image are you using?
+      description: Please specify the exact digest if possible.
```
mthalman, line 37:
```suggestion
      description: Please specify the exact digest if possible; otherwise, provide the tag name.
```

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,94 @@
+name: 🐞 Bug report 2
```
mthalman, line 1:
```suggestion
name: 🐞 Bug report
```

.github/ISSUE_TEMPLATE/bug-report.yml
lbussell: Yeah, I went back and forth on this. Probably just having the field at all will encourage people to fill it out, required or not. But it is just a text box after all, nothing stops you from putting in `image1, image2, image3` or `all 8.0 sdk images`, for what it's worth.

mthalman (approved):

Issue #6023, opened 2024-11-02
Globalization issues when using dotnet/aspnet:9.0-azurelinux3.0-distroless-extra


craigktreasure:
### Describe the bug

I have a simple Blazor application (using Blazorise in this case) that ran just fine on .NET 8 in either the `8.0-azurelinux3.0-distroless-extra` or `8.0-cbl-mariner2.0-distroless-extra` containers. After upgrading the application to .NET 9 RC 2 using the `9.0-azurelinux3.0-distroless-extra` container image (Mariner 2.0 is not available for .NET 9), I began encountering globalization errors that I haven't been able to resolve. There appears to be something different between the globalization resources installed for `9.0-azurelinux3.0-distroless-extra` compared to the others, which is unexpected when upgrading to .NET 9 or switching from another .NET 9 container image.

### Which .NET image(s) are you using?

mcr.microsoft.com/dotnet/aspnet:9.0-azurelinux3.0-distroless-extra with sha256:e006bfc2276280f9f768c555a2870d3484b85a871ef9d3635eab29e9a5795c97

### Steps to reproduce

See the minimal repro [here](https://github.com/[craigktreasure/azurelinux3-invariant-repro](https://github.com/craigktreasure/azurelinux3-invariant-repro)).

### Other information

The same code works in both .NET 8 and .NET 9 with the following container images (not exhaustive):

- `mcr.microsoft.com/dotnet/aspnet:8.0-cbl-mariner2.0-distroless-extra`
- `mcr.microsoft.com/dotnet/aspnet:8.0-azurelinux3.0-distroless-extra`
- `mcr.microsoft.com/dotnet/aspnet:9.0-noble-chiseled-extra`

The same code does not work on .NET 9 with the following container images (not exhaustive):

- `mcr.microsoft.com/dotnet/aspnet:9.0-azurelinux3.0-distroless-extra`

The error encountered is:

```
System.Globalization.CultureNotFoundException: Only the invariant culture is supported in globalization-invariant mode. See https://aka.ms/GlobalizationInvariantMode for more information. (Parameter 'name')
cs is an invalid culture identifier.
   at System.Globalization.CultureInfo..ctor(String name, Boolean useUserOverride)
   at Blazorise.Localization.TextLocalizerService.AddLanguageResource(String cultureName)
   at Blazorise.Localization.TextLocalizerService.ReadResource()
   at Blazorise.Localization.TextLocalizerService..ctor()
[...]

PR #6030, opened 2024-11-06
Run globalization tests on aspnet images

lbussell:
Related: #6023


lbussell:
I have prepared the fix for #6023 here: https://github.com/dotnet/dotnet-docker/commit/2e591b80e358d897eabdb145619ed1697d298869, but I am letting this test run to validate that it catches the issue. If it doesn't, then I'll need to refine the test some.

lbussell:
The trx files for the 9.0-azurelinux3.0-distroless arm64 legs do not contain any references to `extra` images, so I suspect the test is not running. My suspicion is that this is a path filtering issue. I'll investigate tomorrow!

lbussell:
I have filed an issue for the root cause of this and am working on a solution - https://github.com/dotnet/dotnet-docker/issues/6033. This draft PR is no longer necessary

PR #6048, opened 2024-11-12
Merge main to nightly for November 2024 .NET updates

lbussell:
Related: #5983


manifest.versions.json
```diff
@@ -147,7 +137,7 @@
     "monitor|6.3|base-url|main": "$(base-url|public|maintenance|main)",
     "monitor|6.3|base-url|nightly": "$(base-url|public|preview|nightly)",
```
jander-msft, line 138: I think you'll need to flip all of the base-urls for .NET Monitor to be `$(base-url|public|maintenance|nightly)`:
- For 6.3.10 and 8.0.6, the stable version folder doesn't exist on the dotnetbuilds storage account.
- For 9.0.0, the corresponding released version doesn't exist at all on the dotnetbuilds storage account.

jander-msft (approved):

mthalman (approved):

lbussell:
I have re-ran the pipeline for the other failures since they were due to packages not being mirrored back to our nuget.org mirror yet.

8.0 noble runtime deps images are failing package tests because they still haven't been rebuilt to pick up the latest changes to the slices. If they are the only failures left today, then I'll merge and do a no-cache rebuild of all images just to fix that issue.

@mthalman, it also looks like EOL annotation is running in the no-cache build and potentially hitting some unexpected behavior? https://dev.azure.com/dnceng-public/public/_build/results?buildId=868328&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=f999747d-9c54-5468-549c-0f64892ca96d&l=98

mthalman:
> @mthalman, it also looks like EOL annotation is running in the no-cache build and potentially hitting some unexpected behavior? https://dev.azure.com/dnceng-public/public/_build/results?buildId=868328&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=f999747d-9c54-5468-549c-0f64892ca96d&l=98

Looks like a bug in the dry-run logic. I'll investigate further. Don't let that block this from merging.

mthalman:
Actually, maybe don't merge yet. Let me investigate more.

lbussell:
> Actually, maybe don't merge yet. Let me investigate more.

OK. The other failures are Noble package tests only as described above.

mthalman:
Ok, this will be safe to merge. The error occurs because we still have 6.0 Dockerfiles in the repo and since 6.0 is now EOL, it's including those images to be annotated. It applies those annotations to the images produced by the build. But since this is a dry run and nothing is published, there are no digests in the image info file. It's not handling that gracefully. So this will only occur in PR builds until I get this fixed.

lbussell:
@mthalman please file a follow-up issue. Thanks

mthalman:
> @mthalman please file a follow-up issue. Thanks

https://github.com/dotnet/docker-tools/pull/1498

PR #6055, opened 2024-11-13
Pass paths to test script from build-and-test

lbussell:
This fixes a bug which prevented you from running the correct tests when passing paths to `build-and-test.ps1`

Other changes:

- Avoid running the `pre-build` tests twice by removing it from the test categories if it is ran before the build step.
- Don't run the build twice if we aren't building samples.
- Adjust the default test categories to make building and testing faster by default.

### DIFF
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -36,9 +36,20 @@ if ($Version -notmatch '^\d+\.\d+(\.[\d*])?|\*$') {
     }
 }
 
-if (($Mode -eq "BuildAndTest" -or $Mode -eq "Test") -and $TestCategories.Contains("pre-build")) {
-    & ./tests/run-tests.ps1 -TestCategories "pre-build" -Version "*"
-    $TestCategories.Remove("pre-build")
+if (($Mode -eq "BuildAndTest" -or $Mode -eq "Test")) {
+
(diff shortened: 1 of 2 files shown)


build-and-test.ps1
```diff
@@ -21,9 +21,11 @@ param(
 
     # Categories of tests to run
     [ValidateSet("runtime", "runtime-deps", "aspnet", "sdk", "pre-build", "sample", "image-size", "monitor", "aspire-dashboard")]
-    [string[]]$TestCategories = @("runtime", "runtime-deps", "aspnet", "sdk", "pre-build", "sample", "image-size", "monitor", "aspire-dashboard")
+    [string[]]$TestCategories = @("runtime", "runtime-deps", "aspnet", "sdk", "monitor", "aspire-dashboard")
```
mthalman, line 24: It might be helpful as a reminder to the dev of what is actually getting run. Maybe it would be useful to output which categories are being run.

lbussell:
Added an extra feature that I had been trying for a while - you can run very specific tests by passing `-CustomTestFilter "VerifyFxDependentAppScenario"`. Or more specific, `"AspnetImageTests.VerifyFxDependentAppScenario"`. It is based on the fully qualified name of the test method.

PR #6058, opened 2024-11-14
Update PowerShell for .NET 9 images

adityapatwardhan:


lbussell:
The Publish failure here is a known issue. See https://github.com/dotnet/dotnet-docker/pull/6048#issuecomment-2474425538 for context. This requires an ImageBuilder update, which I will handle.

lbussell (approved):

PR #6089, opened 2024-12-04
Update graceful-shutdown.md - Avoid memory leak on combined CancellationTokenSource

JakobOoms:
Avoid memory leak on combined CancellationTokenSource


lbussell (approved):
Thanks for the contribution!

PR #6103, opened 2024-12-09
Enable product EOL support for annotations

mthalman:
The latest 6.0 images do not currently have EOL annotations set. This is because those 6.0 Dockerfiles had been rebuilt due to a base image update and the pipeline was not configured to take into account 6.0's EOL status. This fixes this by enabling the `annotate-eol-products` option.

Related: https://github.com/dotnet/dotnet-docker-internal/issues/6927


lbussell (approved):

lbussell:
I put the `do-not-merge` on this until later today or tomorrow.

PR #6108, opened 2024-12-11
Merge main to nightly

mthalman:
This also cleans up usage of `publishEolAnnotations` in nightly pipeline. These are obsolete now that it's incorporated into `core.yml` which is referenced by all those pipelines.


lbussell (approved):
Oops, looks like I squashed https://github.com/dotnet/dotnet-docker/pull/6107 instead of doing a regular merge. Can you make sure this PR is merged without squashing?

Issue #5990, opened 2024-10-16
Update-dependencies pipeline: create separate PRs per upgraded image component


lbussell:
### Describe the Problem

The Update-dependencies tool, via our pipelines, creates one pull request per day if there are any components that should be upgraded. This usually bundles together .NET updates along with updates to other tools or image components, like [syft](https://github.com/dotnet/dotnet-docker/blob/3a34bc22bbddd555e87d1f25f05aba85755b9763/manifest.versions.json#L281), [mingit](https://github.com/dotnet/dotnet-docker/blob/3a34bc22bbddd555e87d1f25f05aba85755b9763/manifest.versions.json#L121), or [chisel](https://github.com/dotnet/dotnet-docker/blob/3a34bc22bbddd555e87d1f25f05aba85755b9763/manifest.versions.json#L64-L66).

When all components are updated together in one PR and one commit it is more difficult to isolate the cause of PR validation failures. Furthermore, merging all of the changes together provides less detail in the commit history about what versions were updated at which times.

### Describe the Solution

In the update-dependencies pipeline, we should submit separate PRs for each of the tools listed above and keep them separate from .NET updates.

lbussell:
[Triage] This is completed with https://github.com/dotnet/dotnet-docker/pull/6122.

PR #6114, opened 2024-12-13
Allow updating tools/image components separately from .NET product versions

### LINKED ISSUE #5920: Use Canonical-published checksums for installing Chisel tool
Currently in update-dependencies, we manually calculate checksums for Chisel:

https://github.com/dotnet/dotnet-docker/blob/6ffe4d902be6a5efead2031a079fb6cc1ddf4a57/eng/update-dependencies/ChiselUpdater.cs#L50-L59

For future releases, Canonical will compute and publish their own checksums. We should use those once they're available:

- https://github.com/canonical/chisel/pull/155

lbussell:
This gets us most of the way towards [Update-dependencies pipeline: create separate PRs per upgraded image component (#5990)](https://github.com/dotnet/dotnet-docker/issues/5990). There should be follow-up to enable this feature in the pipeline.

This also fixes [Use Canonical-published checksums for installing Chisel tool (#5920)](https://github.com/dotnet/dotnet-docker/issues/5920), since I needed to update the Chisel updater anyways. It is really simple - the shas are just in additional release assets which are just text files.

Changes:
- Add a new option to `update-dependencies` for updating tools/image components.
  - It can be used like `--tool <chisel|mingit|rocks-toolbox|syft>`
  - Tools can be updated individually or together.
  - You don't need to update any .NET versions at the same time. But you can if you want to, just by passing all the arguments at the same time.
  - To do this, I needed to treat these tools more like their own products in the code. It required a fair bit of refactoring but I think the resulting implementation is less hacky than before.

Fixes #5920


eng/update-dependencies/ChiselUpdater.cs
```diff
@@ -43,20 +53,47 @@ public static string GetChiselManifestVariable(string product, string arch, stri
...
+        string arch)
+        : GitHubReleaseUrlUpdater(
+            repoRoot,
+            ChiselUpdater.ToolName,
+            GetChiselManifestVariable("chisel", arch, ShaFucntion, "latest"),
+            ChiselUpdater.Owner,
+            ChiselUpdater.Repo,
+            GetAssetRegex(arch))
     {
+        private const string ShaFucntion = "sha384";
```
mthalman, line 67:
```suggestion
        private const string ShaFunction = "sha384";
```

eng/update-dependencies/ChiselUpdater.cs
```diff
@@ -43,20 +53,47 @@ public static string GetChiselManifestVariable(string product, string arch, stri
...
             }
 
-            return ChecksumHelper.ComputeChecksumShaAsync(s_httpClient, downloadUrl).Result;
+            downloadUrl = $"{downloadUrl}.{ShaFucntion}";
+            return GetChecksumFromUrlAsync(downloadUrl).Result;
+        }
+
+        private static async Task<string?> GetChecksumFromUrlAsync(string downloadUrl)
+        {
+            using HttpResponseMessage response = await s_httpClient.GetAsync(downloadUrl);
```
mthalman, line 85: I'm curious what the download URL would be in this case.

mthalman (approved):

eng/update-dependencies/ChiselUpdater.cs
```diff
@@ -76,7 +77,23 @@ private class ChiselReleaseShaUpdater(
...
+            {
+                Trace.TraceInformation($"Failed to download {downloadUrl}.");
+                return null;
+            }
+
+            // Expected format:
+            // abcdef1234567890  chisel_v1.0.0_linux_amd64.tar.gz
+            string content = await response.Content.ReadAsStringAsync();
+            string sha = content.Split("  ")[0];
+            return sha.ToLowerInvariant();
```
richlander, line 96: I see that this project doesn't have nullable enabled. This code looks like it would benefit from that.

lbussell: @richlander I have the changes ready for enabling nullable annotations on a separate branch. I am planning to address this in a follow-up PR once this is merged. Trying to keep changes separate where possible 🙂

eng/update-dependencies/ChiselUpdater.cs
lbussell: Example:

Binary: https://github.com/canonical/chisel/releases/download/v1.0.0/chisel_v1.0.0_linux_amd64.tar.gz

Checksum:
https://github.com/canonical/chisel/releases/download/v1.0.0/chisel_v1.0.0_linux_amd64.tar.gz.sha384

Exactly the same but with the checksum algorithm added to the end. Is this what you meant?

lbussell: @mthalman OK to merge?

mthalman:
yep

PR #6117, opened 2024-12-16
Re-write ASP.NET Core HTTPS documentation

### LINKED ISSUE #5113: Some sample commands still assume the in-container user will be root

lbussell:
Fixes #5113

While working on #5113 I noticed that the two documents, `run-aspnetcore-https-development.md` and `host-aspnetcore-https.md` contained much of the same information. I tried to combine them and improve both documents in the process.

A couple of other notes:
- I couldn't get `Kestrel:Certificates:Development:Password` or `ASPNETCORE_Kestrel__Certificates__Development__Password` to work for any scenario. Only `Default` worked.
- I couldn't get any sample image to read a certificate from `/home/app/.aspnet/https/aspnetapp.pfx`, so I made all the example commands specific the exact path to the certificate. I think this is clearer anyways.


richlander:
These docs pre-date the official docs: https://learn.microsoft.com/aspnet/core/security/enforcing-ssl

Ideally, we'd rely as much as possible on the official docs. Either way, we should get the ASP.NET Core team to sign off on the changes.

@javiercn @Rick-Anderson

lbussell:
> These docs pre-date the official docs: https://learn.microsoft.com/aspnet/core/security/enforcing-ssl
> 
> Ideally, we'd rely as much as possible on the official docs. Either way, we should get the ASP.NET Core team to sign off on the changes.
> 
> @javiercn @Rick-Anderson

Yes, I did find those, and they seemed very similar to these docs. Worth noting that [Hosting ASP.NET Core images with Docker over HTTPS](https://learn.microsoft.com/en-us/aspnet/core/security/docker-https?view=aspnetcore-9.0#running-pre-built-container-images-with-https) links to `run-aspnetcore-https-development.md` at the bottom of the document.

I would love to throw out `host-aspnetcore-https.md` and link to the learn.microsoft.com documentation. Not sure how others feel about that.

mthalman:
I'm fine with getting rid of these files and updating the existing links to them to point to the docs site instead.

mthalman (commented):
Why are we having two files that have the same content?

lbussell:
> Why are we having two files that have the same content?

I don't know, I didn't create them initially. They don't have the same content anymore.

I chose to leave the files in case there are links to them from other places. For example, [run-aspnetcore-https-development.md](https://github.com/dotnet/dotnet-docker/blob/main/samples/run-aspnetcore-https-development.md) is referenced from [Hosting ASP.NET Core images with Docker over HTTPS](https://learn.microsoft.com/aspnet/core/security/docker-https). If you think we should delete the file and change the docs website then we can also do that.

mthalman:
> They don't have the same content anymore.

Weird. I don't know what I was doing but when I looked at them they were identical.

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
+directory, `/root/`.
+
+For Linux containers on Windows:
+
+```pwsh
+docker run --rm -it `
+    -p 8001:8001 `
+    -e ASPNETCORE_HTTPS_PORTS=8001 `
+    -e ASPNETCORE_ENVIRONMENT=Development `
+    # Use user secrets instead of certificate password environment variable
```
mthalman, line 93: The comments here mess up the multi-line PowerShell. It essentially breaks the line continuations that are used causing an invalid command when pasting this whole thing.

samples/host-aspnetcore-https.md
```diff
@@ -1,98 +1,19 @@
...
-docker run --rm -it -p 8001:8001 -e ASPNETCORE_HTTPS_PORTS=8001 -e ASPNETCORE_Kestrel__Certificates__Default__Password="<CREDENTIAL_PLACEHOLDER>" -e ASPNETCORE_Kestrel__Certificates__Default__Path=\https\aspnetapp.pfx -v $env:USERPROFILE\.aspnet\https:C:\https\ --user ContainerAdministrator mcr.microsoft.com/dotnet/samples:aspnetapp
-```
-
-> [!NOTE]
-> The password must match the password used for the certificate. Running as ContainerAdministrator is not recommended in production scenarios. The `--user ContainerAdministrator` flag has been added in this example since there is a potential and intermittent race condition that throws a `WindowsCryptographicException` at the container/app startup when using self-signed certificates. This is a known bug and it has been [reported here](https://github.com/dotnet/runtime/issues/70386).
+ASP.NET Core uses [HTTPS by
+default](https://docs.microsoft.com/aspnet/core/security/enforcing-ssl).
+[HTTPS](https://en.wikipedia.org/wiki/HTTPS) relies on
+[certificates](https://en.wikipedia.org/wiki/Public_key_certificate) for trust,
+identity, and encryption.
```
Rick-Anderson, line 7:
```suggestion
ASP.NET Core uses [HTTPS by default](https://docs.microsoft.com/aspnet/core/security/enforcing-ssl). [HTTPS](https://en.wikipedia.org/wiki/HTTPS) relies on [certificates](https://en.wikipedia.org/wiki/Public_key_certificate) for trust, identity, and encryption.
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
 
-The Windows examples below are written for PowerShell. CMD users will need to change the format of the environment variables in the instructions from `$env:USERPROFILE` to `%USERPROFILE%`.
+The Windows examples below are written for PowerShell. CMD users will need to
+change the format of the environment variables in the instructions from
+`$env:USERPROFILE` to `%USERPROFILE%`.
 
-This sample requires [Docker 17.06](https://docs.docker.com/release-notes/docker-ce) or later of the [Docker client](https://www.docker.com/products/docker).
+This example requires [Docker
+Desktop](https://www.docker.com/products/docker-desktop/). Make sure you have
+the latest version.
```
Rick-Anderson, line 24:
```suggestion
This example requires [Docker Desktop](https://www.docker.com/products/docker-desktop/). We recommend the latest version.
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
 
-Generate cert and configure local machine:
+## Create and trust a development certificate
 
-```console
-dotnet dev-certs https -ep $env:USERPROFILE\.aspnet\https\aspnetapp.pfx -p <CREDENTIAL_PLACEHOLDER>
-dotnet dev-certs https --trust
-```
+Follow the instructions for creating a [.NET development certificate](https://learn.microsoft.com/dotnet/core/tools/dotnet-dev-certs)
+from [Running pre-built container images with HTTPS](https://learn.microsoft.com/aspnet/core/security/docker-https#running-pre-built-container-images-with-https).
```
Rick-Anderson, line 43:
```suggestion
Follow the instructions for creating a [.NET development certificate](https://learn.microsoft.com/dotnet/core/tools/dotnet-dev-certs) from [Running pre-built container images with HTTPS](https://learn.microsoft.com/aspnet/core/security/docker-https#running-pre-built-container-images-with-https).
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
->
-> - The certificate name, in this case *aspnetapp*.pfx must match the project assembly name.
-> - `<CREDENTIAL_PLACEHOLDER>` is used as a stand-in for a password of your own choosing.
-> - If console returns "A valid HTTPS certificate is already present.", a trusted certificate already exists in your store. It can be exported using MMC Console.
+## Enable HTTPS using environment variables
 
-Configure application secrets, for the certificate:
+See [Hosting ASP.NET Core images with Docker over HTTPS](https://learn.microsoft.com/aspnet/core/security/docker-https).
+If you don't want to use environment variables to store your development
+certificate password, continue reading.
```
Rick-Anderson, line 49:
```suggestion
See [Hosting ASP.NET Core images with Docker over HTTPS](https://learn.microsoft.com/aspnet/core/security/docker-https). See the following section if you don't want to use environment variables to store your development
certificate password.
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
-
-Build a container image:
+## Using user secrets for certificate password
 
-```console
-docker build --pull -t aspnetapp .
-```
+If you don't want to use environment variables to store your development
+certificate password, you can use [.NET user secrets](https://learn.microsoft.com/aspnet/core/security/app-secrets?view=aspnetcore-9.0&tabs=linux)
+instead to store the certificate password on your machine.
```
Rick-Anderson, line 55:
```suggestion
Rather than using environment variable to specify the development certificate password,  use [.NET user secrets](https://learn.microsoft.com/aspnet/core/security/app-secrets) to store the password.
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
 
-Run the container image with ASP.NET Core configured for HTTPS:
+This will require modifications to the `aspnetapp` sample, so make sure you've cloned
+this repo or [download the repository as a zip](https://github.com/dotnet/dotnet-docker/archive/main.zip).
```
Rick-Anderson, line 58:
```suggestion
Using User Secrets requires modifications to the `aspnetapp` sample. Clone
this repo or [download the repository as a zip](https://github.com/dotnet/dotnet-docker/archive/main.zip).
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
-
-```console
-docker run --rm -it -p 8001:8001 -e ASPNETCORE_HTTPS_PORTS=8001 -e ASPNETCORE_ENVIRONMENT=Development -v ${HOME}/.microsoft/usersecrets/:/root/.microsoft/usersecrets -v ${HOME}/.aspnet/https:/root/.aspnet/https/ aspnetapp
+git clone https://github.com/dotnet/dotnet-docker/
 ```
 
-After the application starts, navigate to `https://localhost:8001` in your web browser.
-
-### Linux containers on Linux host
+First, initialize user secrets for your app, and set the certificate password.
```
Rick-Anderson, line 64:
```suggestion
Initialize user secrets for your app, and set the certificate password:
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
-> - The certificate name, in this case *aspnetapp*.pfx must match the project assembly name.
-> - `<CREDENTIAL_PLACEHOLDER>` is used as a stand-in for a password of your own choosing.
-> - If console returns "A valid HTTPS certificate is already present.", a trusted certificate already exists in your store. It can be exported using MMC Console.
-
-Configure application secrets, for the certificate:
-
-```console
 dotnet user-secrets init -p aspnetapp/aspnetapp.csproj
-dotnet user-secrets -p aspnetapp\aspnetapp.csproj set "Kestrel:Certificates:Development:Password" "<CREDENTIAL_PLACEHOLDER>"
+dotnet user-secrets -p aspnetapp/aspnetapp.csproj set "Kestrel:Certificates:Default:Password" "<CREDENTIAL_PLACEHOLDER>"
```
Rick-Anderson, line 69: Per SFI, credential instructions can't be copy/paste ready. Add illegal strings so they can't have a working copy/paste. 

```suggestion
dotnet user-secrets -p aspnetapp/aspnetapp.csproj set "Kestrel:Certificates:Default:Password" "<`"CREDENTIAL" "PLACEHOLDER`>"
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
 ```
 
-> [!NOTE]
-> The password must match the password used for the certificate.
-
-Build a container image:
+Initializing user-secrets for the first time on a project will modify the
+project file, so we need to re-build the sample image:
```
Rick-Anderson, line 73:
```suggestion
Initializing user-secrets for the first time on a project modifies the project file, which requires re-building the sample image:
```

samples/run-aspnetcore-https-development.md
```diff
@@ -1,239 +1,131 @@
...
-After the application starts, navigate to `https://localhost:8001` in your web browser.
-
-> In the case of using https, be sure to check the certificate you're using is trusted on the host. You can start with navigating to `https://localhost:8001` in the browser. If you're looking to test https with a domain name (e.g. `https://contoso.com:8001`), the certificate would also need the appropiate Subject Alternative Name included, and the DNS settings on the host would need to be updated. In the case of using the generated dev certificate, the trusted certificate will be issued from localhost and will not have the SAN added.
+In Linux containers, .NET looks under the `~/.microsoft/usersecrets/` directory
+for user secrets data. All you need to do is bind-mount your host machine's
+user secrets directory to the container's filesystem in the correct location,
+similar to what we did with the certificate above. if you are running your
+container as the `root` user, replace `/home/app/` with the `root` user's home
+directory, `/root/`.
+
```
Rick-Anderson, line 85:
```suggestion
In Linux containers, .NET looks under the `~/.microsoft/usersecrets/` directory for user secrets data. Bind-mount your host machine's user secrets directory to the container's filesystem. This is similar to the preceding certificate instructions. f you are running your container as the `root` user, replace `/home/app/` with the `root` user's home directory, `/root/`.
```

Rick-Anderson (changes requested):

samples/run-aspnetcore-https-development.md
```diff
@@ -4,236 +4,98 @@ ASP.NET Core uses [HTTPS by default](https://docs.microsoft.com/aspnet/core/secu
...
->
-> - The certificate name, in this case *aspnetapp*.pfx must match the project assembly name.
-> - `<CREDENTIAL_PLACEHOLDER>` is used as a stand-in for a password of your own choosing.
-
-Configure application secrets, for the certificate:
-
-```console
 dotnet user-secrets init -p aspnetapp/aspnetapp.csproj
-dotnet user-secrets -p aspnetapp/aspnetapp.csproj set "Kestrel:Certificates:Development:Password" "<CREDENTIAL_PLACEHOLDER>"
+dotnet user-secrets -p aspnetapp/aspnetapp.csproj set "Kestrel:Certificates:Default:Password" "<`"CREDENTIAL" "PLACEHOLDER`>"
```
mthalman, line 51: The `<CREDENTIAL_PLACEHOLDER>` change needs to be reverted. It is specifically defined that way as a magic string to appease the secrets scanner.

lbussell: Can we validate that this new format won't trip the secret scanner? Or find another way to break the command so that it can't be copy-pasted without changes?

mthalman: Don't know. It gets executed automatically on the server side when attempting to push into the AzDO repo.

lbussell: I was able to push this branch internally without issue: https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet-docker?path=/samples/run-aspnetcore-https-development.md&version=GBdev/loganbussell/pulls/6117&line=51&lineEnd=52&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

mthalman: Ok, that's good. But I don't understand the need for the change. It's not formatted correctly. It leads to this error: `Unrecognized command or argument 'PLACEHOLDER>'`

lbussell: My interpretation of Rick's suggestion is that the command _shouldn't_ be formatted correctly. It should error if users copy-paste without changing the placeholder text, to prevent them from using `<CREDENTIAL_PLACEHOLDER>` as the password.

mthalman: > My interpretation of Rick's suggestion is that the command _shouldn't_ be formatted correctly. It should error if users copy-paste without changing the placeholder text, to prevent them from using `<CREDENTIAL_PLACEHOLDER>` as the password.

@Rick-Anderson - Is that the case here and if so, is that a common pattern used in our docs? I'm concerned that users will get confused by this syntax and think that things need to be formatted in special way rather than just a simple string.

Rick-Anderson: > > My interpretation of Rick's suggestion is that the command _shouldn't_ be formatted correctly. It should error if users copy-paste without changing the placeholder text, to prevent them from using `<CREDENTIAL_PLACEHOLDER>` as the password.
> 
> @Rick-Anderson - Is that the case here and if so, is that a common pattern used in our docs? I'm concerned that users will get confused by this syntax and think that things need to be formatted in special way rather than just a simple string.

That's the new SFI pattern and mandate. Any markup or code that is copy/pasted must throw an error to prevent copy paste. We can monitor the doc if ppl have problems and add some comments helping them.

mthalman (approved):

samples/run-aspnetcore-https-development.md
lbussell: I was playing around with other formats. This one works (or rather, doesn't work):

```
> dotnet user-secrets -p aspnetapp/aspnetapp.csproj set "Kestrel:Certificates:Default:Password" $CREDENTIAL_PLACEHOLDER

Missing parameter value for 'value'.
Use the '--help' flag to see info.
```

It also passes AzDO credential check. Since it looks so much better I'm going to update to that.

mthalman (approved):

Rick-Anderson (approved):

PR #6118, opened 2024-12-17
Enable nullable annotations in update-dependencies project

lbussell:
Related: https://github.com/dotnet/dotnet-docker/issues/5006, https://github.com/dotnet/dotnet-docker/pull/6114#discussion_r1887343464

### DIFF
--- a/eng/update-dependencies/ScriptRunnerUpdater.cs
+++ b/eng/update-dependencies/ScriptRunnerUpdater.cs
@@ -15,26 +15,23 @@ namespace Dotnet.Docker
     /// </summary>
     public class ScriptRunnerUpdater : IDependencyUpdater
     {
-        public required string ScriptPath { get; init; }
+        private readonly string _scriptPath;
 
-        private ScriptRunnerUpdater()
+        private ScriptRunnerUpdater(string scriptPath)
         {
+            _scriptPath = scriptPath;
         }
 
         public static IDependencyUpdater GetDockerfileUpdater(string repoRoot)
         {
-            return new ScriptRunnerUpdater()
-            {
-                ScriptPath = Path.Combine(repoRoot, "eng", "dockerfile-templates", "Get-GeneratedDockerfiles.ps1")
-            };
+            string scriptPath = Path.Combine(repoRoot, "eng", "dockerfile-templates", "Get-GeneratedDockerfiles.ps1");
+            return new ScriptRunnerUpdater(scriptPath);
         }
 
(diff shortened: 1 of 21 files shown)


richlander:
This code is subject to out of bounds exception, yes?

https://github.com/dotnet/dotnet-docker/blob/be05cf5bfd77288e0bb51b736b262b49772000a5/eng/update-dependencies/ChiselUpdater.cs#L95

lbussell:
> This code is subject to out of bounds exception, yes?
> 
> https://github.com/dotnet/dotnet-docker/blob/be05cf5bfd77288e0bb51b736b262b49772000a5/eng/update-dependencies/ChiselUpdater.cs#L95

How so?

If we get an unexpected result from the Chisel checksum file, then

richlander:
Ah. I forgot/didn't realize that string.split worked that way. Makes sense.

eng/update-dependencies/ScriptRunnerUpdater.cs
```diff
@@ -16,7 +15,7 @@ namespace Dotnet.Docker
     /// </summary>
     public class ScriptRunnerUpdater : IDependencyUpdater
     {
-        private string _scriptPath;
+        public required string ScriptPath { get; init; }
```
mthalman, line 18: There's no need for this to be `public` since it's not consumed outside this class. This can stay a field and just be set in the constructor.

eng/update-dependencies/UpdateDependencies.cs
```diff
@@ -414,8 +413,8 @@ private static IEnumerable<IDependencyUpdater> GetProductUpdaters()
 
         private static IEnumerable<IDependencyUpdater> GetGeneratedContentUpdaters() =>
         [
-            ScriptRunnerUpdater.GetDockerfileUpdater(RepoRoot),
-            ScriptRunnerUpdater.GetReadMeUpdater(RepoRoot)
+            // ScriptRunnerUpdater.GetDockerfileUpdater(RepoRoot),
+            // ScriptRunnerUpdater.GetReadMeUpdater(RepoRoot)
```
mthalman, line 417:
```suggestion
            ScriptRunnerUpdater.GetDockerfileUpdater(RepoRoot),
            ScriptRunnerUpdater.GetReadMeUpdater(RepoRoot)
```

lbussell: I should just add a CLI option for this at some point.

mthalman (approved):

PR #6120, opened 2024-12-18
Add NLS test scenario

### LINKED ISSUE #6034: Check for Windows NLS vs. ICU behavior differences in Globalization tests
Nano Server 2025 does not include ICU by default. 

When ICU is missing on Linux, Globalization Invariant Mode is disabled, and you try to use globalization features, the .NET Runtime throws an exception.

In the same scenario on Windows, instead of throwing an error, the Runtime falls back to using Windows' NLS (National Language Support) feature. This has some breaking changes which are documented here: https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu

We should check for those breaking changes in the recently added [globalization test scenario](https://github.com/dotnet/dotnet-docker/pull/6012).

---

Related Issues: #5944, #5635 

The following table summarizes my best understanding of the .NET Runtime's behavior under different globalization scenarios:

| OS | ICU Present | Invariant Mode | Behavior when using globalization features |
| --- | --- | --- | --- |
| Linux | ✅ | ✅ | Exception (expected) |
| Linux | ✅ | ❌ | Works as expected |
| Linux | ❌ | ✅ | Exception (expected) |
| Linux | ❌ | ❌ | Exception - I consider this scenario a misconfiguration |
| --- | --- | --- | --- |
| Windows | ✅ | ✅ | Exception (expected) |
| Windows | ✅ | ❌ | Works as expected |
| Windows | ❌ | ✅ | Exception (expected) |
| Windows | ❌ | ❌ | **Falls back to NLS behavior!** |

lbussell:
- Fixes #6034 

NLS is Windows-only, so I defined a completely separate scenario for it that runs only on Windows.

The test uses the [recommended method](https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#determine-if-your-app-is-using-icu) for determining whether or not we're using ICU, and then validates some of the differences between NLS and ICU behavior based on what it finds.

For now, we will expect that NLS is used on all Nano Server images, and ICU is used on all Windows Server Core images. ICU has been the default on Server Core since .NET 7 and Server Core LTSC 2019. See https://learn.microsoft.com/en-us/dotnet/core/compatibility/globalization/7.0/icu-globalization-api for more context there.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NLSTest.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NLSTest.cs
@@ -1,0 +1,87 @@
+using System.Globalization;
+using static System.Console;
+
+const string nlsVar = "DOTNET_SYSTEM_GLOBALIZATION_USENLS";
+const string icuVar = "ICU_EXPECTED";
+
+GetEnvironmentVariableValue(nlsVar);
+bool icuExpected = GetEnvironmentVariableValue(icuVar);
+
+bool icuMode = ICUMode();
+WriteLine($"Detected ICU mode: {icuMode}");
+
+if (icuMode != icuExpected)
+{
+    throw new Exception($"ICU mode detected as {icuMode}, expected {icuExpected}");
+}
+
+// https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#stringendswith
+Assert(
+    """ "abc".EndsWith("\0")""",
+    "abc".EndsWith("\0").ToString(),
+    icuResult: true.ToString(),
+    nlsResult: false.ToString(),
+    icuMode);
+Assert(
+    """ "abc".EndsWith("\0", StringComparison.CurrentCulture)""",
+    "abc".EndsWith("\0", StringComparison.CurrentCulture).ToString(),
+    icuResult: true.ToString(),
+    nlsResult: false.ToString(),
+    icuMode);
+
+// https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#stringstartswith
+Assert(
+    """ "foo".StartsWith("\0")""",
+    "foo".StartsWith("\0").ToString(),
+    icuResult: true.ToString(),
+    nlsResult: false.ToString(),
+    icuMode);
+Assert(
+    """ "foo".StartsWith("\0", StringComparison.CurrentCulture)""",
+    "foo".StartsWith("\0", StringComparison.CurrentCulture).ToString(),
+    icuResult: true.ToString(),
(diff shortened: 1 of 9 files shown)


tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NLSTest.cs
```diff
@@ -0,0 +1,87 @@
...
+    """ "Hel\0lo".IndexOf("\0", StringComparison.CurrentCulture)""",
+    "Hel\0lo".IndexOf("\0", StringComparison.CurrentCulture).ToString(),
+    icuResult: "0",
+    nlsResult: "3",
+    icuMode);
+
+WriteLine("All assertions passed!");
+
+// https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#determine-if-your-app-is-using-icu
+bool ICUMode()
```
mthalman, line 63:
```suggestion
bool IsIcuMode()
```

tests/Microsoft.DotNet.Docker.Tests/TestScenarios/NLSScenario.cs
```diff
@@ -0,0 +1,82 @@
...
+        DotNetImageRepo repo,
+        DockerHelper dockerHelper)
+    {
+        _imageData = imageData;
+        _repo = repo;
+        _dockerHelper = dockerHelper;
+    }
+
+    // ICU is not supported on Nano Server
+    private bool ICUSupported => _imageData.OS.Contains(OS.ServerCore);
```
mthalman, line 38:
```suggestion
    private bool IsIcuSupported => _imageData.OS.Contains(OS.ServerCore);
```

tests/Microsoft.DotNet.Docker.Tests/TestScenarios/NLSScenario.cs
```diff
@@ -0,0 +1,82 @@
...
+            contextDir: _tempFolderContext.Path,
+            pull: Config.PullImages,
+            buildArgs: buildArgs);
+
+        string containerName = ImageData.GenerateContainerName(nameof(NLSScenario));
+        Func<string> runImage = () => _dockerHelper.Run(tag, containerName);
+
+        string justification = $"image {runtimeImage} should{(ICUSupported ? "" : " not")} support ICU";
+        runImage.Should().NotThrow(because: justification)
+            .Which.Should().NotBeNullOrWhiteSpace()
```
mthalman, line 77: Isn't `NotBeNullOrWhiteSpace` redundant based on the next `ContainEquivalentOf` assertion?

mthalman (approved):

PR #6151, opened 2025-01-16
Update dotnetcli domain in samples

richlander:
Removing last bits of "dotnetcli." usage.

Related: https://github.com/dotnet/core/issues/9671


lbussell (approved):

lbussell (changes requested):
Actually, can you split this PR into two @richlander.

update-dependencies changes should be targeted at the nightly branch, and samples can be updated in the main branch.

lbussell (approved):

PR #6176, opened 2025-01-28
Add chisel manifest slice to .NET 10 noble images

### LINKED ISSUE #6135: Add Chisel manifest to Ubuntu Chiseled images
This is a follow-up to https://github.com/dotnet/dotnet-docker/issues/5973.

The Chisel Manifest is eventually supposed to be the main way to scan Chisel images for CVEs.

We must continue using [chisel-wrapper](https://github.com/canonical/rocks-toolbox/blob/main/chisel-wrapper) to generate dpkg cache files until the Chisel manifest is widely supported by scanning tools.

Since existing images already have a dpkg cache file, and Chisel manifests aren't widely supported yet, we should only add the Chisel manifest to new images (.NET 10+).

lbussell:
Fixes #6135


mthalman (commented):
What do you think of including a test that validates the installation of the manifest?

lbussell:
> What do you think of including a test that validates the installation of the manifest?

I think that's a good idea, I'll add that today.

PR #6177, opened 2025-01-28
Update Dockerfile

byron-perez:
change deprecated "-o /app" for "--property:PublishDir=/app"

### DIFF
--- a/samples/aspnetapp/Dockerfile
+++ b/samples/aspnetapp/Dockerfile
@@ -10,7 +10,7 @@ RUN dotnet restore -a $TARGETARCH
 
 # Copy source code and publish app
 COPY --link aspnetapp/. .
-RUN dotnet publish -a $TARGETARCH --no-restore -o /app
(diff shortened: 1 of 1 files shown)


baronfel:
We shouldn't take this change. There is no difference between `-o` and `-p:PublishDir` when publishing, and the thing that is being warned on here is publishing at the solution level at all. It's inherently racy, and we have a long history of build problems that all come down to folks building multiple projects to the same output directory.

Fortunately, in this repo we do not need the solution to publish - a real fix here would change the Dockerfile from publishing the .sln file and instead don't copy the solution into the Docker build, publishing only the csproj.

lbussell:
@byron-perez would you like to implement @baronfel's suggestion? I'm thinking we would either want to add `*.sln` to the `.dockerignore`, or run `dotnet restore` on the specific `.csproj` file.

Otherwise, I'll close this PR and I'll follow up on this in a separate PR/issue later. Thanks!

Issue #6199, opened 2025-02-04
No tests are being executed in the `Linux_amd64 8.0-azurelinux3.0-runtime-deps` job


mthalman:
[Example pipeline run](https://dev.azure.com/dnceng-public/public/_build/results?buildId=933703&view=results)

Test Summary:

```
Skipped! - Failed:     0, Passed:     0, Skipped:    58, Total:    58, Duration: 40 ms - Microsoft.DotNet.Docker.Tests.dll (net9.0)
```

I looked at an internal pipeline run where images were actually built, not just pulled as cached images, and the same thing was happening there. This is the only job where this behavior seems to be occurring.

lbussell:
The test data is missing: https://github.com/dotnet/dotnet-docker/blob/f528525ecaa407f6ba09384c321d4611afc461e0/tests/Microsoft.DotNet.Docker.Tests/TestData.cs#L17-L53

This would be avoided by implementing https://github.com/dotnet/dotnet-docker/issues/5337.

PR #6203, opened 2025-02-04
Add missing TestData for Azure Linux 3.0 Amd64

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/6199


mthalman (approved):
In lieu of fixing https://github.com/dotnet/dotnet-docker/issues/5337, an easy change to make would be to make a helper method that would produce this common set of `ProductImageData` objects. I know there's variation in some cases but there's a lot of commonality here that could be captured in a method. I'm not suggesting that this be done in this PR.

PR #6204, opened 2025-02-05
Scrub servicing versions for internal Dockerfile snapshot tests

lbussell:
This PR fixes the test failures in:
- https://github.com/dotnet/dotnet-docker/pull/6200
- https://github.com/dotnet/dotnet-docker/pull/6201
- https://github.com/dotnet/dotnet-docker/pull/6202

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/DockerfileHelper.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/DockerfileHelper.cs
@@ -19,12 +19,9 @@ public static partial class DockerfileHelper
     [GeneratedRegex("[A-Fa-f0-9]{64}")]
     public static partial Regex Sha256Regex { get; }
 
-    // Match 1.2.3-servicing.45678.9
-    [GeneratedRegex(@"\d+\.\d+\.\d+-servicing\.\d+.\d+")]
-    public static partial Regex ServicingVersionRegex { get; }
-
-    [GeneratedRegex(@"\d+\.\d+\.\d+")]
-    public static partial Regex SemanticVersionRegex { get; }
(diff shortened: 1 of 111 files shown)


tests/Microsoft.DotNet.Docker.Tests/DockerfileHelper.cs
```diff
@@ -19,7 +19,11 @@ public static partial class DockerfileHelper
     [GeneratedRegex("[A-Fa-f0-9]{64}")]
     public static partial Regex Sha256Regex { get; }
 
-    [GeneratedRegex(@"\d+\.\d+\.\d+(?!\.windows)")]
+    // Match 1.2.3-servicing.45678.9
+    [GeneratedRegex(@"\d+\.\d+\.\d+-servicing\.\d+.\d+")]
+    public static partial Regex ServicingVersionRegex { get; }
+
+    [GeneratedRegex(@"\d+\.\d+\.\d+")]
```
jander-msft, line 26: I think the existing semantic version regex should account for an optional label e.g. `\d+\.\d+\.\d+(?:-[a-zA-Z0-9]+\.\d+\.\d+)`. This allows the version to mutate over time without having to regenerate the baselines. For example, for .NET Monitor, we typically insert a servicing version into nightly, insert the same version into main as a final version, and then merge that back to nightly. Then the process starts over for each upcoming release. This would cause us to update the baseline at the start of the next version with a label in it and when it is merged from main to nightly. We can avoid this need to constantly update the baseline if the regex considers an optional label.

lbussell: Good call - this should account for `preview.X` releases as well.

jander-msft (approved):

PR #6207, opened 2025-02-06
Test testdata

lbussell:
Issues like https://github.com/dotnet/dotnet-docker/issues/6199 should not be possible. We need to have confidence that we're testing all of our images.

There is an issue, https://github.com/dotnet/dotnet-docker/issues/5337, which tracks eliminating the test data by using the manifest/image info files instead, but that work is more complicated. @mthalman suggested [generating the test data](https://github.com/dotnet/dotnet-docker/pull/6203#pullrequestreview-2595818985). However, we would first need a way to have confidence that what we generate covers all of the images we have.

With all that in mind, I implemented this low-cost test which reads the tags from the manifest, and validates that each image in the manifest has an associated entry in the TestData. It already found several missing test datas which is a win in itself.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
@@ -136,10 +136,15 @@ namespace Microsoft.DotNet.Docker.Tests
 
         public static string GetImageName(string tag, string repoName, string repoNameModifier = null)
         {
-            string repo = $"dotnet{repoNameModifier ?? GetRepoNameModifier()}/{repoName}";
+            string repo = GetRepoName(repoName, repoNameModifier);
             string registry = GetRegistryName(repo, tag);
 
             return $"{registry}{repo}:{tag}";
+        }
+
+        public static string GetRepoName(string repoName, string repoNameModifier = null)
+        {
+            return $"dotnet{repoNameModifier ?? GetRepoNameModifier()}/{repoName}";
         }
 
         protected string GetTagName(string tagPrefix, string os, string tagPostfix = null)
@@ -149,9 +154,9 @@ namespace Microsoft.DotNet.Docker.Tests
             return string.Join('-', tagParts);
         }
 
-        protected virtual string GetArchTagSuffix() => (Arch == Arch.Amd64 && !DockerHelper.IsLinuxContainerModeEnabled)
-            ? string.Empty
-            : GetArchLabel();
+        protected virtual string GetArchTagSuffix() => (Arch == Arch.Amd64 && IsWindows)
+                ? string.Empty
+                : GetArchLabel();
 
         protected string GetArchLabel() =>
             Arch switch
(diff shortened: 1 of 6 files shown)


lbussell:
@jander-msft, there's an issue with the 9.0 Monitor images. It seems like the tags weren't updated to reflect the lack of OS in the tags. I need to also check if this affects the other appliance images, but they didn't throw any test failures.

I looked at a couple of test result files and I'm not convinced that the 9.0 monitor tests are running.

Also, there are no `monitor/base` tests whatsoever.

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -136,22 +136,27 @@ protected void PullImageIfNecessary(string imageName, DockerHelper dockerHelper,
...
             tagParts = tagParts.Where(s => !string.IsNullOrEmpty(s));
             return string.Join('-', tagParts);
         }
 
-        protected virtual string GetArchTagSuffix() => (Arch == Arch.Amd64 && !DockerHelper.IsLinuxContainerModeEnabled)
-            ? string.Empty
-            : GetArchLabel();
+        protected virtual string GetArchTagSuffix() => (Arch == Arch.Amd64 && IsWindows)
+                ? string.Empty
+                : GetArchLabel();
```
mthalman, line 159: revert?

tests/Microsoft.DotNet.Docker.Tests/ImageData.cs
```diff
@@ -136,22 +136,27 @@ protected void PullImageIfNecessary(string imageName, DockerHelper dockerHelper,
...
+            string repo = GetRepoName(repoName, repoNameModifier);
             string registry = GetRegistryName(repo, tag);
 
             return $"{registry}{repo}:{tag}";
         }
 
+        public static string GetRepoName(string repoName, string repoNameModifier = null)
+        {
+            return $"dotnet{repoNameModifier ?? GetRepoNameModifier()}/{repoName}";
+        }
```
mthalman, line 148:
```suggestion
        public static string GetRepoName(string repoName, string repoNameModifier = null) =>
            $"dotnet{repoNameModifier ?? GetRepoNameModifier()}/{repoName}";
```

tests/Microsoft.DotNet.Docker.Tests/TestDataTests.cs
```diff
@@ -0,0 +1,82 @@
...
+    [Theory]
+    [MemberData(nameof(ImageRepos))]
+    public void VerifyTestData(DotNetImageRepo imageRepo)
+    {
+        string repoName = ImageData.GetRepoName(ProductImageData.GetDotNetImageRepoName(imageRepo));
+
+        if (repoName.Contains("base"))
+        {
+            return;
+        }
```
mthalman, line 41: Can you add a comment why this is skipped here? It's not clear to me why that is.

lbussell: It is because there are no `monitor/base` tests whatsoever. It needs a better comment.

lbussell: It was actually not an issue because we are getting the test datas from the DotNetImageRepo Enum values and there is no entry for monitor/base.

lbussell:
> It seems like the [monitor test datas] weren't updated to reflect the lack of OS in the tags.

I fixed this issue in the test data, so the test data validation passes now. I am still concerned because it hasn't been affecting the tests.

mthalman (approved):

PR #6213, opened 2025-02-07
Updates for .NET February 2025 Servicing

lbussell:


tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/monitor-8.0-cbl-mariner-distroless-amd64-Dockerfile.approved.txt
```diff
@@ -0,0 +1,32 @@
...
+
+RUN tdnf install -y \
+        ca-certificates \
+        gzip \
+        tar \
+    && tdnf clean all
+
+# Retrieve .NET Monitor extensions
+RUN dotnet_monitor_extension_version=0.0.0 \
+    && curl -u :$ACCESSTOKEN --basic -fSL --output dotnet-monitor-egress-azureblobstorage.tar.gz "https://artifacts.visualstudio.com/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-azureblobstorage-$dotnet_monitor_extension_version-linux-x64.tar.gz" \
```
jander-msft, line 14: Is this commit correct? https://github.com/lbussell/dotnet-docker/commit/03069a41dccc67f02bd0aaf1e2d6ebf224b60da7

It seems to be undoing the changes from https://github.com/lbussell/dotnet-docker/commit/38fb3b04b15e9285fc37d9c741e60c2bb49a793c

I expect the baseline to not have the `$dotnet_monitor_extension_version` version variable "directory" in the download URI. The `nightly` branch does not have it: https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/monitor-8.0-cbl-mariner-distroless-amd64-Dockerfile.approved.txt#L14

jander-msft: Although, it seems that the runtime baselines do have this variable in their baselines in both `main` and `nightly`.

lbussell: I think that, whenever the monitor version is stable, it uses the variable, otherwise it uses the full servicing version (e.g. 8.0.8-servicing.25104.1).

My commits were always just applying whatever baseline changes the tool created so that the test would pass.

jander-msft (approved):

mthalman (approved):

Issue #6212, opened 2025-02-06
Replace dotnet/reverse-proxy images to dotnet/yarp


benjaminpetit:
Hi,

Since the yarp project migrated from `microsoft/reverse-proxy` to `dotnet/yarp`, we want to reflect this image change in the container. The new container image name should be `dotnet/yarp`.

We only shipped a preview version in the nightly branch.

Steps needed:

- [x] Add a new entry for the new images in https://github.com/microsoft/mcr (https://github.com/microsoft/mcr/pull/3848)
- [x] Rename the existing doc and docker container in this repo to `dotnet/yarp` (https://github.com/dotnet/dotnet-docker/pull/6215)
- [x] Update docs links to reflect the github change (https://github.com/dotnet/dotnet-docker/pull/6215)

To be confirmed (if possible to do):

- [x] De-list `dotnet/reverse-proxy` in mcr so it doesn't show up on the portal  (https://github.com/microsoft/mcr/pull/3848)

Related issue: https://github.com/dotnet/yarp/issues/2736

lbussell:
This looks complete to me

PR #6215, opened 2025-02-07
Rename reverse-proxy image to yarp

benjaminpetit:
Fix https://github.com/dotnet/dotnet-docker/issues/6212

Related issue on MCR: https://github.com/microsoft/mcr/pull/3848

Binairies used haven't changed, the only relevant change in the Dockerfile is the use of the config file `/etc/yarp.config` instead of `/etc/reverse-proxy.config`.

The links to the yarp repos, docs, and samples were updated.


lbussell (approved):
Broken links are all to `https://github.com/dotnet/dotnet-docker/blob/nightly/README.yarp.md` which is expected.

LGTM but needs to wait on https://github.com/microsoft/mcr/pull/3848 to be merged.

PR #6230, opened 2025-02-12
Initial .NET Monitor 9.1 Preview Images

jander-msft:


tests/Microsoft.DotNet.Docker.Tests/ImageVersion.cs
```diff
@@ -14,6 +14,7 @@ public readonly record struct ImageVersion
         public static readonly ImageVersion V8_1 = new(new Version(8, 1), isPreview: false);
         public static readonly ImageVersion V9_0 = new(new Version(9, 0), isPreview: false);
         public static readonly ImageVersion V9_0_Preview = new(new Version(9, 0), isPreview: true);
```
mthalman, line 16: Is this in use anymore?

jander-msft: It's used by the Aspire dashboard

lbussell:
The pre-build validation tests are saying that in nightly, the `9` floating tag should be on 9.1.

jander-msft:
> The pre-build validation tests are saying that in nightly, the `9` floating tag should be on 9.1.

I see that now. I didn't run all of the correct tests before pushing. However, is this the desired behavior? The 9.1 image is a preview image, so it probably shouldn't have the `9` major tag or the `latest` tag on it (in fact, the `latest` tag tests probably aren't correct because they are not failing while its still on the 9.0 images). This assertion is the behavior in the `main` branch.

Should the nightly branch exclusion [here](https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs#L555) be removed? I can just blindly follow what the test current is, but I want to double-check that this is the desired behavior in the nightly branch.

lbussell:
> is this the desired behavior? The 9.1 image is a preview image, so it probably shouldn't have the `9` major tag or the `latest` tag on it 

It is my opinion that if you're on the nightly branch, using the latest tag, you should get the most bleeding edge thing we offer. This is what I've done with .NET 10 previews as well.

I see what you're saying though, since this image would have a `9-preview` and not a `9` tag?
If you are able to amend the tests to fix that, I think that'd be acceptable. Something like this?

- nightly
  - .NET Monitor 9.0: `9`, `9.0`
  - .NET Monitor 9.1: `9-preview`, `9.1-preview`, `latest`
- main
  - .NET Monitor 9.0: `9`, `9.0`, `latest`
  - .NET Monitor 9.1: `9-preview`, `9.1-preview`

> the `latest` tag tests probably aren't correct because they are not failing while its still on the 9.0 images

Hmm, yeah, perhaps `LatestTag_OnCorrectMajorVersion` should instead be `LatestTag_OnCorrectMajorMinorVersion`. Good catch. Can you file an issue for that?

> Should the nightly branch exclusion [here](https://github.com/dotnet/dotnet-docker/blob/nightly/tests/Microsoft.DotNet.Docker.Tests/StaticTagTests.cs#L555) be removed? I can just blindly follow what the test current is, but I want to double-check that this is the desired behavior in the nightly branch.
>
> If I do remove that condition, thus making the policy the same between nightly and main regarding the major tags, then everything passes for all images.

Does it pass for .NET 10? I would like to keep `latest` on .NET 10 in `nightly`.

lbussell (approved):

PR #6238, opened 2025-02-18
Add devcontainer configuration

lbussell:
This will enable:

- Codespaces
- GitHub Copilot Workspaces (preview)

I validated that I can run comands like `Get-GeneratedDockerfiles.ps1` successfully. Non-root user is not currently working with the Docker-in-Docker feature. I will file an issue on https://github.com/devcontainers/features for that.


lbussell:
Got non-root user working. Node is also not needed. Stole some other configs from dotnet/aspire.

Issue #6244, opened 2025-02-19
Install all .NET binaries using an installer stage in Dockerfiles


lbussell:
### Describe the Problem

We need to build images internally using private .NET binaries. We'd like for them to have a similar/identical `docker history` output to images that are published from public .NET binaries. We also don't want any secrets to be leaked through `docker history`.

### Describe the Solution

All .NET binaries should be installed via a separate "installer" stage. We already follow this pattern in some Dockerfiles, so we just need to apply it to the rest of the images. As a bonus, it should simplify the Dockerfile templates a little.

Images

- Linux .NET images - https://github.com/dotnet/dotnet-docker/pull/6245
- Windows .NET images - https://github.com/dotnet/dotnet-docker/pull/6249
- Appliance images
  - .NET Monitor - nothing to do
  - .NET Aspire Dashboard - nothing to do
  - YARP - nothing to do

### Additional Context

Part of https://github.com/dotnet/dotnet-docker-internal/issues/6782. Follow-up to https://github.com/dotnet/dotnet-docker-internal/issues/6844.

lbussell:
I filed https://github.com/dotnet/dotnet-docker/issues/6251 as clean-up work.

PR #6245, opened 2025-02-19
Update Linux .NET images to use an installer stage

lbussell:
Part of https://github.com/dotnet/dotnet-docker/issues/6244


src/aspnet/10.0/alpine3.21/amd64/Dockerfile
```diff
@@ -1,15 +1,21 @@
...
     && rm aspnetcore.tar.gz
+
+
+# ASP.NET Core image
+FROM $REPO:10.0.0-preview.1-alpine3.21-amd64
+
+# ASP.NET Core version
+ENV ASPNET_VERSION=10.0.0-preview.1.25107.4
+
+COPY --from=installer ["/shared/Microsoft.AspNetCore.App", "/usr/share/dotnet/shared/Microsoft.AspNetCore.App"]
```
mthalman, line 21: aspnet is actually being installed at `/usr/share/dotnet/shared/Microsoft.AspNetCore.App` in the installer stage, not `/shared/Microsoft.AspNetCore.App`.

lbussell:
There is a LOT of simplification and clean-up that can happen in these templates. I think we can remove lots of unused conditions and sub-templates since with this change there will essentially be only one single pattern for Dockerfiles. 

I intentionally didn't start going down that path as I didn't want to get carried away - I made the smallest changes possible to get the Dockerfiles looking correct.

@mthalman or @MichaelSimons what are your thoughts on addressing clean-up in this PR vs. follow-up(s)?

mthalman:
I'm fine with addressing cleanup in separate PR.

mthalman (approved):

MichaelSimons:
> @mthalman or @MichaelSimons what are your thoughts on addressing clean-up in this PR vs. follow-up(s)?

I would prefer we complete the e2e, meaning we can pre-stage, then circle back and do these type of clean-up tasks.

PR #6249, opened 2025-02-20
Use installer stage for Windows Server Core Dockerfiles

lbussell:
Part of https://github.com/dotnet/dotnet-docker/issues/6244

### DIFF
--- a/src/sdk/10.0/windowsservercore-ltsc2019/amd64/Dockerfile
+++ b/src/sdk/10.0/windowsservercore-ltsc2019/amd64/Dockerfile
@@ -1,23 +1,9 @@
 # escape=`
 
 ARG REPO=mcr.microsoft.com/dotnet/aspnet
-FROM $REPO:10.0.0-preview.1-windowsservercore-ltsc2019
 
-ENV `
-    # Do not generate certificate
-    DOTNET_GENERATE_ASPNET_CERTIFICATE=false `
-    # Do not show first run text
-    DOTNET_NOLOGO=true `
-    # SDK version
-    DOTNET_SDK_VERSION=10.0.100-preview.1.25112.18 `
-    # Enable correct mode for dotnet watch (only mode supported in a container)
-    DOTNET_USE_POLLING_FILE_WATCHER=true `
-    # Skip extraction of XML docs - generally not useful within an image/container - helps performance
-    NUGET_XMLDOC_MODE=skip `
-    # PowerShell telemetry for docker image usage
-    POWERSHELL_DISTRIBUTION_CHANNEL=PSDocker-DotnetSDK-WindowsServerCore-ltsc2019 `
-    # Workaround for https://github.com/PowerShell/PowerShell/issues/20685
-    DOTNET_ROLL_FORWARD=Major
+# Installer image
+FROM mcr.microsoft.com/windows/servercore:ltsc2019-amd64 AS installer
 
 # Download MinGit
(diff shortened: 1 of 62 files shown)


src/sdk/10.0/windowsservercore-ltsc2019/amd64/Dockerfile
```diff
@@ -61,7 +49,30 @@ RUN powershell -Command " `
         Remove-Item -Force PowerShell.Windows.x64.$powershell_version.nupkg; `
         Remove-Item -Path $Env:ProgramFiles\powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force;"
```
mthalman, line 51: There should be another line separating the stages.

src/sdk/10.0/windowsservercore-ltsc2019/amd64/Dockerfile
```diff
@@ -44,10 +44,17 @@ RUN powershell -Command " `
...
         }; `
-        & $Env:ProgramFiles\dotnet\dotnet tool install --add-source . --tool-path $Env:ProgramFiles\powershell --version $powershell_version PowerShell.Windows.x64; `
-        & $Env:ProgramFiles\dotnet\dotnet nuget locals all --clear; `
+        & \dotnet\dotnet tool install --add-source . --tool-path \powershell --version $powershell_version PowerShell.Windows.x64; `
+        & \dotnet\dotnet nuget locals all --clear; `
         Remove-Item -Force PowerShell.Windows.x64.$powershell_version.nupkg; `
-        Remove-Item -Path $Env:ProgramFiles\powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force;"
+        Remove-Item -Path \powershell\.store\powershell.windows.x64\$powershell_version\powershell.windows.x64\$powershell_version\powershell.windows.x64.$powershell_version.nupkg -Force; `
+        `
+        # Delete everything in the dotnet folder that's not needed in the SDK layer but will instead be derived from base layers
```
mthalman, line 52: I don't know why we do this rather than what it was originally with extracting from the tarball only what we wanted. I concede that I probably authored this originally. But I don't like it. There must have been a reason.

lbussell: I'm happy to try and switch to the `tar` method for both images, or switch back just for windowsservercore to keep things as-is.

mthalman: No, that's ok. I didn't realize that this logic has been around forever. I thought perhaps it was a thing only for internal Dockerfiles that was now getting used here, but it's not.

lbussell: Oh, I know. It's because we need to extract the whole SDK in order to run the `dotnet tool install ...` commands. If we didn't extract the runtime too, then those commands would fail. But in a single stage scenario, when you're extracting on top of the aspnet image, the runtime is already there so it's not a problem to extract fewer files.

mthalman: I knew there must have been a sensible reason. 🙂

mthalman (approved):

Issue #6255, opened 2025-02-21
Add internal build support to YARP Dockerfile templates


lbussell:
See https://github.com/dotnet/dotnet-docker/issues/6252 for context.

PR #6286, opened 2025-02-27
Add ACCESSTOKEN support to Aspire Dashboard and YARP internal Dockerfiles 

lbussell:
This is part of #6254 and #6255. 

As there isn't yet support for Aspire Dashboard or YARP in the update-dependencies tooling, I can't tell what the precise download paths will be, so I didn't change them yet. However, this PR sets up the plumbing for using the ACCESSTOKEN arg for these Dockerfile builds so we don't have to do it later.


tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/aspire-dashboard-9.1-azurelinux-distroless-arm64v8-Dockerfile.approved.txt
```diff
@@ -1,4 +1,5 @@
 ARG REPO=mcr.microsoft.com/dotnet/aspnet
+ARG ACCESSTOKEN
```
mthalman, line 2: This should be defined within the `installer` stage, just as was done for the yarp Dockerfile. In fact, it needs to be. Otherwise, you'd can't access the ARG within the stage unless it is redeclared within that stage.

mthalman (approved):

PR #6298, opened 2025-03-05
Add .NET 10 AOT SDK images

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/5020


mthalman (approved):
Note that https://github.com/dotnet/dotnet-docker/issues/5020 shouldn't be closed until this is rolled out to main.

What is the plan for the .NET 8 & 9 AOT images in nightly?

tests/Microsoft.DotNet.Docker.Tests/TestData.cs
```diff
@@ -279,6 +278,11 @@ public static class TestData
                     ImageVariant = DotNetImageVariant.Composite | DotNetImageVariant.Extra, SupportedImageRepos = DotNetImageRepo.Aspnet },
             new ProductImageData { Version = V10_0_Preview, OS = OS.TrixieSlim,        Arch = Arch.Amd64 },
 
+            new() { Version = V10_0_Preview, Arch = Arch.Amd64, SdkImageVariant = DotNetImageVariant.AOT, SupportedImageRepos = DotNetImageRepo.Runtime_Deps, OS = OS.Alpine321 },
+            new() { Version = V10_0_Preview, Arch = Arch.Amd64, SdkImageVariant = DotNetImageVariant.AOT, SupportedImageRepos = DotNetImageRepo.Runtime_Deps, OS = OS.AzureLinux30Distroless, SdkOS = OS.AzureLinux30 },
+            new() { Version = V10_0_Preview, Arch = Arch.Amd64, SdkImageVariant = DotNetImageVariant.AOT, SupportedImageRepos = DotNetImageRepo.Runtime_Deps, OS = OS.NobleChiseled, SdkOS = OS.Noble },
+            new() { Version = V10_0_Preview, Arch = Arch.Amd64, SdkImageVariant = DotNetImageVariant.AOT, SupportedImageRepos = DotNetImageRepo.Runtime_Deps, OS = OS.TrixieSlim },
```
lbussell, line 284: Just acknowledging here that I arranged these TestDatas in a different format in order to make it more obvious what was different between each entry. I didn't apply it to the whole file so that there wasn't a huge diff. We can apply that later to the rest of the file.

lbussell:
> What is the plan for the .NET 8 & 9 AOT images in nightly?

AFAIK there are users using .NET 8 and 9 AOT nightly images in production. Because of that I don't think we can/should remove them.

I see our options as:

1. Do nothing
2. Port them to main as-is
3. Port them to main in this new format we have for .NET 10
    - I imagine if we did this, we'd remove runtime-deps AOT images in nightly, but add the `-aot` tag as undocumented to the existing runtime-deps images so that users with those tags don't need to take any action.

PR #6331, opened 2025-03-24
Yarp/otlp

benjaminpetit:
Update YARP binaries and add docs on OpenTelemetry support.


samples/otlptestlistener/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+
+# Runtime stage
+FROM mcr.microsoft.com/dotnet/aspnet:9.0-azurelinux3.0-distroless
+EXPOSE 8080
+EXPOSE 4317
+WORKDIR /app
+COPY --link --from=build /app .
+USER $APP_UID
+ENTRYPOINT ["./OtlpTestListener"]
```
lbussell, line 24: It doesn't look like this sample is showcasing any containers-specific .NET functionality. Since it's used in the YARP test, I think it would fit better under `tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts`.

benjaminpetit: Moved

benjaminpetit:
Unrelated test failures

eng/readme-templates/Use.yarp.md
```diff
@@ -54,4 +56,16 @@ This example will proxy every requests from `http://localhost:5000/aspnetapp` to
...
+
+{{ARGS["top-header"]}}# OpenTelemetry support
+
+This image supports OpenTelemetry. It can be configured by passing environment variables to the container:
+
+```bash
+docker run --rm -v $(pwd)/my-config.config:/etc/yarp.config -p 5000:5000 -e OTEL_EXPORTER_OTLP_ENDPOINT=https://otlp-endpoint.internal:4317 mcr.microsoft.com/dotnet/yarp:latest
+```
+
+See the [OTLP Exporter Configuration](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) for all available variables.
```
lbussell, line 69:
```suggestion
See the [OTLP Exporter Configuration](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) for all supported environment variables.
```

eng/readme-templates/Use.yarp.md
```diff
@@ -54,4 +56,16 @@ This example will proxy every requests from `http://localhost:5000/aspnetapp` to
...
+
+You can skip https validation for the OTLP endpoint only by passing the environment variable `YARP_UNSAFE_OLTP_CERT_ACCEPT_ANY_SERVER_CERTIFICATE`.
```
lbussell, line 71:
```suggestion
You can skip HTTPS validation for the OTLP endpoint only by passing the environment variable `YARP_UNSAFE_OLTP_CERT_ACCEPT_ANY_SERVER_CERTIFICATE`.
```

manifest.versions.json
```diff
@@ -215,13 +215,13 @@
     "powershell|10.0|Linux|x64|sha": "ac4eb5cb83e130aa7cc9c856245635d3f4e624b8d4e12933f5cd534f02545e2f4c61b7f63bc899d7ad797a70ec5c2d7bca33e57ea6eb7da0a410808d3ad2cf94",
     "powershell|10.0|Windows|x64|sha": "6f4518a04d9fd3ade4c7fa76574b29f6abb7e8a37793fd2c14a123700dfd8fa69ff4a200287ac909942bd687209ead999f53fbb2727ae96200d669564e1935eb",
 
-    "yarp|2.3|build-version": "2.3.0-preview.1.25060.1",
-    "yarp|2.3|product-version": "2.3.0-preview.1",
+    "yarp|2.3|build-version": "3.0.0-preview.1.25171.2",
+    "yarp|2.3|product-version": "2.3.0-preview.2",
```
lbussell, line 219: Looks like this version should be updated:

```suggestion
    "yarp|3.0|build-version": "3.0.0-preview.1.25171.2",
    "yarp|3.0|product-version": "3.0.0-preview.1",
```

Also applies to all other `yarp|*` variables below.

src/yarp/2.3/azurelinux-distroless/amd64/Dockerfile
```diff
@@ -9,9 +9,9 @@ RUN tdnf install -y \
     && tdnf clean all
```
lbussell, line 9: Since the version changed, the Dockerfiles should be moved to `src/yarp/3.0/...`

tests/Microsoft.DotNet.Docker.Tests/Microsoft.DotNet.Docker.Tests.csproj
```diff
@@ -30,4 +30,14 @@
...
+  <ItemGroup>
+    <Content Include="TestAppArtifacts/otlptestlistener/**/*">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="TestAppArtifacts/otlptestlistener/**/*.cs" />
+  </ItemGroup>
+
```
lbussell, line 42: Can these ItemGroups be combined with the ones above? e.g. one for all `<Content .../>` elements and one for all `<Compile .../>` elements.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/otlptestlistener/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Learn about building .NET container images:
+# https://github.com/dotnet/dotnet-docker/blob/main/samples/README.md
+FROM mcr.microsoft.com/dotnet/sdk:9.0-azurelinux3.0 AS build
+WORKDIR /source
+
+# Copy project file and restore as distinct layers
+COPY --link *.sln .
```
lbussell, line 7: Solution file is not necessary since there's just one project.

lbussell (commented):
Will review the rest of the files shortly

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/otlptestlistener/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY --link OtlpTestListener/*.csproj ./OtlpTestListener/
+RUN dotnet restore
```
lbussell, line 9: This project needs support for using internal NuGet feeds when building with internal .NET versions. There are some prior examples of this for other tests - the internal NuGet config file is present in TestAppArtifacts but you'll probably need to copy it into the project folder during the test.

lbussell (changes requested):
In general, this is a whole lot of code added just to test YARP's OpenTelemetry support.

For testing purposes, I think it's acceptable to use a library for the OTLP spec. Or even if there is an off the shelf container image we can use, that'd greatly simplify things. Could Aspire Dashboard do this? or will CORS get in the way?

Otherwise, we need documentation and/or automation to keep this project up-to-date.

manifest.versions.json
benjaminpetit: The YARP binaries are still from the 2.3.0 YARP release, it's just the daily build definition of yarp automatically naming the zip file 3.0.

src/yarp/2.3/azurelinux-distroless/amd64/Dockerfile
benjaminpetit: The YARP binaries are still from the 2.3.0 YARP release, it's just the daily build definition of yarp automatically naming the zip file 3.0.

benjaminpetit:
> In general, this is a whole lot of code added just to test YARP's OpenTelemetry support.
> 
> For testing purposes, I think it's acceptable to use a library for the OTLP spec. Or even if there is an off the shelf container image we can use, that'd greatly simplify things. Could Aspire Dashboard do this? or will CORS get in the way?
> 
> Otherwise, we need documentation and/or automation to keep this project up-to-date.

Aspire Dashboard can act as an OTLP endpoint, but there is no API to query the logs/metrics.

I could use an existing docker image like `grafana/otel-lgtm`, but it's a **1.7 GB** container. Your call.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/otlptestlistener/Dockerfile
lbussell: The project should use the appropriate NuGet config in the Dockerfile for security/compliance reasons. For an example, see: https://github.com/dotnet/dotnet-docker/blob/a93e9cc9fae5540a2108ab30cd4c83cf0132069f/tests/Microsoft.DotNet.Docker.Tests/TestSolution.cs#L60-L74

Feel free to factor out some of the highlighted logic to share it between tests.

W.r.t preview/internal .NET versions - all the other test Dockerfiles use parametrized SDK/Runtime base images, and the test code selects which image to use. But since this code is only meant to test the YARP image and not the .NET SDK image, leaving the reference to the live version of `mcr.microsoft.com/dotnet/sdk:9.0-azurelinux3.0` is probably more stable in this case. I am fine with that as-is. But we should still use the NuGet config.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/otlptestlistener/OtlpTestListener/Services/DefaultLogsService.cs
```diff
@@ -1,5 +1,3 @@
-namespace OtlpTestListener.Services;
-
 public class DefaultLogsService : LogsService.LogsServiceBase
```
lbussell, line 1: What's the reason for removing the namespace here?

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/otlptestlistener/OtlpTestListener/otlp/opentelemetry/proto/README.md
```diff
@@ -0,0 +1,5 @@
+# OpenTelemetry Proto
+
+See [open-telemetry on GitHub](https://github.com/open-telemetry/opentelemetry-proto) for all the sources.
+
+We only support v1 at this point.
```
lbussell, line 5: Who supports V1? YARP? Is there a place in the YARP repo where we can see what version this needs to be kept in line with?

And perhaps it'd be good to note which specific commit sha or tag the proto was copied from?

manifest.versions.json
lbussell: Is this a temporary situation? I don't understand which version of YARP we are delivering previews for. Is the next version a 2.3.0 stable release, 2.3.1, 2.4, or 3.0? The versions here should reflect that.

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/otlptestlistener/OtlpTestListener/Services/DefaultLogsService.cs
benjaminpetit: Because it was not needed?

tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/otlptestlistener/Dockerfile
lbussell: We can handle this one in a follow-up issue if you want.

benjaminpetit: If we could merge this update and fix it it a next PR that would be great yes

lbussell (approved):

PR #6341, opened 2025-03-27
Update mcrdocs access token variable

lbussell:
See https://github.com/dotnet/dotnet-docker-internal/issues/7655 for context.

If this works out, then I'll backport this to docker-tools so the change can flow to the other repos.


mthalman (approved):

lbussell:
Merging around PR validation since 1) this is blocking publishing and 2) affected code won't be exercised during PR validation.

PR #6352, opened 2025-04-01
Remove .NET 8 and 9 AOT runtime-deps images

lbussell:
See https://github.com/dotnet/dotnet-docker/discussions/6312

> .NET 8 and 9 nightly -aot images will no longer be produced. AOT-specific runtime-deps tags will re-tagged to non-AOT specific tags.


mthalman (commented):
Looks like there are some tag validation failures.

mthalman (approved):

PR #6372, opened 2025-04-14
Remove .NET Monitor 8.0

jander-msft:
.NET Monitor 8.1 superseded .NET Monitor 8.0 on February 11, 2025 as the latest .NET Monitor 8 release. As announced in https://github.com/dotnet/dotnet-monitor/discussions/7938, .NET Monitor 8.0 will be EOL on May 11, 2025, which is before the next planned servicing release. This change removes the Dockerfiles and related infrastructure for .NET Monitor 8.0 since the Docker images will no longer be maintained at the next planned servicing release.


jander-msft:
> Remove the test data:
> 
> https://github.com/dotnet/dotnet-docker/blob/d7c2d2c95c917865a388a39af8978c36b20829a5/tests/Microsoft.DotNet.Docker.Tests/TestData.cs#L350-L351
> 
> Also why isn't this failing? 😕 There might be an issue with monitor tests running:

I agree that the .NET Monitor test data needs to be removed (oops!), but this isn't indicative of the .NET Monitor tests not running. The tests are filtered using file paths to Dockerfiles based on the manifest. If the manifest doesn't have entries for the .NET Monitor 8.0 paths, then the .NET Monitor 8.0 image data is filtered out so they aren't run.

As for why Azure DevOps is saying they aren't executed (if you hover on the icon it has a tooltip of "Not executed"), I'm not sure. Here are the same visualizations for .NET 10 Azure Linux (which there aren't any .NET Monitor images yet):

Screenshot 2025-04-15 085938

and .NET 10 Windows Nano Server (where .NET Monitor has _never_ had any Windows images):

Screenshot 2025-04-15 090057

So in my mind, it showing "Not executed" for some test doesn't really mean anything.

In any event, I will remove the .NET Monitor 8.0 image test data.

lbussell:
> The tests are filtered using file paths to Dockerfiles based on the manifest. If the manifest doesn't have entries for the .NET Monitor 8.0 paths, then the .NET Monitor 8.0 image data is filtered out so they aren't run.

Ah, yeah, makes sense with the current setup. No concern from me then.

lbussell (approved):

Issue #6363, opened 2025-04-03
Consume .NET 10 builds from VMR instead of dotnet/sdk


lbussell:
For .NET 10 Preview 4, .NET will be built from the VMR as part of the unified build migration.
We need to update the logic in update-dependencies to consume new builds from https://github.com/dotnet/dotnet instead of https://github.com/dotnet/sdk like we do now.

UB documentation: https://github.com/dotnet/arcade/tree/main/Documentation/UnifiedBuild

lbussell:
Hey @martincostello, in regards to using the shortlink, that is _more or less_ to what this repo is doing [today](https://github.com/dotnet/dotnet-docker/blob/6a269116aaa1dcace32637417a04fd903bc76548/eng/Get-DropVersions.ps1#L265), except we also now need to support:

- Consuming arbitrary fixed builds that aren't just the latest version
- Consuming internal-only builds
- Resolving other product versions besides just the SDK
  - But I do see that the other shortlinks exist, e.g. https://aka.ms/dotnet/10.0/preview/runtime-productVersion.txt redirects to https://builds.dotnet.microsoft.com/dotnet/Runtime/10.0.0-preview.2.25163.2/runtime-productVersion.txt
  - Discoverability of those short URLs is difficult, aka.ms/bar is easier to reason about IMO

We have basically hacked in support for those other things in our existing PowerShell scripts but using [DarcLib](https://github.com/dotnet/arcade-services/tree/main/src/Microsoft.DotNet.Darc/DarcLib) is going to be a more scalable solution for us.

I think https://github.com/dotnet/arcade-services/ may be a better place to ask about your question, since it hosts all the code for the build asset registry. /cc @premun @mmitche

martincostello:
Thanks for the information. Sounds like the changes are mainly "behind the curtain" and what I currently do from outside where I only care about public builds isn't really changing for the VMR?

If there's an _easier_ way to consume things over HTTP (as my action is JavaScript so anything in NuGet isn't useful to it) that's publically accessible then I'm all ears for possible improvements.

lbussell:
Yeah, I think there should be no breaking changes for .NET consumers. If you notice something, then please file an issue 🙂

lbussell:
This is completed with #6379. If we need to roll-back, the old pipeline templates are still around for the time being, so we can just change which template we call in the update-dependencies pipeline. I will file an issue for cleaning up the old templates once they're no longer needed.

[...]

PR #6378, opened 2025-04-16
Refactor Update Dependencies project to facilitate adding additional commands

### LINKED ISSUE #3899: Upgrade to System.CommandLine 2.0.0 beta4
The update-dependencies project is currently implemented on beta1 of System.CommandLine. It should be updated to beta4 to remain current. There are breaking changes between these versions that need to be accounted for.

Current docs are available at https://docs.microsoft.com/dotnet/standard/commandline/.

lbussell:
### Motivation

- Update-dependencies tool relies too heavily on `Set-DotNetVersions.ps1` and `Get-DropVersions.ps1` to construct the correct set of arguments for update operations. It is desirable to use .NET for this instead.
- As part of #6363, we need to write/rewrite new logic to enable consuming builds from the VMR. The options were to write more PowerShell scripts or to refactor parts of this tool to enable consuming those VMR builds. Re-factoring this tool means we can easily use pre-existing .NET libraries written by the product construction team, so this is the preferred approach.
- So ultimately, update-dependencies needs to provide a way to define additional entrypoints (new subcommand(s)) that can call into the existing update-dependencies logic.

### Details

- The old root command (update-dependencies) is now a sub-command called "specific"
  - Intended usage:
    - `update-dependencies specific 10.0 --product-version foo=bar ...`
  - Potential future usage:
    - `update-dependencies from-channel 4567 https://github.com/dotnet/dotnet`
    - `update-dependencies from-builds 123456`
    - `update-dependencies from-builds 123456 133337 ...`
    - ...We can just infer product versions and Dockerfile versions from the BAR builds
- New sub-commands can instantiate `SpecificCommand` and call it using arguments, or use any of the other existing classes to facilitate updating versions, e.g. `DockerfileShaUpdater.CreateUpdaters(...) ...`

### Other changes

- `System.CommandLine` 2.0.0 Beta 5
  - No, this is not out yet, but presumably it will be soon. Once I learned there was a version 5, I didn't want to write new code in an older version. This project now references the dotnet-libraries nuget feed in order to get it.
  - Standard dependency injection setup which will allow services to be injected into individual command classes
  - Some boilerplate to handle command setup and registration
  - Use `System.CommandLine.NamingConventionBinder` to bind options to properties

Fixes #3899... sorta.


eng/update-dependencies/NuGet.config
mthalman, line 1: A nuget.config already exists at the repo root with two of these feeds defined there. Why not update that instead?

lbussell: It needs to be present when built from the Dockerfile. ...Which was my initial justification, but I just realized the Dockerfile is built with the repo root as the context. So you're right, I can use the root Nuget.config instead.

PR #6393, opened 2025-04-25
Use productCommit assets to resolve build versions when updating dependencies

lbussell:
In #6379, an incorrect assumption was made about the productVersion* text files. The versions of the Runtime and ASP.NET Core can indeed be different in the same VMR build.

We were using the `Runtime/*/productVersion.txt` file, but instead we should have been using the `Runtime/*/runtime-productVersion.txt` file. However there is no equivalent for ASP.NET.

This PR changes the from-channel command to use the `*productCommit.json` file directly from the SDK instead, which includes the correct product versions for the major SDK components.

### DIFF
--- a/eng/update-dependencies/BuildAssetService.cs
+++ b/eng/update-dependencies/BuildAssetService.cs
@@ -12,7 +12,7 @@ namespace Dotnet.Docker;
 
 internal interface IBuildAssetService
 {
-    Task<string> GetAssetContentsAsync(Asset asset);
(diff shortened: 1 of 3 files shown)


mthalman (approved):
Can you check to make sure this will work for multi-SDK releases? I know we don't have that for .NET 10 right now. I'm just suggesting to examine that darc output from one of the .NET 9 releases that has multiple SDKs and check the code against that.

PR #6398, opened 2025-04-29
Add .NET 8 Debian 13 "Trixie" images to nightly branch

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/6174


mthalman (approved):

mthalman (approved):

lbussell:
8.0 Trixie arm32 failed 3 times in a row with the same issue. It fails to restore NuGet packages during tests. This is interesting because we do not encounter this problem in 9.0 or 10.0 Trixie images.

lbussell:
I separated out .NET 9 images from this PR 

Also, I'm pretty sure this is ["Breaking change: .NET 9 is Y2038 compatible (Arm32 glibc); .NET 8 is not (dotnet/core#9285)"](https://github.com/dotnet/core/discussions/9285). We will not be able to offer ARM32 .NET 8 Debian Trixie images.

mthalman (approved):

PR #6402, opened 2025-04-29
Show help output when there are unmatched tokens passed to update-dependencies CLI

### LINKED ISSUE #6394: Update dependencies tool does not fail when passing incorrect/no sub-command
To reproduce, run something like `dotnet run --project .\eng\update-dependencies\ -- foo`

`foo` is not a valid subcommand, yet the app starts and no-ops. Instead, it should show the help output.

Seems related to https://github.com/dotnet/command-line-api/issues/2525

lbussell:
Fixes #6394

`System.Commandline.Hosting` forcibly sets `yourCommand.TreatUnmatchedTokensAsErrors = false` and also overrides the command's action, so we can't fix this behavior by either of those means.

However, unmatched tokens are passed to the `hostBuilderFactory`, which allows you to configure additional hosting options based on those extra tokens. Since we don't need those extra arguments, we can use that opportunity to print the help output and throw an exception.

https://github.com/dotnet/command-line-api/blob/a91bcf488ba7d0112b2637b505f8de32aded6825/src/System.CommandLine.Hosting/HostingAction.cs#L26-L27

### DIFF
--- a/eng/update-dependencies/Program.cs
+++ b/eng/update-dependencies/Program.cs
@@ -30,7 +30,7 @@ config.UseHost(
             if (unmatchedArgs.Length > 0)
             {
                 var helpBuilder = new HelpBuilder();
-                var stringWriter = new StringWriter();
+                using var stringWriter = new StringWriter();
                 helpBuilder.Write(rootCommand, stringWriter);
                 Console.WriteLine(stringWriter.ToString());
                 throw new InvalidOperationException($"Unmatched tokens: {string.Join(" ", unmatchedArgs)}");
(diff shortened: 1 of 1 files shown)


eng/update-dependencies/Program.cs
```diff
@@ -22,8 +25,20 @@
...
 
 config.UseHost(
-    _ => Host.CreateDefaultBuilder(),
-    host => host.ConfigureServices(services =>
+    hostBuilderFactory: unmatchedArgs =>
+        {
+            if (unmatchedArgs.Length > 0)
+            {
+                var helpBuilder = new HelpBuilder();
+                var stringWriter = new StringWriter();
```
mthalman, line 33: This is disposable. 

```suggestion
                using var stringWriter = new StringWriter();
```

mthalman (approved):

PR #6409, opened 2025-05-01
Update some usage and samples links

lbussell:
- Updated main .NET Docker tutorial link
- Added explicit link to [Host ASP.NET Core in Docker containers](https://learn.microsoft.com/aspnet/core/host-and-deploy/docker) for repos that it's relevant for
- Added link to native AOT sample for runtime-deps

### DIFF
--- a/eng/readme-templates/Use.runtime-deps.md
+++ b/eng/readme-templates/Use.runtime-deps.md
@@ -2,4 +2,5 @@
     _ ARGS:
       top-header: The string to use as the top-level header.
       readme-host: Moniker of the site that will host the readme
-}}* [.NET self-contained Sample](https://github.com/dotnet/dotnet-docker/blob/main/samples/dotnetapp/README.md) - This [sample](https://github.com/dotnet/dotnet-docker/blob/main/samples/dotnetapp/Dockerfile.debian) builds and runs an application as a self-contained application.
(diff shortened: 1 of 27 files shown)


eng/readme-templates/Use.runtime-deps.md
```diff
@@ -2,4 +2,5 @@
     _ ARGS:
       top-header: The string to use as the top-level header.
       readme-host: Moniker of the site that will host the readme
-}}* [.NET self-contained Sample](https://github.com/dotnet/dotnet-docker/blob/main/samples/dotnetapp/README.md) - This [sample](https://github.com/dotnet/dotnet-docker/blob/main/samples/dotnetapp/Dockerfile.debian) builds and runs an application as a self-contained application.
+}}* [Self-contained sample](https://github.com/dotnet/dotnet-docker/blob/main/samples/releasesapp/README.md) - [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/main/samples/releasesapp/Dockerfile)
+* [Native AOT sample](https://github.com/dotnet/dotnet-docker/blob/main/samples/releasesapi/README.md) (Preview) - [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/main/samples/releasesapi/Dockerfile)
```
mthalman, line 6: Is it necessary to label this as "Preview" here? We'd then have to update it when .NET 10 is out of preview. I also am unsure whether readers will understand what Preview means in this context.

lbussell: Fair criticism. I also considered "(.NET 10+)". But thinking about it some more, neither of those caveats apply to the runtime-deps image itself so this should be removed.

mthalman (approved):

Issue #6389, opened 2025-04-22
Add dry-run option to update-dependencies pipeline


lbussell:
Sometimes it's desirable to check the functionality of the update-dependencies pipeline without actually submitting a pull request.

Related: #6305, #6387

lbussell:
Completed with https://github.com/dotnet/dotnet-docker/pull/6414

PR #6414, opened 2025-05-02
Add dry-run option to update-dependencies tool and pipelines

lbussell:
Fixes https://github.com/dotnet/dotnet-docker/issues/6389


eng/pipelines/steps/update-dependencies-specific.yml
```diff
@@ -27,7 +36,7 @@ steps:
       # If this is the last iteration, include the credentials to cause a PR to be generated.
       # For internal builds this will cause branch to be pushed to AzDO, but no PRs will be generated.
       if ($customArgs -eq $customArgsArray[-1]) {
-        $customArgs += " $credArgs"
+        $customArgs += " $credArgs $dryRunArg"
```
jander-msft, line 39: Should dry run be applied to each iteration of the loop, not just the last?

eng/update-dependencies/CreatePullRequestOptions.cs
```diff
@@ -35,6 +36,7 @@ public string TargetBranch
         new Option<string>("--version-source-name") { Description = "The name of the source from which the version information was acquired." },
         new Option<string>("--source-branch") { Description = "Branch where the Dockerfiles are hosted" },
         new Option<string>("--target-branch") { Description = "Target branch of the generated PR (defaults to value of source-branch)" },
+        new Option<bool>("--dry-run") { Description = "Execute all steps but do not create a pull request" },
```
jander-msft, line 39: nit: Maybe this should be called `--no-pr` instead of `--dry-run` to better describe the use of the flag. To me, `--dry-run` means "no side effects" which includes writing files to disk.

eng/pipelines/steps/update-dependencies-specific.yml
jander-msft, line 39: If the intent is to just generate the files without opening a PR, how about conditionally applying the credential arguments instead of conditionally applying dry run?

jander-msft, line 39: Where is $credArgs supplied? It was deleted from the former line 20.

eng/update-dependencies/CreatePullRequestOptions.cs
lbussell: I think you are right, since the intention of this option is to validate everything all the way up to actually having _external_ side effects (like pushing code or submitting a PR). So `--no-pr` or `--skip-pr` makes more sense.

eng/pipelines/steps/update-dependencies-specific.yml
lbussell: Skipping the credential arguments short circuits the code long before PRs are created. By passing the new argument, we should search for existing PRs, compute the commit message, commit author and email, PR title, etc. all before exiting.

lbussell: Done in https://github.com/dotnet/dotnet-docker/pull/6414/commits/a40950675e2ede769e8168b363c8d8094deccb50.

mthalman:
The descriptions of these options already states that it will skip the PR creation if they are not provided: https://github.com/dotnet/dotnet-docker/blob/438dab25f6a4387dfd4adf90a45b818bc0ab175f/eng/update-dependencies/CreatePullRequestOptions.cs#L29-L31

So it seems we already have a means to skip the PR creation and don't need a new option. Are those descriptions not true? If not, we should consider providing that functionality anyway since it would make sense when executing the tool from a dev machine where you wouldn't be providing the creds.

lbussell:
> The descriptions of these options already states that it will skip the PR creation if they are not provided:
> 
> https://github.com/dotnet/dotnet-docker/blob/438dab25f6a4387dfd4adf90a45b818bc0ab175f/eng/update-dependencies/CreatePullRequestOptions.cs#L29-L31
> 
> So it seems we already have a means to skip the PR creation and don't need a new option. Are those descriptions not true? If not, we should consider providing that functionality anyway since it would make sense when executing the tool from a dev machine where you wouldn't be providing the creds.

See https://github.com/dotnet/dotnet-docker/pull/6414#discussion_r2073603809. Although, after thinking about it some more, the driver behind this change is not to test the code between the credentials short-circuit and the PR creation. So in order to keep things simpler I think yours and Justin's idea makes more sense.

mthalman (approved):

PR #6435, opened 2025-05-08
Implement release staging pipeline

lbussell:
Part 2/2 of https://github.com/dotnet/dotnet-docker-internal/issues/7414

This PR is a draft because it contains changes from (and depends on) https://github.com/dotnet/docker-tools/pull/1704. Once it is merged I will rebase this PR.

### DIFF
--- a/eng/pipelines/release-promotion.yml
+++ b/eng/pipelines/release-promotion.yml
@@ -22,12 +22,11 @@ extends:
       parameters:
         internalProjectName: ${{ variables.internalProjectName }}
         publicProjectName: ${{ variables.publicProjectName }}
-
+        isStandalonePublish: true
         # Only allow the dotnet-docker-release-staging pipeline as a source for publishing
         # https://dev.azure.com/dnceng/internal/_build?definitionId=1434
         sourceBuildPipelineDefinitionId: "1434"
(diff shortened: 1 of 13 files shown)


eng/pipelines/release-promotion.yml
```diff
@@ -1,11 +1,39 @@
...
+    - name: $(marStatus.serviceConnectionName)
+    stages:
+    - template: /eng/pipelines/stages/publish.yml@self
+      parameters:
+        internalProjectName: ${{ variables.internalProjectName }}
+        publicProjectName: ${{ variables.publicProjectName }}
+        isStandalonePublish: true
+        # Only allow the dotnet-docker-release-staging pipeline as a source for publishing
+        # https://dev.azure.com/dnceng/internal/_build?definitionId=1434
+        sourceBuildPipelineDefinitionId: "1434"
```
mthalman: Use this instead?

```suggestion
        sourceBuildPipelineDefinitionId: "$(resources.pipeline.dotnet-docker-release-staging.pipelineID)"
```

lbussell: Good point, I think this was probably a hold-over from when I started implementing this with a string parameter instead of a pipeline resource.

PR #6460, opened 2025-05-29
Update documentation for new OS support timelines

lbussell:


documentation/supported-platforms.md
```diff
@@ -6,10 +6,14 @@ This document describes the platforms (OS and architectures) supported by the of
...
 
-- New versions of operating systems that we support are added on a best-effort basis, typically within one month.
-- New OSes are always available in the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md) first, and then are added to the officially supported repos afterwards.
+- .NET container images for new OS versions are added according to the following policies:
+  - Azure Linux, Debian, Ubuntu LTS, and Windows — new images released the same day as the new OS release.
+  - Alpine Linux — new images are released as soon as possible except when the Alpine Linux release aligns closely with a .NET servicing release, in which case new images will be released alongside the .NET servicing updates or shortly thereafter.
+- New OSes are always available in the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md) first, and are added to the officially supported repos afterwards.
```
richlander, line 12:
```suggestion
- Images for new OSes are typically released within one month of the new OS release, with a goal to release same-day when possible.
- New OSes are always available in the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md) first, and are added to the officially supported repos afterwards.
```

documentation/supported-platforms.md
```diff
@@ -6,10 +6,12 @@ This document describes the platforms (OS and architectures) supported by the of
...
 
-- New versions of operating systems that we support are added on a best-effort basis, typically within one month.
-- New OSes are always available in the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md) first, and then are added to the officially supported repos afterwards.
+- Images for new OSes are typically released within one month of the new OS release, with a goal to release same-day when possible.
+- New OSes are always available in the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md) first, and are added to the officially supported repos afterwards.
```
mthalman: I'm wondering whether we need to be more crisp with the terminology. Rather than "new OSes", it's most accurate to say "new OS versions". But even that is open to interpretation. Does a new patch version fall into this bucket? No. New major version? Mostly. But that doesn't easily apply for something like Alpine where it's the minor version that is changing. So I'm wondering whether we need to list out all the supported OSes/distros and explain what we mean by new version.

richlander: Major and minor versions are not useful terms.

Alpine 3.22 and Fedora 42 (and .NET 10) are the same kind of release, whatever word you want to use to describe it. The fact that Alpine uses the minor version number doesn't mean anything.

Yes, "new OS versions" sounds good.

lbussell:
```suggestion
- Images for new OS versions are typically released within one month of the new OS release, with a goal to release same-day when possible.
- New OS versions are always available in the [nightly repositories](https://github.com/dotnet/dotnet-docker/blob/nightly/README.md) first, and are added to the officially supported repos afterwards.
```

richlander: Can you remove the word "always"?

lbussell: Applied in https://github.com/dotnet/dotnet-docker/pull/6460/commits/2ff63e2c442ed12d8612ffb2831527049069ccb5

mthalman (approved):

richlander:
LGTM

PR #6462, opened 2025-05-29
[main] Ignore https://aka.ms/ in markdown link checker

lbussell:
This ignores https://aka.ms/ in the markdown link checker since it leads to an internal site. Please note that it does not ignore all aka.ms/* links, only https://aka.ms/ itself.


.github/linters/.linkspector.yml
```diff
@@ -13,4 +13,5 @@ excludedDirs:
 ignorePatterns:
   - pattern: "^https://github.com/microsoft/mcr.*$"
   - pattern: "^https://github.com/dotnet/release.*$"
+  - pattern: "^https://aka.ms/$"
```
mthalman, line 16: In what circumstances would it attempt to check `https://aka.ms/` by itself?

PR #6472, opened 2025-06-03
Update internal Dockerfile templates to support managed identity OAuth token

lbussell:
Part of https://github.com/dotnet/dotnet-docker-internal/issues/7413

The internal Dockerfile templates have been updated to use the new staging storage account. We will use managed identity to access it instead of the azure pipelines system access token.

The existing internal Dockerfile templates did not work with managed identity tokens, so this PR also contains updates to the templates and download commands that are required.

### DIFF
--- a/eng/dockerfile-templates/Dockerfile.linux.download-file
+++ b/eng/dockerfile-templates/Dockerfile.linux.download-file
@@ -8,14 +8,15 @@
         sha-var-name: Name of variable that stores the checksum
         sha-function: SHA function to use ^
 
-    set isInternal to find(ARGS["url"], "artifacts.visualstudio.com") >= 0 ^
-    set alpineSetTokenCmd to 'token=$(echo "${ACCESSTOKEN}:${ACCESSTOKEN}" | base64 -w 0)' ^
-    set additionalWgetArgs to when(isInternal, '--header="Authorization: Basic $token" ', '') ^
-    set additionalCurlArgs to when(isInternal, '-u :$ACCESSTOKEN --basic ', '') ^
+    _ Documentation for using Managed Identity/OAuth tokens to access Azure storage accounts:
+        https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-with-azure-active-directory#call-storage-operations-with-oauth-tokens ^
+
+    set isInternal to find(ARGS["url"], "dotnetstage") >= 0 ^
+    set additionalWgetArgs to when(isInternal, '--header "Authorization: Bearer $ACCESSTOKEN" --header "x-ms-version: 2017-11-09" ', '') ^
+    set additionalCurlArgs to when(isInternal, '-H "Authorization: Bearer $ACCESSTOKEN" -H "x-ms-version: 2017-11-09" ', '') ^
     set isAlpine to find(OS_VERSION, "alpine") >= 0 ^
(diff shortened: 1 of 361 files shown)


eng/dockerfile-templates/Dockerfile.linux.download-file
```diff
@@ -8,14 +8,15 @@
...
 
-    set isInternal to find(ARGS["url"], "artifacts.visualstudio.com") >= 0 ^
-    set alpineSetTokenCmd to 'token=$(echo "${ACCESSTOKEN}:${ACCESSTOKEN}" | base64 -w 0)' ^
-    set additionalWgetArgs to when(isInternal, '--header="Authorization: Basic $token" ', '') ^
-    set additionalCurlArgs to when(isInternal, '-u :$ACCESSTOKEN --basic ', '') ^
+    _ Documentation for using Managed Identity/OAuth tokens to access Azure storage accounts:
+        https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-with-azure-active-directory#call-storage-operations-with-oauth-tokens ^
+
+    set isInternal to find(ARGS["url"], "dotnetstage") >= 0 ^
+    set additionalWgetArgs to when(isInternal, '--header "Authorization: Bearer $ACCESSTOKEN" --header "x-ms-version: 2017-11-09" ', '') ^
```
mthalman: Consider setting a common variable for the header value.

lbussell: Turns out you can use the same argument (`--header`) for both

tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/aspnet-10.0-trixie-slim-amd64-Dockerfile.approved.txt
```diff
@@ -6,8 +6,8 @@ FROM amd64/buildpack-deps:trixie-curl AS installer
 ARG ACCESSTOKEN
 
 # Retrieve ASP.NET Core
-RUN aspnetcore_version=0.0.0 \
-    && curl -u :$ACCESSTOKEN --basic -fSL --output aspnetcore.tar.gz "https://artifacts.visualstudio.com/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-linux-x64.tar.gz" \
+ RUN aspnetcore_version=0.0.0 \
+    && curl --header "Authorization: Bearer $ACCESSTOKEN" --header "x-ms-version: 2017-11-09" -fSL --output aspnetcore.tar.gz "https://dotnetstage.blob.core.windows.net/aspnetcore/Runtime/0.0.0/aspnetcore-runtime-0.0.0-linux-x64.tar.gz" \
```
lbussell: This is a bit of a regression, the `*_version` variables are currently unused in the internal Dockerfile templates. I have an idea about how to generalize this logic so that the download URL always does The Right Thing.

lbussell:
Yes, this PR touches *all* Dockerfiles because _all_ of them had extra whitespace. I *can* separate the whitespace fix out into a separate PR if necessary.

PR #6473, opened 2025-06-05
Add staging pipeline support to update-dependencies tool

lbussell:
- New command: `from-staging-pipeline`
  - This command supplants `Get-DropVersions.ps1 -BuildId ...` and accomplishes the same things.
  - This command does _not_ yet update dependencies with internal versions. This means the base urls won't be updated and the Dockerfiles won't be generated with the internal versions of the templates just yet.
  - The reason for that is the old update-dependencies code depends on using the base URL for accessing the sha files, so that will need to be updated or replaced separately.

### DIFF
--- a/eng/update-dependencies/Model/Release/BuildManifest.cs
+++ b/eng/update-dependencies/Model/Release/BuildManifest.cs
@@ -9,7 +9,7 @@ using System.Text.Json;
 
 namespace Dotnet.Docker.Model.Release;
 
-public record ReleaseManifest
+public record BuildManifest
 {
     private static readonly JsonSerializerOptions s_jsonOptions = new()
     {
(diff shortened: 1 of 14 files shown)


eng/update-dependencies/Model/Release/ReleaseManifest.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+
+namespace Dotnet.Docker.Model.Release;
+
+public record ReleaseManifest
```
mthalman: Is there no library we scan reference for this API?

lbussell: Negative. By the way, the model is from [here](https://dev.azure.com/dnceng/internal/_git/dotnet-release?path=/src/Microsoft.DotNet.Release/Microsoft.DotNet.ReleaseLib/Models/BuildManifest.cs). I think I forgot to link to that in a doc comment. It's not public :/

mthalman (approved):

eng/update-dependencies/Model/Release/ReleaseManifest.cs
lbussell: Added comments

eng/update-dependencies/Model/Release/BuildManifest.cs
```diff
@@ -8,6 +8,15 @@
 
 namespace Dotnet.Docker.Model.Release;
 
+// The models in this file are a subset of the BuildManifest models used by the .NET release infrastructure.
```
mthalman, line 11: Consider filing an issue on the release repo to make this model available.

mthalman (approved):

eng/update-dependencies/Model/Release/BuildManifest.cs
lbussell: Filed https://github.com/dotnet/release/issues/1336

PR #6485, opened 2025-06-10
Update release staging pipeline to allow internal builds

lbussell:
Tests are still broken but that can be handled in a follow-up PR.

This PR also includes one very small fix for update-dependencies that I discovered was necessary while validating this.

Depends on ImageBuilder version >= [`2727330`](https://github.com/dotnet/docker-tools/pull/1729/files)

### DIFF
--- a/eng/pipelines/release-staging.yml
+++ b/eng/pipelines/release-staging.yml
@@ -13,8 +13,20 @@ resources:
     endpoint: dotnet
     name: dotnet/versions
 
+parameters:
+- name: sourceBuildPipelineRunId
+  displayName: >
+    Source build pipeline run ID. This refers to runs of *this pipeline*.
+    Override this parameter in combination with disabling the `Build` stage to
+    test images that were built in a different pipeline run. When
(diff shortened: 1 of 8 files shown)


eng/pipelines/release-staging.yml
```diff
@@ -13,20 +13,35 @@ resources:
...
+  parameters:
+    sourceBuildPipelineRunId: ${{ parameters.sourceBuildPipelineRunId }}
 
 extends:
   template: /eng/common/templates/1es-official.yml@self
   parameters:
     serviceConnections:
     - name: $(internal-mirror.serviceConnectionName)
     - name: $(build.serviceConnectionName)
+    - name: $(dotnetstaging.serviceConnection.name)
```
mthalman: Wouldn't it better to follow the same naming pattern for all the service connections?

```suggestion
    - name: $(dotnetstaging.serviceConnectionName)
```

lbussell: Sure. I updated it.

PR #6496, opened 2025-06-19
Use more descriptive long-form options for curl, tar, and checksum validation

lbussell:
Reading through our Dockerfiles can be a little confusing if you're not familiar with specific command line option abbreviations for `curl`, `tar`, etc. 

This PR attempts to make them easier to read by using the long-form options:

- `curl -fSLO` -> `curl --fail --silent --location --output`
- `tar -oxzf` -> `tar --gzip --extract --no-same-owner --file`
- `sha512sum -c ...` -> `sha512sum --check ...`


mthalman (approved):
This is limited to the three commands you described. Are there plans to expand this everywhere? At a glance, I can see `ln` and `mkdir` as candidates. I guess what's the overall plan here?

richlander:
Here's a HN post that describes much the same thing: https://news.ycombinator.com/item?id=43440184. It's useful context for folks (including us, later) who want to understand the motivation.

lbussell:
Yes, I acknowledge that I left out `mkdir` and `ln` (and probably others). I can add them. I will also update CONTRIBUTING.md accordingly.

mthalman (commented):
Other commands I see that can use long form are `rm -rf` and `useradd -l`.

eng/dockerfile-templates/Dockerfile.linux.install-pkgs
```diff
@@ -18,14 +18,14 @@
 }}{{
 if isDnf:dnf install -y{{ARGS["pkg-mgr-opts"]}} \^
 elif isApk:apk add --upgrade --no-cache{{ARGS["pkg-mgr-opts"]}} \^
-elif isTdnf:tdnf install -y{{ARGS["pkg-mgr-opts"]}} \^
+elif isTdnf:tdnf install --assumeyes{{ARGS["pkg-mgr-opts"]}} \^
```
mthalman: Can we make an exception for this? I feel like `-y` is so common it doesn't require long form.

lbussell: Sure. That's also how I felt about `rm -rf`.

mthalman: That's fair. I could go either way on `rm`.

lbussell:
OK, here's what I did:

- `CONTRIBUTING.md` now says the cli option choice is a preference, not a directive
- Reverted `rm -rf` because it is obvious
- Reverted `ln -s` and `sha512sum -c` because the alpine/busybox versions of those only support the shorthand arguments, and it keeps the templates simple and the Dockerfiles consistent
- Reverted package manager assume-yes changes per your comment.

eng/dockerfile-templates/Dockerfile.linux.install-pkgs
lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/6496/commits/f59f78be3cda736154b7758f00c701301382dd51

mthalman (approved):
This strikes a good balance of short and long form, I think.

lbussell:
(Markdown link check failure is a false positive)

Issue #6495, opened 2025-06-18
Consider updating SHA checks to new SHA files


richlander:
Was just looking at https://github.com/dotnet/dotnet-docker/blob/d2401d0f2ae676c1dbd0d685ccfe917d2e406674/src/runtime/10.0/alpine3.22/amd64/Dockerfile.

The following now works:

```bash
$ curl -O https://builds.dotnet.microsoft.com/dotnet/Runtime/10.0.0-preview.5.25277.114/dotnet-runtime-10.0.0-preview.5.25277.114-linux-musl-x64.tar.gz
$ curl -O https://builds.dotnet.microsoft.com/dotnet/Runtime/10.0.0-preview.5.25277.114/dotnet-runtime-10.0.0-preview.5.25277.114-linux-musl-x64.tar.gz.sha512
$ cat dotnet-runtime-10.0.0-preview.5.25277.114-linux-musl-x64.tar.gz.sha512
716B01559B44320C124459C461DC9BE7A2391B680EC9BA46A22F77A737EB93C4E3323A73B1138DDD1574DCC50934C82ADF7ABE792D6BBD953B8AEA790619941F
$ dotnet_runtime_sha512=$(cat dotnet-runtime-10.0.0-preview.5.25277.114-linux-musl-x64.tar.gz.sha512)
$ echo "$dotnet_runtime_sha512 dotnet-runtime-10.0.0-preview.5.25277.114-linux-musl-x64.tar.gz" | sha512sum -c -
dotnet-runtime-10.0.0-preview.5.25277.114-linux-musl-x64.tar.gz: OK
```

This approach could significantly simplify the workflow for building .NET images. I always delete the SHA checking lines when I update these Dockerfiles. I am sure others do the same. It would be nice to not need to do so.

AFAIK, this only works for .NET 10, currently.

lbussell:
Amazing!!

So in combination with [Refactor how version numbers are specified after the introduction of installer stages (dotnet/dotnet-docker#6266)](https://github.com/dotnet/dotnet-docker/issues/6266), our new Dockerfile could look something like this:

```Dockerfile
ARG REPO=mcr.microsoft.com/dotnet/runtime-deps
ARG DOTNET_VERSION=10.0.0-preview.5.25277.114

# Installer image
FROM $REPO:10.0.0-preview.5-alpine3.22-amd64 AS installer
ARG DOTNET_VERSION

# Retrieve .NET Runtime
[...]

Issue #6266, opened 2025-02-24
Refactor how version numbers are specified after the introduction of installer stages


MichaelSimons:
Now that [every Dockerfile is utilizing an installer stage](https://github.com/dotnet/dotnet-docker/pull/6245), consider changing how the .NET version numbers are being handled.

Current Pattern (.NET version is being specified in multiple places):

```Dockerfile
ARG REPO=mcr.microsoft.com/dotnet/runtime-deps

# Installer image
FROM amd64/buildpack-deps:noble-curl AS installer

# Retrieve .NET Runtime
RUN dotnet_version=10.0.0-preview.1.25080.5 \
    && curl -fSL --output dotnet.tar.gz https://ci.dot.net/public/Runtime/$dotnet_version/dotnet-runtime-$dotnet_version-linux-x64.tar.gz \
...

# .NET runtime image
FROM $REPO:10.0.0-preview.1-noble-amd64

# .NET Runtime version
ENV DOTNET_VERSION=10.0.0-preview.1.25080.5
```

Proposal (Central the .NET version number):

```Dockerfile
ARG REPO=mcr.microsoft.com/dotnet/runtime-deps
ARG DOTNET_VERSION=10.0.0-preview.1.25080.5

# Installer image
FROM amd64/buildpack-deps:noble-curl AS installer
ARG DOTNET_VERSION

# Retrieve .NET Runtime
RUN curl -fSL --output dotnet.tar.gz https://ci.dot.net/public/Runtime/${DOTNET_VERSION}/dotnet-runtime-$dotnet_version-linux-x64.tar.gz \
...

# .NET runtime image
FROM $REPO:10.0.0-preview.1-noble-amd64
ARG DOTNET_VERSION

# .NET Runtime version
ENV DOTNET_VERSION=${DOTNET_VERSION}
```

Improves readability and maintainability of the Dockerfiles which is a benefit to the maintainers of this repo as well as customers who copy the Dockerfiles which is sometimes necessary when they can't make use of the .NET images themselves.

PR #6500, opened 2025-06-24
Refactor Linux Dockerfile templates to make changing shas easier

lbussell:
The goal of this PR is to refactor Dockerfile templates to make the following two tasks easier:

- #6495 
- #6266

Summary:

- Simplified/re-implemented templates for downloading and extracting .NET archives
  - Added support for "sha-url" and "sha-url-aggregate" arguments to the new `download-dotnet` template. So when we want to switch to online shas, we only need to change the arguments to `download-dotnet`.
- Removed lots of unused code and conditions
- Inlined some templates in some of the simpler appliance Dockerfiles

There should be no functional changes here. There are a few minor stylistic changes:

- ASP.NET Core Composite images are now always multi-stage, just like every other .NET Dockerfile
- .NET Monitor extension files use a combined `curl` command to download both extension binaries
- Internal Dockerfiles lost the quotes around the download URL 🤷‍♂️


eng/dockerfile-templates/Dockerfile.linux.download-file
```diff
@@ -2,28 +2,45 @@
...
 
-}}{{downloadCommand}} {{ARGS["out-file"]}} {{ARGS["url"]}}{{if ARGS["sha"]: \
+    _ Add authentication header for internal builds, if appropriate. The same
+      header works for both wget and curl. ^
+    if (isInternal):{{
+        set downloadCommand to cat(downloadCommand, ['--header "Authorization: Bearer $ACCESSTOKEN" --header "x-ms-version: 2017-11-09"'])
+    }}^
+
+    if (useWget):{{
+        set downloadCommand to cat(downloadCommand, ["--output-document", ARGS["out-file"]])
```
mthalman: `out-file` is optional and may be unset. It should share the same logic as below.

eng/dockerfile-templates/Dockerfile.linux.download-file
```diff
@@ -2,12 +2,15 @@
...
-        out-file: (optional) name of the output file
-        sha: (optional) Expected checksum of the downloaded file - mutually exclusive with sha-url
-        sha-url: (optional) URL to download the checksum file - mutually exclusive with sha
-        sha-var-name (optional): Name of variable that stores the checksum
-        sha-function (optional): SHA function to use (e.g. 256, 384, 512) ^
+        url                 : URL to download the file from
+        out-file            : (optional) name of the output file
+        sha                 : (optional) Expected checksum of the downloaded file - mutually exclusive with sha-url
+        sha-url             : (optional) URL to download the checksum file - mutually exclusive with sha
+        sha-url-aggregate   : (optional) URL to download the aggregate checksum file - mutually exclusive with sha
```
mthalman: These are all mutually exclusive with each other, right?

eng/dockerfile-templates/runtime/Dockerfile.linux
mthalman, line 1: None of the changes in this file seem necessary except for the `InsertTemplate("Dockerfile.linux.install-runtime")` change.

lbussell: You're correct, I initially started to inline that template until I realized it was used by the composite template as well, and accidentally left some changes behind.

eng/dockerfile-templates/Dockerfile.linux.download-file
lbussell: Seems like this comment is for an outdated version of the file. Please check https://github.com/lbussell/dotnet-docker/blob/72de167713bf01ef8dbe4fc2bb45c9c85948879c/eng/dockerfile-templates/Dockerfile.linux.download-files#L38-L52

eng/dockerfile-templates/Dockerfile.linux.download-file
lbussell: Seems like this comment is for an outdated version of the file, but I went ahead and gave the args documentation a second pass.

eng/dockerfile-templates/runtime/Dockerfile.linux
mthalman: The `platform` variable is also unused.

mthalman (approved):

eng/dockerfile-templates/runtime/Dockerfile.linux
lbussell: Removed in https://github.com/dotnet/dotnet-docker/pull/6500/commits/6c9655e97090711769d08077779462f10256e336

PR #6502, opened 2025-06-25
Add update-dependencies support for resolving .NET build manifest from pipeline artifacts

### LINKED ISSUE #6484: update-dependencies from-staging-pipeline command always relies on staging storage account access
Currently, the update-dependencies `from-staging-pipeline` command always relies on having Entra ID/Managed Identity authentication to the staging storage account. Individual devs do not have read access to the storage account (for good reason).

It is still useful/necessary at times to run the `from-staging-pipeline` command locally when preparing for a release. To make that possible, the `from-staging-pipeline` command should behave different when running locally. It should use an Azure CLI token to reach out to the staging pipeline itself and download the ReleaseManifest/BuildManifest that way. We can use that file to grab the versions for the release. Since we still won't have access to the staging storage account locally, this will only work for finalized releases where checksums are already available publicly.

After this is implemented, https://github.com/dotnet/dotnet-docker/blob/main/eng/Get-DropVersions.ps1 can be deleted.

Related:
- https://github.com/dotnet/dotnet-docker/issues/6483
- https://github.com/dotnet/dotnet-docker/issues/6305

lbussell:
Fixes #6484

This PR allows you to choose between resolving the .NET Build Manifest from either the staging storage account or the staging pipeline artifacts, based on whether you passed the CLI option `--staging-storage-account`. This allows devs to target staged builds from their local machine without needing access to the production staging storage account.

### DIFF
--- a/eng/update-dependencies/AzdoAuthProvider.cs
+++ b/eng/update-dependencies/AzdoAuthProvider.cs
@@ -2,9 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.Net.Http;
-using System.Net.Http.Headers;
-using System.Text;
 using Azure.Identity;
 using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Services.WebApi;
@@ -34,18 +31,6 @@ public class AzdoAuthProvider
         var credential = new VssBasicCredential(userName: string.Empty, password: AccessToken);
         var connection = new VssConnection(baseUrl, credential);
         return connection;
-    }
-
-    public HttpClient GetAuthenticatedHttpClient()
-    {
-        // Create an HttpClient that uses the Azure Developer CLI credential for authentication.
-        var httpClient = new HttpClient();
-        var header = new AuthenticationHeaderValue(
-            "Basic",
-            Convert.ToBase64String(Encoding.ASCII.GetBytes(string.Format("{0}:{1}", "", _accessToken.Value))));
-        httpClient.DefaultRequestHeaders.Authorization = header;
-
-        return httpClient;
     }
 
     private static string GetAccessTokenInternal()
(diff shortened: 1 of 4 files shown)


eng/update-dependencies/AzdoAuthProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        // Create an HttpClient that uses the Azure Developer CLI credential for authentication.
+        var httpClient = new HttpClient();
+        var header = new AuthenticationHeaderValue(
+            "Basic",
+            Convert.ToBase64String(Encoding.ASCII.GetBytes(string.Format("{0}:{1}", "", _accessToken.Value))));
+        httpClient.DefaultRequestHeaders.Authorization = header;
+
+        return httpClient;
+    }
```
lbussell: I know doing `HttpClient`s this way is bad. I'll come back and fix it in a bit.

eng/update-dependencies/StorageAccountBuildManifestProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+    public async Task<BuildManifest> GetBuildManifestAsync(
+        string storageAccountUrl,
+        int stagingPipelineRunId,
+        CancellationToken cancellationToken = default)
+    {
+        // Each pipeline run has a corresponding blob container named stage-${options.StagingPipelineRunId}.
+        // Release metadata is stored in metadata/ReleaseManifest.json.
+        // Release assets are stored individually under in assets/shipping/assets/[Sdk|Runtime|aspnetcore|...].
+        string containerName = $"stage-{stagingPipelineRunId}";
+        string blobName = "metadata/ReleaseManifest.json";
```
mthalman, line 45: Is this going to work if the staging pipeline is using the old manifest version?

lbussell: Yes, that's my understanding. Here's an example of the old release pipeline uploading to metadata/ReleaseManifest.json - https://dev.azure.com/dnceng/internal/_build/results?buildId=2713778&view=logs&j=7e6c29a1-b657-5eef-a7a0-1248622e458e&t=bf8bad21-ba0f-56d0-19cd-874f40dccaa6&l=28

mthalman (approved):

PR #6520, opened 2025-07-07
Fix internal download URLs for stable product versions

lbussell:
Follow up to #6518. This PR updates the Dockerfile templates to make use of the full build versions.

Example pipeline where all internal Dockerfiles were built successfully using this pattern: https://dev.azure.com/dnceng/internal/_build/results?buildId=2742659&view=results (testing is still being worked on).

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/aspnet-8.0-alpine3.21-amd64-Dockerfile.approved.txt
+++ b/tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/aspnet-8.0-alpine3.21-amd64-Dockerfile.approved.txt
@@ -7,15 +7,16 @@ ARG ACCESSTOKEN
 
 # Retrieve ASP.NET Core
 RUN aspnetcore_version=0.0.0 \
+    && aspnetcore_build_version=$aspnetcore_version \
     && wget --header "Authorization: Bearer $ACCESSTOKEN" --header "x-ms-version: 2017-11-09" \
-        https://dotnetstage.blob.core.windows.net/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz \
-        https://dotnetstage.blob.core.windows.net/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz.sha512 \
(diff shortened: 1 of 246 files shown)


tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/aspnet-8.0-alpine3.21-amd64-Dockerfile.approved.txt
```diff
@@ -7,15 +7,16 @@ ARG ACCESSTOKEN
 
 # Retrieve ASP.NET Core
 RUN aspnetcore_version=0.0.0 \
+    && aspnetcore_build_version=$aspnetcore_version \
     && wget --header "Authorization: Bearer $ACCESSTOKEN" --header "x-ms-version: 2017-11-09" \
-        https://dotnetstage.blob.core.windows.net/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz \
-        https://dotnetstage.blob.core.windows.net/aspnetcore/Runtime/$aspnetcore_version/aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz.sha512 \
-    && echo "$(cat aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz.sha512)  aspnetcore-runtime-$aspnetcore_version-linux-musl-x64.tar.gz" | sha512sum -c - \
+        https://dotnetstage.blob.core.windows.net/aspnetcore/Runtime/$aspnetcore_build_version/aspnetcore-runtime-$aspnetcore_build_version-linux-musl-x64.tar.gz \
```
mthalman, line 12: Is this right? Isn't the file name supposed to be using `aspnetcore_version`?

lbussell: This happens because of the way we scrub versions from the Dockerfiles. We replace all versions (including preview, serivcing, etc.) with `0.0.0` here:

https://github.com/dotnet/dotnet-docker/blob/a846cb4f827603f38366cce4714ee50851412430/tests/Microsoft.DotNet.Docker.Tests/GeneratedArtifactTests.cs#L90

The Dockerfile template automatically picks which variables to substitute. It looks through all the defined variables in the order of most recently defined. It sees that `aspnetcore_build_version` was just set to `0.0.0`, so it substitutes that one.

In a real internal Dockerfile we'd have e.g. `aspnetcore_version=8.0.0` and `aspnetcore_build_version=8.0.0-servicing.12345.6` and it would behave the way you described. But this is just the new templates behaving as expected.

PR #6544, opened 2025-07-14
When testing staged builds, use public NuGet feeds for public preview releases

lbussell:
- Added SemanticVersion (based on official semver spec) and DotNetVersion classes that can slowly start to be used instead of primitives/strings throughout update-dependencies
- Do not use an internal NuGet feed for public preview releases, because they don't exist. Public preview releases use public NuGet feeds. Use that public feed instead.

### DIFF
--- a/eng/update-dependencies/SemanticVersion.cs
+++ b/eng/update-dependencies/SemanticVersion.cs
@@ -1,0 +1,54 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Text.RegularExpressions;
+
+namespace Dotnet.Docker;
+
+internal partial record SemanticVersion
+{
+    private readonly Match _match;
+
+    public SemanticVersion(string versionString)
+    {
+        VersionString = versionString;
+        _match = SemanticVersionRegex.Match(VersionString);
+
+        Major = int.Parse(_match.Groups["major"].Value);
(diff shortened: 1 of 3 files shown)


eng/update-dependencies/SemanticVersion.cs
```diff
@@ -0,0 +1,54 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Text.RegularExpressions;
+
+namespace Dotnet.Docker;
+
+internal partial record SemanticVersion
```
mthalman: Can we just use the `SemanticVersion` type from `NuGet.Versioning` instead?

https://github.com/NuGet/NuGet.Client/blob/c08c2c29435d9ae4a8b738096033cae175485c03/src/NuGet.Core/NuGet.Versioning/SemanticVersion.cs#L16

lbussell: I could; I simply didn't know it existed. It looks like it doesn't have any implicit conversions to/from string which was the primary reason I introduced these classes - to use them in place of strings without any extra work. So I would probably need to add that functionality to `DotNetVersion` or extension methods if possible.

lbussell: https://github.com/dotnet/dotnet-docker/pull/6544/commits/f033b52e87d3ff3a0f121ee8942370c9dd69a466

PR #6552, opened 2025-07-16
Replace ServiceMonitor.exe source which was taken offline with recommended replacement

Benjamin-Displayr:
As detailed in https://github.com/dotnet/announcements/issues/351 the old url https://dotnetbinaries.blob.core.windows.net/servicemonitor/2.0.1.10/ServiceMonitor.exe has been taken offline and replaced with a new link sourced from github.

This PR simply replaces the link in the sample dockerfile with the new link, to get the dockerfile working again.


lbussell (approved):
Thanks for catching this!

PR #6554, opened 2025-07-16
Add `dnx` symlink to `/usr/bin` in .NET SDK images

lbussell:
This allows you to run `dnx` directly in Linux containers:

`docker run --rm mcr.microsoft.com/dotnet/sdk:10.0-preview dnx`

On Windows containers, there is nothing to change. `dnx` already works (sort of) because the entire `/Program Files/dotnet/` directory is already on the PATH.

- `docker run --rm mcr.microsoft.com/dotnet/sdk:10.0-preview-nanoserver-ltsc2025 dnx.cmd` - works because `dnx.cmd` is in the PATH
- `docker run --rm mcr.microsoft.com/dotnet/sdk:10.0-preview-nanoserver-ltsc2025 cmd /c 'dnx'` - works because `cmd.exe` resolves `dnx` to `dnx.cmd`
- `docker run --rm mcr.microsoft.com/dotnet/sdk:10.0-preview-nanoserver-ltsc2025 dnx` - does not work (by design), because there is not a `dnx` executable in the PATH

### DIFF
--- a/src/sdk/10.0/alpine3.22/amd64/Dockerfile
+++ b/src/sdk/10.0/alpine3.22/amd64/Dockerfile
@@ -46,8 +46,9 @@ RUN apk add --upgrade --no-cache \
 
 COPY --from=installer ["/dotnet", "/usr/share/dotnet"]
 
-# Trigger first run experience by running arbitrary cmd
-RUN dotnet help
+RUN ln -s /usr/share/dotnet/dnx /usr/bin/dnx \
+    # Trigger first run experience by running arbitrary cmd
+    && dotnet help
 
 # Install PowerShell global tool
 RUN powershell_version=7.6.0-preview.4 \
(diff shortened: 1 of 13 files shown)


src/sdk/10.0/alpine3.22/amd64/Dockerfile
```diff
@@ -46,8 +46,9 @@ RUN apk add --upgrade --no-cache \
 
 COPY --from=installer ["/dotnet", "/usr/share/dotnet"]
 
-# Trigger first run experience by running arbitrary cmd
-RUN dotnet help
+RUN ln -s /usr/share/dotnet/dnx /usr/bin/dnx \
+    # Trigger first run experience by running arbitrary cmd
+    && dotnet help
```
mthalman, line 51: I don't think it's particularly important that these commands be included in the same `RUN` instructions.

lbussell: I was trying to save a layer. I would categorize these commands together since they're both necessary to properly set up the SDK.

Issue #6560, opened 2025-07-18
`aspnet:8.0-alpine` - OpenSSL 3.5.0 appears to cause frequent connection failures with Azure CosmosDB SDK running in Direct connection mode


nickwb:
### Describe the bug

Also raised here: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5302

After upgrading to the `.NET 8.0.18` edition of the `aspnet:8.0-alpine` tag, we found that frequent connection issues began to occur when using Azure CosmosDB in Direct connection mode.

We found that either downgrading to `aspnet:8.0.18-alpine3.21` OR upgrading `libssl3` to `3.5.1` caused the issue to cease.

### Which .NET image(s) are you using?

8.0-alpine a6695199ce879ae4bcfc92f3dfd55b7db5e46e3bdc128a690c6594444f0e300a

### Steps to reproduce

ASP.NET Core app, running .NET 8.0.18, connecting to CosmosDB in Direct connection mode using the latest .NET SDK 3.45.0.

### Other information

More details are available here: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5302

### Output of `docker version`

```console

```

### Output of `docker info`

```console

```

DineshUmasankar-AVEVA:
We are also facing the same issue, would love to receive an update. +1

lbussell:
Hi @nickwb, sorry to hear you're having problems with the Alpine image. The version of libssl in the image you described above is already 3.5.1:

```
> docker run --rm mcr.microsoft.com/dotnet/aspnet:8.0-alpine@sha256:a6695199ce879ae4bcfc92f3dfd55b7db5e46e3bdc128a690c6594444f0e300a /bin/sh -c 'apk list -I | grep openssl'

libcrypto3-3.5.1-r0 x86_64 {openssl} (Apache-2.0) [installed]
libssl3-3.5.1-r0 x86_64 {openssl} (Apache-2.0) [installed]
```

Did you do anything else with the libssl/openssl installation that caused the problem stop happening?

richlander:
@dotnet/ncl @bartonjs @vcsjones @Pilchie

vcsjones:
@nickwb that image hash you specified was built 2 days on Wednesday:

```
docker inspect mcr.microsoft.com/dotnet/aspnet:8.0-alpine@sha256:a6695199ce879ae4bcfc92f3dfd55b7db5e46e3bdc128a690c6594444f0e300a
```

[...]

PR #6565, opened 2025-07-21
Roll back Alpine floating tags from 3.21 to 3.22

lbussell:
This PR moves the Alpine floating tags from `3.22` to `3.21` due to this issue with OpenSSL 3.5 and CosmosDB:
- https://github.com/dotnet/dotnet-docker/issues/6560
- https://github.com/Azure/azure-cosmos-dotnet-v3/issues/5302

Related:
- https://github.com/dotnet/dotnet-docker/discussions/6488
- https://github.com/dotnet/dotnet-docker/discussions/6536

### DIFF
--- a/README.aspnet.md
+++ b/README.aspnet.md
@@ -81,10 +81,10 @@ For more information, see the [composite images section in the Image Variants do
 Tags | Dockerfile | OS Version
 -----------| -------------| -------------
 9.0.7-bookworm-slim-amd64, 9.0-bookworm-slim-amd64, 9.0.7-bookworm-slim, 9.0-bookworm-slim, 9.0.7, 9.0, latest | [Dockerfile](src/aspnet/9.0/bookworm-slim/amd64/Dockerfile) | Debian 12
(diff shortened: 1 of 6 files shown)


mthalman (approved):

richlander:
We are reverting this change because there is a significant incompatibility between Azure CosmosDB and OpenSSL 3.5. The team needs more time to investigate and respond. [OpenSSL 3.5](https://openssl-library.org/post/2025-04-08-openssl-35-final-release/) is the biggest OpenSSL release since OpenSSL 3.0. Alpine 3.22 is the first distro version to include it that we publish in our container images.

The `alpine` floating time is a convenience service we offer users. We do the following:

- Update the floating tag to reference the newest Alpine version one month after it ships.
- We wait for a month to ensure that there are no known issues with the new release, generally, or with .NET in particular.
- We have delayed updating the floating tag based on known issues multiple times in the past.
- If we learn that there are issues after changing the floating tag, we will revert that change.

In short, users rely on this tag precicely because we'll move it foward and _backward_ based on signal.

We offer fixed version tags for users that want more certainy.

Examples (of both styles):

- `8.0-alpine`
- `8.0-alpine3.21`
- `8.0-alpine3.22`

PR #6572, opened 2025-07-22
Clean up unnecessary and unused pipelines

lbussell:
- `eng/common` - removed left over files that were deleted from docker-tools upstream
- `eng/pipelines/dotnet-core-internal-nightly.yml`, `dotnet-core-internal-pr.yml`, `dotnet-core-internal-testing.yml` - superseded by `release-staging` pipeline
- `eng/pipelines/update-readmes.yml` - unused since October 2023, no documentation for when it should be used.


eng/pipelines/update-readmes.yml
mthalman, line 1: This is intended to be used whenever we want to publish any readme changes without having to publish new images. I think it's worthwhile to keep this.

lbussell: Running the dotnet-docker pipeline wouldn't accomplish this?

mthalman: Perhaps. What is the behavior nowadays in the Publish stage if there are no images to push?

lbussell: Not entirely sure. I'll defer this potential removal to a later date.

mthalman (approved):

PR #6584, opened 2025-07-29
fixed typos, linux/amd64 linux/arm64

WillEhrendreich:
these were swapped.


lbussell:
It's actually not swapped! This example is supposed to show how the aspnetapp sample Dockerfile supports cross-platform builds. If you're on an amd64 machine, you can pass the `--platform linux/arm64` argument to build an arm64 image and vice versa.

PR #6593, opened 2025-08-01
Refactor update-dependencies pipelines to have unofficial versions

lbussell:
Apologies that this PR is kind of large. I attempted to re-arrange the changes into digestible commits.

General approach for official/unofficial pipelines:
- There are shared pipeline implementations under **eng/pipelines/pipelines**. Each one takes different arguments. The official and unofficial pipelines (under **eng/pipelines/**) pass different parameters to make the base pipeline behave differently. These are **extends** templates rather than stage/job/step templates.

Changes for update-dependencies:
- No longer runs in a container. This makes it much easier to reason about the pipeline as a whole.
- **update-dependencies-internal** is now based on the new command, **from-staging-pipeline**. It uses **$(System.AccessToken)** from the Azure pipeline to gather update info from the .NET staging pipeline.

Other notes:
- **update-dependencies-internal-unofficial** uses a new managed identity to push changes to **unofficial/*** branches in Azure devops.
- **update-dependencies-unofficial** has no side effects, but later down the line we may want to make it submit pull requests to a private GitHub repo, or Azure devops.
- Includes changes from:
  - https://github.com/dotnet/docker-tools/pull/1758
  - https://github.com/dotnet/docker-tools/pull/1759

Follow-up work:
- Give **update-dependencies-monitor** a once over and also implement an unofficial version of it.


eng/pipelines/update-dependencies-internal.yml
```diff
@@ -2,24 +2,21 @@ trigger: none
...
-    - template: steps/update-dotnet-dependencies-specific.yml
-      parameters:
-        skipPullRequest: ${{ parameters.skipPullRequest }}
-        useInternalBuild: true
-        buildId: $(buildId)
+
+extends:
+  template: /eng/pipelines/pipelines/update-dependencies-internal.yml@self
+  parameters:
+    official: false
```
mthalman: Shouldn't this be `true`? Instead of using an `official` parameter consider this value to be derived from the pipeline name and using a convention of `-official` and `-unofficial` in the pipeline name.

lbussell: Yes, it should be `true`. I don't see a way to derive this from the pipeline name (e.g. from [predefined variables](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml)) - did you have something specific in mind?

mthalman: This technique is used in the VMR to vary the behavior based on the pipeline name: https://github.com/dotnet/dotnet/blob/0cbe593e702c0a8f4ddef3bd3c39e3eb50666ace/eng/pipelines/pr.yml#L43

lbussell:
Rebased to pick up changes from https://github.com/dotnet/dotnet-docker/pull/6592 which will allow PR validation to run. ([diff](https://github.com/dotnet/dotnet-docker/compare/69b4c61066816692c42d02d82e6bf87605452984..0e88e301ad34f2df98811b4b5b3954acfbbb2058))

PR #6604, opened 2025-08-07
Undocument CBL Mariner 2.0 tags

lbussell:
Related: #5018


richlander:
Context: Mariner 2.0 is now EOL. We will continue to ship the images a bit longer for a small set of specific users that still need them.

PR #6611, opened 2025-08-13
Use Syft inside a Dockerfile build to scan images during tests

lbussell:
Related:
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1435#issuecomment-2946447028

Syft images >= v1.27 use a non-root user by default, event for the `-debug` versions of images. That's incompatible with the way we were using the syft image. Instead of writing to a directory inside the syft image, mount the syft image and the image to scan in the same layer and copy syft's output to a `scratch` layer. This has the benefit of being much faster _and_ compatible with distroless + non-root images on both sides (syft and the target images).

Other changes:
- Fixed warning due to unused otel variable
- Update Syft image to latest version v1.31.0
- Some minor refactoring/better encapsulation for syft related code.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/SyftHelper.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/SyftHelper.cs
@@ -112,19 +112,22 @@ public sealed class SyftHelper(DockerHelper dockerHelper, ITestOutputHelper outp
         ];
 
         excludePaths ??= [];
-        IEnumerable<string> excludeArgs = excludePaths.SelectMany(path => new List<string> { "--exclude", path });
+        IEnumerable<string> excludeArgs = excludePaths.SelectMany(path => new[] { "--exclude", path });
 
         syftCommand = [.. syftCommand, .. excludeArgs];
-
         var syftCommandString = string.Join(", ", syftCommand.Select(a => $"\"{a}\""));
+
+        // Since syft is running outside its container, it can't access its copy of CA certificates,
+        // meaning it can't make a secure HTTPS connection to its update server, so we need to
+        // disable its update check. Also, we don't need it checking for updates during tests anyways.
+        const string DisableSyftUpdateCheck = "ENV SYFT_CHECK_FOR_APP_UPDATE=0";
 
         return $"""
             FROM {syftImage} AS syft
             FROM {imageToScan} AS scan-image
 
             FROM scratch AS run-scan
-            ARG SCAN_IMAGE
-            ENV SYFT_CHECK_FOR_APP_UPDATE=0
+            {DisableSyftUpdateCheck}
             RUN --mount=from=syft,source=/,target=/syft \
                 --mount=from=scan-image,source=/,target=/rootfs \
                 [{syftCommandString}]
--- a/tests/Microsoft.DotNet.Docker.Tests/SyftHelper.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/SyftHelper.cs
@@ -1,0 +1,135 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using Xunit.Abstractions;
+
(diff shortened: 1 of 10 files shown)


tests/Microsoft.DotNet.Docker.Tests/SyftHelper.cs
```diff
@@ -0,0 +1,136 @@
...
+            "scan",
+            "/rootfs/",
+            "--source-name",
+            imageToScan,
+            "--output",
+            $"json=/{OutputFileName}",
+        ];
+
+        excludePaths ??= [];
+        IEnumerable<string> excludeArgs = excludePaths.SelectMany(path => new List<string> { "--exclude", path });
```
mthalman:
```suggestion
        IEnumerable<string> excludeArgs = excludePaths.SelectMany(path => new[] { "--exclude", path });
```

tests/Microsoft.DotNet.Docker.Tests/SyftHelper.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        var syftCommandString = string.Join(", ", syftCommand.Select(a => $"\"{a}\""));
+
+        return $"""
+            FROM {syftImage} AS syft
+            FROM {imageToScan} AS scan-image
+
+            FROM scratch AS run-scan
+            ARG SCAN_IMAGE
+            ENV SYFT_CHECK_FOR_APP_UPDATE=0
```
mthalman: I'm guessing this prevents syft from checking whether it's on the latest version? A comment explaining this would be best here.

lbussell: Correct - because we pull syft out of its own container into a `scratch` layer, it's unable to resolve the system CA certificates to make an authenticated request to its update server (also it's just unnecessary). I'll add a comment.

tests/Microsoft.DotNet.Docker.Tests/SyftHelper.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        using var tempDir = FileHelper.UseTempFolder();
+        var tempDirPath = tempDir.Path;
+
+        var syftImage = _dockerHelper.PullDockerHubImage(s_syftImageTag.Value);
+        var imageToInspect = imageData.GetImage(imageRepo, _dockerHelper);
+
+        // Ignore the dotnet folder, or else syft will report all the packages in the .NET Runtime.
+        // We only care about the packages from the linux distro for these tests.
+        extraExcludePaths ??= [];
```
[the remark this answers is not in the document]
lbussell: Collection expressions are preferred where possible.

lbussell:
The current PR failures are because `syft` is interpreting the directory as, well, a directory. It's not treating it like a container filesystem, and thus it's creating a different output. I'm experimenting with other ways to get `syft` to scan an image inside a Dockerfile build, but I might have to look for another way around this.

One option is to bind-mount the docker socket into the syft image/Dockerfile build so that it can access the image directly, but giving syft privileged access to the host is not desirable.

mthalman:
> It's not treating it like a container filesystem, and thus it's creating a different output.

Can you expand more on that?

lbussell:
> > It's not treating it like a container filesystem, and thus it's creating a different output.
> 
> Can you expand more on that?

Sure. Syft scans images and directories using "catalogers". It uses a different set of default catalogers for directories vs. images. So without any extra configuration, the output will be different.

You can select different catalogers with arguments like described in this wiki page: https://github.com/anchore/syft/wiki/Package-Cataloger-Selection

So you can select the "image" catalogers by passing **--select-catalogers image**. I'm trying that now.

One other option we have is passing in a token to the staging registry and scanning directly from the registry: https://github.com/anchore/syft/wiki/Supported-Sources#performance-recommendations. Haven't tried it yet so it's unclear how the performance compares to scanning in the Dockerfile build.

Anyways, I was just comparing the old syft json output to the new output and found the culprit. The rpm-db cataloger is failing:

```json
{
    "id": "6e7cbd3b924fc6a9",
    "location": {
        "path": "/var/lib/rpm/rpmdb.sqlite"
    },
    "unknowns": [
        "rpm-db-cataloger: failed to create temp rpmdb file: open /tmp/rpmdb4274246351: no such file or directory"
    ]
}
```

Obviously `/tmp` is not in the `scratch` base image. We can easily fix that. A little concerning that it fails silently though.

PR #6633, opened 2025-08-28
Fix appliance app test filtering due to incorrect OS directories in TestData

lbussell:
Some appliance images were not being properly tested in internal pipeline runs due to the `OSDir` property being set incorrectly in TestData.

This is a band-aid fix - https://github.com/dotnet/dotnet-docker/issues/5337 is a better long term solution.


mthalman (approved):

lbussell:
Filed https://github.com/dotnet/dotnet-docker/issues/6638, https://github.com/dotnet/dotnet-docker/issues/6639 for failing YARP image tests.

Issue #6567, opened 2025-07-22
Use online checksums for .NET Monitor Dockerfiles


lbussell:
This is part of #6548. Instead of using hard-coded checksums, they should be downloaded from the build/release storage account. This makes the Dockerfiles more flexible and simplifies this repo's maintenance, since we will no longer need to calculate or keep track of checksums.

Example:
https://github.com/dotnet/dotnet-docker/blob/157267f8403b1985704cb6810a5413fc7e3a1bc2/src/runtime/10.0/alpine3.22/amd64/Dockerfile#L7-L16

lbussell:
CI build checksums are under ci.dot.net as expected:

```
https://ci.dot.net/public/diagnostics/monitor/9.1.0-preview.3.25363.1/dotnet-monitor-base-9.1.0-preview.3.25363.1-linux-x64.tar.gz
https://ci.dot.net/public-checksums/diagnostics/monitor/9.1.0-preview.3.25363.1/dotnet-monitor-base-9.1.0-preview.3.25363.1-linux-x64.tar.gz.sha512
```

Official builds are located in dotnetclichecksums storage account:

```
https://builds.dotnet.microsoft.com/dotnet-checksums/diagnostics/monitor/8.1.1/dotnet-monitor-base-8.1.1-linux-x64.tar.gz
https://dotnetclichecksums.blob.core.windows.net/dotnet/diagnostics/monitor/8.1.1/dotnet-monitor-base-8.1.1-linux-arm64.tar.gz.sha512
```

I filed https://github.com/dotnet/dotnet-monitor/issues/8436 to track publishing these to dotnetcli (which maps to builds.dotnet.microsoft.com).

lbussell:
https://github.com/dotnet/dotnet-docker/pull/6634 fixes this for .NET 10.

PR #6634, opened 2025-08-29
[Experimental] Use downloaded sha for dotnet-monitor

wiktork:


eng/dockerfile-templates/monitor-base/Dockerfile.linux
```diff
@@ -12,7 +13,9 @@
             when(find(OS_VERSION_NUMBER, "3.0") >= 0, "azurelinux", "cbl-mariner"),
             "/base/core:",
             OS_VERSION_NUMBER),
-        cat(ARCH_VERSIONED, "/buildpack-deps:", osVersionBase, "-curl"))
+        cat(ARCH_VERSIONED, "/buildpack-deps:", osVersionBase, "-curl")) ^
+    set isVersion10OrLater to monitorMajor = "10" ^
```
wiktork: This check is really a "is it 10 check". I'll look into expressing equal to or greater than here

lbussell: We prefer to use expressions like `monitorMajor != "8" && monitorMajor != "9"` because it's more easily searchable when are cleaning up old .NET versions later.

eng/dockerfile-templates/monitor-base/Dockerfile.linux
```diff
@@ -12,7 +13,9 @@
...
+    set monitorBaseTemplate to when(isVersion10OrLater, "Dockerfile.linux.install-monitor-base.10", "Dockerfile.linux.install-monitor-base")
```
lbussell: So I would express this like:

```suggestion
    set monitorBaseTemplate to when(monitorMajor != "8" && monitorMajor != "9",
        "Dockerfile.linux.install-monitor-base.10",
        "Dockerfile.linux.install-monitor-base") ^
```

Then once 8 and 9 reach end of support, it's really obvious that the condition can be removed and the file renamed.

src/monitor-base/10.0/azurelinux-distroless/amd64/Dockerfile
```diff
@@ -11,12 +11,15 @@ RUN tdnf install -y \
...
-    && echo "$dotnet_monitor_base_sha512  dotnet-monitor-base.tar.gz" | sha512sum -c - \
+    && curl --fail --show-error --location \
+        --remote-name https://ci.dot.net/public/diagnostics/monitor/$dotnet_monitor_version/dotnet-monitor-base-$dotnet_monitor_version-linux-x64.tar.gz \
+        --remote-name https://ci.dot.net/public-checksums/diagnostics/monitor/$dotnet_monitor_version/dotnet-monitor-base-$dotnet_monitor_version-linux-x64.tar.gz.sha512 \
+    && echo "$(cat dotnet-monitor-base-$dotnet_monitor_version-linux-x64.tar.gz.sha512)  dotnet-monitor-base-$dotnet_monitor_version-linux-x64.tar.gz" | sha512sum -c - \
     && mkdir --parents /app \
-    && tar --gzip --extract --no-same-owner --file dotnet-monitor-base.tar.gz --directory /app \
-    && rm dotnet-monitor-base.tar.gz
+    && tar --gzip --extract --no-same-owner --file dotnet-monitor-base-$dotnet_monitor_version-linux-x64.tar.gz --directory /app \
+    && rm  \
```
lbussell, line 20: Nit: whitespace

```suggestion
    && rm \
```

wiktork:
> Great work!
> 
> The internal Dockerfiles are missing the `ACCESSTOKEN` build arg. You can reference other templates/internal baseline files for that.

Should be added now, let me know if this looks correct

src/monitor-base/10.0/azurelinux-distroless/amd64/Dockerfile
wiktork: I believe this is actually because of the download-appliance template (also used by yarp and aspire). I will leave this alone for now.

lbussell: That doesn't surprise me - it's fine for now then.

eng/dockerfile-templates/monitor-base/Dockerfile.linux
```diff
@@ -12,16 +13,19 @@
...
+        cat(ARCH_VERSIONED, "/buildpack-deps:", osVersionBase, "-curl")) ^
+    set monitorBaseTemplate to when(monitorMajor != "8" && monitorMajor != "9", "Dockerfile.linux.install-monitor-base.10", "Dockerfile.linux.install-monitor-base")
 }}ARG REPO=mcr.microsoft.com/dotnet/aspnet
 
 # Installer image
 FROM {{installerImageTag}} AS installer
-{{if isAzureLinux:
+{{if isInternal:
+ARG ACCESSTOKEN
+}}{{if isAzureLinux:
```
lbussell, line 31: `isInternal` needs to be set earlier in the template. Other Dockerfile templates key off of the base URL:

```
set isInternal to find(baseUrl, "artifacts.visualstudio.com") >= 0 ^
```

"artifacts.visualstudio.com" is the correct URL for pipeline artifacts.

If the ARG is set up correctly then it should show up in the internal baseline file (`tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/monitor-base-10.0-azurelinux-distroless-amd64-Dockerfile.approved.txt`).

eng/dockerfile-templates/monitor-base/Dockerfile.linux
```diff
@@ -8,6 +8,7 @@
     set aspnetBaseTag to
         cat("$REPO:", VARIABLES[cat("dotnet|", dotnetMajorMinor, "|product-version")], "-", OS_VERSION, ARCH_TAG_SUFFIX) ^
     set osVersionBase to match(OS_VERSION, ".+(?=.*-)")[0] ^
+    set isInternal to find(baseUrl, "dotnetstage") >= 0 ^
```
lbussell: `baseUrl` is not defined in this template, so this condition will never trigger.

lbussell (approved):

lbussell:
Related: #6567

PR #6640, opened 2025-09-02
Use pipeline resource instead of parameter for update-dependencies-internal-unofficial pipeline

lbussell:
Using a pipeline resource for referencing the release staging pipeline allows us to set pipeline-based triggers.

Successful test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2784879&view=results

### DIFF
--- a/eng/pipelines/update-dependencies-internal-unofficial.yml
+++ b/eng/pipelines/update-dependencies-internal-unofficial.yml
@@ -1,18 +1,24 @@
 trigger: none
 pr: none
 
-parameters:
-- name: buildId
-  type: string
-  default: ""
-
 variables:
 - template: /eng/common/templates/variables/dotnet/common.yml@self
+
+resources:
+  pipelines:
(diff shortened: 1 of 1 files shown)


eng/pipelines/pipelines/update-dependencies-internal.yml
```diff
@@ -33,7 +33,7 @@ parameters:
   default: ""
 
 extends:
-  template: /eng/common/templates/1es-${{ iif(contains(variables['Build.DefinitionName'], '-official'), 'official', 'unofficial') }}.yml@self
+  template: /eng/common/templates/1es.yml@self
```
mthalman, line 36: Was this inadvertent? There doesn't exist a `1es.yml` file.

lbussell: It's there: https://github.com/dotnet/dotnet-docker/blob/nightly/eng/common/templates/1es.yml

It's the version which automatically switches between official and unofficial. I hope to eventually remove the specific official/unofficial versions.

mthalman: Ah, I was looking in the main branch.

PR #6649, opened 2025-09-05
Updates for September 2025 releases

lbussell:


manifest.versions.json
```diff
@@ -150,6 +152,19 @@
     "monitor-ext-s3storage|9.1|linux|x64|sha": "82b550ad9432b86ff58f0bb0f959cd52341b2225d3806fd7211b28462bb216b99ddd828fc36146fca2e92c6c545bc6d81c629ee8dd691d69cd2232528b81d48b",
     "monitor-ext-s3storage|9.1|linux|arm64|sha": "3854991fe4a6bbbd18f7504d8e4da8ade88133a8db0172facd8a50b8057967fbc8ad5643776c68f04b5bcdd5fd901bc2d298d2007a88a2046261c0cfb57226aa",
 
+    "monitor|10.0|build-version": "10.0.0-rc.1.25453.1",
+    "monitor|10.0|product-version": "10.0.0-rc.1",
+    "monitor|10.0|fixed-tag": "$(monitor|10.0|product-version)",
```
wiktork, line 157: Will fixed tag being product-version produce the wrong tag?

lbussell: No, it should be the product version. It looks correct to me. From README.monitor.md:

> `10.0.0-rc.1-amd64, 10.0-preview-amd64, 10.0.0-rc.1, 10.0-preview, 10-preview`

`10.0.0-preview` as a tag instead of `10.0.0-rc.1-amd64` doesn't make sense, because it's less specific. Is that what you're asking?

lbussell: But I did notice another inconsistency with the Monitor RC heading in the readme, which I'll fix really quick.

lbussell:
.NET version updates will come in a follow-up PR.

manifest.versions.json
```diff
@@ -97,8 +97,8 @@
     "libssl|jammy": "3",
     "libssl|noble": "3t64",
 
-    "mingit|latest|x64|url": "https://github.com/git-for-windows/git/releases/download/v2.50.1.windows.1/MinGit-2.50.1-64-bit.zip",
-    "mingit|latest|x64|sha": "6f672aebe9e488a246efd6875f9197dbc0d9a40100e218acc3877cba2b206c45",
+    "mingit|latest|x64|url": "https://github.com/git-for-windows/git/releases/download/v2.51.0.windows.1/MinGit-2.51.0-64-bit.zip",
```
mthalman, line 100: I notice 10.0 is using an older version.

lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/6649/commits/abdc19fc34b76ce64cdf1c12e33d34f15a3024c1

Issue #6632, opened 2025-08-28
Update samples to .NET 10


lbussell:
- Consider creating new samples to feature new .NET or .NET Docker features
- [ ] Update documentation as necessary to reference the new .NET version
- [ ] Update IIS sample URL to preview version of `dotnet-hosting-win.exe` (`/dotnet/<version>/preview/`): <https://github.com/dotnet/dotnet-docker/blob/main/samples/aspnetapp/Dockerfile.windowsservercore-iis#L19>

PR #6658, opened 2025-09-10
Update `dotnetapp` sample to .NET 10

lbussell:
Part of #6632


samples/dotnetapp/Dockerfile.nanoserver
```diff
@@ -2,7 +2,7 @@
 # Learn about building .NET container images:
 # https://github.com/dotnet/dotnet-docker/blob/main/samples/README.md
 ARG TAG=ltsc2025
```
mthalman: Since we're only targeting one Windows version of samples, I think we should simplify things and get rid of this `ARG`. Also applies to WSC.

manifest.samples.json
```diff
@@ -66,9 +66,6 @@
               "variant": "v8"
             },
             {
-              "buildArgs": {
-                "TAG": "ltsc2022"
```
mthalman, line 82: Need to also remove this for `aspnetapp` sample.

lbussell: `aspnetapp` sample isn't updated in this PR.

PR #6672, opened 2025-09-16
Add from-component command and interface for updating arbitrary image component

lbussell:
* Added the `FromComponentCommand` and related options (`FromComponentOptions`) to allow updating a single image component using an `IDependencyVersionSource`
* Added documentation for onboarding new dependencies to the tool
* Added `ChiselVersionSource` as an example implementation of `IDependencyVersionSource` for `chisel` versions.
* Made `BaseUrlUpdater` do nothing when it doesn't find a base url - this was needed to allow `latest` as the "Dockerfile version" when calling the CLI (instead of `8.0`/`9.0`/etc.)

/cc @adityapatwardhan


eng/update-dependencies/FromComponentCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+            "Resolving component '{Component}' with channel '{Channel}'",
+            options.Component, options.Channel);
+
+        bool versionSourceExists = _serviceProvider.TryGetRequiredKeyedService(
+            serviceKey: options.Component,
+            out IDependencyVersionSource? versionSource);
+
+        if (!versionSourceExists || versionSource is null)
+        {
+            throw new InvalidOperationException("No version source registered for component '{Component}'");
```
mthalman:
```suggestion
            throw new InvalidOperationException($"No version source registered for component '{options.Component}'");
```

lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/6672/commits/8a990656fba6d67d0d8253b19332ec769916fa69

PR #6693, opened 2025-09-29
Record internal build version when using --internal flag

lbussell:
Part of https://github.com/dotnet/dotnet-docker-internal/issues/6741.

It's described pretty well in the code change, but we need to record all internal versions that we apply to the repo so that we can rewind and replay updates whenever we need to update the internal/release branch. Using this method will avoid merge conflicts.

### DIFF
--- a/eng/update-dependencies/FromStagingPipelineCommand.cs
+++ b/eng/update-dependencies/FromStagingPipelineCommand.cs
@@ -1,12 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
-using System.Collections.Generic;
-using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
-using System.Text;
-using System.Linq;
 
 namespace Dotnet.Docker;
 
--- a/eng/update-dependencies/FromStagingPipelineCommand.cs
+++ b/eng/update-dependencies/FromStagingPipelineCommand.cs
@@ -48,6 +48,9 @@ internal partial class FromStagingPipelineCommand(
         string dotnetProductVersion = VersionHelper.ResolveProductVersion(releaseConfig.RuntimeBuild);
         string dockerfileVersion = VersionHelper.ResolveMajorMinorVersion(releaseConfig.RuntimeBuild).ToString();
 
+        // Record pipeline run ID for this dockerfileVersion, for later use by sync-internal-release command
+        RecordInternalVersion(dockerfileVersion, options.StagingPipelineRunId.ToString());
+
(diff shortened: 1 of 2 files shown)


eng/update-dependencies/FromStagingPipelineCommand.cs
```diff
@@ -111,6 +109,59 @@ public override async Task<int> ExecuteAsync(FromStagingPipelineOptions options)
...
+                .ToDictionary(parts => parts[0], parts => parts[1]);
+        }
+        catch (FileNotFoundException)
+        {
+            // File doesn't exist - it will be created
+        }
+
+        versions[dockerfileVersion] = stagingPipelineRunId;
+        var versionLines = versions.Select(kv => $"{kv.Key}={kv.Value}");
+        File.WriteAllLines(internalVersionFile, versionLines);
```
mthalman, line 162: Can't it just do a file append instead of reparsing the file contents?

lbussell: If we did that, then new versions would not overwrite any old versions. For example if our branch had the following `internal-versions.txt` file:

```
8.0=1111111
9.0=3333333
10.0=4444444
```

And we wanted to update 8.0 with build 2222222, we want the file to be this:

```
8.0=2222222
9.0=3333333
10.0=4444444
```

And not this:

```
8.0=1111111
9.0=3333333
10.0=4444444
8.0=2222222
```

mthalman: Is that to accommodate dev scenarios? Because the pipeline should never run into such a thing, right?

lbussell: We want to automatically trigger the update-dependencies-internal pipeline every time there's a new staged build of .NET, which will run this command to update to internal versions. We don't control how many times the release team decides to stage new builds. So we will very likely encounter this regularly.

mthalman: Oh, so this will actually be committed into the branch along with the version updates?

lbussell: Correct - in the internal release branch only. It'll never get committed publicly.

For the sync-internal-release branch, the flow will look like this:

1. Public `release/*` branch updates and triggers the sync-internal-release pipeline.
2. SyncInternalReleaseCommand reads this new `internal-versions.txt` file.
3. SyncInternalReleaseCommand runs `git reset` on the `internal/release/*` branch to reset it to the state of `release/*`.
4. SyncInternalReleaseCommand uses the versions it read earlier to re-apply the internal version updates and submits a PR. The PR contains one commit for the `git reset` and one commit for each internal version.

PR #6699, opened 2025-10-02
Re-apply internal staging versions of .NET in sync-internal-release command

lbussell:
This is the biggest part of https://github.com/dotnet/dotnet-docker-internal/issues/6741. The functionality added in this PR ensures that the internal/release/* branch can be updated with the state of the release/* branch even when the two branches diverge.

I have attempted to make commits reviewable individually.


eng/update-dependencies/Sync/SyncInternalReleaseCommand.cs
```diff
@@ -89,33 +89,141 @@ await repo.Remote.CreateRemoteBranchAsync(
...
+            Internal = true,
+            StagingPipelineRunId = stagingPipelineRunId,
+            StagingStorageAccount = stagingStorageAccount,
+        };
+
+        await _updateFromStagingPipeline.ExecuteAsync(fromStagingPipelineOptions);
+        _logger.LogInformation("Applied internal build {BuildNumber}", stagingPipelineRunId);
+
+        await localRepo.StageAsync(".");
+        await localRepo.CommitAsync($"Update dependencies from build {stagingPipelineRunId}", committerIdentity);
```
mthalman: Would it be possible to include a URL that links to the build when we refer to the run ID?

lbussell: Yes, that's trivial, I'll add it to the log.

lbussell: Added in https://github.com/dotnet/dotnet-docker/pull/6699/commits/5545b067cec47630569c01ee6bd749633a498831

lbussell:
The biggest change from the last ~3 commits has to do with the `SourceBranch` argument. It was used by various update-dependencies commands to determine which base url variables to update. However the new sync command needs to use the `SourceBranch` option because it operates on two branches.

The "source" branch (or "upstream" branch) is just a property of the repo. For example, our `Get-Branch` PowerShell function reads the name of the first repo in the manifest to determine the upstream branch. That is passed to the Dockerfile/Readme generation scripts as a variable override. I figured everything would be simpler if we just put it in `manifest.versions.json`.

The various update-dependencies can now just read it from the manifest variables directly instead of taking it as an argument, which frees up `SourceBranch` for the sync command to use.

I actually already made this change when working on https://github.com/dotnet/docker-tools/pull/1813 to simplify things. There will need to be a little bit of cleanup in various scripts that still override the branch variable when generating dockerfiles/readmes.

eng/update-dependencies/CreatePullRequestOptionsExtensions.cs
```diff
@@ -7,4 +7,18 @@ internal static class CreatePullRequestOptionsExtensions
...
+        // Validate that we have all the required pieces to construct the repo URL.
+        ArgumentException.ThrowIfNullOrWhiteSpace(options.AzdoOrganization);
+        ArgumentException.ThrowIfNullOrWhiteSpace(options.AzdoProject);
+        ArgumentException.ThrowIfNullOrWhiteSpace(options.AzdoRepo);
+        ArgumentException.ThrowIfNullOrWhiteSpace(options.TargetBranch);
+        ArgumentException.ThrowIfNullOrWhiteSpace(options.SourceBranch);
+
+        // AzdoOrganization is a URL like https://dev.azure.com/<org>
+        // A valid Azure DevOps repository URL is formatted like https://dev.azure.com/<org>/<project>/_git/<repo>
+        return $"{options.AzdoOrganization.TrimEnd('/')}/{options.AzdoProject}/_git/{options.AzdoRepo}";
```
mthalman: This shares the same logic as `SyncInternalReleaseCommand.ExecuteAsync`. It'd be nice if there was one method to format this URL.

lbussell: I made sure that `SpecificCommand` and `SyncInternalReleaseCommand` both use the extension method. Also centralized the trimming into the options class.

PR #6701, opened 2025-10-03
Implement sync-internal-release unofficial and official pipelines

lbussell:
Fixes https://github.com/dotnet/dotnet-docker-internal/issues/6741.

The unofficial version of the pipeline allows you to specify the source and target branch for testing purposes. The official pipeline automatically triggers when the public `release/*` branch is updated and updates (via pull request) the corresponding `internal/release/*` branch.

Example successful unofficial pipeline runs:
- build#2807907
- build#2807923

I think that the official pipeline will likely run into an issue with creating the `internal/release/*` branch initially since it should be protected, but it's hard to know without letting it run.

### DIFF
--- a/eng/pipelines/pipelines/sync-internal-release.yml
+++ b/eng/pipelines/pipelines/sync-internal-release.yml
@@ -1,0 +1,57 @@
+parameters:
+# Service connection used to push new branches and submit pull requests.
+- name: serviceConnection
+  type: string
+# The branch that changes will be pulled from.
+- name: sourceBranch
+  type: string
+# The branch that needs to be updated with changes from sourceBranch.
+# If this branch doesn't exist, it will be created. Otherwise, a pull request
+# will be created targeting this branch.
+- name: targetBranch
+  type: string
+
+extends:
+  template: /eng/common/templates/1es.yml@self
+  parameters:
+    stages:
+    - stage: SyncReleaseBranchesStage
+      displayName: Sync Release Branches
+
+      jobs:
+      - job: SyncReleaseBranchesJob
+        displayName: Sync Release Branches
+        pool:
+          name: $(default1ESInternalPoolName)
+          image: $(default1ESInternalPoolImage)
+          os: linux
+
+        steps:
+        - checkout: self
+          fetchDepth: 1
+
+        - template: /eng/pipelines/steps/install-dotnet.yml@self
+
+        - task: AzureCLI@2
+          displayName: Sync Release Branches
+          inputs:
+            # Used to push new, non-protected branches and submit pull requests.
+            azureSubscription: "${{ parameters.serviceConnection }}"
+            scriptType: "pscore"
+            scriptLocation: "inlineScript"
+            addSpnToEnvironment: true
+            inlineScript: >-
+              dotnet run --configuration Release --project ./eng/update-dependencies/update-dependencies.csproj --
+              sync-internal-release
+              --azdo-organization "$env:SYSTEM_COLLECTIONURI"
+              --azdo-project "$env:SYSTEM_TEAMPROJECT"
+              --azdo-repo "$env:BUILD_REPOSITORY_NAME"
+              --source-branch "${{ parameters.sourceBranch }}"
+              --target-branch "${{ parameters.targetBranch }}"
+              --pr-branch-prefix "pr"
+              --user "$env:DOTNETDOCKERBOT_USERNAME"
+              --email "$env:DOTNETDOCKERBOT_EMAIL"
+              --staging-storage-account "dotnetstage"
+          env:
+            # Used to download build artifacts containing .NET version information.
+            SYSTEM_ACCESSTOKEN: $(System.AccessToken)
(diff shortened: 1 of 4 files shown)


eng/pipelines/sync-internal-release-official.yml
```diff
@@ -0,0 +1,27 @@
...
+    - release/*
+pr: none
+
+variables:
+- template: /eng/common/templates/variables/dotnet/common.yml@self
+- group: DotNet-Docker-Common-2
+# Always run unofficial pipelines in debug mode
+- name: "System.Debug"
+  value: true
+  readonly: true
```
mthalman: Remove since this is the official pipeline?

eng/pipelines/sync-internal-release-official.yml
```diff
@@ -0,0 +1,27 @@
...
+  parameters:
+    # Source branch will always be a release/* branch due to the pipeline trigger above
+    sourceBranch: "$(Build.SourceBranchName)"
+    # Target branch should be the internal version of the release branch
+    targetBranch: "internal/$(Build.SourceBranchName)"
+    # Service connection used to push new branches and submit pull requests.
+    # The "test" service connection works here in the official pipeline because
+    # we don't need to force push to protected branches. This pipeline only
+    # submits pull requests to them.
+    serviceConnection: "$(updateDepsInt-test.serviceConnectionName)"
```
mthalman: Shouldn't use a test connection here.

eng/pipelines/pipelines/sync-internal-release.yml
```diff
@@ -0,0 +1,57 @@
...
+        - checkout: self
+          fetchDepth: 1
+
+        - template: /eng/pipelines/steps/install-dotnet.yml@self
+
+        - task: AzureCLI@2
+          displayName: Sync Release Branches
+          inputs:
+            # Used to push new, non-protected branches and submit pull requests.
+            azureSubscription: "${{ parameters.serviceConnection }}"
```
MiYanni, line 45: If someone is inputting a service connection, is it assumed that person also has approval to use that service connection on that pipeline?

eng/pipelines/pipelines/sync-internal-release.yml
```diff
@@ -0,0 +1,57 @@
...
+            scriptType: "pscore"
```
MiYanni: nit: `pscore` is a string, so it doesn't need explicit declaration (meaning, using `"`) to tell it to handle this value as a string. Just like your value for `displayName:` above. AzDO just follows standard yaml/json handling for type inference.

eng/pipelines/sync-internal-release-official.yml
MiYanni: Hmm. I wasn't aware you could provide a service connection as a variable expression `$()` like that, because I thought resources like service connections needed to be evaluated at compile time, so AzDO can request permissions for the service connection prior to usage/running the pipeline.

eng/pipelines/pipelines/sync-internal-release.yml
lbussell: I was trying to build the habit of quoting more strings to reduce ambiguity. But I didn't go all in on quotes in this file. I'll update the file to be more consistent.

Related: https://github.com/dotnet/docker-tools/issues/1799

eng/pipelines/sync-internal-release-official.yml
lbussell: It's a variable group variable, so it's evaluated at pipeline queue time (after template expansion, I think). It requests service connection approval when the pipeline starts. Also, variable group variables are not available in template expressions, so this is the only way we can reference it.

We use this pattern everywhere to hide service connection names that aren't necessary in public source code.

eng/pipelines/pipelines/sync-internal-release.yml
lbussell: Correct. This template is only intended to be used as an `extends` pipeline base by the official/unofficial definitions of the same name. I added a comment to make this clearer in https://github.com/dotnet/dotnet-docker/pull/6701/commits/c1492dd7f9882e5f562307ef010f5d6f7b1aec1b.

eng/pipelines/pipelines/sync-internal-release.yml
lbussell: Updated in https://github.com/dotnet/dotnet-docker/pull/6701/commits/c82f502cd69c9bf0d1d4cb0d3dfa4fd40becdb7d

eng/pipelines/sync-internal-release-official.yml
lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/6701/commits/4dc65061d8b24a20e09d0baf8cabad8f33eb40a8

eng/pipelines/sync-internal-release-official.yml
lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/6701/commits/5ea84394be029ab786757c58165f97930a012a00

PR #6708, opened 2025-10-09
[Experiment] Try using NetCore-Public pool for amd64 build images

lbussell:


lbussell:
This was an experiment to see if:

1. The `NetCore-Public` build pool starts faster than the Microsoft-hosted images
1. The Ubuntu 22.04 image in `NetCore-Public` is compatible with our build/test suite without issue (experiment shows that it is)

It is hard to quantify No.1, because Azure Pipelines does not record the time spent waiting in queue (as far as I know). However, anecdotally, I can say that the build started right away without waiting for agents. dnceng-public can only use up to 200 Microsoft-hosted agents at a time, whereas it can use an unlimited number of selfhosted `NetCore-Public` jobs.

The build was also slightly slower on `NetCore-Public`, Our internal builds on `NetCore1ESPool-Internal` use `Standard_D8a_v4` machines, whereas `NetCore-Public` uses `Standard_D4a_v4` machines. The `NetCore-Public-XL` pool uses the same `Standard_D8a_v4` which would have the same performance.

Therefore, if we have the budget, we should switch public Linux AMD64 builds to `NetCore-Public-XL`.

Issue #6568, opened 2025-07-22
Use online checksums for Aspire Dashboard Dockerfiles


lbussell:
This is part of #6548. Instead of using hard-coded checksums, they should be downloaded from the build/release storage account. This makes the Dockerfiles more flexible and simplifies this repo's maintenance, since we will no longer need to calculate or keep track of checksums.

Example:
https://github.com/dotnet/dotnet-docker/blob/157267f8403b1985704cb6810a5413fc7e3a1bc2/src/runtime/10.0/alpine3.22/amd64/Dockerfile#L7-L16

Related: #5154

PR #6726, opened 2025-10-14
Add Aspire Dashboard support to update-dependencies tooling

lbussell:
Part 1/2 of https://github.com/dotnet/dotnet-docker/issues/5154

## Usage

Running the following command successfully updates the dashboard to version `13.0.0-preview.1.25514.9` (current daily build) with 0 interaction.

```console
dotnet run --project ./eng/update-dependencies -- from-channel 8103 https://github.com/dotnet/aspire
```

Current Aspire channels that I see at https://aka.ms/bar:
- `8103`: daily builds
- `8104`: release candidates
- `8105`: GA

They all work fine with this new code.

## Summary of changes

* Allow multiple implementations of `IBuildUpdaterService` - `BuildUpdaterService` is now `VmrBuildUpdaterService`
* `FromBuildCommand` and `FromChannelCommand` now support multiple repos by resolving different updater services using DI
* Added a new `AspireUpdaterService`
* Allow `AspireUpdaterService` to compute its own variable names and values while still using existing infrastructure. This was done by adding custom updater support to `SpecificCommand`. While it looks like a lot of changes, this keeps the new code decoupled from the old code so that https://github.com/dotnet/docker-tools/issues/1658 will be easier (that change will be in the near future).

**Dockerfile and metadata template simplification:**

Aspire has a tip-only release/support model and versions quickly. To better support that the following changes were made:
* Removed hardcoded versions from Aspire variable names
* Moved Aspire Dashboard Dockerfiles out of version/OS directories

## Follow-ups

- Add a stage to `eng/pipelines/pipelines/update-dependencies.yml` which already runs every day targeting the nightly branch.


eng/update-dependencies/AspireBuildUpdaterService.cs
```diff
@@ -0,0 +1,126 @@
...
+    private string GetDashboardUrl(string baseUrl, Asset dashboardAsset) => $"{baseUrl}/{dashboardAsset.Name}";
+
+    /// <summary>
+    /// Manually compute the checksum of an Aspire Dashboard asset by downloading it from
+    /// <paramref name="url"/> and hashing the contents.
+    /// </summary>
+    /// <remarks>
+    /// Remove once https://github.com/dotnet/dotnet-docker/issues/6568 is completed.
+    /// </remarks>
+    private Task<string?> GetChecksumAsync(string url) => ChecksumHelper.ComputeChecksumShaAsync(_httpClient, url);
```
joperezr, line 110: As FYI, we should be able to upload checksums for the dashboard very easily if you prefer.

eng/update-dependencies/BuildExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+/// <summary>
+/// Extensions for .NET build asset registry <see cref="Build"/>s.
+/// </summary>
+internal static class BuildExtensions
+{
+    /// <summary>
+    /// Given a <see cref="Build"/>, maps its source repository (either GitHub
+    /// or Azure Devops) to a supported <see cref="BuildRepo"/> enum value.
+    /// </summary>
+    public static BuildRepo GetBuildRepo(this Build build)
```
joperezr, line 17: Is this logic just for Aspire? If so, perhaps we can remove the VMR as aspire isn't part of the VMR

joperezr (approved):
Left a couple of comments but this looks super awesome @lbussell thank you so much for working on this!

eng/update-dependencies/AspireBuildUpdaterService.cs
lbussell: That's tracked by https://github.com/dotnet/aspire/issues/10605. If that could be done soon that'd be awesome.

We still need to leave this code in until both https://github.com/dotnet/aspire/issues/10605 and #6568 are complete.

eng/update-dependencies/BuildExtensions.cs
lbussell: No, the repo already gets updates from the VMR for .NET SDK, Runtime, aspnet, etc.

This PR adds Aspire support to what already existed for the VMR.

eng/update-dependencies/AspireBuildUpdaterService.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        _logger.LogInformation("Found Aspire build version: {version}", dashboardAssets.First().Version);
+
+        IEnumerable<string?> dashboardChecksums = await dashboardAssets
+            .Select(asset => GetDashboardUrl(dashboardBaseUrl, asset))
+            .ToAsyncEnumerable()
+            .SelectAwait(async url => await GetChecksumAsync(url))
+            // Null checksums indicate that they weren't found.
+            // An error should have been logged in the terminal if so.
```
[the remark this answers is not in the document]
lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/6726/commits/c107be9198537cac0bde35b75258e10cdec8649a

eng/update-dependencies/AspireBuildUpdaterService.cs
```diff
@@ -84,7 +83,30 @@ public async Task<int> UpdateFrom(Build build, CreatePullRequestOptions pullRequ
...
+            User = pullRequestOptions.User,
+            Email = pullRequestOptions.Email,
+            Password = pullRequestOptions.Password,
+            AzdoOrganization = pullRequestOptions.AzdoOrganization,
+            AzdoProject = pullRequestOptions.AzdoProject,
+            AzdoRepo = pullRequestOptions.AzdoRepo,
+            VersionSourceName = pullRequestOptions.VersionSourceName,
+            SourceBranch = pullRequestOptions.SourceBranch,
+            TargetBranch = pullRequestOptions.TargetBranch,
+        };
```
mthalman: This is repeated a lot through the codebase. I would recommend creating a static `FromPullRequestOptions` method on `SpecificCommandOptions` to encapsulate this.

eng/shared/DotNetVersion.cs
```diff
@@ -0,0 +1,91 @@
...
+        if (SemanticVersion.TryParse(versionString, out var fullSemanticVersion))
+        {
+            return new DotNetVersion(fullSemanticVersion);
+        }
+
+        // We also need to allow major-only and major.minor versions, so fall
+        // back to simple parsing when that happens.
+        // A valid major.minor.patch version would have been handled above so
+        // we don't need to account for that in our parsing.
+        var matchGroups = MajorMinorVersionRegex.Match(versionString).Groups;
```
mthalman, line 45: Is it possible to use `System.Version.TryParse` here instead of our own regex?

lbussell: `System.Version` defaults unmatched version numbers to -1 instead of 0, which I believe is why it wasn't used in the first place. We could write some logic to translate that to a version that has zeros instead of -1, but IMO that code would be more difficult to understand. I'll add a comment explaining the reasoning.

eng/update-dependencies/AspireBuildUpdaterService.cs
lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/6726/commits/8a78c1d282c09570186924267270a155c25a2681

eng/shared/DotNetVersion.cs
lbussell: Comment added in https://github.com/dotnet/dotnet-docker/pull/6726/commits/5770f23de8b1d952d27f6c0925b7b20c41896bd7

lbussell:
`DotNetVersion` was pretty much copied wholesale from update-dependencies. I moved it to a shared project.

PR #6738, opened 2025-10-21
Automatically submit update PRs to internal release branch

lbussell:
Fixes https://github.com/dotnet/dotnet-docker-internal/issues/6741.

Summary of changes:

- Automatically trigger the update-dependencies-internal-official pipeline every time there is a new staged .NET version.
- Only use service connections/managed identity for submitting PRs to Azure DevOps.
- Both official and unofficial pipelines use pipeline resources to reference the staging pipeline, making them more similar.
- Use the Darc backend to submit pull requests in the `from-staging-pipeline` command.
- `from-staging-pipeline` command now has two explicit "modes" - `Local` and `Remote`.
  - `Local` makes changes in your working repo. This is for the local dev workflow.
  - `Remote` can target any branch and does not touch your local git repo. It works in a temporary git repo. It always submits a pull request. This is the same way the sync pipeline works. Example scenarios where this is useful:
    - From your local machine, run one command to submit an update PR to any internal branch without touching your local repo.
    - Use the latest code from the nightly branch to update dependencies in any other internal branch.
  - All the git functionality is encapsulated in the private `GitRepoContext` class so that it required minimal changes to the logic in `FromStagingPipelineCommand.Execute`.
- Share some code between `FromStagingPipelineCommand` and `SyncInternalReleaseCommand`

Example pull request generated by these changes: https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet-docker/pullrequest/54519

### DIFF
--- a/eng/update-dependencies/DirectoryStack.cs
+++ b/eng/update-dependencies/DirectoryStack.cs
@@ -14,6 +14,10 @@ internal static class DirectoryStack
     /// directory stack and returns an <see cref="IDisposable"/> that
     /// restores the previous working directory when disposed.
     /// </summary>
+    /// <remarks>
+    /// This method is not safe to use in multi-threaded scenarios. It affects
+    /// the current working directory of the entire process.
+    /// </remarks>
     public static IDisposable Push(string path) => new DirectoryContext(path);
 
     /// <summary>


eng/update-dependencies/CreatePullRequestOptionsExtensions.cs
```diff
@@ -14,11 +14,54 @@ public static string GetAzdoRepoUrl(this CreatePullRequestOptions options)
...
         ArgumentException.ThrowIfNullOrWhiteSpace(options.AzdoRepo);
-        ArgumentException.ThrowIfNullOrWhiteSpace(options.TargetBranch);
-        ArgumentException.ThrowIfNullOrWhiteSpace(options.SourceBranch);
 
         // AzdoOrganization is a URL like https://dev.azure.com/<org>
         // A valid Azure DevOps repository URL is formatted like https://dev.azure.com/<org>/<project>/_git/<repo>
         return $"{options.AzdoOrganization}/{options.AzdoProject}/_git/{options.AzdoRepo}";
     }
+
+    public static bool TryGetCommitterIdentity(
```
mthalman: This looks to be unused.

eng/update-dependencies/DirectoryStack.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Dotnet.Docker;
+
+/// <summary>
+/// Helper class to manage a stack of directories for changing the current
+/// working directory.
+/// </summary>
+internal static class DirectoryStack
```
mthalman, line 10: Should this use the `IDisposable` pattern? Consider what happens if you push and an exception gets thrown before you pop.

lbussell: Yeah, that'd probably be better than relying on callers to put a call to 'Pop()' in a 'finally' block.

eng/update-dependencies/CreatePullRequestOptionsExtensions.cs
lbussell: Fixed in https://github.com/dotnet/dotnet-docker/pull/6738/commits/d3b66f0df8e6d28179e4a0de922126af230c7ee8

eng/update-dependencies/DirectoryStack.cs
lbussell: I gave this a shot in https://github.com/dotnet/dotnet-docker/pull/6738/commits/d2891d283e92e2c8391abc2edbc17762f3987390.

eng/update-dependencies/DirectoryStack.cs
```diff
@@ -0,0 +1,73 @@
...
```
mthalman, line 10: It might be worthwhile to add a comment that states this class should not be used in multi-threaded scenarios because of its use of `Directory.SetCurrentDirectory` which is static and process-wide.

lbussell: Done in https://github.com/dotnet/dotnet-docker/pull/6738/commits/fc22ceab41b368c316a710213ac8c98dcf8c0af5

PR #6752, opened 2025-10-24
[release/2025-11B] Move `latest` tags to .NET 10

lbussell:
This PR includes 3 changes:
- Move `latest` tags to .NET 10
- Port the remaining update-dependencies PRs from nightly branch - they are now completely in-sync
- Update .NET 10 to the latest daily build

#### Why update to a daily build of .NET 10 in the release branch?

This is a point-in-time thing that only affects the GA release. Here's why:

- The public release branch currently has the public RC2 build of .NET 10.
- In the public release branch (this branch), the `latest` tags are on .NET 9, which is expected because .NET 10 RC2 is a pre-release version.
- The internal release branch has the same contents except it has been updated to an (in-progress) RTM build of .NET 10.
- The internal release branch is failing pre-build validation because the `latest` tags are not on the latest GA version of .NET. On the internal release branch, the latest GA version is 10.0, not 9.0.
- So, we need to move the `latest` tags to .NET 10 in the _public_ release branch so that the change gets mirrored internally to unblock the internal release build.
- Since we're moving the `latest` tags to .NET 10, we need to pick up any GA/RTM build of .NET 10 in order to pass pre-build validation.

The update-dependencies changes were required in order to resolve the correct build versions of .NET 10 as described in https://github.com/dotnet/dotnet-docker/pull/6732.

This PR will be rebase-merged so that the commits stay intact.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config
+++ b/tests/Microsoft.DotNet.Docker.Tests/TestAppArtifacts/NuGet.config
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet10" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-dotnet-0fc3e418/nuget/v3/index.json" />
   </packageSources>
 </configuration>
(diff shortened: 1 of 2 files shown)


lbussell:
I had to pick up https://github.com/dotnet/dotnet-docker/pull/6727 and https://github.com/dotnet/dotnet-docker/pull/6741 in order to get their changes to Dockerfile scrubbing for internal Dockerfile baselines. Rebased to keep commit history clean.

mthalman (approved):

lbussell:
Also picked up changes from https://github.com/dotnet/dotnet-docker/pull/6755.

PR #6755, opened 2025-10-24
Always use NuGet config during tests

lbussell:
This PR targets the main branch because the change will only affect the main and release branches. This change is needed in order to unblock PR validation in dotnet/dotnet-docker#6752.

This should help for dotnet/dotnet-docker-internal#9547, even though that issue only says that the nightly pipeline was reported.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/TestSolution.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/TestSolution.cs
@@ -59,7 +59,7 @@ public class TestSolution : IDisposable
             string nuGetConfigFileName = "NuGet.config";
             if (Config.IsInternal)
             {
-                nuGetConfigFileName += "NuGet.config.internal";
+                nuGetConfigFileName = "NuGet.config.internal";
             }
             else if (Config.IsNightlyRepo)
             {


lbussell:
> Didn't we attempt this before and we ran into issues on release day due to these packages not being mirrored yet?

Yes.

> If that's the case, wouldn't we then have issues with the test for the Windows images that get rebuilt after the internal images get published?

Yes, potentially. I'm wondering if using NuGet.org will block us on release days due to network isolation policies. Or if it doesn't apply to Docker.

The primary motivation for this change was to get PR validation passing in https://github.com/dotnet/dotnet-docker/pull/6752 because that needed the preview feed for the .NET 10 daily build.

Some potential options that I see are:

- Always use the NuGet config in tests, but leave it empty on the main branch like before: `<configuration></configuration>`
- Have a pipeline parameter that controls whether we use NuGet.org or the mirror on release days only.
- Have a pipeline parameter that disables tests that require NuGet packages.

I am wondering which/how many Windows tests reach out to NuGet.org. I can try and run an experiment to find out.

mthalman (approved):

lbussell:
This PR is stale. Check https://github.com/dotnet/dotnet-docker-internal/issues/9547 for the latest info/status on CFSClean/Network isolation progress.

Issue #6663, opened 2025-08-28
Changes needed for .NET 10 GA


lbussell:
- [ ] Move `latest` tags in `manifest.json` to the new version
- [ ] Update documentation as necessary to reference the new .NET version
- [ ] Update IIS sample URL to final version of `dotnet-hosting-win.exe` (remove "preview" from the URL): <https://github.com/dotnet/dotnet-docker/blob/main/samples/aspnetapp/Dockerfile.windowsservercore-iis#L19>.
- [ ] Remove suffix "(Release Candidate)" from Featured Tags and replace with "(Long-Term Support)" or "(Standard Support)"

PR #6762, opened 2025-10-27
Mark .NET 10 as long-term support in Readmes

lbussell:
This PR is part of [Changes needed for .NET 10 GA (#6663)](https://github.com/dotnet/dotnet-docker/issues/6663).


lbussell:
> It'd be nice if the readme template could use the .NET build version (the suffix part) in the `manifest.versions.json` to construct the appropriate text here.

Filed https://github.com/dotnet/dotnet-docker/issues/6764 for this.

PR #6789, opened 2025-11-10
Update dotnet-monitor download paths for internal builds

lbussell:
This PR updates the download paths for internal builds of dotnet-monitor. This is required to build and stage dotnet/monitor and monitor-base images using internal builds.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/monitor-10.0-azurelinux-distroless-amd64-Dockerfile.approved.txt
+++ b/tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/monitor-10.0-azurelinux-distroless-amd64-Dockerfile.approved.txt
@@ -14,10 +14,10 @@ RUN tdnf install -y \
 # Retrieve .NET Monitor extensions
 RUN dotnet_monitor_extension_version=0.0.0 \
     && curl --header "Authorization: Bearer $ACCESSTOKEN" --header "x-ms-version: 2017-11-09" --fail --show-error --location \
-        --output dotnet-monitor-egress-azureblobstorage.tar.gz https://dotnetstage.blob.core.windows.net/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-azureblobstorage-$dotnet_monitor_extension_version-linux-x64.tar.gz \
-        --output dotnet-monitor-egress-s3storage.tar.gz https://dotnetstage.blob.core.windows.net/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-s3storage-$dotnet_monitor_extension_version-linux-x64.tar.gz \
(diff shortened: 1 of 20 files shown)


mthalman (commented):
Shouldn't this require updates to the internal generated Dockerfiles for test validation?

lbussell:
> Shouldn't this require updates to the internal generated Dockerfiles for test validation?

Fixed in https://github.com/dotnet/dotnet-docker/pull/6789/commits/d6682a9d2dc609388dd9939e5c58e2a0a9578d35

tests/Microsoft.DotNet.Docker.Tests/Baselines/GeneratedArtifactTests/VerifyInternalDockerfilesOutput/monitor-10.0-azurelinux-distroless-amd64-Dockerfile.approved.txt
```diff
@@ -14,10 +14,10 @@ RUN tdnf install -y \
...
-        --output dotnet-monitor-egress-s3storage.tar.gz https://dotnetstage.blob.core.windows.net/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-s3storage-$dotnet_monitor_extension_version-linux-x64.tar.gz \
-        --output dotnet-monitor-egress-azureblobstorage.tar.gz.sha512 https://dotnetstage.blob.core.windows.net/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-azureblobstorage-$dotnet_monitor_extension_version-linux-x64.tar.gz.sha512 \
-        --output dotnet-monitor-egress-s3storage.tar.gz.sha512 https://dotnetstage.blob.core.windows.net/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-s3storage-$dotnet_monitor_extension_version-linux-x64.tar.gz.sha512 \
+        --output dotnet-monitor-egress-azureblobstorage.tar.gz https://dotnetstage.blob.core.windows.net/$dotnet_monitor_extension_version/BlobAssets/dotnet-monitor-egress-azureblobstorage-$dotnet_monitor_extension_version-linux-x64.tar.gz \
```
mthalman, line 17: This doesn't look right. It's removing `diagnostics/monitor`.

lbussell: Here's an example of an actual download URL: https://dev.azure.com/dnceng/internal/internal%20Team/_build/results?buildId=2832969&view=logs&j=9c2ccffd-8e7d-5b34-ccf9-d3f1053322e9&t=7d6a3fb1-2b20-5de3-a297-11dcc8cbf94b&l=603

> https://dotnetstage.blob.core.windows.net/dotnet-monitor/9.0.X-servicing.XXXXX.X/BlobAssets/dotnet-monitor-egress-azureblobstorage-symbols-linux-arm64-9.0.X.tar.gz

The real URL doesn't have `diagnostics/monitor` in the path.

The baseline URL is still missing the `dotnet-monitor/` part of the path because I'm relying on that to be part of the base URL as mentioned in the comment. The URL we pass in is generalized across all .NET artifacts:
    https://github.com/dotnet/dotnet-docker/blob/d6682a9d2dc609388dd9939e5c58e2a0a9578d35/tests/Microsoft.DotNet.Docker.Tests/GeneratedArtifactTests.cs#L70-L75
    The same simplification is made for the .NET runtime/aspnet/sdk artifacts as well - the base URL is missing the stage container path for those. It's not perfect but it gets us close enough.

mthalman (approved):

PR #6802, opened 2025-11-11
Make globalapp single file

richlander:
I made this change for two reasons:

- Alpine only has a runtime-deps `extra` image so targeting that and aligning the sample around that as lowest-common-denominator.
- Adds a nice single-file demo.

### DIFF
--- a/eng/update-dependencies/README.md
+++ b/eng/update-dependencies/README.md
@@ -33,9 +33,16 @@ removal tracked by [this GitHub issue](https://github.com/dotnet/docker-tools/is
 
 ### .NET dependencies
 
-If the .NET dependency can be resolved from a BAR (build asset registry) build,
-you may be able to add support to the [BuildUpdaterService](./BuildUpdaterService.cs),
-or adapt its code for a similar use.
+To add support for new .NET repos to the `from-build` and `from-channel`
+commands:
+
+- Add the new repo to the [`BuildRepo`](./BuildRepo.cs) enum.
+- Update the [GetBuildRepo()](./BuildExtensions.cs) extension method to support
+  the new enum value.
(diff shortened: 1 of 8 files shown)


samples/globalapp/Dockerfile.alpine
lbussell: This can be removed

lbussell:
> Alpine only has a runtime-deps extra image so targeting that and aligning the sample around that as lowest-common-denominator.

IMO we should reconsider adding an Alpine runtime extra image. It doesn't feel right that the other size-focused distros we ship (Chiseled, Azlinux distroless) both have one while Alpine doesn't. Even I got confused by that just the other day.

richlander:
Agreed. That crossed my mind. Nothing stopping that. However, there is a large size reduction with this approach to using runtime-deps so we can still move forward with this change.

samples/globalapp/Dockerfile.alpine
baronfel:
```suggestion
```

remove all of the ICU/tzdata packages because they're part of `-extra`

richlander:
```suggestion
```

samples/globalapp/Dockerfile
```diff
@@ -10,11 +10,11 @@ RUN dotnet restore -a $TARGETARCH
 
 # Copy source code and publish app
 COPY --link . .
-RUN dotnet publish -a $TARGETARCH --no-restore -o /app
```
lbussell, line 13: Why is `--no-restore` removed?

richlander: Due to: https://github.com/dotnet/sdk/issues/51667

I wasn't able to get that flow to work.

PR #6803, opened 2025-11-12
Update aspire-dashboard image to 13.0

joperezr:
cc: @lbussell


lbussell:
Can we also change this to `9.5` to `13` and regenerate readmes?

https://github.com/dotnet/dotnet-docker/blob/1aecb1b0d29823d348ff5a73d17797c5ec344d03/eng/readme-templates/FeaturedTags.md?plain=1#L23-L24

You can leave off the `.0`. Since it'll be updated again eventually.

Issue #6817, opened 2025-11-13
update-dependencies cannot update with an internal SDK-only build


lbussell:
Example of failure: https://dev.azure.com/dnceng/internal/internal%20Team/_build/results?buildId=2838162&view=logs&j=6b443fb8-dee0-558e-a185-67b40a667c8d&t=f9bc539e-e1ec-55fb-e6d6-dc57313d0488&l=7626

Internal build that caused the failure: https://dev.azure.com/dnceng/internal/_build/results?buildId=2838088&view=results

Stack trace:

```
Unhandled exception: System.Collections.Generic.KeyNotFoundException: The given key 'runtime' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Dotnet.Docker.FromStagingPipelineCommand.ExecuteAsync(FromStagingPipelineOptions options) in /mnt/vss/_work/1/s/eng/update-dependencies/FromStagingPipelineCommand.cs:line 149
   at Dotnet.Docker.BaseCommand`1.<>c.<<get_Handler>b__4_0>d.MoveNext() in /mnt/vss/_work/1/s/eng/update-dependencies/BaseCommand.cs:line 37
```

lbussell:
Fixed by https://github.com/dotnet/dotnet-docker/pull/6823

PR #6823, opened 2025-11-17
Fix update dependencies from staging pipeline command for SDK-only releases

### LINKED ISSUE #6085: Update-dependencies-internal pipeline cannot handle SDK-only releases
Related: https://github.com/dotnet/dotnet-docker-internal/issues/5952 [internal link]

The pre-release validation update pipeline runs into an issue with SDK-only releases. Here's what happens:

Let's say we release version 8.0.200, and we follow that up with an SDK-only release of 8.0.201. If we want to test 8.0.201 then we would call the update-dependencies tool would be called with the arguments:

```
Set-DotNetVersions \
-ProductVersion 8.0 \
-RuntimeVersion 8.0.X \
-AspnetVersion 8.0.X \
-SdkVersion 8.0.201 \
-UseInternalBuild $true \
...
```

This will cause the SDK pipeline to look for shas from the SDK build staging pipeline given by the argument `--internal-base-url=$InternalBaseUrl`. Since that build was SDK-only, the pipeline will fail to find any SHAs for the pre-existing ASP.NET and Runtime versions. This causes the pipeline to fail.

The workaround is currently to update dotnet versions manually and push a branch yourself.

For an easy solution, we could potentially add a pipeline parameter for "Is SDK-only release".

lbussell:
Fixes #6085, #6817

### DIFF
--- a/eng/update-dependencies/FromStagingPipelineCommand.cs
+++ b/eng/update-dependencies/FromStagingPipelineCommand.cs
@@ -250,6 +250,9 @@ internal partial class FromStagingPipelineCommand : BaseCommand<FromStagingPipel
                 createPullRequest = async (commitMessage, prTitle, prBody) =>
                 {
                     logger.LogInformation("Skipping commit and pull request creation in {Mode} mode.", options.Mode);
+                    logger.LogInformation("Commit message: {CommitMessage}", commitMessage);
+                    logger.LogInformation("Pull request title: {PullRequestTitle}", prTitle);
+                    logger.LogInformation("Pull request body:\n{PullRequestBody}", prBody);
                 };
             }
 


eng/update-dependencies/FromStagingPipelineCommand.cs
```diff
@@ -146,14 +146,18 @@ public override async Task<int> ExecuteAsync(FromStagingPipelineOptions options)
             return exitCode;
         }
 
-        var commitMessage = $"Update .NET {majorMinorVersionString} to {productVersions["sdk"]} SDK / {productVersions["runtime"]} Runtime";
+        var commitMessage = releaseConfig switch
+        {
+            { SdkOnly: true } => $"Update .NET {majorMinorVersionString} SDK to {productVersions["sdk"]}",
```
MichaelSimons, line 151: Nit: consider a pattern where "Update .NET {majorMinorVersionString} SDK to {productVersions["sdk"]}" is not duplicated.  Benefit increased readability and maintainability.

PR #6851, opened 2025-12-03
Add Alpine 3.23 images

lbussell:


eng/mcr-tags-metadata-templates/runtime-tags.yml
```diff
@@ -3,12 +3,14 @@ $(McrTagsYmlTagGroup:10.0-noble)
 $(McrTagsYmlTagGroup:10.0-noble-chiseled)
 $(McrTagsYmlTagGroup:10.0-noble-chiseled-extra)
 $(McrTagsYmlTagGroup:10.0-alpine3.22)
+$(McrTagsYmlTagGroup:10.0-alpine3.23)
 $(McrTagsYmlTagGroup:10.0-azurelinux3.0)
 $(McrTagsYmlTagGroup:10.0-azurelinux3.0-distroless)
 $(McrTagsYmlTagGroup:10.0-azurelinux3.0-distroless-extra)
 $(McrTagsYmlTagGroup:9.0-bookworm-slim)
 $(McrTagsYmlTagGroup:9.0-alpine3.21)
```
mthalman: This reminds me that we should have an issue logged for removing 3.21.

lbussell: I created https://github.com/dotnet/dotnet-docker/pull/6852.

You are correct that we never had a tracking issue for removing 3.21, even after all the floating tag delays with 3.22 - that's not great.

Before this PR is merged, I will make sure to create issues for removing 3.22 and 3.23, both with due dates.

PR #6097, opened 2024-12-06
Add Alpine 3.21 images

lbussell:
Fixes #6095

PR #6852, opened 2025-12-03
Remove Alpine 3.21 images

lbussell:
The .NET container images policy is to stop publishing the previous version of Alpine 3 months after the new version of Alpine was released.

.NET images based on Alpine 3.22 were [released in June 2025](https://github.com/dotnet/dotnet-docker/discussions/6488). We initially intended to stop publishing .NET images based on Alpine 3.21 in September 2025, 3 months after Alpine 3.22 images were first published.

However, we had to [temporarily roll back .NET Alpine floating tags to 3.21](https://github.com/dotnet/dotnet-docker/discussions/6566) due to a compatibility issue with CosmosDB. This delayed the original plan to stop publishing Alpine 3.21 images.

The floating tags were [moved back to Alpine 3.22 in October 2025](https://github.com/dotnet/dotnet-docker/discussions/6566#discussioncomment-14421260). December 2025 marks 3 months since the floating tags were moved back to Alpine 3.22. This means that we can now stop publishing Alpine 3.21 images.

Related:
- #6097 
- #6464 
- #6851


manifest.versions.json
```diff
@@ -95,8 +95,6 @@
     "libicu|jammy": 70,
     "libicu|noble": 74,
 
-    "libssl|alpine3.20": "3",
```
mthalman, line 98: This shouldn't be deleted.

mthalman: Oh whoops. This is an even older version. Never mind!

richlander (approved):

PR #6886, opened 2025-12-17
Add PowerShell tests for dotnet SDK images

anamnavi:
This PR adds PowerShell tests for dotnet SDK images. The existing PowerShell tests are moved from SdkImageTests.cs to PowerShellTests.cs and new tests are added to verify that `$PSVersionTable` returns a `7.x` version, `Invoke-WebRequest` works as expected, and `POWERSHELL_DISTRIBUTION_CHANNEL` environment variable is defined and matches `PSDocker-*` pattern.

### DIFF
--- a/tests/Microsoft.DotNet.Docker.Tests/PowerShellTests.cs
+++ b/tests/Microsoft.DotNet.Docker.Tests/PowerShellTests.cs
@@ -82,10 +82,10 @@ public class PowerShellTests : ProductImageTests
     [MemberData(nameof(GetImageData))]
     public void VerifyPowerShellScenario_InvokeWebRequest(ProductImageData imageData)
     {
-        string command = "(Invoke-WebRequest -Uri 'https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/metadata.json').Content";
+        string command = "(Invoke-WebRequest -Uri 'https://graph.microsoft.com').StatusCode";
         string output = PowerShellScenario_Execute(imageData, command, string.Empty);
 
-        Assert.Contains("\"StableReleaseTag\":", output);
+        Assert.Equal("200", output);
     }
 
     [DotNetTheory]


anamnavi:
@dotnet-policy-service agree company="Microsoft"

tests/Microsoft.DotNet.Docker.Tests/PowerShellTests.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Threading.Tasks;
+using Polly;
+using Polly.Retry;
+using SharpCompress.Common;
+using SharpCompress.Readers;
+using Xunit;
+using Xunit.Abstractions;
+
+#nullable enable
+namespace Microsoft.DotNet.Docker.Tests
```
lbussell: Let's use file-scoped namespaces since this is a new file.

tests/Microsoft.DotNet.Docker.Tests/PowerShellTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            string output = PowerShellScenario_Execute(imageData, command, string.Empty);
+
+            Assert.StartsWith("7", output, StringComparison.OrdinalIgnoreCase);
+        }
+
+        [DotNetTheory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifyPowerShellScenario_InvokeWebRequest(ProductImageData imageData)
+        {
+            string command = "(Invoke-WebRequest -Uri 'https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/metadata.json').Content";
```
lbussell: GitHub rate-limits unauthenticated requests like this one, which has caused build/test failures for us in the past. Is there a different endpoint we can hit that won't be rate-limited? Anything behind a CDN owned by either PowerShell or .NET?

We could probably stand up a local endpoint but then we won't get to test TLS.

tests/Microsoft.DotNet.Docker.Tests/PowerShellTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            : base(outputHelper)
+        {
+        }
+
+        protected override DotNetImageRepo ImageRepo => DotNetImageRepo.SDK;
+
+        public static IEnumerable<object[]> GetImageData()
+        {
+            return TestData.GetImageData(DotNetImageRepo.SDK)
+                .Where(imageData => !imageData.IsDistroless)
```
lbussell: Consider filtering by PowerShell support here instead of distroless-ness. That would also probably fix the failing Arm32/Arm64 Alpine Linux tests.

```suggestion
                .Where(imageData => imageData.SupportsPowerShell)
```

tests/Microsoft.DotNet.Docker.Tests/PowerShellTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        private string PowerShellScenario_Execute(ProductImageData imageData, string command, string? optionalArgs = null)
+        {
+            string image = imageData.GetImage(DotNetImageRepo.SDK, DockerHelper);
+
+            if (!imageData.SupportsPowerShell)
+            {
+                OutputHelper.WriteLine($"PowerShell is not supproted on {image}");
+                return string.Empty;
+            }
```
lbussell: If we filter TestData by `.SupportsPowerShell` as specified in https://github.com/dotnet/dotnet-docker/pull/6886/changes#r2629017596, then this condition becomes unnecessary.

tests/Microsoft.DotNet.Docker.Tests/PowerShellTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        [DotNetTheory]
+        [MemberData(nameof(GetImageData))]
+        public void VerifyPowerShellScenario_PSDistributionEnvVar(ProductImageData imageData)
+        {
+            string command = "$env:POWERSHELL_DISTRIBUTION_CHANNEL";
+            string output = PowerShellScenario_Execute(imageData, command, string.Empty);
+
+            Assert.StartsWith("PSDocker-", output, StringComparison.OrdinalIgnoreCase);
+        }
```
lbussell: Consider using `DockerHelper.GetEnvironmentVariables` here instead of running PowerShell, which would get you two benefits:

- The test will be faster since it won't rely on starting up the image
- The test will no longer need PowerShell to be functional to have an accurate result

PR #6925, opened 2026-01-12
Updates for January 2026 .NET releases

lbussell:


internal-versions.txt
mthalman, line 1: Exclude this

lbussell: Filed https://github.com/dotnet/dotnet-docker/issues/6926

lbussell: Removed via rebase to keep the file out of git history.

PR #7000, opened 2026-01-20
Add workflow that adds `needs-backport` label to pull requests targeting the `nightly` branch

lbussell:
Adding this label to PRs that have not been backported helps not lose track of changes for servicing releases.

I also pinned actions references in our two labeler workflows (following security best practices).

### DIFF
--- a/.github/labeler-pull-request.yml
+++ b/.github/labeler-pull-request.yml
@@ -1,3 +1,0 @@
-# Add 'needs-backport' label to PRs targeting the nightly branch
-needs-backport:
-  - base-branch: 'nightly'
--- a/.github/workflows/label-pull-request.yml
+++ b/.github/workflows/label-pull-request.yml
@@ -1,20 +1,43 @@
-name: "Label Pull Request"
+# Workflow: Label Merged PRs with "needs-backport"
+#
+# This workflow automatically adds a "needs-backport" label to all PRs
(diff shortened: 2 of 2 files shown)


.github/workflows/label-issue.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
-    - uses: github/issue-labeler@v3.4
+    - uses: github/issue-labeler@c1b0f9f52a63158c4adc09425e858e87b32e9685 # v3.4
```
mthalman, line 14: Do you want Renovate wired up to update these?

mthalman (commented):
I'm thinking about the case where a PR gets closed but not merged. Is it possible to only apply this label when the PR gets merged? Or continue to add the label on open but remove it when closed but not merged? That would prevent such PRs from showing up in the label search for action.

lbussell:
> I'm thinking about the case where a PR gets closed but not merged. Is it possible to only apply this label when the PR gets merged? Or continue to add the label on open but remove it when closed but not merged? That would prevent such PRs from showing up in the label search for action.

The labeler action is very configurable. I will see if I can figure out how to configure it one of those two ways.

lbussell:
Had to drop the labeler action and use `gh` CLI instead.

PR #7155, opened 2026-04-17
Fix releases sample support phase handling

richlander:
## Summary
- use a shared support-phase helper in the releases API and app samples
- treat only `active` and `maintenance` phases as supported
- keep preview releases like 11.0 reported as unsupported while restoring support for maintenance channels


richlander:
Found some spelling mitakes in a sample while using them as a demo for a blog post.

richlander:
The spelling mistake above is ironic so I'll leave it.

mthalman (approved):

lbussell:
The PR validation failure is infrastructure and not a problem with the changes.

PR #7198, opened 2026-05-16
Non-nightly images in samples

erwinkramer:
The samples should be for end-users and consistent across all other samples, thus `nightly` versions should be discarded in the 2 samples found.


erwinkramer:
@dotnet-policy-service agree

lbussell (approved):
Nice catch! Thanks for the contribution.

PR #7207, opened 2026-05-21
Fix nested pwsh invocation for non-root user in SDK images

### LINKED ISSUE #7205: Cannot start nested PWSH due to executable permission problems
### Describe the bug

Using `2-10.0`

It appears there is some kind of shim used to redirect /usr/bin/pwsh to `/usr/share/powershell/.store/powershell.linux.x64/7.6.1/powershell.linux.x64/7.6.1/tools/net10.0/any`

It will start fine initially but if you need to start a nested powershell session, it will fail. This is because the new path gets added to `PATH` automatically and it tries to start from here instead of /usr/bin

Image

# Workaround
Make the "nested" pwsh executable, then it will work fine.
`sudo chmod +x '/usr/share/powershell/.store/powershell.linux.x64/7.6.1/powershell.linux.x64/7.6.1/tools/net10.0/any/pwsh'`

### Which .NET image(s) are you using?

mcr.microsoft.com/dotnet-sdk:2-10.0

### Steps to reproduce

Start bash, then start pwsh, then try to start pwsh again

### Other information

_No response_

### Output of `docker version`

```console

```

### Output of `docker info`

```console

```

lbussell:
Fixes #7205.

Previously, nested `pwsh` invocation in Linux SDK images failed when running as a non-root user. The root cause is that PowerShell prepends the tool's nested path to `PATH`, but the nested `pwsh` shim under `/usr/share/powershell/.store/.../tools/*/any/pwsh` has permissions `744`, not `755` like the main PowerShell executable.

Changing the shim from `744` to `755` permissions aligns with how we already set the main `pwsh` executable to `755`, so I don't think this creates an appreciable difference to security.

### DIFF
--- a/eng/dockerfile-templates/sdk/Dockerfile.linux.install-powershell
+++ b/eng/dockerfile-templates/sdk/Dockerfile.linux.install-powershell
@@ -23,6 +23,7 @@ RUN powershell_version={{VARIABLES[cat("powershell|", dotnetVersion, "|build-ver
     && rm {{nupkgName}} \
     && ln -s /usr/share/powershell/pwsh /usr/bin/pwsh \
     && chmod 755 /usr/share/powershell/pwsh \
+    && chmod 755 /usr/share/powershell/.store/powershell.{{lcase(nupkgPlatform)}}/$powershell_version/powershell.{{lcase(nupkgPlatform)}}/$powershell_version/tools/*/any/pwsh \
     # To reduce image size, remove the copy nupkg that nuget keeps.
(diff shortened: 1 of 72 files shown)


mthalman (approved):

lbussell:
This change uncovered https://github.com/PowerShell/PowerShell/issues/27490 which was causing the Arm64 failures. /cc @adityapatwardhan @TravisEz13

mthalman (approved):

lbussell:
Same issue on Alpine/musl since the shim is linked against glibc. I noted that as a comment on the PowerShell issue.

PR #7231, opened 2026-06-09
Reduce size of AOT SDK images by removing unnecessary packages

lbussell:
Native AOT compilation only needs a linker and something to drive it. The .NET runtime defaults to clang but falls back to gcc when clang isn't present. Both are used to drive the same `bfd` linker under the hood, so clang/llvm is not needed. Here is where the runtime calculates the toolchain/linker: https://github.com/dotnet/runtime/blob/4832002fd28748322015287b8fb3b8126334865f/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets#L18-L26

We also replace the heavyweight `build-base`/`build-essential` meta-packages with just the packages AOT linking actually requires (gcc, binutils, libc dev headers, zlib dev).

This is applied only to .NET 11; changing the toolchain on .NET 10 and earlier would be a breaking change.

Size impact:

| Image | Before | After | Saved |
|---|---:|---:|---:|
| dotnet/sdk:11.0-preview-alpine3.23-aot | 401.2 MB | 277.3 MB | 123.9 MB (30.9%) |
| dotnet/sdk:11.0-preview-azurelinux3.0-aot | 565.5 MB | 415.9 MB | 149.6 MB (26.5%) |
| dotnet/sdk:11.0-preview-resolute-aot | 429.5 MB | 362.7 MB | 66.8 MB (15.6%) |
| dotnet/sdk:11.0-preview-azurelinux4.0-aot | 380.4 MB | 380.4 MB | unchanged (already gcc-only) |


lbussell:
This PR is ready for review, but it should be merged only after #7234 is merged.

mthalman (approved):

PR #7329, opened 2026-08-24
Update Resolute Chiseled GCC package baseline

lbussell:
Resolute's libgcc-s1 package now depends on gcc-16-base, which caused the Chiseled image package baseline to fail.

This PR updates the expected package from gcc-14-base to gcc-16-base.

Related: #7328


mthalman (approved):

lbussell:
Merging because all no-cache runs are passing.

PR #7339, opened 2026-08-27
Update Resolute Chiseled OpenSSL package baseline

lbussell:
Resolute Chiseled package validation reports `openssl-provider-legacy` on amd64 and arm32 images, but not arm64 images. This PR makes the expected package baseline architecture-specific so the .NET 10 and .NET 11 image tests match Syft's package inventory.

Related: https://dev.azure.com/dnceng/internal/_build/results?buildId=3058308&view=results


lbussell:
The only failing PR validation is the cached legs, which is expected until a no-cache rebuild of chiseled images is queued.

If there is more turbulence here, I will turn this test off for chiseled resolute images. It isn't providing enough value to be worth the constant breakage.

mthalman (approved):

PR #7343, opened 2026-09-01
Use a common Resolute OpenSSL package baseline

lbussell:
PR #7339 made `openssl-provider-legacy` architecture-specific because earlier package scans did not report it on ARM64. Current Resolute Chiseled images include the package on all supported architectures, causing the ARM64 validation in #7342 to fail.

This PR moves `openssl-provider-legacy` into the common Resolute Chiseled package baseline and leaves only `gcc-16` as AMD64-specific.

Related: #7339
Related: #7342


lbussell:
PR validation is failing only on cached images, as expected.

mthalman (approved):

PR #7349, opened 2026-09-08
Update PowerShell version to v7.4.20

daxian-dbw:
Update PowerShell version to the latest v7.4.20.


lbussell:
Already has version v7.4.20.

